PR #1, opened 2022-03-08
chore: set up boilerplate code

amateima:


README.md
```diff
@@ -1,2 +1,107 @@
-# Across SDK V2
-Repository for shareable code and libraries for Across V2
+# ACROSS V2 SDK
```
chrismaree, line 1: remove this perhaps?

package.json
```diff
@@ -0,0 +1,55 @@
...
+    }
+  },
+  "prettier": {
+    "printWidth": 80,
+    "semi": true,
+    "singleQuote": true,
+    "trailingComma": "es5"
+  },
+  "name": "sdk-v2",
+  "author": "Alexandru Matei",
```
chrismaree, line 35: this should not be your name

package.json
```diff
@@ -0,0 +1,55 @@
...
+    {
+      "path": "dist/sdk-v2.esm.js",
+      "limit": "10 KB"
+    }
+  ],
+  "devDependencies": {
+    "@size-limit/preset-small-lib": "^7.0.8",
+    "husky": "^7.0.4",
+    "size-limit": "^7.0.8",
+    "tsdx": "^0.14.1",
```
chrismaree, line 51: I dont think we want to do this differently to the other packages (we might want to adopt what you did here!). in other ts packages we use `tsc` without any helpers on it. why do we need `tsdx`? can we not just do it the same way we do it through out all the other Uma packages?

package.json
amateima: fixed

README.md
amateima: fixed

.github/workflows/main.yml
```diff
@@ -0,0 +1,36 @@
+name: CI
```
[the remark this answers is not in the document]
amateima: I re-wrote the workflow files

package.json
amateima: Actually we are already using it for the `uma` SDK. It's an easy way to manage all the configurations/deps for setting up the compiler/IDE integration/linter

.github/workflows/main.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Use Node ${{ matrix.node }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+
+      - name: Install deps and build (with cache)
+        uses: bahmutov/npm-install@v1
```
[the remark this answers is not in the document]
amateima: yep. Fixed

package.json
daywiss: one thing to mention about tsdx, its basically an opinionated way to setup a typescript library, but it does not play well with a monorepo, which is why it kind of sucks in protocol. But here as a standalone repo, i think its fine and does a lot of the boilerplate for you.

daywiss (approved):
lgtm, but im not super opinionated on this, i would listen to chris if he has reservations about tsdx

package.json
```diff
@@ -0,0 +1,55 @@
...
+  "prettier": {
+    "printWidth": 80,
+    "semi": true,
+    "singleQuote": true,
+    "trailingComma": "es5"
+  },
+  "name": "sdk-v2",
+  "author": "across",
+  "module": "dist/sdk-v2.esm.js",
+  "size-limit": [
```
chrismaree, line 37: why do you need this?

package.json
```diff
@@ -0,0 +1,55 @@
...
+    "hooks": {
+      "pre-commit": "tsdx lint"
+    }
+  },
+  "prettier": {
+    "printWidth": 80,
+    "semi": true,
+    "singleQuote": true,
+    "trailingComma": "es5"
+  },
```
chrismaree, line 33: should we follow the same pattern as done in the other across repos: https://github.com/across-protocol/contracts-v2/blob/master/.prettierrc

you can just copy this file I think

chrismaree (approved):
LGTM! I had a few more nits on the linting style but it's not a huge deal.

package.json
amateima: done

package.json
amateima: It just came with tsdx and it performs bundle analysis. I want to keep it for now to see how it plays

PR #2, opened 2022-03-09
fix: add missing linter files

amateima:
Small PR, I forgot to add the linter files to #1 PR


.eslintrc.js
```diff
@@ -0,0 +1,12 @@
...
+  "extends": [
+    "react-app",
+    "prettier/@typescript-eslint",
+    "plugin:prettier/recommended"
+  ],
+  "settings": {
+    "react": {
+      "version": "999.999.999"
+    }
+  }
```
daywiss, line 11: the eslint in protocol is a lot different than this, im ok with not matching that but @chrismaree might have a different opinon

.prettierrc
```diff
@@ -0,0 +1,14 @@
+{
+  "printWidth": 120,
+  "bracketSpacing": true,
+  "explicitTypes": "preserve",
+  "tabWidth": 2,
```
daywiss, line 5: this matches the prettier file in protocol

daywiss (approved):

.prettierrc
amateima: I was just following Chris suggestion [from here](https://github.com/across-protocol/sdk-v2/pull/1#discussion_r821936666)

PR #5, opened 2022-03-22
chore: add pool client with only modifications for building

daywiss:

Migrates v1 pool client over to v2. Imports stuff from uma/sdk as well. code hasnt changed other than that. tests pass.


.eslintrc.js
```diff
@@ -2,8 +2,12 @@ module.exports = {
   "extends": [
     "react-app",
     "prettier/@typescript-eslint",
-    "plugin:prettier/recommended"
+    "plugin:prettier/recommended",
+    "eslint:recommended",
   ],
+  rules: {
+    "prettier/prettier": ["warn"],
```
daywiss, line 9: this was giving me an error on somethign that isnt an error. it was erroring where i use `import type`

```
/home/david/src/across/sdk-v2/src/pool/bridgePool.ts                                                            
 5:14  warning  Parsing error: '=' expected  prettier/prettier  
```

jest-e2e-config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "testMatch": ["**/*.e2e.ts"]
```
daywiss, line 2: this is to run jest e2e tests

package.json
```diff
@@ -39,14 +41,17 @@
     }
   ],
   "devDependencies": {
+    "@eth-optimism/contracts": "^0.5.18",
```
daywiss, line 44: this was required to import `@uma/sdk`

src/utils.ts
```diff
@@ -56,3 +61,131 @@ export function max(a: BigNumberish, b: BigNumberish): BN {
...
+  periodsElapsed: Decimalish,
+  periodsPerYear: Decimalish
+): string => {
+  return new Decimal(endAmount)
+    .sub(startAmount)
+    .div(startAmount)
+    .mul(periodsPerYear)
+    .div(periodsElapsed)
+    .toString();
+};
```
daywiss, line 191: these are all kind of specific to pool client but decided might as well just put it in this util file

.eslintrc.js
```diff
@@ -2,8 +2,12 @@ module.exports = {
   "extends": [
     "react-app",
     "prettier/@typescript-eslint",
-    "plugin:prettier/recommended"
+    "plugin:prettier/recommended",
```
amateima, line 5: I think `plugin:prettier/recommended` should always be used as the last item in the array to make sure all rules that conflict with Prettier are overwritten

amateima (approved):

.eslintrc.js
daywiss: i see ok, i can actually remove the one i added, i was trying to get rid of the error i was getting

src/pool/bridgePool.e2e.ts
```diff
@@ -0,0 +1,149 @@
...
+import { Client, Provider, PoolEventState } from "./bridgePool";
+import { ethers, BigNumber } from "ethers";
+import { TransactionReceipt } from "@ethersproject/abstract-provider";
+import assert from "assert";
+import set from "lodash/set";
+import get from "lodash/get";
+
+const { bridgePool } = uma.clients;
+
+dotenv.config();
```
amateima, line 12: I'm just thinking, if each "sub-package" will require a different set of env variables, would be necessary to have one .env file for each 'sub-package' or keep a single .env file for the whole project?

daywiss: hmm, good question, i would just keep a single env but open to a different way

PR #6, opened 2022-03-24
feat: add transfers history client

amateima:
I re-worked the implementation and used David's suggestion as a starting point. Basically now I'm polling the chains in the background and the users can request the transfers asynchronously.

The central point of this implementation is the `SpokePoolEventsQueryService` which is a stateful service, instantiated per each deposit - provider pair  and it has the responsibility to query the chains (by using the `ContractEventsQuerier` adapter), keep track of the queried blocks and update the repositories which store data.

The users can request the filled and pending transfers at any time, but this client also triggers events when the data is updated.

___

~~This is the first proposal for what should represent the client used by FE to get a list of deposits performed in Across V2 along with their status and other relevant data.
This client has the job to query events from multiple chains (L2s + L1) and aggregate them. One particular challenge I encountered during the implementation was the fact that block time differs on all these chains and I also tried to optimise the response time given the fact that the most requested table page in the FE will be page 1. Here's a summary of my implementation:~~

~~1. The requester instantiate the [TransfersClient](https://github.com/across-protocol/sdk-v2/pull/6/files#diff-916d18a80c514fc8c883139e2cd5e3e3bcad780b37da4caf1e47b58293c3476eR24) with the details of the chains that need to be queried, along with the `id` of the chain that should serve as reference. You will find this "reference chain" many times across the implementation, so let me explain it. 
🔵 If the most requested table page will be the first one, it doesn't make sense to query all the blocks. The blocks will be processed in batches, more batches to be requested only if the number of fetched transfers is not enough (limit + offset). Usually the reference chain id should be the id of the chain that has the highest block time, because each batch of queried blocks covers the biggest time span. The chains other than `refChainId` will be queried until the timestamp of the oldest block from the `refChain` is passed in order to make sure we cover the same time span across all chains.
2. The FundsDeposited and FilledRelay events are queried from a batch of blocks of the reference chain.
3. The FundsDeposited and FilledRelay events are queried from other chains **until** is covered the same time span as on the reference chain
4. Aggregate the events from all chains in a single array and filter only the events that have the timestamp >= timestamp of the lowest block fetched from the reference chain
5. If needed, repeat steps 2, 3 and 4 until the number of the requested events is met or until all the blocks are queried.~~

~~Other mentions:~~
~~- In order to not query the chain to block 0, we can specify [what is the last block](https://github.com/across-protocol/sdk-v2/pull/6/files#diff-b71f8c193d9c0d59b3169ee14400bb65d925558e8959cf76eb13e73c66dd519eR36) that has to be queried on each chain. Usually these are the blocks when we deployed the `SpokePool` contracts~~
~~- [the adapter that queries the events](https://github.com/across-protocol/sdk-v2/pull/6/files#diff-9dc2da77df8e56349e91d911cf074bb1802a24646f875dcbb8aa3e0d900d1058R21) has the ability to adjust the block range automatically in case the JSON-RPC provider has limitations regarding the block range that can be queried or the length of the response~~

~~Tests
I started a series of unit tests by mocking the events data returned by `queryFilter`, but I didn't finish them because the implementation became very verbose when I wanted to test all the edge cases. I tested the implementation with a series of e2e tests by using Infura providers and the testnet contracts, but obviously they work only locally because of the env variables that need to be passed. Not sure how to properly proceed with the testing...~~


src/transfers-history/services/FundsDepositedEventsQueryService.ts
```diff
@@ -0,0 +1,173 @@
...
+
+      do {
+        continueQueryEvents = false;
+        const result = await this._getEvents(isRefChain, toBlock);
+        fromBlock = Math.min(fromBlock, result.fromBlock);
+        toBlock = Math.max(toBlock, result.toBlock);
+        numberOfEvents += result.numberOfEvents;
+
+        const latestRefTimestamp = this.state.progress[this.refChainId].latestFromTimestamp;
+        const latestTimestamp = this.state.progress[this.chainId].latestFromTimestamp;
```
daywiss, line 81: i think this kind of optimization might need to be justified more. Why not just let each chain download all its events in the background, and show whatevers available on the FE as it comes in?

src/transfers-history/client.ts
```diff
@@ -0,0 +1,105 @@
...
+    // query events from the chain used as reference
+    await depositEventsRefQueryService.getEvents();
+    await Promise.all(depositEventsQueryServices.map(([_, service]) => service.getEvents()));
+
+    this.transfersAggregatorService.aggregateTransfers();
+    
+    // continue to query events if limit + offset is not reached and we still have blocks to query
+    while (
+      !this.hasRequiredNumberOfTransfersInState(filters.status, limit, offset) &&
+      this.hasBlocksToQuery(chainsIds)
```
daywiss, line 68: while maybe a good optimization, i think this adds a lot of complexity, vs just letting everything download in the background while returning to the user everything thats available

daywiss (commented):
thanks for putting this up, i know this was a lot of work. take a look at these comments and see what you think

src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
```diff
@@ -0,0 +1,144 @@
...
+          events.push(...newEvents);
+        }
+      } catch (error) {
+        if ((error as Web3Error).error.code === Web3ErrorCode.BLOCK_RANGE_TOO_LARGE) {
+          events = [];
+          Logger.debug(
+            `[SpokePoolEventsQuerier::getEventsInBatches]`,
+            `🔴 lowering block range size from ${this.blockRangeSize} to ${(this.blockRangeSize as number) / 2}`
+          );
+          this.blockRangeSize = (this.blockRangeSize as number) / 2;
```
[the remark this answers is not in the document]
mrice32: Agreed. Is there a specific reason we need this to be dynamic? Or is this just trying to make the system more resilient in the face of changing node providers?

src/utils.ts
```diff
@@ -62,6 +62,28 @@ export function max(a: BigNumberish, b: BigNumberish): BN {
...
 export const fixedPointAdjustment = toBNWei("1");
 
+/**
+ * Takes two values and returns a list of number intervals
+ *
+ * @example
+ * ```js
+ * getSamplesBetween(1, 9, 3) //returns [[1, 3], [4, 7], [8, 9]]
+ * ```
+ */
```
mrice32, line 72: Can you give more context here?

1. These aren't arbitrary intervals in that range? How are they picked?
2. What is this meant to do? Is there an example of how this might be useful?

src/transfers-history/services/TransfersAggregatorService.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    if (fromTimestamps.length === 0) {
+      // We didn't fetch any events from chain yet
+      Logger.debug(`[TransfersAggregatorService::aggregateTransfers]`, `🟠 no events in the state to aggregate`);
+      return;
+    }
+
+    // get the highest fromTimestamp from the fetched events
+    const highestFromTimestamp = Math.max(...fromTimestamps);
+    // console.log(this.state.chainTransfers);
```
mrice32, line 22: nit: comment

src/transfers-history/client.ts
```diff
@@ -0,0 +1,105 @@
...
+    const chainsIds = this.config.chains.map(chain => chain.chainId);
+    Logger.debug("[TransfersHistoryClient::getTransfers]", `start getting ${filters.status} transfers from ${chainsIds} chains`);
+    const depositEventsQueryServices = chainsIds
+      .filter(chainId => chainId !== this.config.refChainId)
+      .reduce<[ChainId, IDepositEventsQueryService][]>((acc, chainId) => {
+        return [
+          ...acc,
+          [chainId, this.depositEventsQueryServiceFactory.getService(this.state, chainId, this.config.refChainId, 100000)],
+        ];
+      }, []);
```
mrice32, line 52: How is this different from 
```
.map(chainId => [chainId, this.depositEventsQueryServiceFactory.getService(this.state, chainId, this.config.refChainId, 100000)])
```

I only ask because I think maps are a bit simpler.

src/transfers-history/model/state.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  public insertQueryProgress(chainId: ChainId, progress: QueryProgress) {
+    this.progress[chainId] = progress;
+  }
+
+  public setLatestFromTimestamp(chainId: ChainId, from: number) {
+    if (!this.progress[chainId]) {
+      this.progress[chainId] = {};
+    }
```
mrice32, line 76: Do the chainIds used in TransfersHistoryClient ever change? Any reason we can't make those constructor params here? If so, we could pre-initialize all of these objects (and maybe the values too) so we wouldn't need to do this check over and over.

src/transfers-history/model/state.ts
```diff
@@ -0,0 +1,102 @@
...
+
+const blockLowerBoundInitialValues = {
+  [ChainId.MAINNET]: 0,
+  [ChainId.RINKEBY]: 0,
+  [ChainId.KOVAN]: 0,
+  [ChainId.OPTIMISM_KOVAN]: 0,
+  [ChainId.ARBITRUM_RINKEBY]: 9828565,
+};
+
+export class State {
```
mrice32, line 44: I think this name is a little nonspecific. What is this state tracking? What is it meant to be used for?

mrice32 (commented):
Looks awesome! A few comments

src/transfers-history/adapters/logger.ts
```diff
@@ -0,0 +1,12 @@
+enum CliColor {
+  Reset = "\x1b[0m",
+  FgGreen = "\x1b[32m",
+}
+
+export class Logger {
+  static debug(caller: string, message: string) {
+    if (process.env.LOG_LEVEL === "debug") {
```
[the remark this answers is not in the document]
amateima: removed the envs usage

src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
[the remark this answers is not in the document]
amateima: >  is this just trying to make the system more resilient in the face of changing node providers

yep

> adjust it at the user level. personally id just let this error

My opinion is that the caller should not care about the block rage. It just wants to fetch the transfers without dealing with implementation details

src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    } while (retryWithLowerBatchSize);
+
+    return {
+      events,
+    };
+  }
+}
+
+// export class SpokePoolEventsQuerierTest implements IContractEventsQuerier {
```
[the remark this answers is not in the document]
amateima: Removed it

src/transfers-history/services/TransfersAggregatorService.ts
```diff
@@ -0,0 +1,58 @@
...
+        ...acc,
+        ...depositIdTransferTuples
+          .filter(([_, transfer]) => transfer.depositTime >= highestFromTimestamp)
+          .map(([_, transfer]) => transfer),
+      ];
+    }, []);
+    Logger.debug(
+      `[TransfersAggregatorService::aggregateTransfers]`,
+      `🟢 found ${transfers.length} with timestamp >= ${highestFromTimestamp}`
+    );
```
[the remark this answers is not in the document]
amateima: This implementation is no longer valid, but basically I was making sure that I aggregate the events from multiple chains over the same time span

src/transfers-history/services/FundsDepositedEventsQueryService.ts
```diff
@@ -0,0 +1,173 @@
...
+    const { amount, originToken, destinationChainId, depositId } = args;
+    const transfer: Transfer = {
+      amount,
+      assetAddr: originToken,
+      depositId: depositId,
+      depositTime: timestamp,
+      depositTxHash: transactionHash,
+      destinationChainId: destinationChainId.toNumber(),
+      filled: BigNumber.from("0"),
+      sourceChainId: ChainId.ARBITRUM_RINKEBY,
```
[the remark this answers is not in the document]
amateima: removed

src/transfers-history/services/TransfersAggregatorService.ts
```diff
@@ -0,0 +1,58 @@
...
+    const filteredTransfers = transfers.reduce(
+      (acc, transfer) => ({
+        pending: transfer.status === "pending" ? [...acc.pending, transfer] : acc.pending,
+        filled: transfer.status === "filled" ? [...acc.filled, transfer] : acc.filled,
+      }),
+      { pending: [] as Transfer[], filled: [] as Transfer[] }
+    );
+
+    this.state.completedTransfers = filteredTransfers.filled;
+    this.state.pendingTransfers = filteredTransfers.pending;
```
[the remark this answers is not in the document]
amateima: what do you mean? The requests from FE will come separately for pending/completed transfers

src/constants.ts
```diff
@@ -0,0 +1,12 @@
+export enum ChainId {
+  ARBITRUM_RINKEBY = 421611 as any,
```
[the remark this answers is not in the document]
amateima: I ditched this approach, but it was an workaround to be able tu use enum values in `switch` cases

src/utils.ts
amateima: This just splits an interval into multiple intervals that have the same length. Usage? To get the block ranges between blockNumber1 and blockNumber2 with each range having a max length of x (the third parameter of the function)

src/transfers-history/services/TransfersAggregatorService.ts
amateima: removed

src/transfers-history/model/state.ts
amateima: I completely change the implementation here. I'm using the Repository pattern now

src/transfers-history/services/SpokePoolEventsQueryService.ts
```diff
@@ -0,0 +1,98 @@
...
+export class SpokePoolEventsQueryService {
+  private latestBlockNumber: number | undefined;
+
+  constructor(
+    private chainId: ChainId,
+    private provider: providers.Provider,
+    private eventsQuerier: ISpokePoolContractEventsQuerier,
+    private logger: Logger,
+    private transfersRepository: TransfersRepository,
+    private depositorAddr?: string
```
daywiss, line 21: i feel like you should not make more than one of these per depositor address, othewrise you will re-query a lot of data you already have

src/transfers-history/services/SpokePoolEventsQueryService.ts
```diff
@@ -0,0 +1,98 @@
...
+      "[SpokePoolEventsQueryService::getEvents]",
+      `🟢 chain ${this.chainId}: fetched events from ${from} to ${to}`
+    );
+    const depositEvents = await this.eventsQuerier.getFundsDepositEvents(from, to, this.depositorAddr);
+    const filledRelayEvents = await this.eventsQuerier.getFilledRelayEvents(from, to, this.depositorAddr);
+    this.logger.debug(
+      "[SpokePoolEventsQueryService::getEvents]",
+      `🟢 chain ${this.chainId}: fetched ${depositEvents.length} FundsDeposited events and ${filledRelayEvents.length} FilledRelayEvents`
+    );
+    const blockTimestampMap = await this.getBlocksTimestamp([...depositEvents, ...filledRelayEvents]);
```
daywiss, line 48: im really surprised no timestamp data is available on the events, also why is timestamp so important?

src/transfers-history/adapters/logger.ts
```diff
@@ -0,0 +1,34 @@
...
+  FgReg = "\x1b[31m",
+}
+
+export type LogLevel = "info" | "debug";
+
+export class Logger {
+  private currentLevel: LogLevel;
+
+  public constructor() {
+    this.currentLevel = "info";
```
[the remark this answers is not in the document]
amateima: done

src/transfers-history/client.ts
```diff
@@ -0,0 +1,121 @@
...
+    this.logger.setLevel(level);
+  }
+
+  public async startFetchingTransfers(depositorAddr: string) {
+    this.initSpokePoolEventsQueryServices(depositorAddr);
+    this.getEventsForDepositor(depositorAddr);
+    const timer = setInterval(() => {
+      this.getEventsForDepositor(depositorAddr);
+    }, this.pollingIntervalSeconds * 1000);
+    this.pollingTimers[depositorAddr] = [...(this.pollingTimers[depositorAddr] || []), timer];
```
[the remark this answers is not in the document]
amateima: good point. added [here](https://github.com/across-protocol/sdk-v2/pull/6/commits/fb1b281e2d1a2737bbd6669f648991ded52db5f5)

src/transfers-history/client.ts
```diff
@@ -0,0 +1,121 @@
...
+export type TransfersHistoryClientParams = {
+  chains: {
+    chainId: ChainId;
+    providerUrl: string;
+  }[];
+  pollingIntervalSeconds?: number;
+};
+
+export class TransfersHistoryClient {
+  private eventEmitter = new EventEmitter();
```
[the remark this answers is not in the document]
amateima: > if you want to use event emitter, i think you could also inherit from it so the class natively supports events.

I wanted to keep it separated in case the events logic becomes more complex and want to migrate to a more complicated event bus/event broker approach

> also you could be less opinionated and just use a callback in the constructor, then the caller can decide if they need an event emitter or not ( or what kind they want)

@daywiss hmm, what if we will have multiple event types in the future? Add all of them to the constructor?
I kind of implemented this way on purpose. Events are a separated concern and I didn't way future events changes to impact the current function/constructor calls. One improvement I can do is instantiate the event emitter only if the user subscribes to events using `.on()` call

src/transfers-history/client.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private initSpokePoolEventsQueryServices(depositorAddr: string) {
+    const chainIds = Object.keys(this.spokePoolInstances).map(chainId => parseInt(chainId));
+
+    for (const chainId of chainIds) {
+      if (!this.eventsServices[depositorAddr]) {
+        this.eventsServices[depositorAddr] = {};
+      }
+      this.eventsServices[depositorAddr][chainId] = new SpokePoolEventsQueryService(
```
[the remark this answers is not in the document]
amateima: added this check

src/transfers-history/services/SpokePoolEventsQueryService.ts
amateima: the `queryFilter` calls contains the depositor address as a filter, so I don't think I query data I already have

src/transfers-history/services/SpokePoolEventsQueryService.ts
amateima: we already discussed this during the meeting. It was the only criteria I found for sorting events coming from multiple chains

src/transfers-history/client.ts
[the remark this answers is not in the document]
daywiss: callbacks would work teh same as emitting an event, theres an event type which can be an enum or a string, then theres some common data interface.  the user then can subscribe to particular event types using an event emitter.

```
const events = new Events()
// this is now a callback which the user can decide if they need an event emitter
const transferHistory = new TransferHistory(config,(type,data)=>events.emit(type,data))

// somewhere we want to subscribe
events.on('pendingTransfers',(transfer)=>// etc)

**Review by daywiss — APPROVED** [review]
nice work, cant wait to see how this looks in the FE

File: src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts [hunk 21]
```diff
@@ -0,0 +1,111 @@
...
+      );
+    }
+    return this.spokePool.filters.FundsDeposited();
+  }
+
+  private async getEvents(
+    from: number,
+    to: number,
+    filters: TypedEventFilter<TypedEvent<any>[], any>
+  ): Promise<TypedEvent<any>[]> {
```
mrice32, line 75: Can we make this function generic so it will pass through these types? Or are the typechain types not specific enough to do that?

src/transfers-history/client.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  public async startFetchingTransfers(depositorAddr: string) {
+    const timers = this.pollingTimers[depositorAddr];
+    if (timers) throw new Error(`Address ${depositorAddr} is already monitored`)
+
+    this.initSpokePoolEventsQueryServices(depositorAddr);
+    this.getEventsForDepositor(depositorAddr);
+    const timer = setInterval(() => {
+      this.getEventsForDepositor(depositorAddr);
+    }, this.pollingIntervalSeconds * 1000);
```
mrice32, line 76: Why does this client include polling?

If this is meant for the frontend, doesn't this go against the philosophy of allowing the frontend to manage the frequency and orchestration of updates in hooks? Otherwise, don't you end up with problems where you can update midway through render cycles?

src/transfers-history/client.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    if (timers) {
+      timers.map(timer => clearInterval(timer));
+      delete this.pollingTimers[depositorAddr];
+    }
+  }
+
+  public on(event: TransfersHistoryEvent, cb: TransfersHistoryClientEventListener) {
+    this.eventEmitter.on(event, cb);
+  }
```
mrice32, line 91: If event emitters are the primary way to communicate state updates with the outside world, they may not play particularly nicely with react-query's async-await usage, as they may need to be wrapped in promises.

src/transfers-history/config.ts
```diff
@@ -0,0 +1,24 @@
...
+  web3ProvidersUrls: {},
+  spokePools: {
+    [CHAIN_IDs.ARBITRUM_RINKEBY]: {
+      addr: "0x68306388c266dce735245A0A6DAe6Dd3b727A640",
+      lowerBoundBlockNumber: 9828565,
+    },
+    [CHAIN_IDs.OPTIMISM_KOVAN]: { addr: "0x99EC530a761E68a377593888D9504002Bd191717", lowerBoundBlockNumber: 0 },
+    [CHAIN_IDs.RINKEBY]: { addr: "0x90743806D7A66b37F31FAfd7b3447210aB55640f", lowerBoundBlockNumber: 0 },
+    [CHAIN_IDs.KOVAN]: { addr: "0x73549B5639B04090033c1E77a22eE9Aa44C2eBa0", lowerBoundBlockNumber: 0 },
+  },
```
mrice32, line 23: Any reason we wouldn't ask that the caller provide this? Preferably, there's one source of truth for these sorts of addresses, either in the FE, where it's passed to the SDK or the SDK holds these address constants and they are used by the FE where it needs them.

Right now, I think both have a consts file, which could lead to issues where one is updated and the other is not.

src/utils.ts
```diff
@@ -189,3 +189,24 @@ export const calcApr = (
     .div(periodsElapsed)
     .toString();
 };
+/**
+ * Takes two values and returns a list of number intervals
```
mrice32, line 193: nit: this explanation is a bit confusing. What is the expected use case for this? In the example below, I don't see an obvious relationship between the inputs and outputs. Maybe a real-world example would help?

mrice32 (commented):
Looks good! A few high-level comments

src/utils.ts
amateima: https://github.com/across-protocol/sdk-v2/pull/6#discussion_r837369366

src/transfers-history/adapters/db/transfers-repository.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  public aggregateTransfers(depositorAddr: string) {
+    const chainsIds = Object.keys(this.transfers).map(chainId => parseInt(chainId));
+    const transfers: Transfer[] = [];
+
+    for (const chainId of chainsIds) {
+      transfers.push(...this.getTransfersByChainAndDepositor(chainId, depositorAddr));
+    }
+
+    transfers.sort((t1, t2) => t2.depositTime - t1.depositTime);
```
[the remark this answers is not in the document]
amateima: done

src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
amateima: I tried something like this, but it errors 

```ts
export class SpokePoolEventsQuerier<T extends TypedEvent<any>>  {
    async getEvents() {
        let events: T[] = [];
        const newEvents = await this.spokePool.queryFilter(filters, blockStart, blockEnd);
        events.push(...newEvents);
        // ^^ see screenshot for error
    }
}
```
Screenshot 2022-04-01 at 02 56 49

src/transfers-history/client.ts
amateima: They said react-query has relatively nice support for this kind of async updates

src/transfers-history/client.ts
amateima: We can disable the polling part (or add a config parameter for enabling it), but my vision for the FE page is to make it more interactive by monitoring and updating the deposits table in real time, not showing just a static table. 
I see this client as a "backend" which hides the implementation details and expose the simplest public interface to the FE. 

I don't think we can end up with these kind of problems because the FE updates its state by calling the getTransfers() functions, like an API endpoint for example. The events emitted are just to inform the FE and the user that new data is available

src/transfers-history/config.ts
amateima: [I changed](https://github.com/across-protocol/sdk-v2/commit/e97b8ce0b4c82c65dfa6d775cd143dccc82770ac#diff-916d18a80c514fc8c883139e2cd5e3e3bcad780b37da4caf1e47b58293c3476eR27) the config passed to the constructor method to contain the addresses of the SpokePool contracts

src/transfers-history/client.ts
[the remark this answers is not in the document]
daywiss: this looks like its still adding multiple timers for an address

mrice32 (approved):

package.json
```diff
@@ -44,17 +44,26 @@
...
     "tslib": "^2.3.1",
+    "typechain": "^7.0.1",
     "typescript": "^4.6.2"
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^0.0.29",
+    "@nomiclabs/hardhat-ethers": "^2.0.5",
+    "@nomiclabs/hardhat-etherscan": "^3.0.3",
+    "@nomiclabs/hardhat-waffle": "^2.0.3",
+    "@typechain/hardhat": "^5.0.0",
```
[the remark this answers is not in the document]
amateima: I moved hardhat plugins to be part of the dev deps 👍

PR #9, opened 2022-03-25
feat: update pool client for new hubpool contract

daywiss:

this converts the old pool client, which had a  bridge pool per erc20 token, to a single hub pool.  It tries to index all data based on the l1 token, and keeps most interfaces the same when dealing with the frontend.

this was e2e tested on kovan againts the currently deployed hubpool contract. some sample output from the tests:
image

I think we are missing ratemodel address, so i will need to look into that


src/pool/poolClient.ts
```diff
@@ -268,8 +292,8 @@ function joinPoolState(
 
   if (rateModel) {
     projectedApr = fromWei(
-      calculateInstantaneousRate(rateModel, poolState.liquidityUtilizationCurrent)
-        .mul(poolState.liquidityUtilizationCurrent)
+      calculateInstantaneousRate(rateModel, poolState.utilizedReserves)
+        .mul(poolState.utilizedReserves)
```
daywiss, line 296: not sure if this is equivalent

src/pool/poolClient.ts
```diff
@@ -278,69 +302,59 @@ function joinPoolState(
...
+    return this.poolState.read(tokenAddress, latestBlock);
   }
 }
 export function validateWithdraw(pool: Pool, user: User, lpTokenAmount: BigNumberish) {
   const l1TokensToReturn = BigNumber.from(lpTokenAmount).mul(pool.exchangeRateCurrent).div(fixedPointAdjustment);
   assert(BigNumber.from(l1TokensToReturn).gt("0"), "Must withdraw amount greater than 0");
-  assert(
-    BigNumber.from(pool.liquidReserves).gte(l1TokensToReturn.add(pool.pendingReserves)),
-    "Utilization too high to remove that amount, try lowering withdraw amount"
-  );
```
daywiss, line 315: i removed this because therse no pending reserves, but this means theres no client side error if a user is trying to remove more than the active liquidity

mrice32 (approved):
LGTM!

amateima (approved):
👍

PR #11, opened 2022-03-31
fix: various issues during UI integration

daywiss:

Small tweaks and bugs fixed during integration with FE


mrice32 (approved):
LGTM!

PR #14, opened 2022-04-07
fix(transfers-history): improve response times

amateima:
This PR fixes a couple of issues identified during testing the FE integration

1. getPendingTransfers() was returning the filled transfers
2. I parallelised the calls for getting FundsDeposited and RelayFilled events
3. I parallelised the queryFilter() calls in case these calls need to be performed in batches (web3 provider limitation)
4. `lowerBoundBlockNumber` optional parameter is now part of the client config object and SHOULD be provided in order to limit the number of the queried blocks

### DIFF
--- a/src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
+++ b/src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
@@ -78,27 +78,40 @@ export class SpokePoolEventsQuerier implements ISpokePoolContractEventsQuerier {
     let retryWithLowerBatchSize;
 
     do {
+      const blockRangeSizeAtStart = this.blockRangeSize;
       try {
         retryWithLowerBatchSize = false;
         events = [];
 
         if (this.blockRangeSize) {
-          for (const [intervalStart, intervalEnd] of getSamplesBetween(from, to, this.blockRangeSize)) {
-            const newEvents = await this.spokePool.queryFilter(filters, intervalStart, intervalEnd);
-            events.push(...newEvents);
-          }
+          const intervals = getSamplesBetween(from, to, this.blockRangeSize);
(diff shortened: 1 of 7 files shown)


src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
```diff
@@ -78,27 +78,40 @@ export class SpokePoolEventsQuerier implements ISpokePoolContractEventsQuerier {
...
         events = [];
 
         if (this.blockRangeSize) {
-          for (const [intervalStart, intervalEnd] of getSamplesBetween(from, to, this.blockRangeSize)) {
-            const newEvents = await this.spokePool.queryFilter(filters, intervalStart, intervalEnd);
-            events.push(...newEvents);
-          }
+          const intervals = getSamplesBetween(from, to, this.blockRangeSize);
+          // query events only for the first interval to make sure block range is fine
+          const [intervalStart, intervalEnd] = intervals[0];
```
daywiss, line 89: theres an assumption here that you will only retry if the range is too large, and that you will know if the range is too large on the first query.  seems like you can simplify the code somewhat by testing the first query, adjusting to correct block range, and then requesting the rest of the ranges without having to add any retry logic for the bulk of the queries. i dont think this materially affects the performance/logic, but might make the code a bit cleaner

daywiss (approved):

src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
amateima: I wanted to make a change, but I've just remembered that the same error code could be thrown if the length of the response is greater than a particular value. So actually all block ranges have to monitored for this particular error

Tulun (approved):

daywiss (approved):

PR #17, opened 2022-04-13
fix: prevent events from being triggered after fetching was stopped 

amateima:
This PR stops events from being triggered if the fetching was stopped. This could happen when the last execution of the `setInterval()` callback was happening after the call of `stopFetchingTransfers()`

Also, the client instantiation requires the providers to be passed, instead of the provider URLs. This way we'll have less providers in our app

### DIFF
--- a/src/transfers-history/client.ts
+++ b/src/transfers-history/client.ts
@@ -38,7 +38,7 @@ export class TransfersHistoryClient {
   private eventsServices: Record<string, Record<ChainId, SpokePoolEventsQueryService>> = {};
   private pollingIntervalSeconds: number = 15;
   private pollingTimers: Record<string, NodeJS.Timer> = {};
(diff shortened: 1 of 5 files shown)


src/transfers-history/client.ts
```diff
@@ -38,6 +38,7 @@ export class TransfersHistoryClient {
   private eventsServices: Record<string, Record<ChainId, SpokePoolEventsQueryService>> = {};
   private pollingIntervalSeconds: number = 15;
   private pollingTimers: Record<string, NodeJS.Timer> = {};
+  private fetchingState: Record<string, "started" | "stoped"> = {};
```
Tulun, line 42: stopped, not stoped :)

daywiss (approved):

daywiss (approved):

Tulun (approved):
LGTM

PR #19, opened 2022-04-13
feat(relayFeeCalculator): basic relay fee calculator

daywiss:

This is a low level fee calculator that works with mainnet. Different implementations of `QueryInterface` may be needed for other chains. This is particulary important for ability to get gas estimations, gas costs, and conversions to eth for tokens on each chain.


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Override this to add different behaviors for different chains.
+export class DefaultQueries implements QueryInterface {
+  private coingecko: uma.Coingecko;
+  constructor(private provider: Provider, private priceSymbol: string = "eth", private defaultGas = "305572") {
+    this.coingecko = new Coingecko();
+  }
+  async getGas(): Promise<BigNumberish> {
+    return this.defaultGas;
```
daywiss, line 24: this used to be hard coded in v1 based on weth, eth or uma token

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -0,0 +1,97 @@
...
+import { BigNumberish, calculateGasFees, toBNWei } from "../utils";
+const Coingecko = uma.Coingecko;
+const { percent, fixedPointAdjustment } = uma.across.utils;
+const { connect: erc20Connect } = uma.clients.erc20;
+
+export interface QueryInterface {
+  getGas: (tokenAddress?: string) => Promise<BigNumberish>;
+  getGasPrice: () => Promise<BigNumberish>;
+  getTokenPrice: (tokenAddress: string) => Promise<number | string>;
+  getTokenDecimals: (tokenAddress: string) => Promise<number>;
```
daywiss, line 14: these are the specific calls needed to calculate the fees, but might be different per chain

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -0,0 +1,97 @@
...
+      amountToRelay: amountToRelay.toString(),
+      relayFeePercent,
+      relayFeeTotal,
+      discountPercent: this.discountPercent,
+      feeLimitPercent: this.feeLimitPercent,
+      tokenAddress,
+      isAmountTooLow,
+    };
+  }
+}
```
daywiss, line 97: i imagine that we would eventually have a factory where we specify the chain to switch between different implementations of the query interface

mrice32 (commented):
This is awesome! One high level comment about the interface, otherwise, looks great!

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -0,0 +1,97 @@
...
+  getTokenDecimals: (tokenAddress: string) => Promise<number>;
+}
+
+// Override this to add different behaviors for different chains.
+export class DefaultQueries implements QueryInterface {
+  private coingecko: uma.Coingecko;
+  constructor(private provider: Provider, private priceSymbol: string = "eth", private defaultGas = "305572") {
+    this.coingecko = new Coingecko();
+  }
+  async getGas(): Promise<BigNumberish> {
```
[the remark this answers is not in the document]
daywiss: let me try to work this in and see how it changes things, but yea im for that if it makes this more flexible. 

Theres some assumptions made, that the fees require a gas amount, gas price in native token, and then a conversion price from gas token to relayed token.  it sounds like some chains may break those assumptions.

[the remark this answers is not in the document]
daywiss: i think the matic polygon case is fine with this interface, the optimism case sounds like it would be difficult. need to know more about how this works. 

for the matic case, we would need to override `getTokenPrice` and pass it in the token address based on the matic chain, and it would need to convert the token to its price in matic. 

for the optimsim case, it sounds like you need to get 2 different gas estimations and 2 different token prices to be able to convert to those tokens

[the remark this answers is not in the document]
daywiss: i've updated the interface to combine the gas calculations.

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+  async relayerFeePercent(amountToRelay: BigNumberish, tokenAddress?: string): Promise<BigNumberish> {
+    const gasCosts = await this.queries.getGasCosts(tokenAddress);
+    // if not provided, its assumed the token is the same denomination as the destination gas token
+    if (!tokenAddress) {
+      return percent(gasCosts, amountToRelay).toString();
+    }
+    const tokenPrice = await this.queries.getTokenPrice(tokenAddress);
+    const decimals = await this.queries.getTokenDecimals(tokenAddress);
+    const gasFeesInToken = nativeToToken(gasCosts, tokenPrice, decimals);
```
mrice32, line 95: How would this work for matic? Who is telling nativeToToken not to convert to ETH, but to convert to Matic?

mrice32 (commented):
Looks great! One question about the matic case!

src/relayFeeCalculator/relayFeeCalculator.ts
daywiss: each chain will have to implement their call to `getTokenPrice`, this price should convert whatever the native gas token is on the receiving chain to the token they are transferring. In most cases this will be eth/token, but in matics case you can change it to matic/token

amateima (approved):

src/utils.ts
```diff
@@ -64,14 +64,14 @@ export function max(a: BigNumberish, b: BigNumberish): BN {
...
  *
- * @param {BigNumberish} fromAmount - Amount of eth to convert.
- * @param {string | number} [ price=1 ] - The price as eth per token, ie how much eth can 1 token buy.
+ * @param {BigNumberish} fromAmount - Amount of native gas token to convert.
+ * @param {string | number} [ price=1 ] - The price as native gas token per token, ie how much native gas token can 1 token buy.
  * @param {} [ toDecimals=18 ] - Number of decimals for the token currency.
  * @returns {string} The number of tokens denominated in token decimals in the smallest unit (wei).
  */
-export function ethToToken(fromAmount: BigNumberish, price: string | number = 1, toDecimals = 18): string {
+export function nativeToToken(fromAmount: BigNumberish, price: string | number = 1, toDecimals = 18): string {
```
mrice32, line 74: Does this function assume that the gas token has 18 decimals?

src/relayFeeCalculator/relayFeeCalculator.ts
mrice32: Okay, I think I understand where you're coming from. Makes sense to me now.

mrice32 (approved):

src/utils.ts
daywiss: it does, yea. will need to change this to not assume native gas decimals

PR #21, opened 2022-04-21
fix: fix processing of fill events

amateima:

This PR fixes a small issue I found. In some cases the RelayFilled events were inserted into the database before the initial deposit event, leading to data inconsistencies. Now, the filling events are inserted into the store only after all FundsDeposited events from all chains were fetched.

### DIFF
--- a/src/transfers-history/client.e2e.ts
+++ b/src/transfers-history/client.e2e.ts
@@ -29,14 +29,22 @@ describe("Client e2e tests", () => {
           spokePoolContractAddr: "0x73549B5639B04090033c1E77a22eE9Aa44C2eBa0",
           lowerBoundBlockNumber: 30475937,
         },
+        {
+          chainId: CHAIN_IDs.RINKEBY,
+          provider: new providers.JsonRpcProvider(process.env[`WEB3_NODE_URL_${CHAIN_IDs.RINKEBY}`] || ""),
+          spokePoolContractAddr: "0xB078bBb35f8E24c2431b9d2a88C0bC0c26CC1F92",
+          lowerBoundBlockNumber: 10485193,
+        },
       ],
     });
     client.setLogLevel("debug");
-    await client.startFetchingTransfers("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D");
+    await client.startFetchingTransfers("0x718648C8c531F91b528A7757dD2bE813c3940608");
     client.on(TransfersHistoryEvent.TransfersUpdated, () => {
-      const pendingTransfers = client.getPendingTransfers("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D", 30, 0);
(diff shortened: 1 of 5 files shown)


src/transfers-history/client.ts
```diff
@@ -115,16 +117,31 @@ export class TransfersHistoryClient {
...
+      .reduce((acc, val) => [...acc, ...val.depositEvents], [] as FundsDepositedEvent[]);
+    const filledRelayEvents = events
+      .flat()
+      .reduce((acc, val) => [...acc, ...val.filledRelayEvents], [] as FilledRelayEvent[]);
+    const blockTimestampMap = events
+      .flat()
+      .reduce((acc, val) => ({ ...acc, ...val.blockTimestampMap }), {} as { [blockNumber: number]: number });
+
+    depositEvents.map(e => this.insertFundsDepositedEvent(e, blockTimestampMap[e.blockNumber]));
+    filledRelayEvents.map(e => this.insertFilledRelayEvent(e));
```
amateima, line 142: Now I insert the events data into the store only after all events are fetched from the chain

daywiss: not quite sure how the problem gets fixed but the changes look good. one small nit is about using map instead of forEach

src/transfers-history/client.e2e.ts
```diff
@@ -29,14 +29,22 @@ describe("Client e2e tests", () => {
...
-    await client.startFetchingTransfers("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D");
+    await client.startFetchingTransfers("0x718648C8c531F91b528A7757dD2bE813c3940608");
     client.on(TransfersHistoryEvent.TransfersUpdated, () => {
-      const pendingTransfers = client.getPendingTransfers("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D", 30, 0);
-      client.stopFetchingTransfers("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D");
+      const pendingTransfers = client.getPendingTransfers("0x718648C8c531F91b528A7757dD2bE813c3940608", 10, 0);
+      const filledTransfers = client.getFilledTransfers("0x718648C8c531F91b528A7757dD2bE813c3940608", 10, 0);
+      client.stopFetchingTransfers("0x718648C8c531F91b528A7757dD2bE813c3940608");
       expect(pendingTransfers.length).toBeGreaterThan(0);
+      expect(filledTransfers.length).toBeGreaterThan(0);
```
daywiss, line 47: does this test the condition that was having a problem?

daywiss (approved):
i dont quite see the issue or how its fixed, but the changes look fine to me. not sure if you do this, but you can use `yarn link` and test the code with the frontend, i usually will do that as a sanity check

mrice32 (approved):
LGTM!

PR #24, opened 2022-04-27
fix: handle fee calcs for lp transfers

daywiss:

## motivation
This is an issue where if a user with lp tokens sends them to another address, or receives them from another address, the fee calculation becomes completely broken.

## changes
This fix will look up transfer events on the erc20 contract, ignore mint/burn transactions, and then lookup exchange rates based on the block that those transfers happen. The total value of these transfers are then added/subtracted from your total deposit value (depending if u received or sent tokens).  To enable this feature, you must specify in the config that you have provided an archive provider, and you must have an archive node.

### DIFF
--- a/src/pool/poolClient.e2e.ts
+++ b/src/pool/poolClient.e2e.ts
@@ -14,7 +14,10 @@ const hubPoolAddress = ethers.utils.getAddress("0xD449Af45a032Df413b497A709EeD3E
 const rateModelStoreAddress = ethers.utils.getAddress("0x5923929DF7A2D6E038bb005B167c1E8a86cd13C8");
 const wethAddress = ethers.utils.getAddress("0xd0A1E359811322d97991E03f863a0C30C2cF029C");
 const daiAddress = ethers.utils.getAddress("0x4F96Fe3b7A6Cf9725f59d353F723c1bDb64CA6Aa");
-const users = [ethers.utils.getAddress("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D")];
+const users = [
+  ethers.utils.getAddress("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D"),
+  ethers.utils.getAddress("0x718648C8c531F91b528A7757dD2bE813c3940608"),
+];
 const l1Tokens = [daiAddress, wethAddress];
 const txReceiptHash = "0xb1cad90827baba0d4db5e510fabf12e1bb296f3ab16112d79b8b6af654949d0f";
 const startBlock = 30475928;
@@ -51,8 +54,16 @@ describe("Client", function() {
   let client: Client;
   beforeAll(async () => {
     provider = ethers.getDefaultProvider(process.env.CUSTOM_NODE_URL);
-    client = new Client({ hubPoolAddress, rateModelStoreAddress, wethAddress }, { provider }, (path, data) =>
-      set(state, path, data)
+    client = new Client(
+      {
+        hubPoolAddress,
+        rateModelStoreAddress,
+        wethAddress,
+        // if you have an archive node, set this to true
+        hasArchive: false,
+      },
+      { provider },
+      (path, data) => set(state, path, data)
     );
   });
   test("read users", async function() {
--- a/src/pool/poolClient.ts
+++ b/src/pool/poolClient.ts
@@ -16,6 +16,7 @@ const { loop, exists } = uma.utils;
 const {TransactionManager} = uma.across;
 const { parseAndReturnRateModelFromString } = uma.across.rateModel
 const { SECONDS_PER_YEAR, DEFAULT_BLOCK_DELTA} = uma.across.constants;
+const { AddressZero } = ethers.constants;
 
 export type { Provider };
 
@@ -28,6 +29,7 @@ export type Config = {
   rateModelStoreAddress: string;
   confirmations?: number;
   blockDelta?: number;
+  hasArchive?: boolean;
 };
 export type Dependencies = {
   provider: Provider;
@@ -102,9 +104,7 @@ class PoolState {
   ) {}
   public async read(l1Token:string, latestBlock: number, previousBlock?: number) {
     // typechain does not have complete types for call options, so we have to cast blockTag to any
-    const exchangeRatePrevious = await this.contract.callStatic.exchangeRateCurrent(l1Token,{
-      blockTag: previousBlock || latestBlock - 1,
-    } as any);
+    const exchangeRatePrevious = await this.exchangeRateAtBlock(l1Token, previousBlock || latestBlock - 1);
 
     const exchangeRateCurrent = await this.contract.callStatic.exchangeRateCurrent(l1Token);
 
@@ -119,7 +119,9 @@ class PoolState {
       exchangeRateCurrent,
       ...pooledToken,
     }
-
+  }
+  public async exchangeRateAtBlock(l1Token:string, blockTag:number){
+    return this.contract.callStatic.exchangeRateCurrent(l1Token,{ blockTag } as any);
   }
 }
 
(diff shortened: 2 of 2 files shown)


src/pool/poolClient.e2e.ts
```diff
@@ -51,8 +54,15 @@ describe("Client", function() {
...
   beforeAll(async () => {
     provider = ethers.getDefaultProvider(process.env.CUSTOM_NODE_URL);
-    client = new Client({ hubPoolAddress, rateModelStoreAddress, wethAddress }, { provider }, (path, data) =>
-      set(state, path, data)
+    client = new Client(
+      {
+        hubPoolAddress,
+        rateModelStoreAddress,
+        wethAddress,
+        hasArchive: false,
```
daywiss, line 62: optional parameter to enable the fix with an archive node

src/pool/poolClient.e2e.ts
```diff
@@ -14,7 +14,10 @@ const hubPoolAddress = ethers.utils.getAddress("0xD449Af45a032Df413b497A709EeD3E
 const rateModelStoreAddress = ethers.utils.getAddress("0x5923929DF7A2D6E038bb005B167c1E8a86cd13C8");
 const wethAddress = ethers.utils.getAddress("0xd0A1E359811322d97991E03f863a0C30C2cF029C");
 const daiAddress = ethers.utils.getAddress("0x4F96Fe3b7A6Cf9725f59d353F723c1bDb64CA6Aa");
-const users = [ethers.utils.getAddress("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D")];
+const users = [
+  ethers.utils.getAddress("0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D"),
+  ethers.utils.getAddress("0x718648C8c531F91b528A7757dD2bE813c3940608"),
```
daywiss, line 19: manually transferred lp tokens on kovan between these addresses

src/pool/poolClient.ts
```diff
@@ -217,9 +219,56 @@ export class PoolEventState {
 }
 
 class UserState {
-  constructor(private contract: uma.clients.erc20.Instance) {}
-  public async read(user: string) {
+  private seen = new Set<string>();
```
daywiss, line 222: this class was updated based on copying code from the poolstate class. It now needs to handle events in a similar way as the pool state to find transfer events on the erc20

src/pool/poolClient.ts
```diff
@@ -362,6 +414,46 @@ export class Client {
   private getOrCreatePoolEvents() {
     return this.poolEvents
   }
+  private getOrCreateUserService(userAddress:string,tokenAddress:string){
```
daywiss, line 417: user service used to be stateless, but now queries and caches events from the token contract

src/pool/poolClient.ts
```diff
@@ -362,6 +414,46 @@ export class Client {
...
+    set(this.userServices,[tokenAddress,userAddress],userService)
+    return userService
+  }
+  private updateExchangeRateTable(l1TokenAddress:string, exchangeRateTable:Record<number,BigNumberish>):Record<number,BigNumberish>{
+    if(!this.exchangeRateTable[l1TokenAddress]) this.exchangeRateTable[l1TokenAddress] = {}
+    this.exchangeRateTable[l1TokenAddress] = {...this.exchangeRateTable[l1TokenAddress], ...exchangeRateTable}
+    return this.exchangeRateTable[l1TokenAddress]
+  }
+  // calculates the value of each LP token transfer at the block it was sent. this only works if we have archive node
+  async calculateLpTransferValue(l1TokenAddress:string, userState: Awaited<ReturnType<UserState["read"]>>){
```
daywiss, line 431: this is the main part which allows us to calcluate fees. it looks up exchange rates at block times where the user sent or received lp tokens, then can calculate a value for those transfers

src/pool/poolClient.ts
```diff
@@ -362,6 +414,46 @@ export class Client {
...
+    assert(this.config.hasArchive,'Can only calculate historical lp values with archive node')
+    const contract = this.getOrCreatePoolContract();
+    const pool = new PoolState(contract,this.config.hubPoolAddress);
+    const blockNumbers = userState.transferEvents
+      .map(x=>x.blockNumber)
+      .filter(blockNumber=>!this.exchangeRateTable?.[l1TokenAddress]?.[blockNumber]);
+
+    // new exchange rate lookups
+    const exchangeRateTable = this.updateExchangeRateTable(l1TokenAddress,Object.fromEntries(
```
daywiss, line 440: lookup and cache [token,blocknumber]-> exchange rate

src/pool/poolClient.ts
```diff
@@ -362,6 +414,46 @@ export class Client {
...
+      await Promise.all(blockNumbers.map(async blockNumber=>{
+        return [blockNumber,await pool.exchangeRateAtBlock(l1TokenAddress,blockNumber)]
+      }))
+    ));
+
+    return userState.transferEvents.reduce((result,transfer)=>{
```
daywiss, line 446: calculate the actual value sent/received

src/pool/poolClient.ts
```diff
@@ -119,7 +119,9 @@ class PoolState {
       exchangeRateCurrent,
       ...pooledToken,
     }
-
+  }
+  public async exchangeRateAtBlock(l1Token:string, blockTag:number){
+    return this.contract.callStatic.exchangeRateCurrent(l1Token,{ blockTag } as any);
```
mrice32, line 124: nit: why did you have to cast this as any? What type is it expecting? Is blockTag supposed to be a string rather than a number (even if it is numerical)?

src/pool/poolClient.ts
```diff
@@ -217,9 +219,56 @@ export class PoolEventState {
...
+        if (a.logIndex !== b.logIndex) return a.logIndex - b.logIndex;
+        // if everything is the same, return a, ie maintain order of array
+        return -1;
+      });
+    // ethers queries are inclusive [start,end] unless start === end, then exclusive (start,end). we increment to make sure we dont see same event twice
+    this.startBlock = endBlock + 1;
+    return this.events
+  }
+  public async read(user: string, endBlock:number) {
+    const transferEvents = await this.readEvents(user,endBlock)
```
mrice32, line 266: nit: do we not have any autoformatting in this repo?

src/pool/poolClient.ts
```diff
@@ -217,9 +219,56 @@ export class PoolEventState {
...
+        if (a.transactionIndex !== b.transactionIndex) return a.transactionIndex - b.transactionIndex;
+        if (a.logIndex !== b.logIndex) return a.logIndex - b.logIndex;
+        // if everything is the same, return a, ie maintain order of array
+        return -1;
+      });
+    // ethers queries are inclusive [start,end] unless start === end, then exclusive (start,end). we increment to make sure we dont see same event twice
+    this.startBlock = endBlock + 1;
+    return this.events
+  }
+  public async read(user: string, endBlock:number) {
```
mrice32, line 265: What's the difference between read and readEvents? Just from the interface, it's somewhat unclear what these methods are doing and why they're different, since they have the same args, but a different name.

src/pool/poolClient.ts
```diff
@@ -217,9 +219,56 @@ export class PoolEventState {
...
+    .flat()
+
+    this.events = this.events
+      .concat(events)
+      .sort((a, b) => {
+        if (a.blockNumber !== b.blockNumber) return a.blockNumber - b.blockNumber;
+        if (a.transactionIndex !== b.transactionIndex) return a.transactionIndex - b.transactionIndex;
+        if (a.logIndex !== b.logIndex) return a.logIndex - b.logIndex;
+        // if everything is the same, return a, ie maintain order of array
+        return -1;
```
mrice32, line 259: Returning 0 is the preferred way to maintain the sorting of the array.

But on a more fundamental note: should this be an error? Wouldn't this mean we have duplicate events?

src/pool/poolClient.ts
```diff
@@ -217,9 +219,56 @@ export class PoolEventState {
...
+    if (!seen) this.addEvent(params);
+    return !seen;
+  };
+  public async readEvents(user:string, endBlock: number) {
+    if (endBlock <= this.startBlock) return [];
+    const events = (
+      await Promise.all([
+        ...(await this.contract.queryFilter(this.contract.filters.Transfer(user,undefined), this.startBlock, endBlock)),
+        ...(await this.contract.queryFilter(this.contract.filters.Transfer(undefined,user), this.startBlock, endBlock)),
+      ])
```
mrice32, line 245: We could end up with duplicates where a user transfers to themselves, right? Will that be a problem? Should we do a post-filter for events that are from and to the same address?

src/pool/poolClient.ts
```diff
@@ -362,6 +414,46 @@ export class Client {
...
+    return this.exchangeRateTable[l1TokenAddress]
+  }
+  // calculates the value of each LP token transfer at the block it was sent. this only works if we have archive node
+  async calculateLpTransferValue(l1TokenAddress:string, userState: Awaited<ReturnType<UserState["read"]>>){
+    assert(this.config.hasArchive,'Can only calculate historical lp values with archive node')
+    const contract = this.getOrCreatePoolContract();
+    const pool = new PoolState(contract,this.config.hubPoolAddress);
+    const blockNumbers = userState.transferEvents
+      .map(x=>x.blockNumber)
+      .filter(blockNumber=>!this.exchangeRateTable?.[l1TokenAddress]?.[blockNumber]);
```
mrice32, line 437: Shouldn't the exchange rate table and the object at the l1Token address always exist? I thought that's what the code above was meant to ensure. If that's the case, then why do we need all the optional chaining here?

src/pool/poolClient.ts
```diff
@@ -362,6 +414,46 @@ export class Client {
...
+    return userState.transferEvents.reduce((result,transfer)=>{
+      const exchangeRate = exchangeRateTable[transfer.blockNumber];
+      if(transfer.args.to === userState.address){
+        return result.add(transfer.args.value.mul(exchangeRate).div(fixedPointAdjustment));
+      }
+      if(transfer.args.from === userState.address){
+        return result.sub(transfer.args.value.mul(exchangeRate).div(fixedPointAdjustment));
+      }
+      return result
+    },BigNumber.from(0))
```
mrice32, line 455: This is very cool!

mrice32 (approved):
LGTM! A few minor comments

src/pool/poolClient.ts
daywiss: typechain does not not have the correct teyp annotations for this function, so typescript fails when trying to pass the blockTag in, it doesnt recognize it as a parameter. This was the case at least when originally written

src/pool/poolClient.ts
daywiss: i didnt know that about 0, i can change that. I'm ok with throwing an error here as long as typescript will pass

src/pool/poolClient.ts
daywiss: going to try to remove it and see what happens, looks like its working now

src/pool/poolClient.ts
daywiss: yea can do that!

PR #26, opened 2022-05-10
fix: add eslint and settings from protocol

daywiss:
override tsdx linitng with eslint and rules from protocol linter.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,15 +1,31 @@
 module.exports = {
-  "extends": [
-    "react-app",
-    "prettier/@typescript-eslint",
+  env: {
+    browser: true,
+    es2021: true,
+    node: true
+  },
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/eslint-recommended",
+    "plugin:@typescript-eslint/recommended",
     "plugin:prettier/recommended",
+  ],
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    ecmaVersion: 'latest',
+    sourceType: 'module'
+  },
+  plugins: [
+    "prettier", 
+    '@typescript-eslint'
   ],
   rules: {
     "prettier/prettier": ["warn"],
-  },
-  "settings": {
-    "react": {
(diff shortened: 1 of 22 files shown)


.eslintrc.js
```diff
@@ -1,15 +1,31 @@
 module.exports = {
-  "extends": [
-    "react-app",
-    "prettier/@typescript-eslint",
+  env: {
```
mrice32, line 2: Are these just copied from the protocol repo?

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npm test
```
mrice32, line 4: You want to run all tests before each commit?

package.json
```diff
@@ -18,15 +18,15 @@
     "test": "tsdx test",
     "test:watch": "tsdx test --watch",
     "test:e2e": "tsdx test --config jest-e2e-config.json",
-    "lint": "tsdx lint --fix",
-    "prepare": "tsdx build",
+    "lint": "eslint --fix src",
```
mrice32, line 21: Should we be running prettier as well? Or just eslint?

mrice32, line 21: Do you want to have ci check that the code was linted correctly? Like a lint-check command?

mrice32 (approved):
Looks great!

.eslintrc.js
daywiss: it mostly was, i didnt need to add overrides and removed one rule about flattening brackets which was causing warnings.

.husky/pre-commit
daywiss: yea im ok with that, what do u think?

package.json
daywiss: i think that would be useful, i could add that

amateima (approved):

package.json
daywiss: i added `lint-check`

daywiss: i added `lint-check`

PR #27, opened 2022-05-10
fix: improve tx history client

amateima:

### DIFF
--- a/src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
+++ b/src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
@@ -103,7 +103,10 @@ export class SpokePoolEventsQuerier implements ISpokePoolContractEventsQuerier {
           events.push(...newEvents);
         }
       } catch (error) {
-        if ((error as Web3Error).error.code === Web3ErrorCode.BLOCK_RANGE_TOO_LARGE) {
+        if (
+          (error as Web3Error).error.code === Web3ErrorCode.BLOCK_RANGE_TOO_LARGE ||
+          (error as Web3Error).error.code === Web3ErrorCode.EXCEEDED_MAXIMUM_BLOCK_RANGE
+        ) {
           // make sure the block range size wasn't modified by a parallel function call
           if (this.blockRangeSize === blockRangeSizeAtStart) {
             const newBlockRangeSize = this.blockRangeSize ? this.blockRangeSize / 2 : DEFAULT_BLOCK_RANGE;
@@ -116,6 +119,7 @@ export class SpokePoolEventsQuerier implements ISpokePoolContractEventsQuerier {
           retryWithLowerBatchSize = true;
         } else {
           retryWithLowerBatchSize = false;
+          console.error(error);
           throw error;
         }
       }
--- a/src/transfers-history/services/SpokePoolEventsQueryService.ts
+++ b/src/transfers-history/services/SpokePoolEventsQueryService.ts
@@ -39,17 +39,27 @@ export class SpokePoolEventsQueryService {
         blockTimestampMap: {},
       };
     }
-
-    [depositEvents, filledRelayEvents] = await Promise.all([
+    let start = new Date();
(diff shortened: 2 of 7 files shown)


src/transfers-history/adapters/web3/SpokePoolEventsQuerier.ts
```diff
@@ -116,6 +119,7 @@ export class SpokePoolEventsQuerier implements ISpokePoolContractEventsQuerier {
           retryWithLowerBatchSize = true;
         } else {
           retryWithLowerBatchSize = false;
+          console.error(error);
           throw error;
```
amateima, line 123: log the error here for visibility because this function runs inside Promise.allSettled() and it can error silently

src/transfers-history/services/SpokePoolEventsQueryService.ts
```diff
@@ -39,17 +39,27 @@ export class SpokePoolEventsQueryService {
...
+        filledRelayEvents.length
+      } FilledRelayEvents, blocks: ${from} -> ${to}, ${(end.valueOf() - start.valueOf()) / 1000} seconds`
     );
+    start = new Date();
     blockTimestampMap = await this.getBlocksTimestamp(depositEvents);
+    end = new Date();
+    this.logger.debug(
+      "[SpokePoolEventsQueryService::getEvents]",
+      `🟢 chain ${this.chainId}: fetched block numbers in ${(end.valueOf() - start.valueOf()) / 1000} seconds`
+    );
```
amateima, line 62: Edited the logs for monitoring the performance of fetching the events from each chain

amateima:
One thing I noticed while working on this is that I need to add support for monitoring multiple SpokePool contracts for each chain, right? @daywiss @mrice32 They can change with time, but the TX History page must show all deposits

src/transfers-history/services/SpokePoolEventsQueryService.ts
```diff
@@ -39,17 +39,27 @@ export class SpokePoolEventsQueryService {
...
-
-    [depositEvents, filledRelayEvents] = await Promise.all([
+    let start = new Date();
+    const [depositEventsSettled, filledRelayEventsSettled] = await Promise.allSettled([
       this.eventsQuerier.getFundsDepositEvents(from, to, this.depositorAddr),
       this.eventsQuerier.getFilledRelayEvents(from, to, this.depositorAddr),
     ]);
-
+    depositEvents = depositEventsSettled.status === "fulfilled" ? depositEventsSettled.value : [];
+    filledRelayEvents = filledRelayEventsSettled.status === "fulfilled" ? filledRelayEventsSettled.value : [];
```
mrice32, line 48: Was this the bug? If so, what did the allSettled fix?

mrice32 (commented):
Looks great! I just don't quite see what the problem was/what the fix was.

mrice32:
> One thing I noticed while working on this is that I need to add support for monitoring multiple SpokePool contracts for each chain, right? They can change with time, but the TX History page must show all deposits 🤔 @daywiss @mrice32

Technically there could be, yes. It's not absolutely necessary we handle this now, though, because we have no plans (and may never) upgrade the spoke pools. This would only really be necessary if a particular chain's canonical bridge API were to break or change.

daywiss (approved):
not really sure what fixed the issue, but just make sure this looked ok when testing locally

src/transfers-history/services/SpokePoolEventsQueryService.ts
amateima: The issues were in the [repository file](https://github.com/across-protocol/sdk-v2/pull/27/files/39e2ebd8048b109503205ceca909a577ec56d9be#diff-f833d41d83e455eb24eb6451414108f4d4786a3e59d699f33585de218415aade)

`allSettled` is just to make sure that throwing errors does not prevent getting events which might be fetched successfully

amateima:
> not really sure what fixed the issue, but just make sure this looked ok when testing locally

looked good when tested locally within the FE app

PR #30, opened 2022-05-11
fix: use new across config store instead of rate model store

daywiss:

This basically replaces references to rate model store with the config store.


amateima (approved):
👍

src/contracts/acrossConfigStore.ts
```diff
@@ -0,0 +1,37 @@
...
+import { object, string, Infer, assert } from "superstruct";
+import type { CallOverrides } from "@ethersproject/contracts";
+
+const RateModel = object({
+  UBar: string(), // denote the utilization kink along the rate model where the slope of the interest rate model changes.
+  R0: string(), // is the interest rate charged at 0 utilization
+  R1: string(), // R_0+R_1 is the interest rate charged at UBar
+  R2: string(), // R_0+R_1+R_2 is the interest rate charged at 100% utilization
+});
+const L1TokenConfig = object({
```
mrice32, line 12: nit: I don't love aliasing variables and types. When reading I kept thinking you were passing the RateModel and L1TokenConfig _types_ into functions. feel free to ignore if this is the way the library suggests you name things, just felt a little odd to me

mrice32 (approved):
LGTM!

src/contracts/acrossConfigStore.ts
daywiss: ok sure i can adjust that, not sure whats recommended here

PR #31, opened 2022-05-11
feat: add query objects for 4 chains

mrice32:


src/relayFeeCalculator/chain-queries/ethereum.ts
```diff
@@ -0,0 +1,70 @@
+import { QueryInterface } from "../relayFeeCalculator";
+import { BigNumberish } from "../../utils";
+import { Coingecko } from "../../coingecko/Coingecko";
+import ethers from "ethers";
+import axios from "axios";
+
+const { parseUnits } = ethers.utils;
+
+// Note: these are the mainnet addresses for these symbols meant to be used for pricing.
+export const SymbolMapping: { [symbol: string]: { address: string; decimals: number } } = {
```
daywiss, line 10: we have a similar mapping in the frontend constants. i wonder if maybe we can override this constant by optionally passing the structure in from the FE, this way we only have to change the constant in one place

src/relayFeeCalculator/chain-queries/arbitrum.ts
```diff
@@ -0,0 +1,47 @@
...
+import { SymbolMapping } from "./ethereum";
+import { Coingecko } from "../../coingecko/Coingecko";
+import { ArbitrumSpokePool__factory, ArbitrumSpokePool } from "@across-protocol/contracts-v2";
+
+class ArbitrumQueryInterface implements QueryInterface {
+  private spokePool: ArbitrumSpokePool;
+
+  constructor(readonly provider: ethers.providers.Provider) {
+    // TODO: replace with address getter.
+    this.spokePool = ArbitrumSpokePool__factory.connect("0xe1C367e2b576Ac421a9f46C9cC624935730c36aa", provider);
```
daywiss, line 13: address shoudl be an argument so we can pass it in from the frontend. this will change depending on if we are running testnets

src/relayFeeCalculator/chain-queries/ethereum.ts
```diff
@@ -0,0 +1,70 @@
...
+    address: "0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0",
+    decimals: 18,
+  },
+  WMATIC: {
+    address: "0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0",
+    decimals: 18,
+  },
+};
+
+export const evmRelayAverageGas = 50000;
```
daywiss, line 53: would suggest making this a constructor param (with default value) so we can set envs to tweak on FE if needed

src/relayFeeCalculator/chain-queries/ethereum.ts
```diff
@@ -0,0 +1,70 @@
...
+    return parseUnits(FastGasPrice, 9).mul(evmRelayAverageGas).toString();
+  }
+
+  async getTokenPrice(tokenSymbol: string): Promise<string | number> {
+    const [, price] = await Coingecko.get().getCurrentPriceByContract(SymbolMapping[tokenSymbol].address, "eth");
+    return price;
+  }
+
+  async getTokenDecimals(tokenSymbol: string): Promise<number> {
+    return SymbolMapping[tokenSymbol].decimals;
```
daywiss, line 68: surprised typescript does not force you to prove the value exists here. should probably throw an error if symbol is not supported

daywiss (commented):
this looks great, just a few comments to make this easy to consume on the FE

src/relayFeeCalculator/chain-queries/arbitrum.ts
mrice32: Done.

src/relayFeeCalculator/chain-queries/ethereum.ts
mrice32: Done.

src/relayFeeCalculator/chain-queries/ethereum.ts
mrice32: Done.

src/relayFeeCalculator/chain-queries/polygon.ts
```diff
@@ -0,0 +1,25 @@
+import { QueryInterface } from "../relayFeeCalculator";
+import { BigNumberish } from "../../utils";
+import ethers from "ethers";
+import { evmRelayAverageGas, SymbolMapping } from "./ethereum";
+import { Coingecko } from "../../coingecko/Coingecko";
+
+const { parseUnits } = ethers.utils;
+
+class PolygonQueryInterface implements QueryInterface {
```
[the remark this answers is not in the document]
mrice32: Done.

src/coingecko/Coingecko.e2e.ts
```diff
@@ -16,6 +16,7 @@ describe("coingecko", function () {
   test("getCurrentPriceByContract", async function () {
     const address = "0x04fa0d235c4abf4bcf4787af4cf447de572ef828";
     const result = await cg.getCurrentPriceByContract(address);
+    console.log([typeof result[0], typeof result[1]]);
```
daywiss, line 19: theres a log here

amateima, line 19: do we want to keep this?

daywiss (approved):
i think theres a bit more work to be done for making sure the  simulated tx works on testnet, but we could tackle this later, i dont really have any ideas other than making more parameters.

amateima (approved):

PR #36, opened 2022-05-16
improve: narrow event queries to specific user

daywiss:

There was an issue in across v1, once there were so many events on a pool, queries would fail. This is solved by filtering only events for the current user, and also by setting a startblock for the hubpool, so that queries do not go back further than the start block.

### DIFF
--- a/src/pool/poolClient.ts
+++ b/src/pool/poolClient.ts
@@ -12,7 +12,7 @@ import { calculateInstantaneousRate } from "../lpFeeCalculator";
 import { hubPool, acrossConfigStore } from "../contracts";
 
 const { erc20 } = uma.clients;
-const { loop, exists } = uma.utils;
+const { loop } = uma.utils;
 const { TransactionManager } = uma.across;
 const { SECONDS_PER_YEAR, DEFAULT_BLOCK_DELTA } = uma.across.constants;
 const { AddressZero } = ethers.constants;
@@ -29,6 +29,7 @@ export type Config = {
   confirmations?: number;
   blockDelta?: number;
   hasArchive?: boolean;
+  hubPoolStartBlock?: number;
 };
 export type Dependencies = {
   provider: Provider;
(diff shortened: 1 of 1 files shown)


src/pool/poolClient.ts
```diff
@@ -125,49 +126,50 @@ type EventIdParams = { blockNumber: number; transactionIndex: number; logIndex:
 export class PoolEventState {
   private seen = new Set<string>();
   private iface: ethers.utils.Interface;
-  constructor(
-    private contract: hubPool.Instance,
-    private startBlock = 0,
-    private state: hubPool.EventState = hubPool.eventStateDefaults()
-  ) {
+  // maintain ordered unique list of events so we can calculate state
+  private events: uma.SerializableEvent[] = [];
```
daywiss, line 130: The logic in this class has changed a lot. It previously assumed all events were queried, such that we could optimize by continuously updating the last block seen, and only storing the final state from the events.  Now that each user is queried separately, we basically query from the earliest block every time, then just storing the events, and rebuilding state from the full list of events every time.  We have multiple sources of events, from provider queries, and transaction receipts, so these are all merged into the same list.

src/pool/poolClient.ts
```diff
@@ -125,49 +126,50 @@ type EventIdParams = { blockNumber: number; transactionIndex: number; logIndex:
...
-    // ethers queries are inclusive [start,end] unless start === end, then exclusive (start,end). we increment to make sure we dont see same event twice
-    this.startBlock = endBlock + 1;
-    this.state = hubPool.getEventState(events, this.state);
-    return this.state;
-  }
-  makeEventFromLog(log: Log) {
+  private processEvent = (event: uma.SerializableEvent | undefined): void => {
+    if (event == undefined) return;
+    if (!this.filterSeen(event)) return;
+    this.events = uma.oracle.utils.insertOrderedAscending(this.events, event, this.makeId);
```
daywiss, line 151: oracle utils has some event managing logic

src/pool/poolClient.ts
```diff
@@ -177,57 +179,49 @@ export class PoolEventState {
...
-      .filter(exists);
-
-    const eventState = hubPool.getEventState(events);
+    const events = receipt.logs.map((log) => {
+      try {
+        return this.makeEventFromLog(log);
+      } catch (err) {
+        // return nothing, this throws a lot because logs from other contracts are included in receipt
+        return undefined;
+      }
```
mrice32, line 188: Instead of ignoring errors, could we use the address associated with each log to filter events that were from our contract and only try to parse those? If we only parse those, this would only error if our abi was wrong (which is something we'd want to know!).

Something like:

```ts
const events = receipt.logs
    .filter((log) => 
        ethers.utils.getAddress(log.address) === ethers.utils.getAddress(this.contract.address)
    )
    .map((log) => {
...
```

src/pool/poolClient.ts
```diff
@@ -177,57 +179,49 @@ export class PoolEventState {
...
-    return this.state;
+  readTxReceipt(receipt: TransactionReceipt): hubPool.EventState {
+    const events = receipt.logs.map((log) => {
+      try {
+        return this.makeEventFromLog(log);
+      } catch (err) {
+        // return nothing, this throws a lot because logs from other contracts are included in receipt
+        return undefined;
+      }
+    });
```
mrice32, line 205: Same comment here.

mrice32 (approved):
Looks good! Just a few comments about swallowing errors.

src/pool/poolClient.ts
daywiss: good idea

src/pool/poolClient.ts
```diff
@@ -125,49 +126,50 @@ type EventIdParams = { blockNumber: number; transactionIndex: number; logIndex:
...
-        // if everything is the same, return a, ie maintain order of array
-        return -1;
-      });
-    // ethers queries are inclusive [start,end] unless start === end, then exclusive (start,end). we increment to make sure we dont see same event twice
-    this.startBlock = endBlock + 1;
-    this.state = hubPool.getEventState(events, this.state);
-    return this.state;
-  }
-  makeEventFromLog(log: Log) {
+  private processEvent = (event: uma.SerializableEvent | undefined): void => {
```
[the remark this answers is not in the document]
daywiss: when dealing with transaction receipts, decoding the logs sometimes results in logs we cant decode, so mapping over them will return undefined. I just wanted to be able to pass the results of the map directly to this function, so sometimes the events could be undefined

src/pool/poolClient.ts
daywiss: i've added this in both spots

PR #37, opened 2022-05-16
feat(tx-history): add fill txs to transfers

amateima:

Quick change in order to get the hashes of the fill transactions for displaying them in Across frontend


daywiss (approved):
looks good, glad it was an easy change

mrice32 (approved):
LGTM!

PR #46, opened 2022-05-25
feat: get all txs history

amateima:

This PR improves the TX History client to be able to fetch all Across V2 deposits by calling `client.startFetchingTransfers("all")` instead of a particular ETH address

### DIFF
--- a/src/transfers-history/client.ts
+++ b/src/transfers-history/client.ts
@@ -20,6 +20,8 @@ export type TransfersUpdatedEventListenerParams = {
   pendingTransfersCount: number;
 };
 
+export type TrackableAddress = "all" | string;
+
 export type TransfersUpdatedEventListener = (params: TransfersUpdatedEventListenerParams) => void;
 export type TransfersHistoryClientEventListener = TransfersUpdatedEventListener;
 export type TransfersHistoryClientParams = {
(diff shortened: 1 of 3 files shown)


src/transfers-history/client.ts
```diff
@@ -185,11 +196,17 @@ export class TransfersHistoryClient {
     );
   }
 
-  public getFilledTransfers(depositorAddr: string, limit?: number, offset?: number) {
+  public getFilledTransfers(depositorAddr: TrackableAddress, limit?: number, offset?: number) {
+    if (depositorAddr === "all") {
+      return this.transfersRepository.getAllFilledTransfers();
```
daywiss, line 201: why are limit and offset not passed in

src/transfers-history/client.ts
```diff
@@ -185,11 +196,17 @@ export class TransfersHistoryClient {
     );
   }
 
-  public getFilledTransfers(depositorAddr: string, limit?: number, offset?: number) {
+  public getFilledTransfers(depositorAddr: TrackableAddress, limit?: number, offset?: number) {
+    if (depositorAddr === "all") {
```
daywiss, line 200: this might be cleaner as a new function because you will just have to end up switching something on the frontend, might as well just be a different function rather than address or "all"

daywiss (approved):
might want to update the docs if you have some api docs, but looks good.

src/transfers-history/client.ts
amateima: I started with this thought in mind, but this implied creating separate functions for other things too, for example: stop fetching transfers. For me, passing "all" seemed a cleaner implementation.

PR #53, opened 2022-06-02
fix(tx-history): optimise fetching blocks & release

amateima:

Fetching blocks info should be done in batches

### DIFF
--- a/src/transfers-history/adapters/db/transfers-repository.ts
+++ b/src/transfers-history/adapters/db/transfers-repository.ts
@@ -64,7 +64,7 @@ export class TransfersRepository {
       ...transfer,
       filled,
       status: transfer.amount.eq(filled) ? "filled" : "pending",
-      fillTxs: [...transfer.fillTxs, fillTxHash],
+      fillTxs: Array.from(new Set([...transfer.fillTxs, fillTxHash])),
     };
   }
 
--- a/src/transfers-history/services/SpokePoolEventsQueryService.ts
+++ b/src/transfers-history/services/SpokePoolEventsQueryService.ts
@@ -78,7 +78,16 @@ export class SpokePoolEventsQueryService {
       return { ...acc, [event.blockNumber]: true };
     }, {} as Record<number, any>);
     const uniqueBlockNumbersList = Object.keys(uniqueBlockNumbers).map((blockNumber) => parseInt(blockNumber));
-    const blocks = await Promise.all(uniqueBlockNumbersList.map((blockNumber) => this.provider.getBlock(blockNumber)));
+    this.logger.debug(
+      "[getBlocksTimestamp]",
+      `🟢 chain ${this.chainId}: fetching ${uniqueBlockNumbersList.length} blocks`
+    );
+    const blocksChunks = this.getArrayChunks(uniqueBlockNumbersList);
+    const blocks = [];
+    for (const blocksChunk of blocksChunks) {
+      const newBlocks = await Promise.all(blocksChunk.map((blockNumber) => this.provider.getBlock(blockNumber)));
+      blocks.push(...newBlocks);
+    }
     const timestamps = await Promise.all(blocks.map((block) => block.timestamp));
     const blockTimestampMap = uniqueBlockNumbersList.reduce(
       (acc, blockNumber, idx) => ({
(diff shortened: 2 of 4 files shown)


src/transfers-history/adapters/db/transfers-repository.ts
```diff
@@ -64,7 +64,7 @@ export class TransfersRepository {
       ...transfer,
       filled,
       status: transfer.amount.eq(filled) ? "filled" : "pending",
-      fillTxs: [...transfer.fillTxs, fillTxHash],
+      fillTxs: Array.from(new Set([...transfer.fillTxs, fillTxHash])),
     };
```
amateima, line 68: In some cases (request All transactions followed by requesting only My transactions) `fillTxs` were considered multiple times. Keep only the distinct values

src/transfers-history/services/SpokePoolEventsQueryService.ts
```diff
@@ -78,7 +78,16 @@ export class SpokePoolEventsQueryService {
...
+    this.logger.debug(
+      "[getBlocksTimestamp]",
+      `🟢 chain ${this.chainId}: fetching ${uniqueBlockNumbersList.length} blocks`
+    );
+    const blocksChunks = this.getArrayChunks(uniqueBlockNumbersList);
+    const blocks = [];
+    for (const blocksChunk of blocksChunks) {
+      const newBlocks = await Promise.all(blocksChunk.map((blockNumber) => this.provider.getBlock(blockNumber)));
+      blocks.push(...newBlocks);
+    }
```
amateima, line 90: fetch blocks info in batches

amateima: I want to build a small API service which serves block metadata. Something like `https://BLOCKS_API_SERVICE/blocks?number=BLOCK_NUMBER&chainId=CHAIN_ID`. It'll store the response in a cache and all further requests will use the cache instead of hitting the web3 providers again

daywiss: that would be nice, how do you plan on doing this

daywiss (approved):

src/transfers-history/services/SpokePoolEventsQueryService.ts
amateima: A simple ExpressJS app (eventually managed from Zion) would do it. It'll just receive the providers urls by ENV vars and store in memory the requested blocks

mrice32: Relatedly, we just created a Redis db for timestamp -> blocknumber (and a few other expensive bot queries) lookups on mainnet. It works pretty well from cloud run. We could use a similar strategy here.

@daywiss this might be something that dev support could devise a strategy for that would work for many applications.

src/transfers-history/services/SpokePoolEventsQueryService.ts
```diff
@@ -78,7 +78,16 @@ export class SpokePoolEventsQueryService {
...
     const uniqueBlockNumbersList = Object.keys(uniqueBlockNumbers).map((blockNumber) => parseInt(blockNumber));
-    const blocks = await Promise.all(uniqueBlockNumbersList.map((blockNumber) => this.provider.getBlock(blockNumber)));
+    this.logger.debug(
+      "[getBlocksTimestamp]",
+      `🟢 chain ${this.chainId}: fetching ${uniqueBlockNumbersList.length} blocks`
+    );
+    const blocksChunks = this.getArrayChunks(uniqueBlockNumbersList);
+    const blocks = [];
+    for (const blocksChunk of blocksChunks) {
+      const newBlocks = await Promise.all(blocksChunk.map((blockNumber) => this.provider.getBlock(blockNumber)));
```
mrice32, line 88: Why not use bluebird promises? AFAIK, you can specify the max number of promises you want in flight. I think bluebird might be more efficient (no need to wait on _all_ the requests in the prev group to finish). It also might be a bit clearer since it looks just like a typical Promise.all.

Not strongly opinionated though. If there's some reason the current strategy is preferable, that's totally fine too.

mrice32 (approved):
LGTM!

src/transfers-history/services/SpokePoolEventsQueryService.ts
amateima: I can use bluebird ofc, but this time I actually preferred to wait on all the requests in the prev group to finish because I got fluctuating performances from Infura on some chains (especially on optimism). Some times a huge number of requests went through, but other times the performance was quite bad, so I preferred to not force it by having the max number of promises in flight all the time

PR #54, opened 2022-06-10
feat: Truncate realized LP fee %'s to match UMIP

nicholaspai:
Note: The [across-v2](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-157.md#finding-slow-relays) UMIP doesn't reflect this truncation language yet but this PR can be merged in before hand since its backwards compatible. The `relayer-v2` code will then just need to pass in a flag to turn on truncation. 

The relayer and dataworker might want to only truncate %'s after a certain quote time so that it can validate older realized LP fee %'s

### DIFF
--- a/src/lpFeeCalculator/lpFeeCalculator.ts
+++ b/src/lpFeeCalculator/lpFeeCalculator.ts
@@ -80,8 +80,17 @@ export function calculateApyFromUtilization(
 export function calculateRealizedLpFeePct(
   rateModel: RateModel,
   utilizationBeforeDeposit: BigNumberish,
-  utilizationAfterDeposit: BigNumberish
+  utilizationAfterDeposit: BigNumberish,
+  truncateDecimals = false
 ) {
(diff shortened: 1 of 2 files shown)


src/lpFeeCalculator/lpFeeCalculator.ts
```diff
@@ -80,8 +80,17 @@ export function calculateApyFromUtilization(
...
-  return convertApyToWeeklyFee(apy);
+
+  // ACROSS-V2 UMIP requires that the realized fee percent is floor rounded as decimal to 6 decimals.
+  return truncateDecimals ? truncate18DecimalBN(convertApyToWeeklyFee(apy), 6) : convertApyToWeeklyFee(apy);
+}
+
+export function truncate18DecimalBN(input: BN, digits: number) {
+  const digitsToDrop = 18 - digits;
+  const multiplier = toBN(10).pow(digitsToDrop);
+  return input.div(multiplier).mul(multiplier);
```
daywiss, line 95: interesting

daywiss (approved):
nice backward compatibility

PR #55, opened 2022-06-13
feat: Make relayer fee curve more dynamic

nicholaspai:
This PR is supposed to be backwards compatible and not break any existing clients of the SDK.

This new relayer fee curve is best first described graphically. The top triangle is the marginal fee %, and the bottom is the actual fee  charged. The fee should always be between the lower and upper bound. For any amounts < `cutoff`, the fee % should linearly increase from lower bound to (upper bound + lower bound) / 2. After the cutoff, the fee % should approach to upper bound.

relayer_fee_example

Python implementation that inspired this code:
```py
def relayer_fee(rm, y):
    # Charge at bps_lb on entire amount
    min_charge = rm.bps_lb * y

    # Charge an increasing amount up to
    y_triangle = min(rm.cutoff, y)
    triangle_slope = (rm.bps_ub - rm.bps_lb) / rm.cutoff
    triangle_height = triangle_slope * y_triangle
    triangle_charge = 0.5*triangle_height*y_triangle

    # Charge bps_ub on anything above the cutoff
    y_remainder = max(0, y - rm.cutoff)
    remainder_charge = y_remainder * (rm.bps_ub - rm.bps_lb)

    return min_charge + triangle_charge + remainder_charge


def relayer_fee_in_bps(rm, y):
    if y < 1e-10:
        return rm.bps_lb

    fee = relayer_fee(rm, y)
    fee_pct = fee / y

    return fee_pct * 100 * 100
```

scratch work
Image from iOS

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -1,7 +1,7 @@
 import assert from "assert";
 import * as uma from "@uma/sdk";
 import { BigNumber } from "ethers";
-import { BigNumberish, toBNWei, nativeToToken } from "../utils";
+import { BigNumberish, toBNWei, nativeToToken, toBN, min, max } from "../utils";
 const { percent, fixedPointAdjustment } = uma.across.utils;
 
 // This needs to be implemented for every chain and passed into RelayFeeCalculator
@@ -17,7 +17,16 @@ export interface RelayFeeCalculatorConfig {
   capitalDiscountPercent?: number;
   feeLimitPercent?: number;
   capitalCostsPercent?: number;
+  capitalCostsConfig?: string;
   queries: QueryInterface;
+}
+
+export const expectedCapitalCostsKeys = ["lowerBound", "upperBound", "cutoff", "decimals"];
+export interface CapitalCostConfig {
+  lowerBound: BigNumber;
+  upperBound: BigNumber;
+  cutoff: BigNumber;
+  decimals: number;
 }
 
 export class RelayFeeCalculator {
@@ -27,13 +36,15 @@ export class RelayFeeCalculator {
   private feeLimitPercent: Required<RelayFeeCalculatorConfig>["feeLimitPercent"];
   private nativeTokenDecimals: Required<RelayFeeCalculatorConfig>["nativeTokenDecimals"];
   private capitalCostsPercent: Required<RelayFeeCalculatorConfig>["capitalCostsPercent"];
+  private capitalCostsConfig: { [token: string]: CapitalCostConfig };
   constructor(config: RelayFeeCalculatorConfig) {
     this.queries = config.queries;
     this.gasDiscountPercent = config.gasDiscountPercent || 0;
-    this.capitalDiscountPercent = config.capitalCostsPercent || 0;
+    this.capitalDiscountPercent = config.capitalDiscountPercent || 0;
     this.feeLimitPercent = config.feeLimitPercent || 0;
     this.nativeTokenDecimals = config.nativeTokenDecimals || 18;
     this.capitalCostsPercent = config.capitalCostsPercent || 0;
+    this.capitalCostsConfig = config.capitalCostsConfig ? JSON.parse(config.capitalCostsConfig) : {};
     assert(
       this.gasDiscountPercent >= 0 && this.gasDiscountPercent <= 100,
       "gasDiscountPercent must be between 0 and 100 percent"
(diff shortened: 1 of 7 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -50,7 +60,52 @@ export class RelayFeeCalculator {
...
+    this.capitalCostsConfig = {};
+    if (config.capitalCostsConfig) {
+      const parsed = JSON.parse(config.capitalCostsConfig);
+      for (const token of Object.keys(parsed)) {
+        this.capitalCostsConfig[token] = this.parseAndReturnCapitalCostsFromString(parsed[token]);
+      }
+    }
   }
+
+  parseAndReturnCapitalCostsFromString(capitalCosts: any): CapitalCostConfig {
```
daywiss, line 72: If you think this is a useful function, id provide it as a `static` method on the class, so a user can call it to create a validated config file from string, and so that you can strongly type the configuration parameters to the class constructor. 

id also highly suggest using `superstruct`, which is already included in the sdk. You can see an example of this being [parsed with the config store](https://github.com/across-protocol/sdk-v2/blob/master/src/contracts/acrossConfigStore.ts#L6). Superstruct will let you derive the type as well as prove to typescript an unknown type conforms to it. 

The frontend also has precendent for using superstruct, so its not a great loss to move the parsing and validation logic to the frontend constants file before instantiating your class.

one last thing, you should probably use `unknown` rather than `any`

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -17,20 +17,30 @@ export interface RelayFeeCalculatorConfig {
   capitalDiscountPercent?: number;
   feeLimitPercent?: number;
   capitalCostsPercent?: number;
```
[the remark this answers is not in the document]
daywiss: you can make a breaking change, just properly bump the version number when you do. conventional commits also has a method for signalling this when you append `BREAKING CHANGE: what the change is` in the line below your regular commit message.  then you will need to make an issue for bumping and migrating the frontend

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -50,7 +60,52 @@ export class RelayFeeCalculator {
       this.capitalCostsPercent >= 0 && this.capitalCostsPercent <= 100,
       "capitalCostsPercent must be between 0 and 100 percent"
     );
+    this.capitalCostsConfig = {};
```
daywiss, line 63: i would highly suggest letting the caller of the class do the parsing of the json, and have types defined for these config parameters, rather than passing a json in as a string. that way they can use intellisense to discover what the config fields should be.  Its also a bit wierd from a UX perspective to pass stringified json into a class constructor.

daywiss (commented):
nice work, just a few suggestions

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -64,7 +119,39 @@ export class RelayFeeCalculator {
...
+      const triangleSlope = config.upperBound.sub(config.lowerBound).mul(fixedPointAdjustment).div(config.cutoff);
+      const triangleHeight = triangleSlope.mul(yTriangle).div(fixedPointAdjustment);
+      const triangleCharge = triangleHeight.mul(yTriangle).div(toBNWei(2));
+
+      // For any amounts above the cutoff, the marginal fee % will not increase but will be fixed at the upper bound
+      // value.
+      const yRemainder = max(toBN(0), y.sub(config.cutoff));
+      const remainderCharge = yRemainder.mul(config.upperBound.sub(config.lowerBound)).div(fixedPointAdjustment);
+
+      return minCharge.add(triangleCharge).add(remainderCharge).mul(fixedPointAdjustment).div(y);
```
mrice32, line 151: @cc7768 can you validate this math?

mrice32 (commented):
Looks great -- agree with @daywiss's comments on config parsing and it would be nice to have @cc7768 review the math.

src/relayFeeCalculator/relayFeeCalculator.ts
cc7768: I didn't carefully check things that were big number specific (i.e. decimal places and fixed point adjustments), but the  mathematical operations are all exactly what I would expect.

Let me know if you want me to give the other things a closer look

cc7768 (approved):

src/relayFeeCalculator/relayFeeCalculator.ts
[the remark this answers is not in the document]
nicholaspai: ah i see good point, i might experiment with this

src/relayFeeCalculator/relayFeeCalculator.ts
nicholaspai: > so a user can call it to create a validated config file from string, and so that you can strongly type the configuration parameters to the class constructor.

Is this not strongly typed right now?

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -50,7 +60,29 @@ export class RelayFeeCalculator {
       this.capitalCostsPercent >= 0 && this.capitalCostsPercent <= 100,
       "capitalCostsPercent must be between 0 and 100 percent"
     );
+    this.capitalCostsConfig = config.capitalCostsConfig ? JSON.parse(config.capitalCostsConfig) : {};
+    for (const token of Object.keys(this.capitalCostsConfig)) {
+      RelayFeeCalculator.validateCapitalCostsConfig(this.capitalCostsConfig[token]);
+    }
+  }
+
+  static validateCapitalCostsConfig(_capitalCosts: unknown) {
```
daywiss, line 69: when you do something like this, you can `static validateCapitalCostsConfig(_capitalCosts: unknown) as { [token: string]: CapitalCostConfig }` and it will "prove" to typescript that if this returns the type is correct ( i think) @mrice32

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -50,7 +60,29 @@ export class RelayFeeCalculator {
       this.capitalCostsPercent >= 0 && this.capitalCostsPercent <= 100,
       "capitalCostsPercent must be between 0 and 100 percent"
     );
+    this.capitalCostsConfig = config.capitalCostsConfig ? JSON.parse(config.capitalCostsConfig) : {};
```
daywiss, line 63: you shouldnt need to parse this now that its typed

src/relayFeeCalculator/relayFeeCalculator.ts
nicholaspai: I ended up removing the `unknown` type as its not needed anymore

src/relayFeeCalculator/relayFeeCalculator.ts
nicholaspai: agreed

daywiss (approved):

PR #64, opened 2022-08-08
improve: update sdk queries to accurately estimate gas costs

james-a-morris:
This PR aims to reflect the gas cost estimation changes proposed by [PR 212](https://github.com/across-protocol/relayer-v2/pull/212) at the SDK level. 

During the development of this PR, I found a pattern in the call structure across all chain queries to estimate gas costs. I encapsulated it in two utility functions. 

Finally, the SDK [e2e test ](https://github.com/across-protocol/sdk-v2/runs/7641513096?check_suite_focus=true) was found to have an error on production, so this PR includes a patch.

### DIFF
--- a/src/relayFeeCalculator/chain-queries/arbitrum.ts
+++ b/src/relayFeeCalculator/chain-queries/arbitrum.ts
@@ -1,6 +1,10 @@
 import { QueryInterface } from "../relayFeeCalculator";
-import { BigNumberish } from "../../utils";
-import { BigNumber, providers } from "ethers";
+import {
+  BigNumberish,
+  createUnsignedFillRelayTransaction,
+  estimateTotalGasRequiredByUnsignedTransaction,
+} from "../../utils";
+import { providers } from "ethers";
 import { SymbolMapping } from "./ethereum";
 import { Coingecko } from "../../coingecko/Coingecko";
 import { ArbitrumSpokePool__factory, ArbitrumSpokePool } from "@across-protocol/contracts-v2";
@@ -19,9 +23,8 @@ export class ArbitrumQueries implements QueryInterface {
   }
 
   async getGasCosts(_tokenSymbol: string): Promise<BigNumberish> {
-    const gasEstimate = await this.estimateGas();
-    const gasPrice = BigNumber.from(await this.provider.getGasPrice());
-    return gasPrice.mul(gasEstimate).toString();
+    const tx = await createUnsignedFillRelayTransaction(this.spokePool, this.usdcAddress, this.simulatedRelayerAddress);
+    return estimateTotalGasRequiredByUnsignedTransaction(tx, this.simulatedRelayerAddress, this.provider);
   }
 
   async getTokenPrice(tokenSymbol: string): Promise<string | number> {
@@ -33,22 +36,5 @@ export class ArbitrumQueries implements QueryInterface {
   async getTokenDecimals(tokenSymbol: string): Promise<number> {
     if (!this.symbolMapping[tokenSymbol]) throw new Error(`${tokenSymbol} does not exist in mapping`);
     return this.symbolMapping[tokenSymbol].decimals;
-  }
-
-  estimateGas() {
-    // Create a dummy transaction to estimate. Note: the simulated caller would need to be holding weth and have approved the contract.
-    return this.spokePool.estimateGas.fillRelay(
-      "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B",
-      "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B",
-      this.usdcAddress,
-      "10",
-      "10",
-      "1",
-      "1",
-      "1",
-      "1",
-      "1",
-      { from: this.simulatedRelayerAddress }
-    );
   }
 }
--- a/src/relayFeeCalculator/chain-queries/boba.ts
+++ b/src/relayFeeCalculator/chain-queries/boba.ts
@@ -1,5 +1,9 @@
 import { QueryInterface } from "../relayFeeCalculator";
-import { BigNumberish } from "../../utils";
+import {
+  BigNumberish,
+  createUnsignedFillRelayTransaction,
+  estimateTotalGasRequiredByUnsignedTransaction,
+} from "../../utils";
 import { utils, providers } from "ethers";
 import { SymbolMapping } from "./ethereum";
 import { Coingecko } from "../../coingecko/Coingecko";
@@ -11,7 +15,7 @@ export class BobaQueries implements QueryInterface {
   private spokePool: OptimismSpokePool;
 
   constructor(
-    provider: providers.Provider,
+    private provider: providers.Provider,
     readonly symbolMapping = SymbolMapping,
     spokePoolAddress = "0xBbc6009fEfFc27ce705322832Cb2068F8C1e0A58",
     private readonly usdcAddress = "0x66a2A913e447d6b4BF33EFbec43aAeF87890FBbc",
@@ -22,24 +26,13 @@ export class BobaQueries implements QueryInterface {
   }
 
   async getGasCosts(_tokenSymbol: string): Promise<BigNumberish> {
-    // Create a dummy transaction to estimate. Note: the simulated caller would need to be holding weth and have approved the contract.
-    const gasEstimate = await this.spokePool.estimateGas.fillRelay(
-      "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B",
-      "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B",
-      this.usdcAddress,
-      "10",
-      "10",
-      "1",
-      "1",
-      "1",
-      "1",
-      "1",
-      { from: this.simulatedRelayerAddress }
+    const tx = await createUnsignedFillRelayTransaction(this.spokePool, this.usdcAddress, this.simulatedRelayerAddress);
+    return estimateTotalGasRequiredByUnsignedTransaction(
+      tx,
+      this.simulatedRelayerAddress,
+      this.provider,
+      parseUnits("1", 9)
     );
-
-    // Boba's gas price is hardcoded to 1 gwei.
-    const bobaGasPrice = parseUnits("1", 9);
-    return gasEstimate.mul(bobaGasPrice).toString();
   }
 
   async getTokenPrice(tokenSymbol: string): Promise<number> {
--- a/src/relayFeeCalculator/chain-queries/ethereum.ts
+++ b/src/relayFeeCalculator/chain-queries/ethereum.ts
@@ -1,7 +1,12 @@
 import { QueryInterface } from "../relayFeeCalculator";
-import { BigNumberish } from "../../utils";
+import {
+  BigNumberish,
+  createUnsignedFillRelayTransaction,
+  estimateTotalGasRequiredByUnsignedTransaction,
+} from "../../utils";
 import { Coingecko } from "../../coingecko/Coingecko";
-import { providers, BigNumber } from "ethers";
+import { providers } from "ethers";
+import { EthereumSpokePool, EthereumSpokePool__factory } from "@across-protocol/contracts-v2";
 
 // Note: these are the mainnet addresses for these symbols meant to be used for pricing.
 export const SymbolMapping: { [symbol: string]: { address: string; decimals: number } } = {
@@ -70,15 +75,20 @@ export const SymbolMapping: { [symbol: string]: { address: string; decimals: num
 export const defaultAverageGas = 116006;
 
 export class EthereumQueries implements QueryInterface {
+  private spokePool: EthereumSpokePool;
+
   constructor(
     public readonly provider: providers.Provider,
-    public readonly averageGas = defaultAverageGas,
-    readonly symbolMapping = SymbolMapping
-  ) {}
+    readonly symbolMapping = SymbolMapping,
+    readonly spokePoolAddress = "0x4D9079Bb4165aeb4084c526a32695dCfd2F77381",
+    readonly usdcAddress = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+    readonly simulatedRelayerAddress = "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D"
+  ) {
+    this.spokePool = EthereumSpokePool__factory.connect(this.spokePoolAddress, this.provider);
+  }
   async getGasCosts(_tokenSymbol: string): Promise<BigNumberish> {
-    return BigNumber.from(await this.provider.getGasPrice())
-      .mul(this.averageGas)
-      .toString();
+    const tx = await createUnsignedFillRelayTransaction(this.spokePool, this.usdcAddress, this.simulatedRelayerAddress);
+    return estimateTotalGasRequiredByUnsignedTransaction(tx, this.simulatedRelayerAddress, this.provider);
   }
 
   async getTokenPrice(tokenSymbol: string): Promise<number> {
--- a/src/relayFeeCalculator/chain-queries/queries.e2e.ts
+++ b/src/relayFeeCalculator/chain-queries/queries.e2e.ts
@@ -2,6 +2,8 @@
 // NODE_URL_42161
 // NODE_URL_288
 // NODE_URL_10
+// NODE_URL_1
+// NODE_URL_137
 
 import dotenv from "dotenv";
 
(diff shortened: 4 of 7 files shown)


src/relayFeeCalculator/chain-queries/queries.e2e.ts
```diff
@@ -2,6 +2,8 @@
 // NODE_URL_42161
 // NODE_URL_288
 // NODE_URL_10
+// NODE_URL_1
+// NODE_URL_137
```
james-a-morris, line 6: These two environment variables may need to be added to the e2e test file.

src/utils.ts
```diff
@@ -222,3 +223,63 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
+  } else {
+    // Estimate the Gas units required to submit this transaction
+    const estimatedGasUnits = await voidSigner.estimateGas(unsignedTx);
+    // Provide a default gas price of the market rate if this condition has not been set
+    const resolvedGasPrice = gasPrice ?? (await provider.getGasPrice());
+    // Find the total gas cost by taking the product of the gas
+    // price & the estimated number of gas units needed
+    return BigNumber.from(resolvedGasPrice).mul(estimatedGasUnits).toString();
+  }
+}
```
james-a-morris, line 255: All queries used a form of generating a `new VoidSigner(...)` to estimate the gas units consumed by the `fillRelay` function. Specifically, all but those wrapped by `L2Provider<providers.Provider>`. This function accounts for that case and the case where a specified gas price is provided (for boba).

src/utils.ts
```diff
@@ -222,3 +223,63 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
+    "1",
+    "1",
+    "1",
+    "1",
+    "1",
+    { from: simulatedRelayerAddress },
+  ];
+  // Populate and return an unsigned tx as per the given spoke pool
+  return await spokePool.populateTransaction.fillRelay(...contractFunctionParams);
+}
```
james-a-morris, line 285: All queries used a form of generating a `new VoidSigner(...)` to estimate the gas units consumed by the `fillRelay` function. Specifically, all but those wrapped by `L2Provider<providers.Provider>`. This function accounts for that case and the case where a specified gas price is provided (for boba).

src/relayFeeCalculator/chain-queries/queries.e2e.ts
nicholaspai: [That seems easy to do](https://docs.github.com/en/actions/learn-github-actions/environment-variables)

Alternatively we could use the public [infura key](https://github.com/UMAprotocol/protocol/blob/master/packages/common/src/ProviderUtils.ts#L38) here but I do think setting env vars in github actions is more sustainable

src/relayFeeCalculator/chain-queries/boba.ts
```diff
@@ -11,7 +15,7 @@ export class BobaQueries implements QueryInterface {
   private spokePool: OptimismSpokePool;
 
   constructor(
-    provider: providers.Provider,
+    private provider: providers.Provider,
```
nicholaspai, line 18: out of curiosity, why make it private?

src/relayFeeCalculator/chain-queries/queries.e2e.ts
nicholaspai: Do you want to try to add these to the github actions file in `./.github/`?

src/relayFeeCalculator/chain-queries/queries.e2e.ts
```diff
@@ -35,7 +37,8 @@ describe("Queries", function () {
     ]);
   });
   test("Ethereum", async function () {
-    const ethereumQueries = new EthereumQueries();
+    const provider = new providers.JsonRpcProvider(process.env.NODE_URL_1);
+    const ethereumQueries = new EthereumQueries(provider);
     await Promise.all([
       ethereumQueries.getGasCosts("USDC"),
```
nicholaspai, line 43: Should we test that the return values are sensible values? For example, the gas costs are a wei value, decimals are 6, etc?

src/utils.ts
```diff
@@ -222,3 +223,63 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
+ * @returns A populated (but unsigned) transaction that can be signed/sent or used for estimating gas costs
+ */
+export async function createUnsignedFillRelayTransaction(
+  spokePool: BaseContract,
+  destinationTokenAddress: string,
+  simulatedRelayerAddress: string
+): Promise<PopulatedTransaction> {
+  // Generate a baseline set of function parameters for the fillRelay contract function
+  const contractFunctionParams = [
+    "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B",
```
nicholaspai, line 271: Is `0xBb` the dummy account? A small comment here would be useful

src/relayFeeCalculator/chain-queries/arbitrum.ts
```diff
@@ -19,9 +23,8 @@ export class ArbitrumQueries implements QueryInterface {
   }
 
   async getGasCosts(_tokenSymbol: string): Promise<BigNumberish> {
```
kevinuma, line 25: There's quite a lot of duplication across these different classes (arbitrum, polygon, etc.). Can we potentially change the QueryInterface they currently implement to be a class they can inherit all these common methods from? I think currently only Polygon has a different implementation of getTokenPrice() and all other ones look exactly the same (just different addresses which can easily be passed in the constructor)

nicholaspai (approved):
This is a great refactor! I left some comments on constructor variable visibility consistency and suggestions for the e2e tests. I also want to make sure that @kevinuma reviews this to make sure its consistent with the potential changes in his [relayer-v2](https://github.com/across-protocol/relayer-v2/pull/212) PR

src/utils.ts
james-a-morris: Yes, 0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B is/was the dummy account used to query each chain. I've added the note in this [commit](https://github.com/across-protocol/sdk-v2/pull/64/commits/cf521fa5ec51d8db9a8c7a36a8029b735201c3a5).

src/relayFeeCalculator/chain-queries/ethereum.ts
```diff
@@ -70,15 +75,20 @@ export const SymbolMapping: { [symbol: string]: { address: string; decimals: num
 export const defaultAverageGas = 116006;
 
 export class EthereumQueries implements QueryInterface {
+  private spokePool: EthereumSpokePool;
+
   constructor(
     public readonly provider: providers.Provider,
```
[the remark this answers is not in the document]
james-a-morris: Good point. I've modified each provider to have the `readonly` access modifier instead of just private (this [commit](https://github.com/across-protocol/sdk-v2/pull/64/commits/1e82c4e296ca903ffcaf57ec7377f76ee2c9aa32)). I believe these new modifiers align more closely with master and are now consistent.

src/relayFeeCalculator/chain-queries/boba.ts
james-a-morris: I tend toward encapsulating the data - however, upon further review, the master branch does use `readonly` for several providers. I've pushed a commit that has a consistent `readonly` modifier on all `provider` instance variables.

src/relayFeeCalculator/chain-queries/arbitrum.ts
james-a-morris: I agree - I did notice code-duplication within the chain queries. In this PR I focused primarily on refactoring the duplication within `getGasCosts`

But, I think changing QueryInterface to a class structure (assuming there aren't any frontend/relayer dependencies) would be a great move. As you mentioned, the inheritance would help refactor the code into a more succinct format.

kevinuma: It would actually remove all the code for all chains, if I'm not mistaken, except for Polygon, which needs to override getTokenPrice().

src/relayFeeCalculator/chain-queries/ethereum.ts
```diff
@@ -70,15 +75,20 @@ export const SymbolMapping: { [symbol: string]: { address: string; decimals: num
 export const defaultAverageGas = 116006;
```
kevinuma, line 75: Can this be removed now?

james-a-morris: Yep! I left it in as an artifact for the event that the frontend-v2 / relayer-v2 / calling application referenced the `defaultAverageGas` variable. However, it has no functional use in the code anymore. (Updated: [here](https://github.com/across-protocol/sdk-v2/pull/64/commits/ffa6704b82773bcbeedf8e5e525d188dd11b2be5))

src/relayFeeCalculator/chain-queries/arbitrum.ts
james-a-morris: Yep! And if we pass a `coinGeckoBaseCurrency` variable to the base class, we can use the shared implementation of `super.getTokenPrice()` for the first half of the Polygon `getTokenPrice` custom implementation (further cutting the code reuse).

src/utils.ts
```diff
@@ -222,3 +223,64 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
+ * Estimates the total gas cost required to submit an unsigned (populated) transaction on-chain
+ * @param unsignedTx The unsigned transaction that this function will estimate
+ * @param relayerAddress The address that the transaction will be submitted from ()
+ * @param provider A valid ethers provider - will be used to reason the gas price
+ * @param gasPrice A manually provided gas price - if set, this function will not resolve the current gas price
+ * @returns The total gas cost to submit this transaction - i.e. gasPrice * estimatedGasUnits
+ */
+export async function estimateTotalGasRequiredByUnsignedTransaction(
+  unsignedTx: PopulatedTransaction,
+  relayerAddress: string,
```
mrice32, line 237: tiny nit: nothing else in this function seems specific to a relay, so you might consider changing this variable name to senderAddress since it could be used for _any_ transaction.

src/utils.ts
```diff
@@ -222,3 +223,64 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
+}
+
+/**
+ * Create an unsigned transaction of a fillRelay contract call
+ * @param spokePool The specific spokepool that will populate this tx
+ * @param destinationTokenAddress A valid ERC20 token (system-wide default is UDSC)
+ * @param simulatedRelayerAddress The relayer address that relays this transaction
+ * @returns A populated (but unsigned) transaction that can be signed/sent or used for estimating gas costs
+ */
+export async function createUnsignedFillRelayTransaction(
```
mrice32, line 264: This is great! +1 very nice refactor

src/utils.ts
```diff
@@ -222,3 +223,64 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
+  spokePool: BaseContract,
```
mrice32, line 265: nit: you didn't originally design it this way, but I see that each implementation uses PolygonSpokePool, OptimismSpokePool, etc. However, the interfaces we're using across these classes come from the base SpokePool class.

Anyway, to get better typing, we could change all of those to just SpokePool and take that type in here. Might make things a little cleaner and give us better type-checking.

src/relayFeeCalculator/chain-queries/ethereum.ts
```diff
@@ -67,18 +72,21 @@ export const SymbolMapping: { [symbol: string]: { address: string; decimals: num
...
   constructor(
-    public readonly provider: providers.Provider,
-    public readonly averageGas = defaultAverageGas,
-    readonly symbolMapping = SymbolMapping
-  ) {}
+    readonly provider: providers.Provider,
+    readonly symbolMapping = SymbolMapping,
+    readonly spokePoolAddress = "0x4D9079Bb4165aeb4084c526a32695dCfd2F77381",
+    readonly usdcAddress = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
+    readonly simulatedRelayerAddress = "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D"
```
mrice32, line 83: This is perfect for testing!

Before we merge, we should make sure all the default addresses are moved to `0x893d0D70AD97717052E3AA8903D9615804167759` and that address has the right funding and approvals on all chains.

src/utils.ts
```diff
@@ -222,3 +223,64 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
+ */
+export async function estimateTotalGasRequiredByUnsignedTransaction(
+  unsignedTx: PopulatedTransaction,
+  relayerAddress: string,
+  provider: providers.Provider | L2Provider<providers.Provider>,
+  gasPrice?: BigNumberish
+): Promise<BigNumberish> {
+  const voidSigner = new VoidSigner(relayerAddress, provider);
+  // Verify if this provider has been L2Provider wrapped
+  if (isL2Provider(provider)) {
```
mrice32, line 243: nit: can we make this comment more specific that this detects whether this is an Optimism-specific provider? The way it's written now, it almost seems like all L2s get captured in this if, whereas this logic is Optimism-specific.

Alternatively, you could rename isL2Provider/L2Provider on import to something more optimism specific. Either way is cool with me!

mrice32 (commented):
This all looks really great! Nice work @james-a-morris!

src/utils.ts
james-a-morris: That makes total sense! I've updated it [here](https://github.com/across-protocol/sdk-v2/pull/64/commits/451c80e04daffbffb6d2999e873731a4b4c46ba3).

src/utils.ts
james-a-morris: 100% agree. I've updated the files to use SpokePool instead of BaseContract [here](https://github.com/across-protocol/sdk-v2/pull/64/commits/ede380a86add898ec16ee34c058fd156eae7a086).

src/utils.ts
james-a-morris: Updated [here](https://github.com/across-protocol/sdk-v2/pull/64/commits/d0824419500f03a0dcd73ae6c9f5dd5e96160fcf). To really highlight the difference, I added both a comment & changed the function alias

src/relayFeeCalculator/chain-queries/ethereum.ts
james-a-morris: Not a problem! Currently `0x893d0D70AD97717052E3AA8903D9615804167759` doesn't have the proper access on ETH mainnet & Polygon. I will work to set it properly

james-a-morris: Thanks for the help! Updated [here](https://github.com/across-protocol/sdk-v2/pull/64/commits/45df24b38a73b80e4d830afa74ce4572505ea1d5).

src/relayFeeCalculator/chain-queries/arbitrum.ts
```diff
@@ -1,12 +1,16 @@
...
+import {
+  BigNumberish,
+  createUnsignedFillRelayTransaction,
+  estimateTotalGasRequiredByUnsignedTransaction,
+} from "../../utils";
+import { providers } from "ethers";
 import { SymbolMapping } from "./ethereum";
 import { Coingecko } from "../../coingecko/Coingecko";
-import { ArbitrumSpokePool__factory, ArbitrumSpokePool } from "@across-protocol/contracts-v2";
+import { ArbitrumSpokePool__factory, SpokePool } from "@across-protocol/contracts-v2";
```
mrice32, line 10: nit: shouldn't these factories all be changed to be SpokePool__factory? I'm a bit surprised that the types are working (I didn't think that on-chain inheritance implies typechain type inheritance, but that's cool!). Anyway, it's probably a good idea to use the same factory type as the contract type you're using just to avoid any type issues in the future.

src/utils.ts
```diff
@@ -222,3 +224,64 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
+export async function estimateTotalGasRequiredByUnsignedTransaction(
+  unsignedTx: PopulatedTransaction,
+  senderAddress: string,
+  provider: providers.Provider | L2Provider<providers.Provider>,
+  gasPrice?: BigNumberish
+): Promise<BigNumberish> {
+  const voidSigner = new VoidSigner(senderAddress, provider);
+  // Verify if this provider has been L2Provider wrapped
+  // NOTE: In this case, this will be true if the provider is
+  //       using the Optimism blockchain
```
mrice32, line 245:
```suggestion
  // This branches in the Optimism case because they use a special provider, called L2Provider, and special gas logic
  // to compute gas costs on Optimism.
```

mrice32 (approved):
Looks great! Two minor nits/suggestions

PR #66, opened 2022-08-09
improve(relayerFeeCalculator): Add logging around relayer fee calculations

kevinuma:
This is a temporary solution to test Vercel integration with Datadog. We can switch to a longer term solution like winston logger once this is confirmed to work.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -8,7 +8,7 @@ const { percent, fixedPointAdjustment } = uma.across.utils;
 export interface QueryInterface {
   getGasCosts: (tokenSymbol: string) => Promise<BigNumberish>;
   getTokenPrice: (tokenSymbol: string) => Promise<number | string>;
-  getTokenDecimals: (tokenSymbol: string) => Promise<number>;
+  getTokenDecimals: (tokenSymbol: string) => number;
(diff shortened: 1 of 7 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -72,11 +72,14 @@ export class RelayFeeCalculator {
   }
 
   async gasFeePercent(amountToRelay: BigNumberish, tokenSymbol: string): Promise<BigNumber> {
-    const [gasCosts, tokenPrice, decimals] = await Promise.all([
-      this.queries.getGasCosts(tokenSymbol),
-      this.queries.getTokenPrice(tokenSymbol),
-      this.queries.getTokenDecimals(tokenSymbol),
-    ]);
+    const getGasCosts = this.queries.getGasCosts(tokenSymbol).catch((error) => {
```
nicholaspai, line 75: does this .catch() combination into Promise.all() work? never seen this before but I've never tried

nicholaspai (approved):

src/relayFeeCalculator/relayFeeCalculator.ts
kevinuma: I think so from googling around :) catch() can't be added to Promise.all() as that would obscure which call the error comes from specifically

PR #70, opened 2022-08-09
feat: Allow relayer fee user to pass in token price to use

nicholaspai:
This will allow Vercel API for example to precompute Coingecko price and pass into SDK so that SDK doesn't have to make API call

### DIFF
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -33,7 +33,12 @@ export class Coingecko {
     return this.instance;
   }
 
-  private constructor(private readonly host = "https://api.coingecko.com/api/v3") {}
+  public static getPro(apiKey?: string) {
+    if (!this.instance) this.instance = new Coingecko("https://pro-api.coingecko.com/api/v3", apiKey);
+    return this.instance;
+  }
+
+  private constructor(private readonly host = "https://api.coingecko.com/api/v3", private readonly apiKey?: string) {}
 
   // Fetch historic prices for a `contract` denominated in `currency` between timestamp `from` and `to`. Note timestamps
   // are assumed to be js timestamps and are converted to unixtimestamps by dividing by 1000.
@@ -106,7 +111,7 @@ export class Coingecko {
       try {
         const { host } = this;
         const url = `${host}/${path}`;
-        const result = await axios(url);
+        const result = await axios(url, { params: { x_cg_pro_api_key: this.apiKey } });
         return result.data;
       } catch (err) {
         const msg = get(err, "response.data.error", get(err, "response.statusText", "Unknown Coingecko Error"));
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -28,17 +28,16 @@ export class Coingecko {
   private retryDelay = 1;
   private numRetries = 3;
 
-  public static get(host?: string) {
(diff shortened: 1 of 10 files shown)


kevinuma (approved):

src/relayFeeCalculator/chain-queries/arbitrum.ts
```diff
@@ -27,9 +27,10 @@ export class ArbitrumQueries implements QueryInterface {
     return estimateTotalGasRequiredByUnsignedTransaction(tx, this.simulatedRelayerAddress, this.provider);
   }
 
-  async getTokenPrice(tokenSymbol: string): Promise<number> {
+  async getTokenPrice(tokenSymbol: string, coingeckoProApiKey?: string): Promise<number> {
```
kevinuma, line 30: Can this key be passed in the relayerFeeCalculator config?

src/coingecko/Coingecko.ts
```diff
@@ -33,7 +33,12 @@ export class Coingecko {
     return this.instance;
   }
 
-  private constructor(private readonly host = "https://api.coingecko.com/api/v3") {}
+  public static getPro(apiKey?: string) {
+    if (!this.instance) this.instance = new Coingecko("https://pro-api.coingecko.com/api/v3", apiKey);
+    return this.instance;
+  }
```
mrice32, line 39: nit: can we just drop the host param from the get method and combine get and getPro?

mrice32 (commented):
Looks great! One small question

src/coingecko/Coingecko.ts
nicholaspai: How would that work?

src/relayFeeCalculator/chain-queries/arbitrum.ts
nicholaspai: Even better, I can inject into the Query constructor

src/coingecko/Coingecko.ts
nicholaspai: Like this? https://github.com/across-protocol/sdk-v2/pull/70/commits/6b6fa44da99abf43859950696a68d284c794af59

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -80,7 +84,9 @@ export class RelayFeeCalculator {
       console.error(`ERROR(gasFeePercent): Error while fetching token price ${error}`);
       throw error;
     });
-    const [gasCosts, tokenPrice] = await Promise.all([getGasCosts, getTokenPrice]);
+    const results = await Promise.all(_tokenPrice !== undefined ? [getGasCosts, getTokenPrice] : [getGasCosts]);
+    const gasCosts = results[0];
+    const tokenPrice = (_tokenPrice !== undefined ? _tokenPrice : results[1]) as number;
```
mrice32, line 89: Promise.all works on non-promises, too!

```suggestion
    const [gasCosts, tokenPrice] = await Promise.all([getGasCosts, _tokenPrice !== undefined ? _tokenPrice : getTokenPrice]);
```

mrice32 (approved):
Looks great!

src/coingecko/Coingecko.ts
mrice32: Yep, that's what I was thinking!

kevinuma (approved):

PR #72, opened 2022-08-10
fix(e2e-tests): Fix false positive queries.e2e.ts test file 

nicholaspai:
Test isn't correctly testing when coingecko api pro key is defined. Does not affect code in prod as logic does work, but e2e test wasn't specifically testing the logic

### DIFF
--- a/src/relayFeeCalculator/chain-queries/queries.e2e.ts
+++ b/src/relayFeeCalculator/chain-queries/queries.e2e.ts
@@ -11,7 +11,7 @@ import { ArbitrumQueries } from "./arbitrum";
 
 import { providers } from "ethers";
 import { BobaQueries } from "./boba";
-import { EthereumQueries, SymbolMapping } from "./ethereum";
+import { EthereumQueries } from "./ethereum";
 import { OptimismQueries } from "./optimism";
 import { PolygonQueries } from "./polygon";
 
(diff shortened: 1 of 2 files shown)


kevinuma (approved):

src/relayFeeCalculator/chain-queries/queries.e2e.ts
```diff
@@ -20,92 +20,82 @@ dotenv.config();
...
     ]);
-    const queriesWithCoingeckoProApi = new ArbitrumQueries(
-      provider,
-      SymbolMapping,
-      "0xB88690461dDbaB6f04Dfad7df66B7725942FEb9C",
-      "0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8",
-      "0x893d0d70ad97717052e3aa8903d9615804167759",
-      process.env.COINGECKO_PRO_API_KEY
-    );
-    await queriesWithCoingeckoProApi.getTokenPrice("USDC");
```
kevinuma, line 37: Why wasn't this correctly testing when the api key is set?

nicholaspai: Because the `cg` object is only created once by `Coingecko.get()` [here](https://github.com/across-protocol/sdk-v2/blob/master/src/relayFeeCalculator/chain-queries/arbitrum.ts#L33) and [here](https://github.com/across-protocol/sdk-v2/blob/master/src/coingecko/Coingecko.ts#L32) so the first time we create it without the API key

src/relayFeeCalculator/chain-queries/queries.e2e.ts
```diff
@@ -20,92 +20,82 @@ dotenv.config();
...
     const provider = new providers.JsonRpcProvider(process.env.NODE_URL_42161);
-    const arbitrumQueries = new ArbitrumQueries(provider);
+    const arbitrumQueries = new ArbitrumQueries(
+      provider,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      process.env.COINGECKO_PRO_API_KEY
+    );
```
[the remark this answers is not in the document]
nicholaspai: yep!

PR #75, opened 2022-08-10
improve: add gas buffer for each chain query

james-a-morris:
This PR aims to add a variable gas buffer to each query chain. The buffer is a `markup` that takes the formula of `totalGas = totalEstimatedGas * (1.0 + gasMarkup)` where `gasMarkup = 0.0` in the default case. Additionally, a constraint is applied such that `gasMarkup ∈ [-1.0, 4.0]`

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -242,7 +242,7 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
   gasMarkup: number,
   gasPrice?: BigNumberish
 ): Promise<BigNumberish> {
-  assert(gasMarkup >= -1 && gasMarkup <= 4, "Gas Markup must be within the range of [-1.0, +4.0]");
+  assert(gasMarkup > -1 && gasMarkup <= 4, "Gas Markup must be within the range of (-1.0, +4.0] so that total gas multiplier is between (0, +5.0]");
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -230,29 +231,34 @@ export async function retry<T>(call: () => Promise<T>, times: number, delayS: nu
...
  * @returns The total gas cost to submit this transaction - i.e. gasPrice * estimatedGasUnits
  */
 export async function estimateTotalGasRequiredByUnsignedTransaction(
   unsignedTx: PopulatedTransaction,
   senderAddress: string,
   provider: providers.Provider | L2Provider<providers.Provider>,
+  gasMarkup: number,
   gasPrice?: BigNumberish
 ): Promise<BigNumberish> {
+  assert(gasMarkup >= -1 && gasMarkup <= 4, "Gas Markup must be within the range of [-1.0, +4.0]");
```
nicholaspai, line 245:
```suggestion
  assert(gasMarkup > -1 && gasMarkup <= 4, "Gas Markup must be within the range of (-1.0, +4.0] so that total gas multiplier is between (0, +5.0]");
```

nicholaspai (approved):

PR #76, opened 2022-08-15
fix: Change broken Coingecko log, add winston to fix e2e test

nicholaspai:
Seeing lots of logs in Vercel like:

```
2022-08-15T11:38:18.877Z	ca0a5971-a196-45a0-9a90-e2cf706153c4	DEBUG	[
  {
    at: 'sdk-v2/coingecko',
    message: 'Sent GET request to url undefined'
  }
]
```

So this PR fixes the `undefined`

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
     "tsdx": "^0.14.1",
     "tslib": "^2.3.1",
     "typechain": "^7.0.1",
-    "typescript": "^4.6.2"
+    "typescript": "^4.6.2",
+    "winston": "^3.8.1"
   },
   "dependencies": {
     "@across-protocol/contracts-v2": "^0.0.50",
--- a/src/coingecko/Coingecko.e2e.ts
+++ b/src/coingecko/Coingecko.e2e.ts
@@ -1,14 +1,24 @@
 import assert from "assert";
 import Coingecko from "./Coingecko";
 import dotenv from "dotenv";
+import winston from "winston";
 dotenv.config({ path: ".env" });
 
+const createDummyLogger = () => {
+  return winston.createLogger({
+    silent: true,
+    level: "debug",
+    format: winston.format.json(),
+    transports: [new winston.transports.Console()],
+  });
+};
 // this requires e2e testing, should only test manually for now
 describe("coingecko", function () {
   let cg: Coingecko;
   test("init", function () {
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -66,7 +66,8 @@
     "tsdx": "^0.14.1",
     "tslib": "^2.3.1",
     "typechain": "^7.0.1",
-    "typescript": "^4.6.2"
+    "typescript": "^4.6.2",
```
nicholaspai, line 69: I don't think adding `winston` to the `devDependencies` would slow the Vercel API down but i'm not positive

kevinuma: Are you just adding winston to try things out?

mrice32: Agree @nicholaspai. devDependencies don't get bundled with the package.

mrice32 (approved):
LGTM!

dohaki (approved):
LGTM 👍

src/coingecko/Coingecko.e2e.ts
```diff
@@ -1,14 +1,24 @@
 import assert from "assert";
 import Coingecko from "./Coingecko";
 import dotenv from "dotenv";
+import winston from "winston";
 dotenv.config({ path: ".env" });
 
+const createDummyLogger = () => {
+  return winston.createLogger({
```
kevinuma, line 8: Why use winston for a test and not just a vanilla mock?

kevinuma (approved):

src/coingecko/Coingecko.ts
```diff
@@ -117,7 +117,7 @@ export class Coingecko {
         const { host } = this;
         const url = `${host}/${path}`;
         const result = await axios(url, { params: { x_cg_pro_api_key: this.apiKey } });
-        this.logger.debug({ at: "sdk-v2/coingecko", message: `Sent GET request to url ${result.request.responseURL}` });
+        this.logger.debug({ at: "sdk-v2/coingecko", message: `Sent GET request to url ${url}` });
```
pxrl, line 120: Ah - I have a [PR](https://github.com/across-protocol/sdk-v2/pull/79) up for this but hadn't realised you'd already proposed to fix it here.

I think it would be nice to migrate the message above the request submission. As-is, if the HTTP request fails then you never see the log message, and that's probably exactly when you want to see it. So I'd propose "Sent" -> "Sending" and move it up 1 line.

nicholaspai:
Closing in favor of https://github.com/across-protocol/sdk-v2/pull/81

PR #81, opened 2022-08-17
improve(Coingecko): Use PRO API as fallback, not first choice

nicholaspai:
We want to reduce coingecko Pro API requests to under 500,000/month, currently we're on track for 1.2million a month.

Some stats:
- A direct request to CG basic API takes 200ms if successful or 100ms if you get rate limited and receive a 429
- Same with CG pro API: Takes 200ms if successful
- CG API needs to warm up, first call always takes slightly longer (say 250-300ms instead of 150-200)

Coingecko API requests can only be sent on cache misses, which should be rare (5% of requests) [following this PR](https://github.com/across-protocol/frontend-v2/commit/995853ea370c29b495a799848ca444412d9e53a4). For cache misses, the response time is already going to be slow because of how the Lambda function works, so saving 100ms on these requests isn't a material improvement.

Therefore, the extra 100ms saved by always hitting the Pro API vs trying the basic API first and then on a fallback, hitting the Pro API, is not worth it IMO versus reducing # of CG pro requests. Moreover, most of the basic API requests succeed, although the rate limit of ~20-30/minute does result in say 10% (empirical observation) of requests to fail, so the fallback will make sure no CG requests fail (<1%).

This PR will result in a 50% latency (+100ms) increase for any basic API requests that fail and fallback to pro, but I think that’s ok as these instances should be rare because. The average response time for a cache miss will still be >1000ms (give or take for the lambda spin up time), with some edge cases taking an extra 300ms to fallback to the pro API, and the average case taking an extra 200ms to hit the basic API.

But, the main benefit of this PR will be that CG pro requests will reduce to well under 500,000/month.

### DIFF
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -122,7 +122,7 @@ export class Coingecko {
       const basicUrl = `${host}/${path}`;
 
       try {
-        const result = await axios(basicUrl, { timeout: 1000 });
+        const result = await axios(basicUrl, { timeout: 500 });
         return result.data;
       } catch (err) {
         this.logger.debug({
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -27,8 +27,8 @@ export class Coingecko {
 
   // Retry configuration.
   private retryDelay = 1;
-  private numRetries = 3;
-  private basicApiTimeout = 130; // ms
+  private numRetries = 0; // Most failures are due to 429 rate-limiting, so there is no point in retrying.
+  private basicApiTimeout = 250; // ms
 
   public static get(logger: Logger, apiKey?: string) {
     if (!this.instance)
@@ -136,7 +136,9 @@ export class Coingecko {
         return await this._callPro(path);
       }
     };
-    return retry(sendRequest, this.numRetries, this.retryDelay);
+
+    // Note: If a pro API key is configured, there is no need to retry as the Pro API will act as the basic's fall back.
+    return retry(sendRequest, this.apiKey === undefined ? this.numRetries : 0, this.retryDelay);
   }
 
   private async _callBasic(path: string, timeout?: number) {
@@ -147,7 +149,7 @@ export class Coingecko {
       const result = await axios(url, { timeout });
       return result.data;
     } catch (err) {
-      const msg = get(err, "response.data.error", get(err, "response.statusText", "Unknown Coingecko Error"));
+      const msg = get(err, "response.data.error", get(err, "response.statusText", (err as AxiosError).message));
       throw new Error(msg);
     }
   }
@@ -160,7 +162,7 @@ export class Coingecko {
       const result = await axios(url, { params: { x_cg_pro_api_key: this.apiKey } });
       return result.data;
     } catch (err) {
-      const msg = get(err, "response.data.error", get(err, "response.statusText", "Unknown Coingecko-Pro Error"));
+      const msg = get(err, "response.data.error", get(err, "response.statusText", (err as AxiosError).message));
       throw new Error(msg);
     }
   }
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -122,7 +122,7 @@ export class Coingecko {
       const basicUrl = `${host}/${path}`;
 
       try {
-        const result = await axios(basicUrl, { timeout: 500 });
+        const result = await axios(basicUrl);
         return result.data;
       } catch (err) {
         this.logger.debug({
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -30,17 +30,19 @@ export class Coingecko {
   private numRetries = 3;
 
   public static get(logger: Logger, apiKey?: string) {
-    if (!this.instance) {
-      this.instance =
-        apiKey === undefined
-          ? new Coingecko("https://api.coingecko.com/api/v3", logger)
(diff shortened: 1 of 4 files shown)


src/coingecko/Coingecko.ts
```diff
@@ -113,15 +115,28 @@ export class Coingecko {
 
   async call(path: string) {
     const sendRequest = async () => {
+      const { host, proHost } = this;
+
+      // Try basic, and then pro as fallback.
+      this.logger.debug({ at: "sdk-v2/coingecko", message: `Sending GET request to host ${host}` });
```
kevinuma, line 121: This can add latency. Did you do some test runs to see how much latency this would add when basic fails? You can simulate that by sending enough traffic that the basic endpoint rate limits you.

nicholaspai: how else would you only use pro API as a fallback without adding latency?

nicholaspai: I can force a 1000ms timeout on the basic request also

kevinuma: You can set a relatively short timeout on the basic request to limit latency added. How important is this change though when Vercel already has caching? An alternative is to lengthen the cache TTL for the coingecko endpoint right?

kevinuma: 1000ms is too high IMO. That adds a whole second to latency in the worst case. We should cap in on P50 or less of the basic coingecko request.

nicholaspai: Actually, when empirically testing the latency isn't affected by setting a timeout on the first call. My local tests show that falling back to pro API changes latency from 150-200ms to 300-350ms

nicholaspai: The caching is only for 5 mins. This change would reduce the cg-pro requests a lot because most of the time the basic api works: How important is this change though when Vercel already has caching?

i can cap it at 200ms, 50% of the requests fall under this

src/coingecko/Coingecko.ts
```diff
@@ -113,15 +115,28 @@ export class Coingecko {
...
+        const result = await axios(basicUrl, { timeout: 200 });
         return result.data;
       } catch (err) {
-        const msg = get(err, "response.data.error", get(err, "response.statusText", "Unknown Coingecko Error"));
-        throw new Error(msg);
+        this.logger.debug({
+          at: "sdk-v2/coingecko",
+          message: `Basic CG url request failed, falling back to CG PRO host ${proHost}`,
+        });
+        const proUrl = `${proHost}/${path}`;
```
pxrl, line 132: It's not guaranteed that apiKey is defined, which would cause a subsequent rejection from the endpoint and would falsely report to the caller that the price lookup failed due to an auth/no-key error.

src/coingecko/Coingecko.ts
kevinuma: Is it worth it though to reduce requests to CG Pro this way? IMO, API performance seems more important than saving some amount of requests that miss the edge cache.

mrice32: If we create very low timeouts, we may _increase_ average latency because it causes us to fallback unnecessarily and force us to start a new request.

How often does this first endpoint fail? 5% of the time? 1% of the time? If it's 5% or 1% then it shouldn't have a huge impact on average latency.

src/coingecko/Coingecko.ts
```diff
@@ -113,15 +115,28 @@ export class Coingecko {
...
+        try {
+          const result = await axios(proUrl, { params: { x_cg_pro_api_key: this.apiKey } });
+          return result.data;
+        } catch (errPro) {
+          const msg = get(err, "response.data.error", get(err, "response.statusText", "Unknown Coingecko Error"));
+          throw new Error(msg);
+        }
       }
     };
     return retry(sendRequest, this.numRetries, this.retryDelay);
```
pxrl, line 144: If I am reading correctly, the inbuilt retry behaviour here could result in 8 requests to CoinGecko occurring (in the worst-case scenario of all requests failing). It looks like the requests would be interleaved between free and pro.

src/coingecko/Coingecko.ts
mrice32: According to our usage data, the caching isn't really reducing requests as much as it should, it seems. We've used 350k requests in a 9 days, right? That is far more than 1 request per token per 5 minutes, no?

mrice32: Do we know why that is?

src/coingecko/Coingecko.ts
nicholaspai: Great point! added logic to only use bbasic api if no pro api key

src/coingecko/Coingecko.ts
nicholaspai: Yeah you're right, should we remove the retry? Or perhaps, only retry against pro api?

src/coingecko/Coingecko.ts
nicholaspai: Screen Shot 2022-08-17 at 12 46 51


Quick look at datadog dashboard shows cache getting hit a lot

src/coingecko/Coingecko.ts
nicholaspai: Actually, it would be tricky to implement the "retry only against pro api", any suggestions?

nicholaspai: we could lower the retry count?

nicholaspai: Changed the retry logic to set retry count to 0 if apiKey is present

kevinuma (approved):

PR #80, opened 2022-08-16
improve: consolidate query logic

james-a-morris:
This PR leverages OOP to consolidate the query logic of each Blockchain Query class. (Arbitrum, Ethereum, Optimism, Boba, Polygon)

PR #84, opened 2022-08-18
fix(queries): Add BAL to TokenMapping

kevinuma:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.1.19",
+  "version": "0.1.20",
```
kevinuma, line 4: Bump version so we don't have to do this in a separate PR.

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -1,7 +1,7 @@
 import { SpokePool, SpokePool__factory } from "@across-protocol/contracts-v2";
 import { L2Provider } from "@eth-optimism/sdk";
 import { providers } from "ethers";
-import Coingecko from "../../coingecko/Coingecko";
+import { Coingecko } from "../../coingecko";
```
kevinuma, line 4: @james-a-morris This fixes an issue introduced in https://github.com/across-protocol/sdk-v2/pull/80 that's fialing lint on master

Tulun (approved):

PR #87, opened 2022-08-24
feat: add merkle distributor

amateima:
This PR adds a wrapper around the Merkle tree used for managing the ACX airdrop recipients. This functionality will be used by the scripts used to generate the root and the proofs for the `MerkleDistributor` contract. The scripts will be added to [this repo](https://github.com/across-protocol/helper-scripts)


src/merkleDistributor/model/Distribution.ts
```diff
@@ -0,0 +1,23 @@
+export type DistributionRecipients = {
+  [address: string]: Pick<DistributionRecipient, "amount" | "accountIndex" | "metadata">;
+};
+export type DistributionRecipientsWithProofs = { [address: string]: DistributionRecipient };
+
+export type DistributionRecipient = {
+  amount: string;
+  accountIndex: number;
+  windowIndex: number;
```
dohaki, line 9: Can you explain why these are needed? I struggle to wrap my around how these indices are used 🤔

src/merkleDistributor/model/Distribution.ts
```diff
@@ -0,0 +1,23 @@
...
+  proofs: string[];
```
dohaki, line 10: Isn't this actually just a single `proof`?

dohaki (commented):
Cool stuff 😎 Left a few questions. And to make sure that I understand the usage from a frontend perspective correctly. The required steps to call `claim` from the UI would be:
- Call `MerkleDistributor.createMerkleDistributionProofs` by passing in static recipients JSON or from IPFS
- Call `claim` of contract `MerkleDistributor` with proof of previous step

?

amateima:
> Cool stuff 😎 Left a few questions. And to make sure that I understand the usage from a frontend perspective correctly. The required steps to call `claim` from the UI would be:
> 
> * Call `MerkleDistributor.createMerkleDistributionProofs` by passing in static recipients JSON or from IPFS
> * Call `claim` of contract `MerkleDistributor` with proof of previous step
> 
> ?

There is not need to call `createMerkleDistributionProofs` on the FE. The FE will get the proofs from the JSON file directly and send them using the `claim` contract call. Also the json file will contain the breakdown of the rewards amount

src/merkleDistributor/model/Distribution.ts
amateima: `windowIndex` is used to allow multiple distributions (airdrops) using the same contract without the need of deploying other MerkleDistributor contracts. 
`accountIndex` is an index used to identify the account in the smart contract, instead of using the account address. The contract has a map [windowIndex][accountId][boolean] which stores if the account claimed their tokens or not

src/merkleDistributor/model/Distribution.ts
amateima: it is an array of strings. Should I rename this to `proof`?

src/merkleDistributor/model/MerkleTree.ts
```diff
@@ -0,0 +1,126 @@
+// This script provides some useful methods for building MerkleTrees. It is taken verbatim from the uniswap implementation
```
mrice32, line 1: Random comment, can you explain the meaning of the "model" naming of this directory? I could see types or utils potentially being more clear?

src/merkleDistributor/MerkleDistributor.ts
```diff
@@ -0,0 +1,30 @@
...
+    const recipientsWithProofs = Object.keys(recipients).reduce((acc, recipientAddress, index) => {
+      return {
+        ...acc,
+        [recipientAddress]: {
+          ...recipients[recipientAddress],
+          proofs: merkleTree.getHexProof(recipientLeafs[index]),
+          windowIndex,
+        },
+      };
+    }, {} as DistributionRecipientsWithProofs);
```
mrice32, line 19: tiny nit: this might be simpler with just mapping over the entries and then using Object.fromEntries to turn that back into an object. But that may just be my preference because reduce operations always feel more complex than chained maps :)

src/merkleDistributor/MerkleDistributor.ts
```diff
@@ -0,0 +1,30 @@
...
+    return { recipientsWithProofs, merkleRoot: merkleTree.getHexRoot() };
+  };
+
+  // keccak256(abi.encode(account, amount))
+  static createLeaf = (account: string, amount: string, accountIndex: number) => {
```
mrice32, line 24: With the merkle distributor class from @across-protocol/contracts-v2, you can just pass a function like this into the class and then the class will take care of creating leaves for you.

mrice32 (commented):
Looks awesome, just a minor comment!

amateima:
> Because this is an SDK, would it make sense (in another PR) to begin to add [JSDoc](https://jsdoc.app/) comments?

Yes, definitely! I'll add some comments to this PR too

src/merkleDistributor/MerkleDistributor.ts
```diff
@@ -0,0 +1,30 @@
+import { ethers } from "ethers";
+import { DistributionRecipientsWithProofs, DistributionRecipients, MerkleTree } from "./model";
+
+export class MerkleDistributor {
+  static createMerkleDistributionProofs = (recipients: DistributionRecipients, windowIndex: number) => {
```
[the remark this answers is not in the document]
amateima: I'll refactor the functions in my code

src/merkleDistributor/model/MerkleTree.ts
```diff
@@ -0,0 +1,126 @@
...
+  private readonly layers: Buffer[][];
+
+  constructor(elements: Buffer[]) {
+    this.elements = [...elements];
+    // Sort elements
+    this.elements.sort(Buffer.compare);
+    // Deduplicate elements
+    this.elements = MerkleTree.bufDedup(this.elements);
+
+    // this.bufferElementPositionIndex = this.elements.reduce<{ [hexElement: string]: number }>((acc, el, index) => {
```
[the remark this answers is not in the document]
amateima: useless comment. I removed it

src/merkleDistributor/model/MerkleTree.ts
amateima: It's a synonym for types, yes, but I like it because it's a more generic word. `type` can be misleading in a JS program. Basically in the `model` dir you can find types, classes, enums, etc

src/merkleDistributor/MerkleDistributor.ts
amateima: 👍

src/merkleDistributor/MerkleDistributor.ts
amateima: After importing the existing MerkleTree implementation from the contracts repo, the chained maps were not needed anymore. I have a simple `.reduce` call now

src/merkleDistributor/model/MerkleTree.ts
```diff
@@ -0,0 +1,126 @@
+// This script provides some useful methods for building MerkleTrees. It is taken verbatim from the uniswap implementation
+// https://github.com/Uniswap/merkle-distributor/blob/master/src/merkle-tree.ts
+import { keccak256 } from "ethereumjs-util";
+import { utils } from "ethers";
+
+export class MerkleTree {
```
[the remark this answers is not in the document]
amateima: 👍

src/merkleDistributor/model/MerkleTree.ts
```diff
@@ -0,0 +1,126 @@
...
+  private readonly elements: Buffer[];
+  private readonly bufferElementPositionIndex: { [hexElement: string]: number };
+  private readonly layers: Buffer[][];
+
+  constructor(elements: Buffer[]) {
+    this.elements = [...elements];
```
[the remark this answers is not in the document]
amateima: @james-a-morris This file was removed anyway since we have an improved version of it in the contracts repo. I imported it from there

nicholaspai (approved):
Extremely well documented code, lgtm!

PR #89, opened 2022-08-31
improve(Coingecko): Faster token price retrievals

pxrl:
By wrapping getContractPrices(), we save ~24kB in response size (~99%).
From limited testing on my end, there's also a minor latency saving of
~100mS per request. Your mileage will inevitably vary.

Ref: ACX-194

### DIFF
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -72,10 +72,8 @@ export class Coingecko {
     currency = "usd",
     platform_id = "ethereum"
   ): Promise<[string, number]> {
-    const result = await this.getContractDetails(contract_address, platform_id);
-    const price = get(result, ["market_data", "current_price", currency], null);
-    assert(price !== null, "No current price available for: " + contract_address);
-    return [result.last_updated, price];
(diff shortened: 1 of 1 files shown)


src/coingecko/Coingecko.ts
```diff
@@ -72,10 +72,8 @@ export class Coingecko {
     currency = "usd",
     platform_id = "ethereum"
   ): Promise<[string, number]> {
-    const result = await this.getContractDetails(contract_address, platform_id);
```
pxrl, line 75: Note that the existing call to getContractDetails retrieved a huge amount of information about the token, only to discard 99% of it.

Using WETH on Ethereum L1 as an example:
https://api.coingecko.com/api/v3/coins/ethereum/contract/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2

https://api.coingecko.com/api/v3/simple/token_price/ethereum?contract_addresses=0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2&vs_currencies=usd

kevinuma: Wow amazing improvement - simple yet effective!

kevinuma (approved):

PR #91, opened 2022-09-01
fix(Coingecko): Add winston for tests

pxrl:
The Coingecko e2e test seems to be dependent on winston.

Ref: ACX-194

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
     "hardhat-deploy": "^0.11.2",
     "hardhat-gas-reporter": "^1.0.8",
     "lodash.get": "^4.4.2",
-    "superstruct": "^0.15.4"
+    "superstruct": "^0.15.4",
+    "winston": "^3.8.1"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -78,7 +78,8 @@
     "hardhat-deploy": "^0.11.2",
     "hardhat-gas-reporter": "^1.0.8",
     "lodash.get": "^4.4.2",
-    "superstruct": "^0.15.4"
+    "superstruct": "^0.15.4",
+    "winston": "^3.8.1"
```
pxrl, line 82: FWIW this commit added the dependency on winston: https://github.com/across-protocol/sdk-v2/commit/ca0de27d29e84b4a6ea42bb4a325ada2c0c63121

daywiss (approved):

package.json
pxrl: Just to wrap it up - I reverted this and added winston as a dev dependency instead.

PR #93, opened 2022-09-06
feat(Coingecko): Cache prices & bundle requests

pxrl:
This PR adds two new capabilities to the Coingecko price feed.

 - Maintain a local cache of all known network/currency/price
   combinations. Allow the cache duration to be configurable,
   defaulting to 300 seconds (5 minutes).
 - For each new price cache miss, perform a new price lookup.
   As part of this lookup, opportunistically bundle in all known
   token addresses for the requested network and currency.

The combination of these new features should ideally help to
significantly reduce the number of CoinGecko price queries
that are submitted by via the API backend.

Ref: ACX-194

### DIFF
--- a/src/coingecko/Coingecko.e2e.ts
+++ b/src/coingecko/Coingecko.e2e.ts
@@ -1,17 +1,41 @@
 import assert from "assert";
-import Coingecko from "./Coingecko";
 import dotenv from "dotenv";
 import winston from "winston";
+import * as coingecko from "./Coingecko";
 dotenv.config({ path: ".env" });
+
+const Coingecko = coingecko.Coingecko;
+const msToS = coingecko.msToS;
+type CoinGeckoPrice = coingecko.CoinGeckoPrice;
 
 const dummyLogger = winston.createLogger({
   level: "debug",
   transports: [new winston.transports.Console()],
 });
 
+class TestGecko extends Coingecko {
+
+  private static testInstance: TestGecko | undefined;
+
+  public static get(logger: winston.Logger) {
+    if (!this.testInstance)
+      this.testInstance = new TestGecko(logger);
+    return this.testInstance;
+  }
+
+  // Hack to return protected getPriceCache.
+  _getPriceCache(currency: string, platform_id: string): { [addr: string]: CoinGeckoPrice } {
+    return this.getPriceCache(currency, platform_id);
+  }
+
+  constructor(logger: winston.Logger) {
+    super("127.0.0.1", "127.0.0.1", logger);
+  };
+}
+
 // this requires e2e testing, should only test manually for now
 describe("coingecko", function () {
-  let cg: Coingecko;
+  let cg: coingecko.Coingecko;
   test("init", function () {
     cg = Coingecko.get(dummyLogger, process.env.COINGECKO_PRO_API_KEY);
     assert.ok(cg);
@@ -72,8 +96,47 @@ describe("coingecko", function () {
     jest.setTimeout(30000);
     // Send tons of basic requests so that we hit pro. Basic has a ~50/min rate limit but this varies. In practice
     // its a bit lower more like ~20-30/min.
+
+    cg.maxPriceAge = 0; // Disable cache to force price lookups
     for (let i = 0; i < 20; i++) {
       assert.ok(await cg.getCurrentPriceByContract("0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", "eth"));
+    }
+  });
+  test("Validate price cache", async function () {
+    // Don't lookup against CoinGecko.
+    let tg: TestGecko = TestGecko.get(dummyLogger);
+    assert.ok(tg);
+
+    const baseCurrency = "eth";
+    const network = "ethereum";
+
+    const priceCache: { [addr: string]: CoinGeckoPrice } = tg._getPriceCache(baseCurrency, network);
+    tg.maxPriceAge = 600; // Bound timestamps by 10 minutes
+    for (let i = 0; i < 10; ++i) {
+      const addr = `0x${i.toString(16).padStart(42, "0")}`; // Non-existent, ensure CG lookup would fail.
+      priceCache[addr] = {
+        address: addr,
+        price: Math.random() * (1 + i),
+        timestamp: msToS(Date.now()) - tg.maxPriceAge + (1 + i),
+      };
+    }
+
+    // Verify cache hit for valid timestamps.
+    for (const expected of Object.values(priceCache)) {
+      const addr: string = expected.address;
+      const result: [timestamp: string, price: number] = await tg.getCurrentPriceByContract(addr, baseCurrency);
+      const timestamp: string = result[0];
+      const price: number = result[1];
+
+      assert.ok(timestamp === expected.timestamp.toString(), `${expected.timestamp} !== ${timestamp}`);
+      assert.ok(price === expected.price, `${expected.price} !== ${price}`);
+    }
+
+    // Invalidate all cached results and verify failed price lookup.
+    tg.maxPriceAge = 1; // seconds
+    for (const expected of Object.values(priceCache)) {
+      const addr: string = expected.address;
+      await expect(tg.getCurrentPriceByContract(addr, baseCurrency)).rejects.toThrow();
     }
   });
 });
--- a/src/coingecko/Coingecko.e2e.ts
+++ b/src/coingecko/Coingecko.e2e.ts
@@ -74,9 +74,10 @@ describe("coingecko", function () {
       "0x6B3595068778DD592e39A122f4f5a5cF09C90fE2",
       "0x5F64Ab1544D28732F0A24F4713c2C8ec0dA089f0",
     ];
-    const result = await cg.getContractPrices(addresses);
+
+    const result: CoinGeckoPrice[] = await cg.getContractPrices(addresses);
     assert.equal(result.length, addresses.length);
-    result.forEach((result) => {
+    result.forEach((result: CoinGeckoPrice) => {
       assert.ok(result.price);
       assert.ok(result.timestamp);
       assert.ok(result.address);
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -21,9 +21,19 @@ export type CoinGeckoPrice = {
   price: number;
 };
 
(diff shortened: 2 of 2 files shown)


src/coingecko/Coingecko.e2e.ts
```diff
@@ -1,17 +1,41 @@
...
+
+  constructor(logger: winston.Logger) {
+    super("127.0.0.1", "127.0.0.1", logger);
+  };
+}
+
 // this requires e2e testing, should only test manually for now
 describe("coingecko", function () {
-  let cg: Coingecko;
+  let cg: coingecko.Coingecko;
```
pxrl, line 38: @daywiss This one is frustrating me..any suggestions on how to import the Coingecko class in a way that permits it to be used as both a type and class?

src/coingecko/Coingecko.e2e.ts
```diff
@@ -50,9 +74,10 @@ describe("coingecko", function () {
       "0x6B3595068778DD592e39A122f4f5a5cF09C90fE2",
       "0x5F64Ab1544D28732F0A24F4713c2C8ec0dA089f0",
     ];
-    const result = await cg.getContractPrices(addresses);
+
+    const result: CoinGeckoPrice[] = await cg.getContractPrices(addresses);
     assert.equal(result.length, addresses.length);
-    result.forEach((result) => {
+    result.forEach((result: CoinGeckoPrice) => {
```
pxrl, line 80: These are just driveby type annotations that I added because the CoinGeckoPrice type is now available; they don't relate to the new functionality here.

pxrl:
Also just dropping a log of the e2e test here, since it doesn't run automagically as part of CI:

[Coingecko.e2e.log](https://github.com/across-protocol/sdk-v2/files/9496077/Coingecko.e2e.log)

src/coingecko/Coingecko.ts
```diff
@@ -82,30 +145,74 @@ export class Coingecko {
     currency = "usd",
     platform_id = "ethereum"
   ): Promise<CoinGeckoPrice[]> {
-    // Generate a unique set with no repeated. join the set with the required coingecko delimiter.
-    const contract_addresses = Array.from(new Set(addresses.filter((n) => n).values()));
+    const priceCache: { [addr: string]: CoinGeckoPrice } = this.getPriceCache(currency, platform_id);
```
daywiss, line 148: i feel like the generic cache should be its own class where you use a generic to set the data type and have a basic interface like `set`, `get`, `has`, `delete` that take an argument for age, ie `has(key,timestamp)` returns false  if the value stored either does not exist or is older than timestamp, true otherwise.  the key could just be  string concatenation of `${currency}.${platform_id}.${address}`, or you could maintain separate caches for currency/platform like you do now.

src/coingecko/Coingecko.ts
```diff
@@ -82,30 +145,74 @@ export class Coingecko {
...
+    contract_addresses.forEach((addr) => {
+      const cgPrice: CGTokenPrice | undefined = result[addr.toLowerCase()];
+
+      if (cgPrice === undefined) {
+        this.logger.debug({
+          at: "Coingecko#getContractPrices",
+          message: `Token ${addr} not included in CoinGecko response.`,
+        });
+      } else if (cgPrice.last_updated_at > priceCache[addr].timestamp) {
+        priceCache[addr] = {
```
daywiss, line 193: something like cache.set(key,data,timestamp) would be nice here and only sets if timestamp is newer

src/coingecko/Coingecko.e2e.ts
pxrl: Yeah, works fine. Was pretty sure that I was previously unable to import CoinGeckoPrice this way, but it seems to work now, so not exactly sure what my issue was. Thanks.

src/coingecko/Coingecko.ts
```diff
@@ -21,9 +21,19 @@ export type CoinGeckoPrice = {
   price: number;
 };
 
+type PriceCache = {
```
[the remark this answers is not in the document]
pxrl: I think this is a fair observation.

To give a bit more context about what I see as the roadmap for the "price lookups" feature:
 - Implement a generic "PriceClient" class that acts as a client-facing implementation to query a price from.
 - During initialisation, the PriceClient instance has _n_ price feed adapters added, with a priority specified for each. For example:
   1. Across API
   2. CoinGecko Free
   2. CoinGecko Pro
   2. CoinMarketCap (??)
   2. On-chain DEX (??)
 - When the caller requests a price from the PriceClient, it looks up internally in its in-memory cache.
   1. For a cache hit, it returns the price immediately.
   2. For a cache miss, it performs a price lookup request incrementally down its ordered list until it gets a newer response than it currently has.

The advantages of this approach, as I see them:
 - Allows PriceClient users (front end, v2 bots) to preferentially leverage the Across API endpoint.
 - In the event of a failed lookup to any price feed, the existence of fallback price feeds should allow a price resolution without blocking the bots.
 - The existence of a preferential ordering in lookups allows users to minimise their reliance on paid endpoints like CoinGecko Pro, and to only fall back to them when they are really needed (i.e. price feed of last resort).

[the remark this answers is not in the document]
nicholaspai: @pxrl I think the `PriceClient` design with built in caching is a great idea, where the cache length can be confirgurable.

src/coingecko/Coingecko.ts
nicholaspai: This is a really good suggestion and will making adding caching to other data clients in the future easier. However, how much extra code will this add? I'm a bit worried about adding generalization but trading off for future work where we need to retrofit a new way of caching data to the existing `Cache` client that we create. 

@daywiss can you give a prototype of this proposed change? Also, @pxrl isn't as familiar with javascript specific things like generics so it'd be helpful to understand more what you mean

nicholaspai (commented):
High level this is a great feature and I think your proposed road map for a generic `PriceClient` makes sense. I do think David has a point with creating a generic caching client

src/coingecko/Coingecko.ts
```diff
@@ -82,30 +145,74 @@ export class Coingecko {
...
-    // Generate a unique set with no repeated. join the set with the required coingecko delimiter.
-    const contract_addresses = Array.from(new Set(addresses.filter((n) => n).values()));
+    const priceCache: { [addr: string]: CoinGeckoPrice } = this.getPriceCache(currency, platform_id);
+
+    // Pre-populate price cache with requested token addresses
+    addresses.forEach((addr: string) => {
+      if (priceCache[addr] === undefined) {
+        priceCache[addr] = { address: addr, price: 0, timestamp: 0 };
+      }
+    });
```
mrice32, line 155: I don't think this causes any obvious issues, but I would suggest doing all the writing synchronously, i.e. don't write before the async CG request and then again after. I would suggest doing all the writing after, so you don't leave the cache in a strange state for others to read while you're waiting on the CG price to come back (if possible).

mrice32 (approved):
LGTM!

src/coingecko/Coingecko.ts
```diff
@@ -41,12 +51,31 @@ export class Coingecko {
...
+  protected constructor(
     private readonly host: string,
     private readonly proHost: string,
     private readonly logger: Logger,
     private readonly apiKey?: string
-  ) {}
+  ) {
+    this.prices = {
+      ethereum: {
+        usd: {},
```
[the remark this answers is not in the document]
pxrl: I just noticed that we have a bias for USD in most cases, but the code works fine if it's not initialised, so there's no hard reason for having it here. Will remove it.

src/coingecko/Coingecko.ts
pxrl: My motivation for pre-populating with any new entries is that it allows the use of priceCache keys as the source for addresses to be requested from CG, so we inherit an array of unique addresses (assuming that the user doesn't request the same token in multiple different cases).  This makes it trivial to generate the list of addresses to be submitted to CoinGecko (i.e. line #158 slightly below this comment).

The existing implementation handled this by creating a temporary set to filter out any duplicate addresses. That's definitely also still valid and achievable; is it preferable to revert to that?

src/coingecko/Coingecko.ts
```diff
@@ -72,8 +101,42 @@ export class Coingecko {
...
+      try {
+        // Force price cache update.
+        await this.getContractPrices([contract_address], currency, platform_id);
+      } catch (err) {
+        const errMsg = `Failed to retrieve ${platform_id}/${currency} price for ${contract_address} (${err})`;
+        this.logger.warn({
+          at: "Coingecko#getCurrentPriceByContract",
+          message: errMsg,
+        });
+        throw new Error(errMsg);
```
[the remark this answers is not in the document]
pxrl: I also wasn't totally sure about the best strategy to apply here, so I landed on the status quo with some additional logging. The existing implementation didn't catch any errors, so they just rippled upwards. This implementation is basically just a continuation of that, with the addition of a warning.

The impact on PD is a good point though - this would unconditionally generate a low-urgency incident, and potentially a high-urgency incident if the exception isn't caught in the upper layers. Based on that, it seems like this should probably revert to a debug message instead. Thoughts?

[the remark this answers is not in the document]
pxrl: OK, so in conclusion I realised it was probably an error to log the exception in getPriceByContract(), so I have relocated that logging to getContractPrices() because it's the focal point for exceptions that might be thrown. Exceptions will unconditionally be rippled upwards, but at least we guarantee that the underlying cause is always logged.

src/coingecko/Coingecko.ts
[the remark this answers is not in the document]
pxrl: Going to mark this one resolved since the description has been migrated over to Linear: https://linear.app/uma/issue/ACX-264/improve-robustness-against-coingecko-rate-limiting

src/coingecko/Coingecko.ts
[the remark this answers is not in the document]
mrice32: That makes sense!

src/coingecko/Coingecko.ts
mrice32: No I think this is fine! I generally get a little scared when I see code that crosses awaits, but I do think this is handled okay elsewhere.

mrice32 (approved):
LGTM!

PR #98, opened 2022-09-12
feat(PriceClient): Support multiple price feeds

pxrl:
This feature adds a new class named PriceClient, which acts like an aggregator for a user-defined list of price feeds. This permits prices to be sourced from a range of sources according to preference (i.e. failover to a lesser-preferred source in the event that a more-preferred source is unavailable).

This feature benefits from some of the functionality that was recently added to the Coingecko class:
 - Basic support for caching of prices.
 - Basic support for bundling token price requests.

This code can be independently used by bots, the frontend, and the backend Across API:
1. Bots can source their prices from the Across API, falling back to CoinGecko. This should eliminate the price retrieval warnings that we see ~several times per day, and which act as a blocker for enabling profitability controls.
2. The frontend _could_ use this for marginally greater robustness, and to abstract away the source of its price feeds (the caller must presently call specific functions to retrieve from Across API or Coingecko). At the moment we are still issuing some price requests directly to CoinGecko by accident.
3. The API is already using an older version of this, so the marginal benefit from using it is less.

The interface exposed by this feature maps very closely to our existing Coingecko API, so I expect that using it should be a small diff from our existing codebase. It's probably easiest (and delivers most benefit) to do this in the bots first.

Depending on whether this class is adopted, it might make sense to transition any additional functionality from the existing Coingecko class. There is however no requirement/need for us to do that.


src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,212 @@
...
+    const skipped: { [token: string]: string } = {}; // Includes reason for skipping
+
+    expected.forEach((addr: string) => {
+      const tokenPrice: TokenPrice | undefined = prices[addr.toLowerCase()];
+      const now: number = msToS(Date.now());
+
+      if (tokenPrice === undefined) {
+        skipped[addr] = "Not included in price feed response.";
+      } else if (tokenPrice.timestamp > now) {
+        skipped[addr] = `Token price timestamp is too new (timestamp ${tokenPrice.timestamp}).`;
```
pxrl, line 182: Missing test cases for these scenarios.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,212 @@
...
+export type TokenPrice = CoinGeckoPrice; // Temporary inversion; CoinGecko should source from here.
+
+// Represents valid source for spot prices (Across API, CoinGecko, ...)
+export interface PriceFeedAdapter {
+  readonly name: string;
+  getTokenPrice(address: string, currency: string, platform: string): Promise<TokenPrice>;
+  getTokenPrices(addresses: string[], currency: string, platform: string): Promise<TokenPrice[]>;
+}
+
+export type PriceCache = {
```
pxrl, line 15: The name PriceCache is a bit unfortunate because it's used to describe a dictionary containing TokenPrice types and has broader usage than in caching. Would like to find a better name for it.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,212 @@
...
+
+export { Logger } from "../relayFeeCalculator"; // Permit adapters to import a common definition.
+export type TokenPrice = CoinGeckoPrice; // Temporary inversion; CoinGecko should source from here.
+
+// Represents valid source for spot prices (Across API, CoinGecko, ...)
+export interface PriceFeedAdapter {
+  readonly name: string;
+  getTokenPrice(address: string, currency: string, platform: string): Promise<TokenPrice>;
+  getTokenPrices(addresses: string[], currency: string, platform: string): Promise<TokenPrice[]>;
+}
```
pxrl, line 13: This interface does not define any support for historical price feeds. The existing Coingecko classes from UMAprotocol/sdk and from this package _do_ support historical price feeds, although I don't know if anyone uses it within Across.

Might be nice to define an optional method that supports retrieving historical prices; I presume only CoinGecko would end up implementing it, so the PriceClient would need to support that only a subset of PriceFeeds can actually support historical prices.

In the event that historical prices are supported, I don't envision implementing any support for caching, because it's such a bespoke use case that caching isn't likely to be used.

pxrl:
Test results as at the current commit:
```
Test Suites: 5 skipped, 2 passed, 2 of 7 total
Tests:       20 skipped, 16 passed, 36 total
Snapshots:   0 total
Time:        7.513s
Ran all test suites with tests matching "priceclient".
Done in 9.99s.
```

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!this.platforms.includes(platform)) throw Error(`Platform ${platform} not supported.`);
+
+    // Note: Assume price is 60 seconds old since the API does not provide a timestamp.
+    const now = msToS(Date.now()) - 60;
+    const acrossPrice: AcrossPrice = await this.query(address, currency);
+
+    return { address: address, price: acrossPrice.price, timestamp: now };
+  }
+
+  // todo: Support bundled prices in the API endpoint.
```
pxrl, line 25: Need to highlight this comment. If we don't implement bundling in the Across API endpoint then we'll probably end up issuing _n_ requests on each price lookup because the PriceClient assumes that the configured PriceFeed supports bundling.

Fortunately, given that we're dogfooding so much here it should be achievable to support bundling via the existing Coingecko implementation, or alternatively via the PriceClient itself (inception...).

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  async getTokenPrices(addresses: string[], currency = "usd", platform = "ethereum"): Promise<TokenPrice[]> {
+    assert(this.priceFeeds.length > 0, "No price feeds were registerted.");
+    const priceCache: PriceCache = this.getPriceCache(currency, platform);
+
+    // Pre-populate price cache with requested token addresses
+    this.initPrices(priceCache, addresses);
+
+    const prices: PriceCache = await this.requestPrices(addresses, currency, platform);
```
pxrl, line 111: Might make sense to guard this by first checking all timestamps, and only proceeding to query if at least one of them are out of date.

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export class PriceFeed implements PriceFeedAdapter {
+  public readonly platforms = ["ethereum"];
+
+  constructor(public readonly logger: Logger, public readonly name: string, public readonly host?: string) {
+    // Allow host to be overridden for test or alternative deployments.
+    if (this.host === undefined) this.host = "across.to";
+  }
+
+  async getTokenPrice(address: string, currency: string, platform: string): Promise<TokenPrice> {
```
pxrl, line 15: Should be marked override (as with getTokenPrices() below, and repeated in coingecko.ts).

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    return prices;
+  }
+
+  private async query(token: string, baseCurrency: string, timeout?: number): Promise<AcrossPrice> {
+    const url = `https://${this.host}/api/coingecko?l1Token=${token}&baseCurrency=${baseCurrency}`;
+
+    try {
+      this.logger.debug({ at: "Across#query", message: "Querying api/coingecko.", query: url });
+      const result = await axios(url, { timeout });
```
pxrl, line 43: Probably need a way to tune the timeout here.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,212 @@
...
+    assert(age >= 0);
+    this.logger.debug({
+      at: "PriceClient#maxPriceAge",
+      message: `Setting maxPriceAge (S) ${this._maxPriceAge} => ${age}.`,
+    });
+    this._maxPriceAge = age;
+  }
+
+  // @todo: Could be nice to specify a priority rather than relying on order of addition.
+  addPriceFeed(priceFeed: PriceFeedAdapter): void {
```
daywiss, line 56: i think this dynamic adding of price feeds makes this code more complicated than it needs to be, does this need to be dynamic?  the alternative could be to just provide an array of price feeds in the constructor

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,212 @@
...
+    const cleared: string[] = this.priceFeeds.map((priceFeed: PriceFeedAdapter) => priceFeed.name);
+    this.priceFeeds = [];
+    this.logger.debug({
+      at: "PriceClient#clearPriceFeeds",
+      message: "Cleared all price feeds.",
+      cleared,
+    });
+  }
+
+  async getTokenPrice(address: string, currency = "usd", platform = "ethereum"): Promise<TokenPrice> {
```
pxrl, line 79: Missing a toBN() conversion somewhere in here; this will incorrectly return a floating-point price. Will add a comment to the tests to catch that as well.

src/priceClient/priceClient.e2e.ts
```diff
@@ -0,0 +1,153 @@
...
+  assert(tokenPrice.price > 0);
+  assert(tokenPrice.timestamp > 0);
+  assert(
+    tokenPrice.timestamp > timestamp - maxPriceAge,
+    `Surprise timestamp received: ${tokenPrice.timestamp} (expected at least ${timestamp - maxPriceAge}).`
+  );
+}
+
+// this requires e2e testing, should only test manually for now
+describe("PriceClient", function () {
```
pxrl, line 45: Should add a test that verifies the token price is returned in the correct numerical format. For example, querying WETH against ETH, or USDC/DAI against USD terms. It should then be OK to confirm that the price returned is accurate +/- 3% or so.

src/priceClient/priceClient.ts
pxrl: > i think the paramers here are really specific to coingecko, which is wierd to me if you are creating an interface for getting prices. coingeckos api is set up for getting price based on an address, but i dont know if thats commonly supported for other sources? I would expect most to be based on a symbol.

Yeah it's definitely true that CoinGecko has influenced the interface a lot - also because the attributes of the interface can be seen in the Across API coingecko endpoint, and those are the only two endpoints that are supported here. I think it's a good point and has also influenced the "platform", which is a CoinGecko concept. Referencing by ChainID and then mapping that to "ethereum", ... for CoinGecko would be more generic.

One challenge that has always bothered me about looking up via a token symbol is that they're not guaranteed to be globally unique. CoinGecko handles this by having their own internal mapping, and I believe it would be possible to dynamically query the correct/intended token by looking up tokens that share a specific symbol, and then find some second parameter to differentiate them on. It definitely risks ending up with a less-certain result though - so the advantage of using an address is that the caller takes full responsibility for identifying the correct token.

src/priceClient/priceClient.ts
pxrl: I think it could be done in the constructor, but since the constructor follows the singleton pattern initially established in the coingecko class, I'd be a little concerned about the impact that might have on the e2e test.

In the e2e test it's quite easy to throw away all existing price feeds in the beforeEach() callback, and then add new price feeds that are relevant to each test being executed. IMO that's quite nice and keeps everything fairly concise, but it's surely not the only way that it could be handled. Would you have an alternative suggestion?

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,50 @@
...
+    for (const address of addresses) {
+      const price: TokenPrice = await this.getTokenPrice(address, currency, platform);
+      prices.push(price);
+    }
+
+    return prices;
+  }
+
+  private async query(token: string, baseCurrency: string, timeout?: number): Promise<AcrossPrice> {
+    const url = `https://${this.host}/api/coingecko?l1Token=${token}&baseCurrency=${baseCurrency}`;
```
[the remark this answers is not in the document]
pxrl: The Across API adapter queries RL infra hosted by Vercel, which is incidentally supplied by CoinGecko (with the advantage of Vercel front-end caching + back-end price retention in the API. This is nice for the RL bots, which are rate-limited by CoinGecko many times per day.

So my proposal for implementation in the bots would probably be:
1. Across API
2. CoinGecko Free
3. CoinGecko Pro

This allows us to increase our usage of Vercel, whilst safely falling back to a reliable solution in the event that Vercel/Across API endpoint isn't working.

One thing to be aware of in this is that there is some potential for "buffer bloat" in the form of caching, so we'd want to look at the end-to-end caching implied by this setup.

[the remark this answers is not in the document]
daywiss: i see, that makes sense

src/priceClient/priceClient.ts
daywiss: personally id prefer simpler business logic to simpler tests, but in the case of testing you can have separate describe blocks for each configuration of the price feed, rather than combining them all into a single describe block. it would be about equivalent to what you have

src/priceClient/priceClient.ts
daywiss: im not suggesting to remove the lookup by token address, simply saying that the interface should be explicit with the function name that it is doing a lookup by token address. when someone who knows nothing about coingecko goes to implement this interface, i dont think they would intuitively think this would be a lookup by token address

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,212 @@
...
+    assert(this.priceFeeds.length > 0, "No price feeds are registered.");
+    const priceCache: PriceCache = this.getPriceCache(currency, platform);
+    const now: number = msToS(Date.now());
+
+    let tokenPrice: TokenPrice | undefined = priceCache[address];
+    const cacheMiss = tokenPrice === undefined || now - this.maxPriceAge > tokenPrice.timestamp;
+
+    if (this.maxPriceAge > 0) {
+      const age: number = tokenPrice ? now - tokenPrice.timestamp : Number.MAX_VALUE;
```
pxrl, line 88: Should probably be Number.MAX_SAFE_INTEGER.

src/priceClient/adapters/across.ts
pxrl: Have confirmed that override is used for extending classes, not implementing interfaces...so this is irrelevant.

src/priceClient/priceClient.ts
pxrl: Followed up on this point - the toBN() conversion occurs in the relayer ProfitClient, and it expected to receive a floating point price from CoinGecko. It should be OK to proceed with the current logic, as long as there is consensus on this implementation handling floating-point prices. Each adapter may eventually need to implement its own normalisation on the price format, though.

src/priceClient/adapters/across.ts
pxrl: Have implemented this + tests in the latest version of the PR.

src/priceClient/priceClient.ts
pxrl: Fair comment :) I've updated the PR with method names:

`getPriceByAddress()`
`getPricesByAddress()`

src/priceClient/priceClient.ts
pxrl: I don't really see the runtime addition of price feeds as being overly complex (the methods are quite small), but I concede that this is a feature that would probably only be used in tests. I looked at adding separate describe() blocks for each instance here, but it would have required a lot of additional boilerplate code. So I have instead implemented your change and have migrated away from the singleton pattern, since it doesn't seem right that the caller of get() will supply an array of price feeds but has no idea whether or not they will actually be used.

IMO this change does fortunately make the code simpler and I'm happy with the result, since I don't think we would actually need the singleton pattern in our usage of this.

src/priceClient/priceClient.e2e.ts
pxrl: Done.

pxrl:
Not much more I can do on this one for now, and there's plenty of other things to do.

I think we could quite easily take this into use in the bots, and I expect that would probably eliminate our price lookup failures.

Happy to hear what others think about it.

src/priceClient/adapters/index.ts
```diff
@@ -0,0 +1,2 @@
+export * as across from "./across";
```
mrice32, line 1: Minor note for the future (nothing to change here!): I know the rest of the sdk works this way, but namespacing all the components like this may have a negative impact on tree-shaking, as I'm not sure webpack can infer which parts of the application are used if they are all exported under namespaces.

In general, we may want to reconsider the approach unless we are confident it won't stop webpack from dropping unused symbols.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,187 @@
...
+        at: "PriceClient#getPriceByAddress",
+        message: `Cache ${cacheMiss ? "miss" : "hit"} (${currency}) for token ${address}.`,
+        age: `${age} S`,
+        price: tokenPrice,
+      });
+    }
+
+    if (cacheMiss) {
+      const prices: TokenPrice[] = await this.getPricesByAddress([address], currency);
+      tokenPrice = prices[0];
```
mrice32, line 77: Just if you want to avoid a temp variable.
```suggestion
      ([tokenPrice] = await this.getPricesByAddress([address], currency));
```

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+    return { address, price: tokenPrice.price, timestamp: tokenPrice.timestamp };
+  }
+
+  async getPricesByAddress(addresses: string[], currency = "usd"): Promise<TokenPrice[]> {
```
mrice32, line 82: I'm a bit confused about how these methods are laid out. They are all public, and it isn't indicated if any are _not_ supposed to be called by a user.

- getPriceByAddress
  - Checks the cache, returns the cache value if valid.
  - If cache is empty, calls getPricesByAddress to get the price
  - Returns the requested price.
- getPricesByAddress
  - Requests the price from the cache and the source.
  - Updates the cache with the source value.
  - Returns the requested prices.

Is getPricesByAddress not expected to be called by an external user? It feels like it's being used as an internal utility function. If so, it should probably be private or protected. However, both seem useful for a user. I could imagine a different setup where getPricesByAddress does all the cache checking and logging and getPriceByAddress is just a convenience function to forward the call.

mrice32 (commented):
Looks great! A few minor comments. I really like this setup.

src/priceClient/adapters/index.ts
pxrl: Interesting - had no idea about this. Is it the `export *` or `as across` that causes issues?

I _think_ we can actually get away without the `export *` statement, since we could just mark the PriceFeed class as the default export. It's still nice to have that exported under the `across` namespace though, because the `PriceFeed` name is reused in the coingecko class (i.e. `across.PriceFeed`, `coingecko.PriceFeed`).

src/priceClient/priceClient.ts
pxrl: The intention is for both to be callable by the user, but it's true that only the singular form (`getPriceByAddress`) checks for a pre-existing valid price. I have an open observation about there [here](https://github.com/across-protocol/sdk-v2/pull/98#discussion_r968995512).

As you've described, it does seem better to migrate the cache liveness check to the plural form (`getPricesByAddress`) and implement a loop over the requested addresses to see whether they all have "live" prices. I'd avoided doing this because I didn't like implementing the extra loop, but it feels awkward that the two methods behave differently in this respect.

pxrl: @mrice32 Per the current version of this PR, I think we should be OK on this point @ :+1:.

src/priceClient/adapters/index.ts
pxrl: @mrice32 To pick up on this one, do we want to note this for future consideration in a Linear issue?

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  private updateCache(priceCache: PriceCache, prices: PriceCache, expected: string[]): void {
+    const updated: TokenPrice[] = [];
+    const skipped: { [token: string]: string } = {}; // Includes reason for skipping
+
+    expected.forEach((address: string) => {
+      const addr = address.toLowerCase(); // for internal priceCache lookup.
+      const tokenPrice: TokenPrice | undefined = prices[addr];
+      const now: number = msToS(Date.now());
```
pxrl, line 152: This is probably dumb, since there's not likely to be any meaningful time difference between loops.

pxrl: Fixed: https://github.com/across-protocol/sdk-v2/pull/98/commits/53b8f48bd5af77693d1ea4e1a4655d627d1ee36d

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,60 @@
+import assert from "assert";
```
mrice32, line 1: nit: might be more appropriate to call this file across-api or something. Just so it's more obvious that we're getting the prices from the _api_.

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  get timeout(): number {
+    return this._timeout;
+  }
+
+  set timeout(timeout: number) {
+    assert(timeout >= 0);
+    this._timeout = timeout;
+  }
```
mrice32, line 23: We don't use getters and setters enough. Good idea.

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,60 @@
...
+  // todo: Support bundled prices in the API endpoint.
+  async getPricesByAddress(addresses: string[], currency = "usd"): Promise<TokenPrice[]> {
+    const prices: TokenPrice[] = [];
+    for (const address of addresses) {
+      const price: TokenPrice = await this.getPriceByAddress(address, currency);
+      prices.push(price);
+    }
+
+    return prices;
+  }
```
mrice32, line 42: I think this code is a bit more idiomatic, and it also has the advantage of making the calls in parallel.

```suggestion
 getPricesByAddress(addresses: string[], currency = "usd"): Promise<TokenPrice[]> {
    return Promise.all(addresses.map(address => this.getPriceByAddress(address, currency)));
  }
```

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,60 @@
...
+    const args = {
+      timeout,
+      params: {
+        l1Token: token,
+        baseCurrency: baseCurrency,
+      },
+    };
+
+    const result = await axios(url, args).catch((err) => {
+      const errMsg: string = err instanceof AxiosError ? err.message : "unknown error";
```
mrice32, line 55: nit: any reason why we can't just do err.toString() in the else case?

src/priceClient/adapters/across.ts
```diff
@@ -0,0 +1,60 @@
...
+    const prices: TokenPrice[] = [];
+    for (const address of addresses) {
+      const price: TokenPrice = await this.getPriceByAddress(address, currency);
+      prices.push(price);
+    }
+
+    return prices;
+  }
+
+  private async query(token: string, baseCurrency = "usd", timeout: number = defaultTimeout): Promise<AcrossPrice> {
```
mrice32, line 44: nit: I don't think we need the default "usd" here, since the method is private and the functions that call this one already do the defaulting.

This is more of a stylistic thing, but I think it's better to eliminate code-paths that are never used.

src/priceClient/adapters/index.ts
mrice32: @pxrl I'm not 100% sure about how tree-shaking works.

My basic understanding is that importing like this allows webpack to shake all parts of the imported code that isn't MyClass:
```ts
import { MyClass } from "@myscope/mypackage";

...

new MyClass();
```

Alternatively, if MyClass, is wrapped up in a namespace with 10 other classes, you would be forced to do this:

```ts
import { mynamespace } from "@myscope/mypackage";

...

new mynamespace.MyClass();
```

I think that would mean that all of mynamespace would be bundled because webpack can't (or just isn't sophisticated enough) to know which part you're going to use.

This is probably something we should test before making any changes, but I know that tree-shaking is very sensitive to things like this.

src/priceClient/adapters/coingecko.ts
```diff
@@ -0,0 +1,68 @@
...
+    return this._timeout;
+  }
+
+  set timeout(timeout: number) {
+    assert(timeout >= 0);
+    this._timeout = timeout;
+  }
+
+  async getPriceByAddress(address: string, currency = "usd"): Promise<TokenPrice> {
+    const price: TokenPrice[] = await this.getPricesByAddress([address], currency);
```
mrice32, line 34: nit: I don't think this type annotation is necessary.

src/priceClient/adapters/coingecko.ts
```diff
@@ -0,0 +1,68 @@
...
+      params: {
+        contract_addresses: addresses.join(","),
+        vs_currencies: baseCurrency,
+        include_last_updated_at: true,
+        x_cg_pro_api_key: this.apiKey ?? "",
+      },
+    };
+
+    const result = await axios(url, args).catch((err) => {
+      const errMsg: string = err instanceof AxiosError ? err.message : "unknown error";
```
mrice32, line 63: Same question as above.

src/priceClient/adapters/coingecko.ts
```diff
@@ -0,0 +1,68 @@
...
+    timeout: number = defaultTimeout
+  ): Promise<CoinGeckoPriceResponse> {
+    const url = `https://${this.host}/api/v3/simple/token_price/ethereum`;
+    const args = {
+      timeout,
+      params: {
+        contract_addresses: addresses.join(","),
+        vs_currencies: baseCurrency,
+        include_last_updated_at: true,
+        x_cg_pro_api_key: this.apiKey ?? "",
```
mrice32, line 58: Is specifying an empty key the same as not specifying a key? Does axios treat these the same? Does CG care?

src/priceClient/adapters/coingecko.ts
```diff
@@ -0,0 +1,68 @@
...
+    const prices: CoinGeckoPriceResponse = await this.query(addresses, currency);
+    return addresses.map((addr: string) => {
+      const price: CoinGeckoTokenPrice = prices[addr.toLowerCase()];
+      return { address: addr, price: price[currency], timestamp: price.last_updated_at };
+    });
+  }
+
+  private async query(
+    addresses: string[],
+    baseCurrency = "usd",
```
mrice32, line 48: Same comment as above, I suspect this default isn't necessary.

src/priceClient/adapters/coingecko.ts
```diff
@@ -0,0 +1,68 @@
...
+    timeout: number = defaultTimeout
```
mrice32, line 49: Does this function not respect the configured timeout in the class?

mrice32 (commented):
Looks great! Flushing some comments -- will take a look at priceClient next!

src/priceClient/adapters/across.ts
pxrl: I'd thought about this, but wasn't sure whether it was a good idea to hit the same host _n_ times concurrently. For our load levels it's probably OK, but I think it'd be better if we updated the API to accept multiple prices in a single request.

I was also curious as to whether this might trigger ~_n_ parallel DNS lookups, but I tested it quickly and it looks like the DNS resolver cache in my local machine seems to recognise that it already has queries in flight and only makes a single request...so I think it's probably fine.

In any case, I've incorporated this change because I guess it'll be fine.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,179 @@
...
+    const now: number = msToS(Date.now());
+    const missed: { [address: string]: number } = {};
+    addresses.forEach((address: string) => {
+      const addr = address.toLowerCase();
+      let tokenPrice: TokenPrice | undefined = priceCache[addr];
+      if (tokenPrice === undefined) {
+        tokenPrice = priceCache[addr] = { address: "unused", price: 0, timestamp: 0 };
+      }
+
+      const age: number = tokenPrice ? now - tokenPrice.timestamp : Number.MAX_SAFE_INTEGER;
```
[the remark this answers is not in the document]
pxrl: Hm, I don't think this is a 1:1 replacement. Line 73 of the change builds in assignment of priceCache[addr]:

```tokenPrice = priceCache[addr] = { address: "unused", price: 0, timestamp: 0 };```

File: src/priceClient/adapters/across.ts (hunk 20 above)
Comment (pxrl) [reply src/priceClient/adapters/across.ts:49]: Just my ignorance about the object that is provided during an exception. My impression is that it sometimes seems like an amorphous blob (!!), so I wasn't confident that I'd be able to interpret it reliably.

File: src/priceClient/adapters/coingecko.ts (hunk 24 above)
Comment (pxrl) [reply src/priceClient/adapters/coingecko.ts]: It works fine; it goes to a different endpoint when the API key is not defined, and that endpoint just ignores it. It is however a bit lazy. Fixed in this commit: https://github.com/across-protocol/sdk-v2/pull/98/commits/fb5780a3c40637f17247612e7bbfc988319d3b9a

File: src/priceClient/adapters/coingecko.ts (hunk 25 above)
Comment (pxrl) [reply src/priceClient/adapters/coingecko.ts]: This has been removed in this commit: https://github.com/across-protocol/sdk-v2/pull/98/commits/fb5780a3c40637f17247612e7bbfc988319d3b9a

The `query()` function has been moved to baseAdapter.ts and is now generic, so it doesn't know anything about the arguments that it's bundling.

File: src/priceClient/adapters/coingecko.ts (hunk 26 above)
Comment (pxrl) [reply src/priceClient/adapters/coingecko.ts]: This has now been consolidated as part of this commit: https://github.com/across-protocol/sdk-v2/pull/98/commits/fb5780a3c40637f17247612e7bbfc988319d3b9a

I'd probably like to have the theoretical ability to override this per-call, but we don't have a need for it right now.

File: src/priceClient/adapters/coingecko.ts (hunk 23 above)
Comment (pxrl) [reply src/priceClient/adapters/coingecko.ts]: Have consolidated the query() function into baseAdapter.ts now, so we only have a single implementation. It now tries to produce a string from err if the error is something other than an AxiosError. The change is here: https://github.com/across-protocol/sdk-v2/pull/98/commits/fb5780a3c40637f17247612e7bbfc988319d3b9a

File: src/priceClient/adapters/across.ts (hunk 21 above)
Comment (pxrl) [reply src/priceClient/adapters/across.ts]: The various `query()` implementations have now been merged into a single implementation in baseAdapter.ts, and they're agnostic to the params being passed to them. So this is now a non-issue :) The commit is here: https://github.com/across-protocol/sdk-v2/pull/98/commits/fb5780a3c40637f17247612e7bbfc988319d3b9a

File: src/priceClient/adapters/across.ts (hunk 20 above)
Comment (pxrl) [reply src/priceClient/adapters/across.ts]: Has been implemented in this commit btw: https://github.com/across-protocol/sdk-v2/pull/98/commits/fb5780a3c40637f17247612e7bbfc988319d3b9a

File: src/priceClient/adapters/across.ts (hunk 17 above)
Comment (pxrl) [reply src/priceClient/adapters/across.ts]: Implemented here: https://github.com/across-protocol/sdk-v2/pull/98/commits/1eb2eda88fdc8ccffa543e78e03d8e1b7bb3baae

File: src/priceClient/priceClient.ts [hunk 28]
```diff
@@ -0,0 +1,179 @@
...
+    for (const priceFeed of this.priceFeeds) {
+      this.logger.debug({
+        at: "PriceClient#getPriceByAddress",
+        message: `Looking up prices via ${priceFeed.name}.`,
+      });
+      try {
+        prices = await priceFeed.getPricesByAddress(addresses, currency);
+        if (prices.length === 0) {
+          throw Error("Zero-length response received");
+        }
```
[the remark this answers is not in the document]
pxrl: As an extension to this work, I think it'd be cool to implement per-token timestamp validation, so we can iterate through the various feeds with an (hopefully) reducing set of addresses, until we've obtained "live" prices for all requested tokens.

Background: The CoinGecko Free tier has been providing prices that are 2 - 4 hours old all day, so being served outdated prices is actually a real possibility.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,179 @@
...
+    });
+    this._maxPriceAge = age;
+  }
+
+  listPriceFeeds(): string[] {
+    return this.priceFeeds.map((priceFeed: PriceFeedAdapter) => priceFeed.name);
+  }
+
+  async getPriceByAddress(address: string, currency = "usd"): Promise<TokenPrice> {
+    const tokenPrices: TokenPrice[] = await this.getPricesByAddress([address], currency);
```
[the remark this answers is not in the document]
pxrl: Addressed in this commit: https://github.com/across-protocol/sdk-v2/pull/98/commits/609cc2a3314c369683dce3631c65f8cb2939c1ab

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,179 @@
...
+    });
+
+    if (Object.keys(missed).length > 0) {
+      const requestAddresses = Object.keys(priceCache);
+      this.logger.debug({
+        at: "PriceClient#getPricesByAddress",
+        message: `${currency.toUpperCase()} cache miss (age > ${this.maxPriceAge} S).`,
+        tokens: missed,
+      });
+      const prices: PriceCache = await this.requestPrices(requestAddresses, currency);
```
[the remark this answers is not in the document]
pxrl: Addressed in this commit: https://github.com/across-protocol/sdk-v2/pull/98/commits/609cc2a3314c369683dce3631c65f8cb2939c1ab

src/priceClient/priceClient.ts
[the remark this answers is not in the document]
pxrl: Addressed in this commit: https://github.com/across-protocol/sdk-v2/pull/98/commits/609cc2a3314c369683dce3631c65f8cb2939c1ab

I will add a comment to the modification to make sure we are sure that it's not insane.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,177 @@
...
+    assert(this.priceFeeds.length > 0, "No price feeds are registered.");
+    const priceCache = this.getPriceCache(currency);
+
+    // Determine whether *all* prices are current.
+    const now = msToS(Date.now());
+    const missed: { [address: string]: number } = {};
+    addresses.forEach((address: string) => {
+      const addr = address.toLowerCase();
+      // note: If priceCache[addr] is undefined, it is assigned here.
+      const tokenPrice = (priceCache[addr] = priceCache[addr] ?? ({ price: 0, timestamp: 0 } as TokenPrice));
```
pxrl, line 72: @mrice32 Tagging this for consideration. The code relies on priceCache[addr] being assigned to a default value if it does not already exist. This is a concise way of doing it, but it does a reassignment of an object member that may already be assigned, and it comes with some risk of being misunderstood and unintentionally broken in future changes. I tried to call that out with a comment.

I also don't think prettier did any favours with its addition of brackets. The formulation works, but it's logically counter-intuitive for me.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,179 @@
...
+// TokenPrice object when returning to the caller.
+export type PriceCache = {
+  [address: string]: TokenPrice;
+};
+
+export function msToS(ms: number): number {
+  return Math.floor(ms / 1000);
+}
+
+export class PriceClient implements PriceFeedAdapter {
```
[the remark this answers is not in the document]
pxrl: Working to add a README.md now (as is the precedent set by all other modules here). Will have it up on Friday :+1:.

src/priceClient/adapters/acrossApi.ts
```diff
@@ -0,0 +1,43 @@
...
+      l1Token: address,
+      baseCurrency: currency,
+    };
+
+    const now = msToS(Date.now()) - 60; // Assume price is 60 seconds old.
+    const response: unknown = await this.query("api/coingecko", queryArgs);
+    if (!this.validateResponse(response))
+      throw new Error(`Unexpected ${this.name} response: ${JSON.stringify(response)}`);
+
+    return { address: address, price: response.price, timestamp: now };
```
pxrl, line 24: nb. Not sure if this completely hits the mark for the intended use of a predicate. At no point does the linter or tsc appear to complain about a lack of typing for response on line 24...I'm not _totally_ confident though.

src/priceClient/adapters/acrossApi.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  // todo: Support bundled prices in the API endpoint.
+  async getPricesByAddress(addresses: string[], currency = "usd"): Promise<TokenPrice[]> {
+    return Promise.all(addresses.map((address) => this.getPriceByAddress(address, currency)));
+  }
+
+  private validateResponse(response: unknown): response is AcrossPrice {
+    if (typeof response !== "object") return false;
+
+    // prettier-ignore
```
pxrl, line 35: Prettier insists on re-orging the && to the rhs of each line, which (IMO) makes the conditional logic less obvious. Seems to be a lot of people complaining that prettier doesn't allow this to be configured.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,177 @@
...
+    const priceCache = this.getPriceCache(currency);
+
+    // Determine whether *all* prices are current.
+    const now = msToS(Date.now());
+    const missed: { [address: string]: number } = {};
+    addresses.forEach((address: string) => {
+      const addr = address.toLowerCase();
+      // note: If priceCache[addr] is undefined, it is assigned here.
+      const tokenPrice = (priceCache[addr] = priceCache[addr] ?? ({ price: 0, timestamp: 0 } as TokenPrice));
+      const age = tokenPrice ? now - tokenPrice.timestamp : Number.MAX_SAFE_INTEGER;
```
mrice32, line 73: nit: I don't think this can ever be undefined, so you should be able to just do:
```ts
const age = now - tokenPrice.timestamp;
```

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,177 @@
...
+        at: "PriceClient#getPricesByAddress",
+        message: `${currency.toUpperCase()} cache miss (age > ${this.maxPriceAge} S).`,
+        tokens: missed,
+      });
+      const prices = await this.requestPrices(requestAddresses, currency);
+      if (Object.keys(prices).length === 0) {
+        this.logger.warn({ at: "PriceClient#getPricesByAddress", message: "Failed to update token prices." });
+        // @todo throw ?
+        return [];
+      }
```
mrice32, line 92: How could this ever happen? It looks like the requestPrices function always throws if the lookup fails (or returns 0 results), right?

Also, do we prefer to warn or throw an uncaught error? requestPrices throws, but this function just seems to want to warn.

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,177 @@
...
+        if (prices.length === 0) {
+          throw Error("Zero-length response received");
+        }
+      } catch (err) {
+        this.logger.debug({
+          at: "PriceClient#requestPrices",
+          message: `Price lookup against ${priceFeed.name} failed (${err}).`,
+          tokens: addresses,
+        });
+        continue; // Failover to the next price feed...
```
mrice32, line 132: nit: is this continue redundant? Won't it loop anyway?

src/priceClient/priceClient.ts
```diff
@@ -0,0 +1,177 @@
...
+      const tokenPrice: TokenPrice | undefined = prices[addr];
+
+      if (tokenPrice === undefined) {
+        skipped[address] = "Not included in price feed response.";
+      } else if (tokenPrice.timestamp > now) {
+        skipped[address] = `Token price timestamp is too new (timestamp ${tokenPrice.timestamp}).`;
+      } else if (tokenPrice.timestamp >= priceCache[addr].timestamp) {
+        const { price, timestamp } = tokenPrice;
+        // @todo: Do we care if the token price is older than maxPriceAge?
+        priceCache[addr] = { address: "unused", price: price, timestamp: timestamp };
```
mrice32, line 159: Why are we assigning the address field "unused" for a real value? Why does this field exist?

mrice32 (approved):
Looks great!

src/priceClient/adapters/acrossApi.ts
```diff
@@ -0,0 +1,43 @@
...
+    return { address: address, price: response.price, timestamp: now };
+  }
+
+  // todo: Support bundled prices in the API endpoint.
+  async getPricesByAddress(addresses: string[], currency = "usd"): Promise<TokenPrice[]> {
+    return Promise.all(addresses.map((address) => this.getPriceByAddress(address, currency)));
+  }
+
+  private validateResponse(response: unknown): response is AcrossPrice {
+    if (typeof response !== "object") return false;
```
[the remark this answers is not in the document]
pxrl: Ah...yes, and no - as it turns out. tsc seems unwilling or unable to progressively update its world-view after each of these statements are progressivefully fulfilled.

The code below produces the following error: `adapters/acrossApi.ts(39,21): semantic error TS2531: Object is possibly 'null'.` This seems pretty counter-intuitive to me; I get the feeling I am asking too much of tsc, and that I might in the land of tsc implementation details and quirks.

```
 33     // if (typeof response !== "object") return false;
 34 
 35     // prettier-ignore
 36     return (
 37       response !== null
 38       && typeof response === "object"
 39       && "price" in response
 40       && typeof (response as { [key: string]: unknown }).price === "number"
 41     );
```

I'll remove the check on `typeof response === object` on line 38 instead.

src/priceClient/priceClient.ts
pxrl: Yeah; and there's no way to break out of the loop when a successful retrieval occurs. That's pretty critical...

src/priceClient/priceClient.ts
pxrl: > Why are we assigning the address field "unused" for a real value? Why does this field exist?

All our existing code is built around a CoinGeckoPrice type, which follows this type definition:
```
export type CoinGeckoPrice = {
  address: string;
  timestamp: number;
  price: number;
};
```
I'm reusing that type for convenience/simplicity in the code, but don't have any need for the address member because I map the prices by their address, instead of storing them in a list or set. I prefer this since I don't like searching each time I need to retrieve a price.

So...the address member is superfluous to my needs, but it means I don't need to define a new type that has only one difference to the existing CoinGeckoPrice type that we already depend on.

src/priceClient/priceClient.ts
pxrl: Have flip-flopped on this a bit...In conclusion I think it probably makes sense to throw if all lookups failed. That is actually what has been implemented, because there's a test that validates that. So the log message and comment were both redundant here.

src/priceClient/priceClient.ts
pxrl: Implemented here: https://github.com/across-protocol/sdk-v2/pull/98/commits/0138242162a90c2574b75b0eab27ee93118fc413

src/priceClient/adapters/coingecko.ts
pxrl: dfgdfg

src/priceClient/priceClient.ts
pxrl: True - fixed: https://github.com/across-protocol/sdk-v2/pull/98/commits/0138242162a90c2574b75b0eab27ee93118fc413

src/priceClient/priceClient.ts
mrice32: Oh, I would use a different type in this situation! You don't have to change in this PR, but I think if that field is just not being used, people will extract it and think that it has some purpose or meaning.

src/priceClient/priceClient.ts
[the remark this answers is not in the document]
pxrl: It's a bit more than a one-liner, but I added some explanation in this commit: https://github.com/across-protocol/sdk-v2/pull/98/commits/c41d850034e71b5af114294be13fd507f22d00fe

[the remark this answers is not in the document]
mrice32: Sounds good! This explanation is great, so definitely leave as is. All I was really hoping for was something like "This class serves as a fallback price provider and cache to make requests as reliably and sparingly as possible".

PR #102, opened 2022-09-27
improve(utils): Improve gas price estimation

pxrl:
Testing has shown that eth_getPrice() consistently returns a value that excludes the chain priority fee. This means that gas price estimation can be lower than it should be on chains where the priority fee is non-zero. In turn, this means that the Across front end and bots are underestimating gas costs for individual transactions.

This change instead uses the ethers helper getFeeData() call in order to determine the gas price (note that ethers hardcodes an assumed 1.5 Gwei for maxPriorityFeePerGas). The impact of this change is _mostly_ felt on Ethereum. Arbitrum, Optimism and Boba are not impacted for various reasons. Polygon's gas price estimation is another special case, and this change does bump the price by 1.5 Gwei MATIC. Investigation shows that we are probably still estimating incorrectly for Polygon, but the costs on Polygon are so low that the impact is minute...so it's not a high priority to correct this.

Note that the use of feeData.gasPrice is a temporary workaround until a future upgrade to ethers v5.7.x (or newer), at which point feeData.lastBaseFeePerGas will become available. Testing shows that lastBaseFeePerGas should be more reliable than gasPrice, and the commit message for the addition of lastBaseFeePerGas indicates that it is intended for custom estimation. The field feeData.maxFeePerGas is deliberately not used for gas cost computation because it will typically to over-estimate the gas price (i.e. 2x feeData.gasPrice) and will therefore artificially inflate the projected gas cost for a transaction.

For follow-up work, I would propose that the code be adapted to support per-chain gas price oracles. For example, Polygon recommend bypassing the provider and querying a specific endpoint instead, and it seems that over time (with additional chains) it may be tricky to maintain generic behaviour across all chains.

This work was prompted by the observations of Across community member _GR_.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -247,22 +247,55 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
     "Gas Markup must be within the range of (-1.0, +4.0] so that total gas multiplier is between (0, +5.0]"
   );
   const gasTotalMultiplier = 1.0 + gasMarkup;
+  const network: providers.Network = await provider.getNetwork(); // Served locally by StaticJsonRpcProvider.
   const voidSigner = new VoidSigner(senderAddress, provider);
-  // Verify if this provider has been L2Provider wrapped
-  // NOTE: In this case, this will be true if the provider is
-  //       using the Optimism blockchain
-  if (isOptimismL2Provider(provider)) {
+
+  if ([10].includes(network.chainId)) {
+    // Verify that this provider has been L2Provider wrapped.
+    assert(isOptimismL2Provider(provider), `Unexpected provider for chain ID ${network.chainId}.`);
+    assert(gasPrice === undefined, `Gas price (${gasPrice}) supplied for Optimism gas estimation (unused).`);
     const populatedTransaction = await voidSigner.populateTransaction(unsignedTx);
     return (await provider.estimateTotalGasCost(populatedTransaction)).mul(gasTotalMultiplier).toString();
-  } else {
-    // Estimate the Gas units required to submit this transaction
-    const estimatedGasUnits = await voidSigner.estimateGas(unsignedTx);
-    // Provide a default gas price of the market rate if this condition has not been set
-    const resolvedGasPrice = gasPrice ?? (await provider.getGasPrice());
-    // Find the total gas cost by taking the product of the gas
-    // price & the estimated number of gas units needed
-    return BigNumber.from(resolvedGasPrice).mul(gasTotalMultiplier).mul(estimatedGasUnits).toString();
   }
+
+  // Estimate the Gas units required to submit this transaction
+  const estimatedGasUnits = await voidSigner.estimateGas(unsignedTx);
+  let resolvedGasPrice: BigNumberish;
+
+  // Use a fixed gas price (if supplied), else query the provider.
+  if (gasPrice) {
+    resolvedGasPrice = gasPrice;
+  } else {
+    // Retrieve feeData; sub in artificially high defaults if anything fails.
+    const feeData: providers.FeeData = await provider
+      .getFeeData()
+      .then((response: providers.FeeData) => {
+        Object.entries(response).forEach(([key, value]) => {
+          if (!BigNumber.isBigNumber(value)) throw new Error(`Unexpected FeeData entry (${key}: ${value})`);
+        });
+        return response;
+      })
+      .catch(() => {
+        return {
+          gasPrice: toBN(Number.MAX_SAFE_INTEGER),
+          maxFeePerGas: toBN(Number.MAX_SAFE_INTEGER),
+          maxPriorityFeePerGas: toBN(0),
+          // lastBaseFeePerGas: toBN(Number.MAX_SAFE_INTEGER), ethers v5.7.x
+        };
+      });
+
+    // @todo: Update to ethers 5.7.x to access FeeData.lastBaseFeePerGas
+    // https://github.com/ethers-io/ethers.js/commit/8314236143a300ae81c1dcc27a7a36640df22061
+    resolvedGasPrice = toBN(feeData.gasPrice || Number.MAX_SAFE_INTEGER);
+    // Arbitrum One ignores maxPriorityFeePerGas.
+    if (![42161].includes(network.chainId)) {
+      resolvedGasPrice = resolvedGasPrice.add(feeData.maxPriorityFeePerGas || 0);
+    }
+  }
+
+  // Find the total gas cost by taking the product of the gas price & the
+  // estimated number of gas units needed.
+  return BigNumber.from(resolvedGasPrice).mul(gasTotalMultiplier).mul(estimatedGasUnits).toString();
 }
 
 /**


src/utils.ts
```diff
@@ -247,22 +247,55 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
   );
   const gasTotalMultiplier = 1.0 + gasMarkup;
+  const network: providers.Network = await provider.getNetwork(); // Served locally by StaticJsonRpcProvider.
   const voidSigner = new VoidSigner(senderAddress, provider);
-  // Verify if this provider has been L2Provider wrapped
-  // NOTE: In this case, this will be true if the provider is
-  //       using the Optimism blockchain
-  if (isOptimismL2Provider(provider)) {
+
+  if ([10].includes(network.chainId)) {
```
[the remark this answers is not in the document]
nicholaspai: I don't think the function call adds any cost here (no external calls, hardly any extra memory cost). However, I think we should rename the [10] array so we can add to it in the future and understand the code a bit more. 

How would you describe this list @pxrl ?

src/utils.ts
```diff
@@ -247,22 +247,55 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+        return response;
+      })
+      .catch(() => {
+        return {
+          gasPrice: toBN(Number.MAX_SAFE_INTEGER),
+          maxFeePerGas: toBN(Number.MAX_SAFE_INTEGER),
+          maxPriorityFeePerGas: toBN(0),
+          // lastBaseFeePerGas: toBN(Number.MAX_SAFE_INTEGER), ethers v5.7.x
+        };
+      });
```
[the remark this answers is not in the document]
nicholaspai: In addition to what James suggested, is  it possible to compare `response` to the `FeeData` type upfront?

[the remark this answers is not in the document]
nicholaspai: Yeah this is a bit more in line with our existing style

src/utils.ts
```diff
@@ -247,22 +247,55 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+
+    // @todo: Update to ethers 5.7.x to access FeeData.lastBaseFeePerGas
+    // https://github.com/ethers-io/ethers.js/commit/8314236143a300ae81c1dcc27a7a36640df22061
+    resolvedGasPrice = toBN(feeData.gasPrice || Number.MAX_SAFE_INTEGER);
```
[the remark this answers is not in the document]
nicholaspai: Agreed! ?? only goes to the RHS if LHS is null or undefined

src/utils.ts
```diff
@@ -247,22 +247,55 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+    // Arbitrum One ignores maxPriorityFeePerGas.
+    if (![42161].includes(network.chainId)) {
```
nicholaspai, line 291: Same comment as https://github.com/across-protocol/sdk-v2/pull/102/files#r981256728 but I would suggest moving the special network cases (i.e. for Optimism and Arbitrum) as early in the function as possible to make it easier for the maintainer to edit and also so we can exit earlier

nicholaspai (commented):
I really appreciate the detailed PR description it made my review very easy and the code diff is small ! James and I left a first round of comments

src/utils.ts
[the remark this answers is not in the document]
pxrl: Do we accept a gas price of 0 though? My gut feel is that I'd want to reject that, since it's probably more indicative of an error.

src/utils.ts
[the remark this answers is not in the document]
pxrl: > Could we do something like:
> 
> ```
> const assertFn = (key, value) => 
>  Object.entries(response).forEach(([key, value]) => assert(BigNumber.isBigNumber(value), `Unexpected FeeData entry (${key}: ${value})`);
> 
> const modifiedResponse = response as { [key: string]: BigNumber };
> ```
> 
> Where we loop through the response and assert that each key/value is a BigNumber. Then, we use an index signature to cast the response object to an object where all key/value pairs are {string, BigNumber}

I implemented a variant on this. This is data received from outside, so I'd rather not use an assert() there. Instead, it throws on invalid data and we proceed with the defaults, which are insanely high and should cause any profitability calculation to evaluate false.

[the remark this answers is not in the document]
pxrl: I chose the .then().catch() construction because it allowed me a single point of assignment, so I could keep feeData const. I've now made it a try ... catch construction, after changing a few other things to make that feasible. It's cleaner and more concise now.

src/utils.ts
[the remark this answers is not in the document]
pxrl: Per the latest change, we'll now silently accept a value of 0. I'm still not sure how I feel about that, though :)

src/utils.ts
pxrl: This block is dependent on gasTotalMultiplier, network and voidSigner - so I think it's elevated as high as it can go.

pxrl: Per currently implementation, we're dependent on the value of resolvedGasPrice, and that's dependent on the retrieval of feeData. So it's hard to see how it could be moved up with the overall structure of the code.

src/utils.ts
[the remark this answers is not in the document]
nicholaspai: I think this suggestion makes most sense and allows us to scale better

src/utils.ts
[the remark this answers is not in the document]
nicholaspai: Hmm i guess gas price 0 is bad cuz it makes everything profitable... I'm now going back and forth and think we should reject 0

nicholaspai (commented):
Went through all threads and reviewed again

src/utils.ts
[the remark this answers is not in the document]
pxrl: Have implemented this now. It lacks tests; will add those tomorrow.

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,98 @@
+import { BigNumber, providers } from "ethers";
+import { toBN } from "../utils";
+
+export type GasPriceEstimate = {
+  maxFeePerGas: BigNumber;
+  maxPriorityFeePerGas?: BigNumber;
+};
+
+type ValidatedFeeData = {
+  [idx: string]: BigNumber;
```
pxrl, line 10: nb. this is annoying, but seems necessary to perform a string-based lookup into a ValidatedFeeObject.

src/utils.ts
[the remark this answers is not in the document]
pxrl: Per the latest update, we now ignore prices that are 0 and fall back to MAX_SAFE_INTEGER.

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,98 @@
...
+
+async function legacy(provider: providers.Provider): Promise<GasPriceEstimate> {
+  const feeData: ValidatedFeeData = await getProviderFeeData(provider);
+
+  return {
+    maxFeePerGas: feeData.gasPrice,
+  } as GasPriceEstimate;
+}
+
+// @todo: Update to ethers 5.7.x to access FeeData.lastBaseFeePerGas
```
mrice32, line 52: Why not go ahead and do this so we don't have to build a workaround?

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,98 @@
...
+    maxPriorityFeePerGas: toBN(Number.MAX_SAFE_INTEGER),
+    // lastBaseFeePerGas: toBN(Number.MAX_SAFE_INTEGER), @todo: ethers 5.7.x
+  };
+
+  try {
+    const response: providers.FeeData = await provider.getFeeData();
+    const expectedKeys: string[] = Object.keys(feeData);
+    Object.entries(response).forEach(([key, value]) => {
+      if (expectedKeys.includes(key) && BigNumber.isBigNumber(value) && value.gt(0)) {
+        feeData[key] = toBN(value);
```
mrice32, line 34: Isn't this already a BigNumber?

src/gasPriceOracle/oracle.ts
mrice32: Sorry, why is this necessary? What call fails?

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,98 @@
...
+interface GasPriceFeed {
+  (provider: providers.Provider): Promise<GasPriceEstimate>;
+}
+
+async function getProviderFeeData(provider: providers.Provider): Promise<ValidatedFeeData> {
+  // Start with safe defaults.
+  const feeData: ValidatedFeeData = {
+    gasPrice: toBN(Number.MAX_SAFE_INTEGER),
+    maxFeePerGas: toBN(Number.MAX_SAFE_INTEGER),
+    maxPriorityFeePerGas: toBN(Number.MAX_SAFE_INTEGER),
```
mrice32, line 25: Why do we default these to a large integer instead of making them optional?

mrice32 (commented):
Looks good -- a few minor comments

src/gasPriceOracle/oracle.ts
pxrl: If I remove the line (as follows), tsc bails out on me:

```
  9 type ValidatedFeeData = {
 10   // [idx: string]: BigNumber;
 11   gasPrice: BigNumber;
 12   maxFeePerGas: BigNumber;
 13   maxPriorityFeePerGas: BigNumber;
 14 };
```

```
✖ Failed to compile
(typescript) Error: /home/pxrl/dev/across-protocol/sdk-v2/src/gasPriceOracle/oracle.ts(34,9): semantic error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'ValidatedFeeData'.
  No index signature with a parameter of type 'string' was found on type 'ValidatedFeeData'.
```

The offending code is the `feeData[key]` lookup on line 32:

```
 29   try {
 30     const response: providers.FeeData = await provider.getFeeData();
 31     const expectedKeys: string[] = Object.keys(feeData);
 32     Object.entries(response).forEach(([key, value]) => {
 33       if (expectedKeys.includes(key) && BigNumber.isBigNumber(value) && value.gt(0)) {
 34         feeData[key] = toBN(value);
 35       }
 36     });
 37   }
```

So it seems like I am prevented from trying to lookup `feeData[key]` because the ValidatedFeeData type defines specific keys, and tsc can't quite recognise that I have already guaranteed that (indirectly) by validating that `key` exists in ValidatedFeeData.

src/gasPriceOracle/oracle.ts
pxrl: I _thought_ I needed to pipe it through toBN() because the FeeData definition permits all members to be null. Reflecting on that, it was obviously wrong because toBN() wouldn't accept a null anyway.

In any case, the logic on line 33 guarantees that it is a BigNumber, so I can actually remove toBN() here. Nice!

```
export interface FeeData {
    lastBaseFeePerGas: null | BigNumber;
    maxFeePerGas: null | BigNumber;
    maxPriorityFeePerGas: null | BigNumber;
    gasPrice: null | BigNumber;
}
```

src/gasPriceOracle/oracle.ts
pxrl: To me, it seems safer to default to a large number in order to ensure that a gas price resolution failure doesn't accidentally make a transaction look profitable. This has the advantage of being drop-in compatible with the calling code, but has the disadvantage of suppressing price lookup failures (could be mitigated if we add logging here). As I see it, the alternative would be to throw exceptions from this code in the event of gas price lookup failures, which is a more explicit communication to the caller that something went wrong.

Extrapolating the impact of a "default high" price into the two consumers of this functionality:
- In the bots, this should be fine because they only use this to filter out unprofitable transactions.
- On the front end, this will prevent users from placing transactions, and does indirectly mask the underlying error, which _might_ make Across look uncompetitive on the rare occasion that this does happen.

In terms of the selection of MAX_SAFE_INTEGER, it is just an easy constant to use here, but it doesn't have to be this specific constant. It does occur to me that it's wasteful to take the number MAX_SAFE_INTEGER, convert it to a string, and then convert it to a BigNumber.

If we choose to proceed with a high default, we'd probably also be fine to use 1 full unit of the underlying gas token (i.e. 1 ETH, 1 MATIC). That might become more complicated if we ever have a gas token that's not 18 decimals, but I guess that's unlikely.

src/gasPriceOracle/oracle.ts
pxrl: We seem to have a little bit of work to do there (https://github.com/across-protocol/sdk-v2/actions/runs/3135735485/jobs/5091770273#step:4:278), so I thought it best to decouple them.

The impact of proceeding as-is is that we will probably have a _minor_ over-estimation in total gas cost on some occasions - but it's within what we would plan to add with gas padding anyway. The necessary changes after upgrading to ethers 5.7.x are also pretty minor - should only be a couple of lines to modify in this file.

src/gasPriceOracle/oracle.ts
pxrl: nb The change implemented in this PR is a departure from the current norm, so I need to confirm what impact it might have in the FE (i.e. if the FE specifically looks for a 0 response then that might be an issue...).

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,92 @@
...
+  maxFeePerGas: BigNumber;
+  maxPriorityFeePerGas: BigNumber;
+};
+
+interface GasPriceFeed {
+  (provider: providers.Provider): Promise<GasPriceEstimate>;
+}
+
+async function getProviderFeeData(provider: providers.Provider): Promise<ValidatedFeeData> {
+  const validateFeeData = (val: BigNumber | null): BigNumber  => {
```
mrice32, line 20: Nit: why not pull this out of the function so it's not re-initialized on each call?

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,92 @@
...
+    return BigNumber.isBigNumber(val) && val.gt(0) ? val : toBN(Number.MAX_SAFE_INTEGER);
+  };
+
+  // Not much to do on a failed retrieval, so just suppress it.
+  const response: providers.FeeData = await provider.getFeeData().catch(() => {
+    return {} as providers.FeeData
+  });
+
+  const feeData: ValidatedFeeData = {
+    "gasPrice": validateFeeData(response["gasPrice"]),
```
mrice32, line 30: Nit: why did you quote these keys?

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,92 @@
...
+  return {
+    maxFeePerGas: feeData.gasPrice,
+  } as GasPriceEstimate;
+}
+
+// @todo: Update to ethers 5.7.x to access FeeData.lastBaseFeePerGas
+// https://github.com/ethers-io/ethers.js/commit/8314236143a300ae81c1dcc27a7a36640df22061
+async function eip1559(provider: providers.Provider): Promise<GasPriceEstimate> {
+  const feeData: ValidatedFeeData = await getProviderFeeData(provider);
+  const maxPriorityFeePerGas: BigNumber = feeData.maxPriorityFeePerGas;
```
mrice32, line 50: Is this type annotation necessary?

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  return {
+    maxPriorityFeePerGas: maxPriorityFeePerGas,
+    maxFeePerGas: maxPriorityFeePerGas.add(feeData.maxFeePerGas),
+  } as GasPriceEstimate;
```
mrice32, line 55: nit: why does this need to be casted? Shouldn't it conform to the return type you specified without a cast?

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  // Not much to do on a failed retrieval, so just suppress it.
+  const response: providers.FeeData = await provider.getFeeData().catch(() => {
+    return {} as providers.FeeData
+  });
+
+  const feeData: ValidatedFeeData = {
+    "gasPrice": validateFeeData(response["gasPrice"]),
+    "maxFeePerGas": validateFeeData(response["maxFeePerGas"]),
+    "maxPriorityFeePerGas": validateFeeData(response["maxPriorityFeePerGas"]),
```
mrice32, line 32: Does this mean that if tat RPC call fails, we default to an extremely high gas price? What are the implications of that? For the FE, would we want to show a crazy high fee or would we prefer to have the call fail and just fail to compute? What about other callers? I feel like throwing (unless we have a retry or fallback) would give the caller the most flexibility, no?

mrice32 (commented):
Looks great! A few comments

src/gasPriceOracle/oracle.ts
pxrl: Looks like a hangover from a previous iteration of producing the feeData object. Good call.

src/gasPriceOracle/oracle.ts
pxrl: > Does this mean that if tat RPC call fails, we default to an extremely high gas price? What are the implications of that? For the FE, would we want to show a crazy high fee or would we prefer to have the call fail and just fail to compute? What about other callers? I feel like throwing (unless we have a retry or fallback) would give the caller the most flexibility, no?

All good questions :) Some of them were raised (without conclusion) here: https://github.com/across-protocol/sdk-v2/pull/102#discussion_r982098175

My current thinking is to live with the status quo (i.e. don't throw on price lookup failure; default instead to a price of 0).

Implication:
 - Callers need to specifically handle that a price of 0 might be returned. Existing Across "users" of this code do handle this correctly.

Throwing an exception here would probably be the _most_ correct thing to do, but require that callers handle that, and none currently do. Are we interested in upgrading them so that they can tolerate an exception?

src/utils.ts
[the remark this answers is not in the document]
pxrl: This should be resolved now.

src/gasPriceOracle/oracle.ts
mrice32: I think we should throw if the RPC call fails (after retry) rather than catching and converting to some fallback value.

In terms of the fee data _shape_, that _could_ differ by chain. Are we covering all the chains we support? If so, shouldn't we just throw an error if something doesn't match our expectations, since that would imply there is a bug in our code, no?

mrice32 (commented):
One remaining comment! https://github.com/across-protocol/sdk-v2/pull/102#discussion_r990413750

src/gasPriceOracle/oracle.ts
```diff
@@ -0,0 +1,95 @@
...
+  maxPriorityFeePerGas: BigNumber;
+};
+
+interface GasPriceFeed {
+  (provider: providers.Provider): Promise<GasPriceEstimate>;
+}
+
+function validateFeeData(val: BigNumber | null): BigNumber {
+  return BigNumber.isBigNumber(val) ? val : toBN(-1);
+}
```
mrice32: Instead of validating each parameter, why not just have a function that validates the entire struct?

Something like

```ts
type EIP1559FeeData = {
  maxFeePerGas: BigNumber;
  maxPriorityFeePerGas: BigNumber;
};

type LegacyFeeData = {
  gasPrice: BigNumber;
};

function isEip1559FeeData(input: providers.FeeData): input is EIP1559FeeData {
  return feeData.maxFeePerGas !== undefined && 
              BigNumber.isBigNumber(feeData.maxFeePerGas)) &&
              feeData.maxPriorityFeePerGas !== undefined &&
              BigNumber.isBigNumber(feeData.maxPriorityFeePerGas));
}

function isLegacyFeeData(input: providers.FeeData): input is LegacyFeeData {
  return feeData.gasPrice !== undefined && 
              BigNumber.isBigNumber(feeData.gasPrice));
}

function sendError(provider: providers.Provider, message: string): never {
  throw new Error("error message here");
}

function getEip1559FeeData(provider: providers.Provider): EIP1559FeeData {
  const feeData = await provider.getFeeData();
  if (!isEip1559FeeData(feeData)) sendError(provider, "Expected EIP1559 fee data");
  return feeData;
}

function getLegacyFeeData(provider: providers.Provider): LegacyFeeData {
  const feeData = await provider.getFeeData();
  if (!isLegacyFeeData(feeData)) sendError(provider, "Expected Legacy fee data");
  return feeData;
}

...
```

mrice32: A few notes on this.

`function isEip1559FeeData(input: providers.FeeData): input is EIP1559FeeData {` is a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates). When returning true, this forces typescript to understand that the input conforms to the EIP1559FeeData type.

`function sendError(provider: providers.Provider, message: string): never {` this never tells typescript that the function can never return, i.e. it will always throw. So it won't be considered in control flow analysis in the calling function.

`if (!isEip1559FeeData(feeData)) sendError(provider, "Expected EIP1559 fee data");` After this check, typescript should be convinced that the feeData object conforms to our EIP1559FeeData type.

mrice32 (approved):
LGTM! This is great! Super clean!

src/gasPriceOracle/oracle.ts
pxrl: > Instead of validating each parameter, why not just have a function that validates the entire struct?
> 
> Something like
> 
> ```ts
> type EIP1559FeeData = {
>   maxFeePerGas: BigNumber;
>   maxPriorityFeePerGas: BigNumber;
> };
> 
> type LegacyFeeData = {
>   gasPrice: BigNumber;
> };
> ```

One question: Do you think it's necessary to have two different types for this?

I was hoping to consolidate underneath an EIP-1559 definition of `{ maxFeePerGas, maxPriorityFeePerGas }`, with the hope that if the caller is producing a type 0 transaction, they can just take the value from maxFeePerGas anyway. Aside from the naming difference, maxFeePerGas maps pretty well onto gasPrice anyway. In this way we could (hopefully) abstract away the transaction type information, so that callers of this don't need to consider the differences.

mrice32: No, I think you've effectively routed around this problem!

mrice32 (approved):

PR #109, opened 2022-10-18
improve(PriceClient): Add DefiLlama adapter

pxrl:
Interestingly, DefiLlama include a "confidence" score to indicate how reliable they consider the price to be. Might be useful.

### DIFF
--- a/src/priceClient/adapters/defiLlama.ts
+++ b/src/priceClient/adapters/defiLlama.ts
@@ -59,7 +59,7 @@ export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
       throw new Error(`Unexpected ${this.name} response: ${JSON.stringify(tokenPrices)}`);
 
     return addresses.map((addr: string) => {
-      const tokenPrice = tokenPrices.coins[`ethereum:${addr.toLowerCase()}`];
+      const tokenPrice = tokenPrices.coins[`ethereum:${addr}`];
       if (tokenPrice === undefined) throw new Error(`Token ${addr} missing from ${this.name} response`);
 
       if (tokenPrice.confidence < this.minConfidence)
@@ -78,7 +78,7 @@ export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
     return Object.entries(coins as object).every(([address, tokenPrice]) => {
       // prettier-ignore
       return (
-        /[a-z]+:0x[0-9a-fA-F]{40}/.exec(address) !== undefined
+        /[a-z]+:0[xX][0-9a-fA-F]{40}/.exec(address) !== undefined
         && typeof tokenPrice === "object"
         && typeof tokenPrice.symbol === "string"
         && !isNaN(tokenPrice.decimals)
--- a/src/priceClient/adapters/defiLlama.ts
+++ b/src/priceClient/adapters/defiLlama.ts
@@ -51,21 +51,21 @@ export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
   }
 
   async getPricesByAddress(addresses: string[], currency = "usd"): Promise<TokenPrice[]> {
-    if (currency != "usd") throw new Error(`Currency ${currency} not supported by DefiLlama`);
+    if (currency != "usd") throw new Error(`Currency ${currency} not supported by ${this.name}`);
 
     const path = "prices/current/" + addresses.map((address) => `ethereum:${address}`).join();
     const tokenPrices: unknown = await this.query(path, {});
     if (!this.validateResponse(tokenPrices))
       throw new Error(`Unexpected ${this.name} response: ${JSON.stringify(tokenPrices)}`);
 
-    return addresses.map((addr: string) => {
-      const tokenPrice = tokenPrices.coins[`ethereum:${addr}`];
-      if (tokenPrice === undefined) throw new Error(`Token ${addr} missing from ${this.name} response`);
+    return addresses.map((address: string) => {
+      const tokenPrice = tokenPrices.coins[`ethereum:${address}`];
+      if (tokenPrice === undefined) throw new Error(`Token ${address} missing from ${this.name} response`);
 
       if (tokenPrice.confidence < this.minConfidence)
-        throw new Error(`Token ${addr} has low confidence score (${tokenPrice.confidence})`);
+        throw new Error(`Token ${address} has low confidence score (${tokenPrice.confidence})`);
 
-      return { address: addr, price: tokenPrice.price, timestamp: tokenPrice.timestamp };
+      return { address, price: tokenPrice.price, timestamp: tokenPrice.timestamp };
     });
   }
 
--- a/src/priceClient/adapters/defiLlama.ts
+++ b/src/priceClient/adapters/defiLlama.ts
@@ -1,0 +1,83 @@
+import assert from "assert";
+import { PriceFeedAdapter, TokenPrice } from "../priceClient";
+import { BaseHTTPAdapter } from "./baseAdapter";
+
+type DefiLlamaTokenPrice = {
+  price: number;
+  symbol: string;
+  timestamp: number;
+  confidence: number;
+};
+
+type DefiLlamaPriceResponse = {
+  coins: {
+    [key: string]: DefiLlamaTokenPrice;
+  };
+};
+
+const defaultTimeout = 5000; // mS
+
+export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
+  private _minConfidence: number;
+
+  constructor(name = "DefiLlama", { timeout, minConfidence = 0.9 }: { timeout?: number; minConfidence?: number }) {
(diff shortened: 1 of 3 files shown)


nicholaspai (commented):
I'm confused high level at this PR, is this something we're going to use in any production code or are you just suggesting we add it in as a nice, optional to have?

pxrl:
> I'm confused high level at this PR, is this something we're going to use in any production code or are you just suggesting we add it in as a nice, optional to have?

I think it's good to have available in case CoinGecko has issues. My proposal would be to add it to as a fallback option for users of the PriceClient - i.e. we would have some variation of:
- Across API
- CoinGecko Free/Pro
- DefiLlama

Note that DefiLlama are also dependent on CoinGecko, but they do write about implementing fallback logic, so I _think_ they might revert to on-chain sources if CoinGecko goes down. It could therefore be a nice way for us to mitigate against the single point of failure that (currently) is CoinGecko.

src/priceClient/adapters/defiLlama.ts
```diff
@@ -0,0 +1,81 @@
...
+  private validateResponse(response: unknown): response is DefiLlamaPriceResponse {
+    if (response === null || typeof response !== "object") return false;
+
+    const coins: unknown = (response as Record<string, unknown>).coins;
+    if (coins === null || typeof coins !== "object") return false;
+
+    return Object.entries(coins as object).every(([address, tokenPrice]) => {
+      // prettier-ignore
+      return (
+        /[a-z]+:0x[0-9a-fA-F]{40}/.exec(address) !== undefined
```
nicholaspai, line 70: I feel like there shoud be a better way to validate an HTTP response. But I see this is consistent with the coingecko adapter. Have you heard of [yup](https://github.com/jquense/yup)? 

@dohaki has used this in the merkle distributor scripts [here](https://github.com/across-protocol/merkle-distributor/pull/1) you should look into it!

nicholaspai (commented):
Its strange the DefiLlama uses CG on the backend, are we sure this is going to provide us real redundancy? It might be worth looking into.

Additionally, we should try using `yup` for response.data validation !

src/priceClient/adapters/defiLlama.ts
mrice32: Yeah, we use yup and superstruct to validate typescript types in different places. It would be great to standardize around one of them! Do people have strong preferences between the two?

@pxrl I do think we generally should prefer using one of those libs over manual verification.

src/priceClient/adapters/defiLlama.ts
```diff
@@ -0,0 +1,81 @@
...
+  async getPricesByAddress(addresses: string[], currency = "usd"): Promise<TokenPrice[]> {
+    if (currency != "usd") throw new Error(`Currency ${currency} not supported by DefiLlama`);
+
+    const path = "prices/current/" + addresses.map((address) => `ethereum:${address}`).join();
+    const tokenPrices: unknown = await this.query(path, {});
+    if (!this.validateResponse(tokenPrices))
+      throw new Error(`Unexpected ${this.name} response: ${JSON.stringify(tokenPrices)}`);
+
+    return addresses.map((addr: string) => {
+      const tokenPrice = tokenPrices.coins[`ethereum:${addr.toLowerCase()}`];
```
mrice32, line 51: I could be wrong, but right here it looks like you assume that the addresses returned are lower-case, but in the validation function you allow for uppercase characters in the address.

mrice32 (commented):
Looks good! Only minor comments.

src/priceClient/adapters/defiLlama.ts
pxrl: Just summing up some thoughts after spending a bit of time investigating this today.

TLDR: I can see some advantages to using a dedicated package for validation, but I haven't yet found a way for them to efficiently validate data structures with a mixture of fixed and dynamic keys.

See the bottom of this message for an overview of the data structures; dynamic keys are used for Eth addresses and currencies. This seems to be a case that isn't handled handle "out of the box" for these packages; there are threads where others have proposed solutions/workarounds, but they seem a bit hackish and require a surprising amount of boilerplate. and (to me) it becomes pretty difficult to understand what's going on.

I did find another package called [Joi](https://joi.dev/api/?v=17.6.1#objectpatternpattern-schema-options) that seems to handle this nicely by allowing dynamic keys to be specified as a regex pattern. This would _probably_ be a viable solution, and it seems cool, but involves yet another package with additional dependencies. We also already have superstruct as a dependency in sdk-v2, so if we decided to go for something else then I think it'd be worthwhile rolling it out to the places where we currently use superstruct.

@dohaki If you have any suggestions on how to efficiently handle the mix of fixed and dynamic keys then I'd be interested to hear more! cc @daywiss

For context on the data structures that I am aiming to validate:

CoinGecko:
```
{
  [address: string]: {
     [currency: string]: number;
    last_updated_at: number;
  };
};
```

DefiLlama:
```
{
  [address: string]: {
    symbol: string;
    price: number;
    timestamp: number;
    confidence: number;
  };
};
```

src/priceClient/adapters/defiLlama.ts
```diff
@@ -0,0 +1,81 @@
...
+      if (tokenPrice === undefined) throw new Error(`Token ${addr} missing from ${this.name} response`);
+
+      if (tokenPrice.confidence < this.minConfidence)
+        throw new Error(`Token ${addr} has low confidence score (${tokenPrice.confidence})`);
```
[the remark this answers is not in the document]
pxrl: Based on the description from their API docs, the confidence score seems to come into play when sourcing prices from fallback sources that _might_ be less reliable (i.e. directly from on-chain pools). This can effectively be used by consumers as a tripwire/shut-off valve in the event that DefiLlama continues to provide valid responses, even though the quality of its source data might be suspect.

I've set it to 0.9 per default in this implementation, but the caller can override the confidence score..so it's always an option to set it to 0.0 and effectively ignore the confidence score.

```
For tokens that we haven't been able to price in any other way, we find the pool with most liquidity for each on uniswap, curve and serum and then use the prices provided on those exchanges.

Unlike all the other tokens, we can't confirm that these prices are correct, so we only ingest the ones that have sufficient liquidity and, even in that case, we attach a confidence value to them that is related to the depth of liquidity and which represents our confidence in the quality of each price. API consumers can choose to filter out prices with low confidence values.
```

src/priceClient/adapters/defiLlama.ts
pxrl: You're spot on - it wasn't correct.

I've done a quick test with DefiLlama - they respect whatever format the address is provided in, and seem to respond with the same address. Therefore, the regex has been updated to accept all possible cases (because it doesn't know what the request was), and the subsequent lookup is: `tokenPrices.coins[\`ethereum:${addr}\`]`.

src/priceClient/adapters/defiLlama.ts
```diff
@@ -0,0 +1,92 @@
...
+  async getPricesByAddress(addresses: string[], currency = "usd"): Promise<TokenPrice[]> {
+    if (currency != "usd") throw new Error(`Currency ${currency} not supported by ${this.name}`);
+
+    const path = "prices/current/" + addresses.map((address) => `ethereum:${address}`).join();
+    const tokenPrices: unknown = await this.query(path, {});
+    if (!this.validateResponse(tokenPrices))
+      throw new Error(`Unexpected ${this.name} response: ${JSON.stringify(tokenPrices)}`);
+
+    return addresses.map((address: string) => {
+      const tokenPrice = tokenPrices.coins[`ethereum:${address}`];
```
mrice32, line 62: Is the request casing -> response casing relationship documented? I think this is fine as is since we'll start seeing errors if the address requested is missing.

An alternative would be to convert all keys to lowercase before processing them.

mrice32 (approved):
LGTM!

src/priceClient/adapters/defiLlama.ts
pxrl: > Is the request casing -> response casing relationship documented? I think this is fine as is since we'll start seeing errors if the address requested is missing.

Third parties don't really document this - it's only been identified via probing their responses, which of course means it's potentially subject to change.

It's not explicitly documented anywhere in the adapters either, but the PriceClient notes that it stores all addresses in lower-case internally: https://github.com/across-protocol/sdk-v2/blob/77269818898a691623e1a567ed55dae3011be9ae/src/priceClient/priceClient.ts#L18

Because the PriceClient generates its list of addresses from its internal cache, this means that the adapters will implicitly always receive lower-case (...per current implementation).

https://github.com/across-protocol/sdk-v2/blob/77269818898a691623e1a567ed55dae3011be9ae/src/priceClient/priceClient.ts#L78

 I'll concede that it's not really obvious, though...and I was scratching my head a bit on exactly this question earlier. I'll think about how I can describe it properly.

pxrl: Has been documented here: https://github.com/across-protocol/sdk-v2/pull/128/commits/12a81b81bec2e4d7f6a1115df5a87a77213629ff.

PR #111, opened 2022-10-19
fix: :ambulance: relax error on additional parameters

james-a-morris:
This change relaxes the requirement that the L1TokenConfigSs conforms exactly to the shape as the expected superstruct object. Now, a mask utility is called which trims off unnecessary parameters before asserting whether the parseL1TokenConfig input is valid.

Breaking Changes: no breaking changes added

### DIFF
--- a/src/contracts/acrossConfigStore.test.ts
+++ b/src/contracts/acrossConfigStore.test.ts
@@ -20,4 +20,10 @@ describe("Contracts Config Store", () => {
     };
     expect(Client.parseL1TokenConfig(JSON.stringify(structure))).toEqual(BASE_TRUTH);
   });
+  it("should fail to parse the data to parseL1TokenConfig with malformed input", () => {
+    const structure = {
+      rateModel: { UBar: "750000000000000000", R0: "21000000000000000", R1: "0", R2: "600000000000000000" },
+    };
+    expect(() => Client.parseL1TokenConfig(JSON.stringify(structure))).toThrow();
+  });
 });


daywiss (approved):
Very good

src/contracts/acrossConfigStore.ts
```diff
@@ -23,8 +23,9 @@ export class Client {
   }
   static parseL1TokenConfig(data: string): L1TokenConfig {
     const l1TokenConfig = JSON.parse(data);
-    assert(l1TokenConfig, L1TokenConfigSs);
-    return l1TokenConfig;
+    const l1TokenConfigMask = mask(l1TokenConfig, L1TokenConfigSs);
+    assert(l1TokenConfigMask, L1TokenConfigSs);
```
mrice32, line 27: Something I didn't quite understand from the [mask docs](https://docs.superstructjs.org/api-reference/core#mask) was whether it guarantees that the input values contains the required properties or not. The typing of this function would make it seems like it does do the check because the returned type is T and not unknown.

If it does, it would seem that the assert isn't necessary. If not, the assert makes sense. Do you have a better understanding of how this works?

mrice32 (commented):
Looks great! One minor question

src/contracts/acrossConfigStore.ts
mrice32: Understood, that seems reasonable!

mrice32 (approved):
LGTM!

PR #112, opened 2022-10-19
fix(gasPriceOracle): Revert to gasPrice for EIP-1559 chains

pxrl:
The existing use of maxFeePerGas gives a pessimistic view on the probable gas cost of the next transaction.

### DIFF
--- a/src/gasPriceOracle/oracle.ts
+++ b/src/gasPriceOracle/oracle.ts
@@ -34,7 +34,7 @@ async function eip1559(provider: providers.Provider, chainId: number): Promise<G
   });
 
   const maxPriorityFeePerGas = feeData.maxPriorityFeePerGas as BigNumber;
-  const maxFeePerGas = feeData.gasPrice as BigNumber; // note gasPrice is used.
+  const maxFeePerGas = maxPriorityFeePerGas.add(feeData.gasPrice as BigNumber); // note gasPrice is used.
 
   return {
     maxPriorityFeePerGas: maxPriorityFeePerGas,
--- a/src/gasPriceOracle/oracle.ts
+++ b/src/gasPriceOracle/oracle.ts
@@ -10,7 +10,7 @@ interface GasPriceFeed {
 }
 
 function error(method: string, chainId: number, data: providers.FeeData | BigNumber): void {
-  throw new Error(`Malformed ${method} response on chain ID ${chainId} (${JSON.stringify(data)}`);
+  throw new Error(`Malformed ${method} response on chain ID ${chainId} (${JSON.stringify(data)})`);
 }
 
 async function legacy(provider: providers.Provider, chainId: number): Promise<GasPriceEstimate> {
--- a/src/gasPriceOracle/oracle.ts
+++ b/src/gasPriceOracle/oracle.ts
@@ -24,17 +24,17 @@ async function legacy(provider: providers.Provider, chainId: number): Promise<Ga
   };
 }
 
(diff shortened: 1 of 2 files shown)


pxrl:
FWIW, here's the output of a short script that queries getFeeData() against an Eth mainnet RPC. Note that maxFeePerGas ~= 2x gasPrice (per EIP-1559 behaviour). I assume that we ultimately want some multiplier that's manually applied by us, and is probably closer to 1.1 or 1.25. So...fall back to gasPrice for now.

```
Got feeData: {"lastBaseFeePerGas":{"type":"BigNumber","hex":"0x06b831255a"},"maxFeePerGas":{"type":"BigNumber","hex":"0x0dc9ca79b4"},"maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x59682f00"},"gasPrice":{"type":"BigNumber","hex":"0x06d0ba5922"}}.
feeData.lastBaseFeePerGas: 28860032346
feeData.maxFeePerGas: 59220064692
feeData.maxPriorityFeePerGas: 1500000000
feeData.gasPrice: 29271677218
```

nicholaspai (approved):

src/gasPriceOracle/oracle.ts
```diff
@@ -24,17 +24,17 @@ async function legacy(provider: providers.Provider, chainId: number): Promise<Ga
...
   const feeData: providers.FeeData = await provider.getFeeData();
 
-  [feeData.maxFeePerGas, feeData.maxPriorityFeePerGas].forEach((field: BigNumber | null) => {
+  [feeData.gasPrice, feeData.maxPriorityFeePerGas].forEach((field: BigNumber | null) => {
     if (!BigNumber.isBigNumber(field) || field.lt(0)) error("getFeeData()", chainId, feeData);
   });
 
   const maxPriorityFeePerGas = feeData.maxPriorityFeePerGas as BigNumber;
-  const maxFeePerGas = feeData.maxFeePerGas as BigNumber;
+  const maxFeePerGas = feeData.gasPrice as BigNumber; // note gasPrice is used.
```
pxrl, line 37: Before submitting this, I want to confirm whether we should be adding maxPriorityFeePerGas here.

mrice32: Judging from [this implementation](https://github.com/ethers-io/ethers.js/blob/c80fcddf50a9023486e9f9acb1848aba4c19f7b6/packages/abstract-provider/src.ts/index.ts#L235-L256), what we ultimately want our estimate to be is: `lastBaseFeePerGas + maxPriorityFeePerGas`. If `lastBaseFeePerGas` is being reported as `gasPrice` (I think this is what you're saying), then we want the caller to estimate their gas costs by doing: `gasPrice + maxPriorityFeePerGas`.

I don't think the caller should ever use maxFeePerGas to _estimate_ their costs, because maxFeePerGas is a huge overestimate (2x base fee ish). If this is what we're currently doing, then maybe we should sub `gasPrice` for maxFeePerGas as you're doing now, but ultimately, we should change the naming because maxFeePerGas isn't what we're intending to use to feed the gas estimate.

Does that make sense?

mrice32 (commented):
https://github.com/across-protocol/sdk-v2/pull/112#discussion_r1000112407

src/gasPriceOracle/oracle.ts
pxrl: > Does that make sense?

Yep - all good. I was thrown off because I'd implemented maxFeePerGas = gasPrice + maxPriorityFeePerGas, and suddenly the tests failed unexpectedly. I've fixed that now.

PR #114, opened 2022-10-25
improve(PriceClient): Simplify PriceFeed constructor arguments

pxrl:
Per this comment to the use of PriceClient in the relayer-v2 repo: https://github.com/across-protocol/relayer-v2/pull/334#discussion_r1003809539.

This reduces some redundancy and makes it easier to instantiate a PriceFeed
instance. Also, name is now an optional argument that can be supplied as part
of the args object; if not supplied then a sensible per-feed default will be used.

### DIFF
--- a/src/priceClient/adapters/acrossApi.ts
+++ b/src/priceClient/adapters/acrossApi.ts
@@ -10,7 +10,7 @@ type AcrossApiArgs = {
 const defaultTimeout = 5000; // mS
 
 export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
-  constructor(args?: AcrossApiArgs) {
+  constructor(args: AcrossApiArgs = {}) {
     // Allow host to be overridden for test or alternative deployments.
     super(args?.name ?? "Across API", args?.host ?? "across.to", { timeout: args?.timeout ?? defaultTimeout });
   }
--- a/src/priceClient/adapters/acrossApi.ts
+++ b/src/priceClient/adapters/acrossApi.ts
@@ -10,9 +10,9 @@ type AcrossApiArgs = {
 const defaultTimeout = 5000; // mS
 
 export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
-  constructor(args: AcrossApiArgs = {}) {
+  constructor({ name = "Across API", host = "across.to", timeout = defaultTimeout }: AcrossApiArgs = {}) {
     // Allow host to be overridden for test or alternative deployments.
-    super(args?.name ?? "Across API", args?.host ?? "across.to", { timeout: args?.timeout ?? defaultTimeout });
+    super(name, host, { timeout });
   }
 
   async getPriceByAddress(address: string, currency = "usd"): Promise<TokenPrice> {
--- a/src/priceClient/adapters/acrossApi.ts
+++ b/src/priceClient/adapters/acrossApi.ts
@@ -2,12 +2,17 @@ import { msToS, PriceFeedAdapter, TokenPrice } from "../priceClient";
 import { BaseHTTPAdapter } from "./baseAdapter";
 
 type AcrossPrice = { price: number };
+type AcrossApiArgs = {
+  name?: string;
+  host?: string;
+  timeout?: number;
+};
 const defaultTimeout = 5000; // mS
 
 export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
-  constructor(name: string, { host, timeout }: { host?: string; timeout?: number }) {
+  constructor(args?: AcrossApiArgs) {
     // Allow host to be overridden for test or alternative deployments.
-    super(name, host ?? "across.to", { timeout: timeout ?? defaultTimeout });
+    super(args?.name ?? "Across API", args?.host ?? "across.to", { timeout: args?.timeout ?? defaultTimeout });
   }
 
   async getPriceByAddress(address: string, currency = "usd"): Promise<TokenPrice> {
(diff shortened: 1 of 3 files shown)


src/priceClient/adapters/acrossApi.ts
```diff
@@ -2,12 +2,17 @@ import { msToS, PriceFeedAdapter, TokenPrice } from "../priceClient";
...
+  timeout?: number;
+};
 const defaultTimeout = 5000; // mS
 
 export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
-  constructor(name: string, { host, timeout }: { host?: string; timeout?: number }) {
+  constructor(args?: AcrossApiArgs) {
     // Allow host to be overridden for test or alternative deployments.
-    super(name, host ?? "across.to", { timeout: timeout ?? defaultTimeout });
+    super(args?.name ?? "Across API", args?.host ?? "across.to", { timeout: args?.timeout ?? defaultTimeout });
```
pxrl, line 15: nb. This does read a _little_ bit complicated; was the most sensible thing I could figure out, but there might be better/cleaner ways of doing it.

src/priceClient/adapters/acrossApi.ts
```diff
@@ -2,12 +2,17 @@ import { msToS, PriceFeedAdapter, TokenPrice } from "../priceClient";
...
+type AcrossApiArgs = {
+  name?: string;
+  host?: string;
+  timeout?: number;
+};
 const defaultTimeout = 5000; // mS
 
 export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
-  constructor(name: string, { host, timeout }: { host?: string; timeout?: number }) {
+  constructor(args?: AcrossApiArgs) {
```
mrice32, line 13: nit: I don't think it needs to be this complicated. I think you can keep the exact same code as before with a default arg of `{}`. Thoughts? Or am I missing something?

```suggestion
  constructor(args: AcrossApiArgs = {}) {
```

src/priceClient/adapters/coingecko.ts
```diff
@@ -12,14 +12,22 @@ type CoinGeckoPriceResponse = {
...
+
 export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
   private readonly apiKey: string | undefined = undefined;
 
-  constructor(name: string, { timeout, apiKey }: { timeout?: number; apiKey?: string }) {
-    super(name, apiKey ? "pro-api.coingecko.com" : "api.coingecko.com", {
-      timeout: timeout ?? defaultTimeout,
-    });
-    this.apiKey = apiKey;
+  constructor(args?: CoinGeckoArgs) {
```
mrice32, line 24: nit: same as above. I think this is the only change that needs to be made.

```suggestion
  constructor(args: CoinGeckoArgs = {}) {
```

src/priceClient/adapters/acrossApi.ts
mrice32: I think there's a simpler way with just a default arg unless I'm missing something!

pxrl:
@mrice32 I'm sure there is a simpler way to achieve this :D

My objective with this adjustment is to allow the caller to override individual fields in the various PriceFeed "args" objects without nullifying/voiding the defaults for the parameters that they _didn't_ override.

For example:

`new coingecko.PriceFeed()` => all defaults
`new coingecko.PriceFeed({ host: "127.0.0.1" })` => override _only_ host, all other members should inherit defaults
`new coingecko.PriceFeed({ apiKey: "<xxx-api-key>" })` => override _only_ apiKey; allow coingecko.PriceFeed to resolve the name itself.
`new coingecko.PriceFeed({ name: "custom-cg-name", apiKey: "<xxx-api-key>" })` => override _both_ name and apiKey.

Does that make sense?

src/priceClient/adapters/acrossApi.ts
```diff
@@ -2,12 +2,17 @@ import { msToS, PriceFeedAdapter, TokenPrice } from "../priceClient";
...
+type AcrossApiArgs = {
+  name?: string;
+  host?: string;
+  timeout?: number;
+};
 const defaultTimeout = 5000; // mS
 
 export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
-  constructor(name: string, { host, timeout }: { host?: string; timeout?: number }) {
+  constructor(args: AcrossApiArgs = {}) {
```
mrice32, line 13:
```suggestion
  constructor({ name = "Across API", host = "across.to", timeout = defaultTimeout }: AcrossApiArgs = {}) {
```

This will allow you to change the call below to (github won't allow me to suggest):
```ts
super(name, host, { timeout });
```

mrice32 (commented):
Sorry for the slow feedback, one more quick round of comments!

mrice32 (approved):

src/priceClient/adapters/coingecko.ts
```diff
@@ -12,14 +12,22 @@ type CoinGeckoPriceResponse = {
...
-      timeout: timeout ?? defaultTimeout,
-    });
-    this.apiKey = apiKey;
+  constructor(args: CoinGeckoArgs = {}) {
+    super(
+      args?.name ?? args?.apiKey ? "CoinGecko Pro" : "CoinGecko Free",
+      args?.apiKey ? "pro-api.coingecko.com" : "api.coingecko.com",
+      { timeout: args?.timeout ?? defaultTimeout }
+    );
+    this.apiKey = args?.apiKey;
```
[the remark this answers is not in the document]
pxrl: Really nice! I like this a lot.

PR #124, opened 2022-11-01
fix(relay-fee-calculator): Divide by 1e18 for Optimism gas cost 

nicholaspai:
https://github.com/across-protocol/sdk-v2/commit/7abcbfa7b4690e9706bd6ae91ae66de041427906

correctly divided by 1e18 after toBNWei-ing the gasMultiplier but forgot to add the same division for the Optimism special case


src/utils.ts
```diff
@@ -269,7 +272,7 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
 
   // Find the total gas cost by taking the product of the gas price & the
   // estimated number of gas units needed.
-  return BigNumber.from(gasPrice).mul(gasTotalMultiplier).div(toBNWei(1)).mul(estimatedGasUnits).toString();
```
nicholaspai, line 272: Should divide in last step to not accidentally remove precision

PR #130, opened 2022-11-29
fix: Fix APR and APY's to 18 precision

nicholaspai:
when utilized reserves is < 0, apr's can get so small they toString() into scientific notation


mrice32 (approved):
LGTM!

PR #131, opened 2022-11-29
feat(utils): Add a simple function to find a block with desired lookback

kevinuma:
This adds a simple method to find a block with desired lookback such as 2 days. Tested in relayer-v2.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -339,7 +339,9 @@ export async function findBlockAtOrOlder(provider: providers.Provider, desiredLo
   const [toBlock, network] = await Promise.all([provider.getBlock("latest"), provider.getNetwork()]);
   let toBlockTimestamp = toBlock.timestamp;
   const desiredTimestamp = toBlockTimestamp - desiredLookback;
+  assert(desiredTimestamp >= 0, "Desired lookback cannot be more than the current block timestamp");
   let skipDistance = BlockScanSkipDistances[network.chainId];
+  assert(skipDistance > 0, "Skip distance must be strictly positive");
 
   // Fetch the first block to get the block production speed estimate before proceeding further.
   let fromBlockNumber = toBlock.number - skipDistance;
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -341,14 +341,15 @@ export async function findBlockAtOrOlder(
   const latestBlock = await provider.getBlock("latest");
   const latestBlockTimestamp = latestBlock.timestamp;
   const desiredTimestamp = latestBlockTimestamp - desiredLookback;
-  const chainId = (await provider.getNetwork()).chainId;
-  let lastSkipDistance = BlockScanSkipDistances[chainId];
+  let lastSkipDistance = BlockScanSkipDistances[(await provider.getNetwork()).chainId];
   let lastBlockNumber = latestBlock.number;
   let previousLastBlockTimestamp = latestBlockTimestamp;
   let lastBlockTimestamp = latestBlockTimestamp;
   while (lastBlockTimestamp > desiredTimestamp) {
     // Skip the first base case where the last looked up block and the latest block are the same.
     if (previousLastBlockTimestamp - lastBlockTimestamp > 0) {
+      // Calculate the block speed based on last block query and use it to calculate how many more blocks to go back
+      // to find the block with desired timestamp.
       const lastBlockSpeed = lastSkipDistance / (previousLastBlockTimestamp - lastBlockTimestamp);
       lastSkipDistance = Math.floor(lastBlockSpeed * (lastBlockTimestamp - desiredTimestamp));
     }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -335,24 +335,25 @@ export async function createUnsignedFillRelayTransaction(
  * @return The block number that's at or older than the desired lookback.
  */
 export async function findBlockAtOrOlder(provider: providers.Provider, desiredLookback: number): Promise<number> {
-  const [latestBlock, network] = await Promise.all([provider.getBlock("latest"), provider.getNetwork()]);
-  let lastBlockTimestamp = latestBlock.timestamp;
-  let previousLastBlockTimestamp = lastBlockTimestamp;
-  const desiredTimestamp = lastBlockTimestamp - desiredLookback;
-  let lastSkipDistance = BlockScanSkipDistances[network.chainId];
-  let lastBlockNumber = latestBlock.number;
-  while (lastBlockTimestamp > desiredTimestamp) {
-    // Skip the first base case where the last looked up block and the latest block are the same.
-    if (previousLastBlockTimestamp - lastBlockTimestamp > 0) {
-      // Calculate the block speed based on last block query and use it to calculate how many more blocks to go back
-      // to find the block with desired timestamp.
-      const lastBlockSpeed = lastSkipDistance / (previousLastBlockTimestamp - lastBlockTimestamp);
-      lastSkipDistance = Math.floor(lastBlockSpeed * (lastBlockTimestamp - desiredTimestamp));
-    }
-    lastBlockNumber = lastBlockNumber - lastSkipDistance;
-    const block = await provider.getBlock(lastBlockNumber);
-    previousLastBlockTimestamp = lastBlockTimestamp;
-    lastBlockTimestamp = block.timestamp;
+  const [toBlock, network] = await Promise.all([provider.getBlock("latest"), provider.getNetwork()]);
+  let toBlockTimestamp = toBlock.timestamp;
+  const desiredTimestamp = toBlockTimestamp - desiredLookback;
+  let skipDistance = BlockScanSkipDistances[network.chainId];
+
+  // Fetch the first block to get the block production speed estimate before proceeding further.
+  let fromBlockNumber = toBlock.number - skipDistance;
+  let fromBlock = await provider.getBlock(fromBlockNumber);
+  let fromBlockTimestamp = fromBlock.timestamp;
+  while (fromBlockTimestamp > desiredTimestamp) {
+    // Calculate the block speed based on last block query and use it to calculate how many more blocks to go back
+    // to find the block with desired timestamp.
+    const blockSpeed = skipDistance / (toBlockTimestamp - fromBlockTimestamp);
+    skipDistance = Math.floor(blockSpeed * (fromBlockTimestamp - desiredTimestamp));
+    fromBlockNumber -= skipDistance;
+    fromBlock = await provider.getBlock(fromBlockNumber);
+    // Set toBlock equal to current fromBlock and then decrement fromBlock
+    toBlockTimestamp = fromBlockTimestamp;
+    fromBlockTimestamp = fromBlock.timestamp;
   }
-  return lastBlockNumber;
+  return fromBlockNumber;
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -12,14 +12,16 @@ export type Decimalish = string | number | Decimal;
 export const AddressZero = ethers.constants.AddressZero;
 
 const { ConvertDecimals } = uma.utils;
-const BlockSpeedSample: { [chainId: number]: number } = {
-  1: 10000,
-  10: 400000,
-  137: 40000,
+// These are distances used to traverse when looking for a block with desired lookback.
+// They're meant to be small enough to allow for granularity but large enough to minimize the number of reqests needed
+// to find the desired block.
+const BlockScanSkipDistances: { [chainId: number]: number } = {
+  1: 1000,
+  10: 100000,
+  137: 10000,
   288: 1000,
-  42161: 400000,
+  42161: 100000,
 };
-const ONE_DAY_SECS = 86400;
 
 /**
  * toBN.
@@ -313,25 +315,47 @@ export async function createUnsignedFillRelayTransaction(
 }
 
 /**
- * Calculates the average block production speed over the past ~1 day.
- * @param provider A valid ethers provider - will be used to calculate block production speed.
+ * Search back in time for the first block at or older than the desired lookback.
+ *
+ * The approach here is as below:
+ * 1. We use the skip distances, which are meant to be a couple of hours or less for each chain.
+ * 2. Fetch the first block one skip distance from the latest block. Check its timestamp to see how far back it is
+ * relatively to the latest block.
+ * 3. Apply a multiplier to the skip distance based on the time difference and find a block with the desired lookback
+ * This assumes the block production speed is roughly constant.
+ * 4. Check the timestamp again, if we still haven't achieved the desired lookback, repeat step 3 and go back further.
+ * Otherwise, return the block number.
+ *
+ * This is only a rough estimate and can go back further than desired. However, this would minimize the number
+ * of requests and is generally accurate unless the rate of block production has been very volatile over a
+ * short period of time.
+ *
+ * @param provider A valid provider - will be used to find the block with desired lookback.
+ * @param desiredLookback Desired lookback in seconds (e.g. 86400 seconds or 1 day).
+ * @return The block number that's at or older than the desired lookback.
  */
-export async function calculateBlockSpeed(
-  provider: providers.Provider | L2Provider<providers.Provider>
+export async function findBlockAtOrOlder(
+  provider: providers.Provider | L2Provider<providers.Provider>,
+  desiredLookback: number
 ): Promise<number> {
   const latestBlock = await provider.getBlock("latest");
   const latestBlockTimestamp = latestBlock.timestamp;
+  const desiredTimestamp = latestBlockTimestamp - desiredLookback;
   const chainId = (await provider.getNetwork()).chainId;
-  const sampleSpeed = BlockSpeedSample[chainId];
+  let lastSkipDistance = BlockScanSkipDistances[chainId];
   let lastBlockNumber = latestBlock.number;
+  let previousLastBlockTimestamp = latestBlockTimestamp;
   let lastBlockTimestamp = latestBlockTimestamp;
-  // Keep searching for the first block older than 1 day.
-  // This is only a rough estimate and can go back further than 1 day. However, this would minimize the number
-  // of requests and is generally accurate unless the rate of block production has been very volatile over a
-  // short period of time.
-  while (latestBlockTimestamp - lastBlockTimestamp < ONE_DAY_SECS) {
-    lastBlockNumber = lastBlockNumber - sampleSpeed;
-    lastBlockTimestamp = (await provider.getBlock(lastBlockNumber)).timestamp;
+  while (lastBlockTimestamp > desiredTimestamp) {
+    // Skip the first base case where the last looked up block and the latest block are the same.
+    if (previousLastBlockTimestamp - lastBlockTimestamp > 0) {
+      const lastBlockSpeed = lastSkipDistance / (previousLastBlockTimestamp - lastBlockTimestamp);
+      lastSkipDistance = Math.floor(lastBlockSpeed * (lastBlockTimestamp - desiredTimestamp));
+    }
+    lastBlockNumber = lastBlockNumber - lastSkipDistance;
+    const block = await provider.getBlock(lastBlockNumber);
+    previousLastBlockTimestamp = lastBlockTimestamp;
+    lastBlockTimestamp = block.timestamp;
   }
-  return ((latestBlock.number - lastBlockNumber) * ONE_DAY_SECS) / (latestBlockTimestamp - lastBlockTimestamp);
+  return lastBlockNumber;
 }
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -12,6 +12,14 @@ export type Decimalish = string | number | Decimal;
 export const AddressZero = ethers.constants.AddressZero;
 
 const { ConvertDecimals } = uma.utils;
+const BlockSpeedSample: { [chainId: number]: number } = {
+  1: 10000,
```
nicholaspai, line 16: How did you pick these samples by the way?

nicholaspai (commented):
Driveby: So doesn't this implementation have the same (but muted) problem as the original issue, whereby block times are dependent on actual block production variability? How does this algo handle the case where the blocktime on rollups can 3x increase in volatile environments?

Is there a way we can:
- get an approximation of block time accurate to within X hours
- mitigate exposure to actual variance in block times
- where X = 6 hours of blocks, or maybe even better, the # of blocks it takes to reach "finalization"?

src/utils.ts
```diff
@@ -303,3 +311,27 @@ export async function createUnsignedFillRelayTransaction(
...
+  const latestBlockTimestamp = latestBlock.timestamp;
+  const chainId = (await provider.getNetwork()).chainId;
+  const sampleSpeed = BlockSpeedSample[chainId];
+  let lastBlockNumber = latestBlock.number;
+  let lastBlockTimestamp = latestBlockTimestamp;
+  // Keep searching for the first block older than 1 day.
+  // This is only a rough estimate and can go back further than 1 day. However, this would minimize the number
+  // of requests and is generally accurate unless the rate of block production has been very volatile over a
+  // short period of time.
+  while (latestBlockTimestamp - lastBlockTimestamp < ONE_DAY_SECS) {
```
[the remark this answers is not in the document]
kevinuma: It might not be long enough. This is just a simple heuristic-based approach - if it's long enough, keep searching back. Ideally this can be more sophisticated by conducting a search similar to binary search to find the closest block to ~1 day from the latest. However, this might not reduce the number of requests as much and is just a lot more complex

kevinuma:
> Driveby: So doesn't this implementation have the same (but muted) problem as the original issue, whereby block times are dependent on actual block production variability? How does this algo handle the case where the blocktime on rollups can 3x increase in volatile environments?
> 
> Is there a way we can:
> 
> * get an approximation of block time accurate to within X hours
> * mitigate exposure to actual variance in block times
> * where X = 6 hours of blocks, or maybe even better, the # of blocks it takes to reach "finalization"?

This is better than what we currently have because:
1. It'd go back further if the block range is not long enough.
2. It returns the estimate speed, which can be used to construct a more precise block range for event query. 

We can choose a more sophisticated alternative with sampling a lot more blocks (hence more requests) to get more data points for averaging. This is the second option I was thinking but it can lead to more requests being sent and is just overall more complex. I can experiment a bit more.

nicholaspai:
> > Driveby: So doesn't this implementation have the same (but muted) problem as the original issue, whereby block times are dependent on actual block production variability? How does this algo handle the case where the blocktime on rollups can 3x increase in volatile environments?
> > Is there a way we can:
> > 
> > * get an approximation of block time accurate to within X hours
> > * mitigate exposure to actual variance in block times
> > * where X = 6 hours of blocks, or maybe even better, the # of blocks it takes to reach "finalization"?
> 
> This is better than what we currently have because:
> 
> 1. It'd go back further if the block range is not long enough.
> 2. It returns the estimate speed, which can be used to construct a more precise block range for event query.
> 
> We can choose a more sophisticated alternative with sampling a lot more blocks (hence more requests) to get more data points for averaging. This is the second option I was thinking but it can lead to more requests being sent and is just overall more complex. I can experiment a bit more.

Why can't we sample exactly one block? Say we hardcode the sample length to something very conservative like 40,000 blocks on Optimism/Arbitrum as you have it. Can't we just use the time since that block to create a reasonable estimate?

mrice32:
> > > Driveby: So doesn't this implementation have the same (but muted) problem as the original issue, whereby block times are dependent on actual block production variability? How does this algo handle the case where the blocktime on rollups can 3x increase in volatile environments?
> > > Is there a way we can:
> > > 
> > > * get an approximation of block time accurate to within X hours
> > > * mitigate exposure to actual variance in block times
> > > * where X = 6 hours of blocks, or maybe even better, the # of blocks it takes to reach "finalization"?
> > 
> > 
> > This is better than what we currently have because:
> > 
> > 1. It'd go back further if the block range is not long enough.
> > 2. It returns the estimate speed, which can be used to construct a more precise block range for event query.
> > 
> > We can choose a more sophisticated alternative with sampling a lot more blocks (hence more requests) to get more data points for averaging. This is the second option I was thinking but it can lead to more requests being sent and is just overall more complex. I can experiment a bit more.
> 
> Why can't we sample exactly one block? Say we hardcode the sample length to something very conservative like 40,000 blocks on Optimism/Arbitrum as you have it. Can't we just use the time since that block to create a reasonable estimate?

@kevinuma @nicholaspai Not sure if we want to go this direction, but we have some code [here](https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/blockFinder.ts) to essentially find the exact block for a timestamp. You could just copy the class entirely. The only thing you would need to tweak is that it currently uses a mainnet-based estimateBlocksElapsed function for figuring out how quickly to jump backwards in time in its search. You would probably want this class to either allow the user to pass in that function or allow the user to pass in the chain id to so that function could use an appropriate heuristic for the connected chain.

The only reason I suggest this is that we already use it pretty extensively (we use it in the process of determining the across LP fees).

kevinuma:
> > > > Driveby: So doesn't this implementation have the same (but muted) problem as the original issue, whereby block times are dependent on actual block production variability? How does this algo handle the case where the blocktime on rollups can 3x increase in volatile environments?
> > > > Is there a way we can:
> > > > 
> > > > * get an approximation of block time accurate to within X hours
> > > > * mitigate exposure to actual variance in block times
> > > > * where X = 6 hours of blocks, or maybe even better, the # of blocks it takes to reach "finalization"?
> > > 
> > > 
> > > This is better than what we currently have because:
> > > 
> > > 1. It'd go back further if the block range is not long enough.
> > > 2. It returns the estimate speed, which can be used to construct a more precise block range for event query.
> > > 
> > > We can choose a more sophisticated alternative with sampling a lot more blocks (hence more requests) to get more data points for averaging. This is the second option I was thinking but it can lead to more requests being sent and is just overall more complex. I can experiment a bit more.
> > 
> > 
> > Why can't we sample exactly one block? Say we hardcode the sample length to something very conservative like 40,000 blocks on Optimism/Arbitrum as you have it. Can't we just use the time since that block to create a reasonable estimate?
> 
> @kevinuma @nicholaspai Not sure if we want to go this direction, but we have some code [here](https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/blockFinder.ts) to essentially find the exact block for a timestamp. You could import (or copy) the class. The only thing you would need to tweak is that it currently uses a mainnet-based estimateBlocksElapsed function for figuring out how quickly to jump backwards in time in its search. You would probably want this class to either allow the user to pass in that function or allow the user to pass in the chain id to so that function could use an appropriate heuristic for the connected chain.
> 
> The only reason I suggest this is that we already use it pretty extensively (we use it in the process of determining the across LP fees).

Oh thanks Matt! This is a really nice finding! I basically reimplemented a lot of the features there. Let me just use this then

kevinuma:
> Referenc

It uses https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/utils.ts#L103 which has some hardcoded numbers actually. Let me savage what can be reused.

src/utils.ts
```diff
@@ -303,3 +313,49 @@ export async function createUnsignedFillRelayTransaction(
...
+ * of requests and is generally accurate unless the rate of block production has been very volatile over a
+ * short period of time.
+ *
+ * @param provider A valid provider - will be used to find the block with desired lookback.
+ * @param desiredLookback Desired lookback in seconds (e.g. 86400 seconds or 1 day).
+ * @return The block number that's at or older than the desired lookback.
+ */
+export async function findBlockAtOrOlder(
+  provider: providers.Provider | L2Provider<providers.Provider>,
+  desiredLookback: number
```
nicholaspai, line 339: I like parameterizing `desiredLookback`!

src/utils.ts
```diff
@@ -303,3 +313,49 @@ export async function createUnsignedFillRelayTransaction(
...
+): Promise<number> {
+  const latestBlock = await provider.getBlock("latest");
+  const latestBlockTimestamp = latestBlock.timestamp;
+  const desiredTimestamp = latestBlockTimestamp - desiredLookback;
+  const chainId = (await provider.getNetwork()).chainId;
```
nicholaspai, line 344: nit: should we Promise.all the latestBlock and chainId calls?

src/utils.ts
```diff
@@ -303,3 +313,49 @@ export async function createUnsignedFillRelayTransaction(
...
+ * This is only a rough estimate and can go back further than desired. However, this would minimize the number
+ * of requests and is generally accurate unless the rate of block production has been very volatile over a
+ * short period of time.
+ *
+ * @param provider A valid provider - will be used to find the block with desired lookback.
+ * @param desiredLookback Desired lookback in seconds (e.g. 86400 seconds or 1 day).
+ * @return The block number that's at or older than the desired lookback.
+ */
+export async function findBlockAtOrOlder(
+  provider: providers.Provider | L2Provider<providers.Provider>,
```
nicholaspai, line 338: Why this double provider type?

src/utils.ts
```diff
@@ -303,3 +313,49 @@ export async function createUnsignedFillRelayTransaction(
...
+  const desiredTimestamp = latestBlockTimestamp - desiredLookback;
+  const chainId = (await provider.getNetwork()).chainId;
+  let lastSkipDistance = BlockScanSkipDistances[chainId];
+  let lastBlockNumber = latestBlock.number;
+  let previousLastBlockTimestamp = latestBlockTimestamp;
+  let lastBlockTimestamp = latestBlockTimestamp;
+  while (lastBlockTimestamp > desiredTimestamp) {
+    // Skip the first base case where the last looked up block and the latest block are the same.
+    if (previousLastBlockTimestamp - lastBlockTimestamp > 0) {
+      const lastBlockSpeed = lastSkipDistance / (previousLastBlockTimestamp - lastBlockTimestamp);
```
nicholaspai, line 352: Can `(previousLastBlockTimestamp - lastBlockTimestamp)` ever equal 0?

src/utils.ts
```diff
@@ -303,3 +313,49 @@ export async function createUnsignedFillRelayTransaction(
...
+export async function findBlockAtOrOlder(
+  provider: providers.Provider | L2Provider<providers.Provider>,
+  desiredLookback: number
+): Promise<number> {
+  const latestBlock = await provider.getBlock("latest");
+  const latestBlockTimestamp = latestBlock.timestamp;
+  const desiredTimestamp = latestBlockTimestamp - desiredLookback;
+  const chainId = (await provider.getNetwork()).chainId;
+  let lastSkipDistance = BlockScanSkipDistances[chainId];
+  let lastBlockNumber = latestBlock.number;
```
nicholaspai, line 346: Why not make `latestBlockTimestamp` a variable instead of a constant instead of setting both `lastBlockNumber` and `latestBlockTimestamp` equal to `latestBlock.number`

I think overall try to condense the number of variables and re-use them in this function where possible to make it more readable

src/utils.ts
```diff
@@ -303,3 +313,49 @@ export async function createUnsignedFillRelayTransaction(
...
 }
+
+/**
+ * Search back in time for the first block at or older than the desired lookback.
+ *
+ * The approach here is as below:
+ * 1. We use the skip distances, which are meant to be a couple of hours or less for each chain.
+ * 2. Fetch the first block one skip distance from the latest block. Check its timestamp to see how far back it is
+ * relatively to the latest block.
+ * 3. Apply a multiplier to the skip distance based on the time difference and find a block with the desired lookback
```
nicholaspai, line 324: This step 3 toook me a while to understand but I think it might be better summarized as. "3. Use the implied block speed from the latest block to one skip distance earlier to approximate how many blocks earlier we need to lookback to find a block older than the target block. This assumes that the block speed is roughly constant from latest block until the target block."

src/utils.ts
```diff
@@ -303,3 +313,49 @@ export async function createUnsignedFillRelayTransaction(
...
+ * This assumes the block production speed is roughly constant.
+ * 4. Check the timestamp again, if we still haven't achieved the desired lookback, repeat step 3 and go back further.
```
nicholaspai, line 326: nit: I think this algo repeats step 2 and goes on from there, not starting at step 3 right? Because step 2 first figures out the `lastBlockSpeed` which is important to approximate the desired block

src/utils.ts
kevinuma: Yes this is the case for the first iteration as they both start out as the latest block's timestamp (no scan has been done).

src/utils.ts
kevinuma: Great suggestion! Updated.

src/utils.ts
kevinuma: Copy pasted from other util functions. We don't need L2Provider so I removed it

src/utils.ts
kevinuma: Great suggestion! Updated.

src/utils.ts
nicholaspai: You could set the variables like this and then remove the section in your `while` loop that "skips the first base case where the last looked up block and the latest block are the same"

```ts
let toBlock = await provider.getBlock("latest");
let toBlockNumber = toBlock.number
let toBlockTimestamp = toBlock.timestamp;
const desiredTimestamp = toBlockTimestamp - desiredLookback;
let skipDistance = BlockScanSkipDistances[chainId];
let fromBlockNumber = toBlockNumber - skipDistance;
let fromBlock = await provider.getBlock(fromBlockNumber);
let fromBlockTimestamp = fromBlock.timestamp

while(fromBlockTimestamp > desiredTimestamp) {
   const blockSpeed = skipDistance / (toBlockTimestamp - fromBlockTimestamp);
   skipDistance = Math.floor(blockSpeed * (fromBlockTimestamp - desiredTimestamp));
   fromBlockNumber -= skipDistance;
   const block = await provider.getBlock(fromBlockNumber);
   // Set toBlock equal to current fromBlock and then decrement fromBlock
   toBlockTimestamp = fromBlockTimestamp
   fromBlockTimestamp = block.timestamp
}
```

nicholaspai (commented):
I agree with this algo highlevel! i think it should result in max 3 rpc calls to `getBlock` and usually 2. What if we added some buffer to the `blockSpeed ` to make it more likely we only need 2 rpc calls to `getBlock`, something like 

```ts
const blockSpeed = (skipDistance / (toBlockTimestamp - fromBlockTimestamp)) * 1.1;
```

src/utils.ts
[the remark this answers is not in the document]
nicholaspai: ah I see I didn't fully understand the algo at first

src/utils.ts
nicholaspai: sweet

src/utils.ts
kevinuma: These don't have to exact and are meant to be a starting point. The way the algo works here is it fetches the first block exactly 1 distance away from the latest block and uses the timestamp difference to calculate the initial block speed. From then on, it'd add a multiplier to try to get as close to the desired timestamp as possible. The heuristic here is to have these be relatively small but not too small as that'd lead to the first step getting a volatile speed and just make the search less efficient. On the other hand, if these values are too large and the desired lookback timestamp is small, we might end up going back too far.

kevinuma:
> I agree with this algo highlevel! i think it should result in max 3 rpc calls to `getBlock` and usually 2. What if we added some buffer to the `blockSpeed ` to make it more likely we only need 2 rpc calls to `getBlock`, something like
> 
> ```ts
> const blockSpeed = (skipDistance / (toBlockTimestamp - fromBlockTimestamp)) * 1.1;
> ```

The drawback of this though is it might overshoot the desired lookback. I've seen some cases where if block production speed really goes down over the course of the lookback window, it'd overshoot by 20%. Adding an extra multiplier would increase this overshoot.

src/utils.ts
kevinuma: This makes sense. Love the variable names too! Updated.

src/utils.ts
kevinuma: Updated

src/utils.ts
kevinuma: Updated

src/utils.ts
```diff
@@ -303,3 +313,48 @@ export async function createUnsignedFillRelayTransaction(
...
+ * short period of time.
+ *
+ * @param provider A valid provider - will be used to find the block with desired lookback.
+ * @param desiredLookback Desired lookback in seconds (e.g. 86400 seconds or 1 day).
+ * @return The block number that's at or older than the desired lookback.
+ */
+export async function findBlockAtOrOlder(provider: providers.Provider, desiredLookback: number): Promise<number> {
+  const [toBlock, network] = await Promise.all([provider.getBlock("latest"), provider.getNetwork()]);
+  let toBlockTimestamp = toBlock.timestamp;
+  const desiredTimestamp = toBlockTimestamp - desiredLookback;
```
nicholaspai, line 341: nit: suggest asserting that `desiredLookback <= toBlockTimestamp`

src/utils.ts
```diff
@@ -303,3 +313,48 @@ export async function createUnsignedFillRelayTransaction(
...
+  let skipDistance = BlockScanSkipDistances[network.chainId];
+
+  // Fetch the first block to get the block production speed estimate before proceeding further.
+  let fromBlockNumber = toBlock.number - skipDistance;
```
nicholaspai, line 345: also nit: suggest asserting that `skipDistance <= toBlock.number`

src/utils.ts
```diff
@@ -303,3 +313,48 @@ export async function createUnsignedFillRelayTransaction(
...
+  let fromBlock = await provider.getBlock(fromBlockNumber);
+  let fromBlockTimestamp = fromBlock.timestamp;
+  while (fromBlockTimestamp > desiredTimestamp) {
+    // Calculate the block speed based on last block query and use it to calculate how many more blocks to go back
+    // to find the block with desired timestamp.
+    const blockSpeed = skipDistance / (toBlockTimestamp - fromBlockTimestamp);
```
nicholaspai, line 351: In order to make sure `(toBlockTimestamp - fromBlockTimestamp) != 0` I think you need to assert above that `skipDistance > 0`

nicholaspai (approved):
This looks great, have left three nit suggestions for adding `assert()` checks that will cause the function to error quickly and clearly

src/utils.ts
```diff
@@ -303,3 +313,48 @@ export async function createUnsignedFillRelayTransaction(
...
+export async function findBlockAtOrOlder(provider: providers.Provider, desiredLookback: number): Promise<number> {
+  const [toBlock, network] = await Promise.all([provider.getBlock("latest"), provider.getNetwork()]);
+  let toBlockTimestamp = toBlock.timestamp;
+  const desiredTimestamp = toBlockTimestamp - desiredLookback;
+  let skipDistance = BlockScanSkipDistances[network.chainId];
+
+  // Fetch the first block to get the block production speed estimate before proceeding further.
+  let fromBlockNumber = toBlock.number - skipDistance;
+  let fromBlock = await provider.getBlock(fromBlockNumber);
+  let fromBlockTimestamp = fromBlock.timestamp;
```
mrice32, line 347: nit: I suspect there might be a way to flip the order that things are done in the while loop so you don't have to repeat some of the logic in the loop up here.

I think this might simplify things a little.

If not possible, no worries.

src/utils.ts
mrice32: +1

src/utils.ts
mrice32: +1

mrice32 (approved):
Looks good! One minor comment about flipping the order so you have to do less work outside the loop.

src/utils.ts
kevinuma: I originally had it that way but this is slightly cleaner - otherwise the first iteration would need to skip apply the multiplier on skip distance (since fromBlockTimestamp would start out being equal to toBlockTimestamp) so multiplier would be 0.

src/utils.ts
kevinuma: Added an assert that skipDistance > 0

src/utils.ts
kevinuma: Updated

src/utils.ts
kevinuma: Updated

PR #132, opened 2022-12-02
improve(pool): ✨ rework pool fee logic

james-a-morris:
This change aims to account for a user's staking position when calculating pool fee accruals.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.1.35",
+  "version": "0.2.0",
   "license": "AGPL-3.0",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",


src/pool/poolClient.ts
```diff
@@ -315,12 +320,13 @@ export function previewRemoval(
 }
 function joinUserState(
   poolState: Pool,
+  stakeData: StakeData,
```
daywiss, line 323: tiny nits, you can simplify this type to a bignumber and just pass in `stakeData.cumulativeBalance` at the caller. also as a general principle, i like to add new parameters to the function at the end, to maintain backward compatibilty.  if you put this at the end you could also default this to 0 in the case where there is no cumulativeBalance.

src/pool/poolClient.ts
```diff
@@ -444,6 +458,18 @@ export class Client {
...
   }
+  async resolveStakingData(
+    l1TokenAddress: string,
+    userState: Awaited<ReturnType<UserState["read"]>>
+  ): Promise<StakeData> {
+    assert(this.config.acceleratingDistributorAddress, "Must have the accelerating distributor address");
+    const contract = this.getOrCreateAcceleratingDistributorContract();
+    const { cumulativeBalance } = await contract.getUserStake(l1TokenAddress, userState.address);
+    return {
+      cumulativeBalance,
```
daywiss, line 469: id just return cumulative balance by itself rather than in an object if you dont care about anything else

src/pool/poolClient.ts
```diff
@@ -315,12 +320,13 @@ export function previewRemoval(
...
   poolState: Pool,
+  stakeData: StakeData,
   tokenEventState: hubPool.TokenEventState,
   userState: Awaited<ReturnType<UserState["read"]>>,
   transferValue: BigNumberish = 0
 ): User {
   const positionValue = BigNumber.from(poolState.exchangeRateCurrent)
-    .mul(userState.balanceOf)
+    .mul(userState.balanceOf.add(stakeData.cumulativeBalance))
     .div(fixedPointAdjustment);
```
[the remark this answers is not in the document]
daywiss: the math seems like it would work, but were u able to sanity check this either in the UI or with an e2e test just to see if the numbers look right?

nicholaspai (approved):

src/pool/poolClient.ts
```diff
@@ -444,6 +464,49 @@ export class Client {
...
+
+    // Logic: We have a list of staked transactions + a list of unstake
+    // transactions + the user's current balance. We know that if we add
+    // back all the unstake transactions to the cumulative staked balance,
+    // we'll have a value that represents the total amount of funds that
+    // has ever been staked on a user's account. If we then subtract out
+    // the total amount of funds the user has specifically staked from the
+    // external LP pool, we'll be left with exactly how much the user has
+    // claimed through an airdrop.
+    const amountAirdropped = stakes.reduce((acc, previous) => acc.add(previous.amount.mul(-1)), cumulativeBalance);
```
daywiss, line 501: this looked weird until i realized u set unstakes to -1, interesting way to do this.

daywiss (approved):
cant vouch for the math but the approach looks good!

PR #134, opened 2022-12-08
feat!: Add config options to discount fees by destination chain ID

nicholaspai:
Adds a required param to relayerFeeCalculator's `relayerFeeDetails` function: `destinationChainId`. (I can see an argument for this being optional but the codes a bit cleaner to require it as optional and the way its used today, it should be easy to upgrade to the next SDK version and change the `relayerFeeDetails` call.

This PR redefines the interface for `gasDiscountPercent` and `capitalDiscountPercent` in the `RelayFeeCalculatorConfig` rather than add new properties. The reason I do this is because neither the Relayer or the Frontend uses these properties currently, and updating these interfaces in-place required less code than introducing new ons.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -20,8 +20,8 @@ export interface CapitalCostConfig {
 }
 export interface RelayFeeCalculatorConfig {
   nativeTokenDecimals?: number;
-  gasDiscountPercent?: number;
-  capitalDiscountPercent?: number;
+  gasDiscountPercent?: { [chainId: number]: number };
+  capitalDiscountPercent?: { [chainId: number]: number };
   feeLimitPercent?: number;
   capitalCostsPercent?: number;
   capitalCostsConfig?: { [token: string]: CapitalCostConfig };
@@ -61,19 +61,11 @@ export class RelayFeeCalculator {
 
   constructor(config: RelayFeeCalculatorConfig, logger: Logger = DEFAULT_LOGGER) {
     this.queries = config.queries;
-    this.gasDiscountPercent = config.gasDiscountPercent || 0;
-    this.capitalDiscountPercent = config.capitalDiscountPercent || 0;
+    this.gasDiscountPercent = config.gasDiscountPercent || {};
+    this.capitalDiscountPercent = config.capitalDiscountPercent || {};
     this.feeLimitPercent = config.feeLimitPercent || 0;
     this.nativeTokenDecimals = config.nativeTokenDecimals || 18;
     this.capitalCostsPercent = config.capitalCostsPercent || 0;
-    assert(
-      this.gasDiscountPercent >= 0 && this.gasDiscountPercent <= 100,
-      "gasDiscountPercent must be between 0 and 100 percent"
-    );
-    assert(
-      this.capitalDiscountPercent >= 0 && this.capitalDiscountPercent <= 100,
-      "capitalDiscountPercent must be between 0 and 100 percent"
-    );
     assert(
       this.feeLimitPercent >= 0 && this.feeLimitPercent <= 100,
       "feeLimitPercent must be between 0 and 100 percent"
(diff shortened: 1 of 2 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -155,11 +159,24 @@ export class RelayFeeCalculator {
...
+    tokenSymbol: string,
+    destinationChainId: number,
+    tokenPrice?: number
+  ) {
     let isAmountTooLow = false;
-    const gasFeePercent = await this.gasFeePercent(amountToRelay, tokenSymbol, tokenPrice);
+    const rawGasFeePercent = await this.gasFeePercent(amountToRelay, tokenSymbol, tokenPrice);
+    const gasFeeDiscountPercent = this.gasDiscountPercent[destinationChainId]
+      ? toBN(100).sub(this.gasDiscountPercent[destinationChainId])
+      : toBN(100);
```
pxrl, line 172: Is there a naming clash here? We seem to specify gasDiscountPercent during the relayerFeeCalculator() constructor, but here we sub that amount from 100. Wouldn't that instead make it a multiplier, rather than a discount?

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -61,19 +61,11 @@ export class RelayFeeCalculator {
 
   constructor(config: RelayFeeCalculatorConfig, logger: Logger = DEFAULT_LOGGER) {
     this.queries = config.queries;
-    this.gasDiscountPercent = config.gasDiscountPercent || 0;
-    this.capitalDiscountPercent = config.capitalDiscountPercent || 0;
+    this.gasDiscountPercent = config.gasDiscountPercent || {};
+    this.capitalDiscountPercent = config.capitalDiscountPercent || {};
```
pxrl, line 65: Does it make sense to discount the relayer fees at all?

The benefit of per-destination pricing is that we can reduce the burden on HubPool funds. The relayer itself is unaffected by this - its costs are still the same, and it is still without its funds until the next bundle is executed.

I think it's more logical to me to apply a discount to the LP fee, rather than the relayer fee.

mrice32: Yeah, +1 to this question. I'm also curious about the motivation.

nicholaspai: Oh wow I hadn't thought about this. So an LP fee discount could be implemented by adding a key to the `l1TokenConfig` state in the [`AcrossConfigStore`](https://etherscan.io/address/0x3b03509645713718b78951126e0a6de6f10043f5#readContract) like:

`lpFeeDiscount: { [chainId]: number >= 0 && number <= 100 }`

And then the relayer could take the lpFeeDiscount for the token and destination chain and apply it to the lpFee [here](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/ConfigStoreClient.ts#L65) in the `ConfigStoreClient`

Thoughts on this design?

pxrl: Storing a discount on-chain will mean that we need to be pretty conservative in applying any discounts, since we won't be reflexive to changes in demand with a static discount. We risk ending up with a huge flood of funds rushing to L2s, and by the time we do react to it, we might have a high HubPool utilisation and a poor LP APY.

I think a nice solution would factor in the SpokePool balance on the destination chain in order to determine whether it's above or below a target balance, and apply an LP fee modifier accordingly. This is probably deceptively complex though, and there are definitely race conditions that might lead to unexpected outcomes.

As a meta question: When we are modifying the LP fee derivation, how do we ensure that it is rolled out amongst all relayers? I checked with one third-party relayer and they are using the SDK, so unless they synchronise an update of their SDK, they will end up making fills with LP fees that we disagree on. We also have various aggregators that are generating deposit transactions; we assume that they are querying the API suggested-fees endpoint, but it's hard to know for sure...

nicholaspai: I think we should deprecate this PR in favor of https://github.com/across-protocol/relayer-v2/pull/391 which is more flexible than discounting the default LP fee. We can set a per route rate model.

Regarding your meta-question: 

>"As a meta question: When we are modifying the LP fee derivation, how do we ensure that it is rolled out amongst all relayers?"

I think that this SDK version bump is the best way to roll this out, so we should make sure to implement the change in #391 change (along with a corresponding SDK change) to be backwards compatible and give time for other relayers to upgrade their versions. The `routeRateModels` won't be added initially so we will give relayers plenty of time to upgrade before the `routeRateModels` start affecting the LP fees

nicholaspai:
https://github.com/across-protocol/sdk-v2/pull/134#discussion_r1050364933

PR #135, opened 2022-12-12
fix: :bug: use merkle distributor to collect claims

james-a-morris:
This change uses the merkle distributor contract to find the total number of claimable funds that the user has. Originally funds were being calculated by just referencing the accelerating distributor, but this new method references the claim directly.

### DIFF
--- a/src/pool/poolClient.ts
+++ b/src/pool/poolClient.ts
@@ -16,6 +16,7 @@ import {
   MerkleDistributor,
   MerkleDistributor__factory,
 } from "@across-protocol/across-token";
+import { TypedEvent } from "@across-protocol/across-token/dist/typechain/common";
 
 const { erc20 } = uma.clients;
 const { loop } = uma.utils;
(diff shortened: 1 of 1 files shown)


mrice32 (commented):
You didn't change these lines in this PR, but I think this make break the transfer accounting that was previously built (but I could be wrong!).

My understanding is that you've added two terms to the math [here](https://github.com/across-protocol/sdk-v2/blob/10de215e1531c4fcb8d45e2fd28cc00c788ef402/src/pool/poolClient.ts#L338-L344).

You add the user's claim to their "deposits" (basically their starting balance). You also add their staking balance to their final position. This seems right.

There's a term in here called transferValue that is added to deposits that I think is now broken. This transfer value is intended to cover cases where positions are opened and closed _without_ an AddLiquidity or RemoveLiquidity event (by transferring in or out of the user's address).

This works correctly in the old world where transfers out of the user's address are the same as closing their LP position (somebody else is now earning the fees!) and transfers in are the same as opening an LP position (the user is now earning those fees).

In the new world where a user's position can be held in the AD contract and/or their wallet, this breaks because a transfer in or out might be to/from the AD contract, which _wouldn't_ be closing/opening an LP position. Does that make sense?

I think the fix is to ignore transfers where the party opposite the user is the AD contract, just like we do for the 0x0 address [here](https://github.com/across-protocol/sdk-v2/blob/10de215e1531c4fcb8d45e2fd28cc00c788ef402/src/pool/poolClient.ts#L267-L269). I think it should be as simple as two more && conditions in that statement.

Note: all of this really breaks if people call stakeFor directly, since they could be transferring _through_ the AD contract (meaning they would be opening and closing positions via staking), but if we just assume the AD is never do that, I think this all works.

@james-a-morris this analysis could be wrong, so please let me know if you think this is already correctly handled!!

PR #138, opened 2022-12-29
feat: export CONFIG_STORE_VERSION

nicholaspai:
See https://github.com/UMAprotocol/UMIPs/pull/567 for corresponding UMIP change

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -11,6 +11,9 @@ export type BN = BigNumber;
 export type Decimalish = string | number | Decimal;
 export const AddressZero = ethers.constants.AddressZero;
 
+// Used as a signal to protect relayers from running code against an outdated version of the on-chain config store.
+export const CONFIG_STORE_VERSION = "1";
+
 const { ConvertDecimals } = uma.utils;
 // These are distances used to traverse when looking for a block with desired lookback.
 // They're meant to be small enough to allow for granularity but large enough to minimize the number of reqests needed
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -12,7 +12,7 @@ export type Decimalish = string | number | Decimal;
 export const AddressZero = ethers.constants.AddressZero;
 
 // Used as a signal to protect relayers from running code against an outdated version of the on-chain config store.
-export const CONFIG_STORE_VERSION = "1";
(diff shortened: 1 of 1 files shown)


mrice32 (approved):
LGTM!

pxrl:
Hm..looks like I was halfway through writing this yesterday and forgot about it :facepalm:

Following on from [this](https://github.com/across-protocol/relayer-v2/pull/398#discussion_r1058617097) thread, what do you think about migrating the actual ConfigStore query implementation here? IMO would be nice for a common implementation that is easy for users to call, that gives them a binary result as to whether they can proceed with a fill or not.

nicholaspai:
> Hm..looks like I was halfway through writing this yesterday and forgot about it :facepalm:
> 
> 
> 
> Following on from [this](https://github.com/across-protocol/relayer-v2/pull/398#discussion_r1058617097) thread, what do you think about migrating the actual ConfigStore query implementation here? IMO would be nice for a common implementation that is easy for users to call, that gives them a binary result as to whether they can proceed with a fill or not.

So do you suggest moving the full config store client into the sdk? I think this has pros and cons. The cons are that a lot of the other helper utils are already in the relayer v2 code. 

I think at least exporting the latest config store version here allows relayers to know which is the latest version

pxrl:
> So I could see there being an argument to moving the entire ConfigStoreClient to the SDK, but that would be a very big move

Yeah; I can see that. In hindsight I do think that (almost?) everything configstore-related belongs in the SDK, but there's no reason for that to stall this current work.

How about we just record a relayer-v2 => sdk-v2 migration as a backlog item, to be addressed when we have some spare cycles?

nicholaspai:
> > So I could see there being an argument to moving the entire ConfigStoreClient to the SDK, but that would be a very big move
> 
> Yeah; I can see that. In hindsight I do think that (almost?) everything configstore-related belongs in the SDK, but there's no reason for that to stall this current work.
> 
> How about we just record a relayer-v2 => sdk-v2 migration as a backlog item, to be addressed when we have some spare cycles?

https://linear.app/uma/issue/ACX-542/migrate-configstoreclient-to-sdk-v2-from-relayer-v2

PR #140, opened 2023-01-06
feat: Add route rate model

nicholaspai:
Origin and Destination chain are new optional params

### DIFF
--- a/src/pool/poolClient.ts
+++ b/src/pool/poolClient.ts
@@ -787,6 +787,9 @@ export class Client {
 
     let rateModel: acrossConfigStore.RateModel | undefined = undefined;
     try {
+      // Use the default rate model (i.e. not any of the routeRateModels to project the Pool's APR). This assumes
+      // that the default rate model is the most often used, but this may change in future if many different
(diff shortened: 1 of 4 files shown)


src/pool/poolClient.ts
```diff
@@ -787,6 +787,9 @@ export class Client {
 
     let rateModel: acrossConfigStore.RateModel | undefined = undefined;
     try {
+      // Use the default rate model (i.e. not any of the routeRateModels to project the Pool's APR). This assumes
+      // that the default rate model is the most often used, but this may change in future if many different
+      // route rate models are set.
```
nicholaspai, line 792: @daywiss you wrote this logic originally, now that we support deposit-path specific rate models, are we still ok using as the projected APR? @mrice32 @pxrl wdyt

mrice32: I think we want to get rid of projected APR anyway. I think this is fine for now and still a decent approximation since it uses the default rate model.

mrice32 (approved):
LGTM!

PR #141, opened 2023-01-10
feat: export token definitions

dohaki:
Effort to centralize token definitions in the SDK to reduce duplication on the FE.


src/relayFeeCalculator/chain-queries/ethereum.ts
```diff
@@ -1,87 +1,12 @@
...
-    decimals: 18,
-  },
-  KOR: {
-    address: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
-    decimals: 18,
-  },
-  ARETH: {
-    address: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
-    decimals: 18,
-  },
```
dohaki, line 38: Not sure if we really need these definitions? I searched through the FE if these symbols are used but couldn't find any places where we use them, so I opted to not include them in `TOKEN_SYMBOLS_MAP`. Correct me if I am wrong.

src/constants.ts
```diff
@@ -0,0 +1,80 @@
+export const TOKEN_SYMBOLS_MAP = {
```
nicholaspai, line 1: Should we add a comment that these should all be checksummed?

nicholaspai (approved):
This looks great! Can you also add some non-mainnet addresses? Perhaps its better to have a property `addresses: { [chainId: number]: string }` where we fill in L2 addresses in addition to mainnet. You can use the following as a guide:

https://github.com/across-protocol/relayer-v2/blob/master/src/common/Constants.ts#L116

src/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+  },
+  ETH: {
+    name: "Ether",
+    symbol: "ETH",
+    decimals: 18,
+    mainnetAddress: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+  },
+  MATIC: {
+    name: "Matic",
+    symbol: "WMATIC",
```
pxrl, line 40: MATIC/WMATIC is an odd one, but the symbol is MATIC on Ethereum mainnet and WMATIC on Polygon.

Do we rely on it being called WMATIC when we are referring to the Ethereum mainnet variant?

dohaki: Hmm good catch. I think this should be `MATIC` yea

src/constants.ts
dohaki: Added here https://github.com/across-protocol/sdk-v2/pull/141/commits/c96a884bc1daf827578340fb87f5e2b36bc5383a#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R12

src/constants.ts
dohaki: Changed here https://github.com/across-protocol/sdk-v2/pull/141/commits/c96a884bc1daf827578340fb87f5e2b36bc5383a#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R85

dohaki:
> This looks great! Can you also add some non-mainnet addresses? Perhaps its better to have a property `addresses: { [chainId: number]: string }` where we fill in L2 addresses in addition to mainnet. You can use the following as a guide:
> 
> https://github.com/across-protocol/relayer-v2/blob/master/src/common/Constants.ts#L116

Thanks for the suggestion. Changed here https://github.com/across-protocol/sdk-v2/pull/141/commits/c96a884bc1daf827578340fb87f5e2b36bc5383a

src/constants.ts
```diff
@@ -0,0 +1,165 @@
...
+};
+
+// NOTE: All addresses should be checksummed
+export const TOKEN_SYMBOLS_MAP = {
+  ACX: {
+    name: "ACX",
+    symbol: "ACX",
+    decimals: 18,
+    addresses: {
+      [CHAIN_IDs.MAINNET]: "0x44108f0223A3C3028F5Fe7AEC7f9bb2E66beF82F",
```
nicholaspai, line 18: +1 on using the CHAIN_IDS constant

src/constants.ts
```diff
@@ -0,0 +1,165 @@
...
+    addresses: {
+      [CHAIN_IDs.MAINNET]: "0x6B175474E89094C44Da98b954EedeAC495271d0F",
+      [CHAIN_IDs.OPTIMISM]: "0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1",
+      [CHAIN_IDs.POLYGON]: "0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063",
+      [CHAIN_IDs.BOBA]: "0xf74195Bb8a5cf652411867c5C2C5b8C2a402be35",
+      [CHAIN_IDs.ARBITRUM]: "0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1",
+      [CHAIN_IDs.GOERLI]: "0x5C221E77624690fff6dd741493D735a17716c26B",
+    },
+  },
+  ETH: {
```
nicholaspai, line 69: Should this be WETH instead of ETH?

nicholaspai: Or should this ETH key be deleted entirely?

nicholaspai (commented):
Really well done and I love the use of the CHAIN_IDS struct so we don't hardcode network ID's. Have one comment on the `ETH` key

src/constants.ts
dohaki: We also have `WETH` as a separate key [here](https://github.com/across-protocol/sdk-v2/pull/141/files/33852a83bc8bbfed0df37de883e603bdee8ece50#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R142).

I am not sure if we can remove this safely because on the FE it seems that we sometimes use both 🤔 I would leave this for now and if we are sure that we don't need this on the FE then I can remove it.

nicholaspai: ok sgtm

PR #148, opened 2023-01-12
feat: :sparkles: include API client in the SDK

james-a-morris:
Multiple repositories currently use the Across Serverless Function API interface to access critical information. Because logic to call this API is duplicated, it stands that we should centralize its definition in the SDK-V2 repository.

### DIFF
--- a/src/apiClient/types.ts
+++ b/src/apiClient/types.ts
@@ -2,7 +2,7 @@ import { ethers } from "ethers";
 
 export type CoingeckoDataReturnType = { price: ethers.BigNumber };
 
-type Fee = {
+export type Fee = {
   total: ethers.BigNumber;
   pct: ethers.BigNumber;
 };


nicholaspai:
This is awesome, will review next week!

src/apiClient/mockedClient.ts
```diff
@@ -0,0 +1,90 @@
...
+  SuggestedFeeReturnType,
+} from "./types";
+
+/**
+ * MockedApiClient is a mocked implementation of the AbstractApiClient.
+ * @class
+ * @extends AbstractApiClient
+ * @note This implementation does not make API calls to RESTful services. Instead, it returns mocked data.
+ */
+export default class MockedApiClient extends AbstractApiClient {
```
daywiss, line 19: what is this going to be used for?

src/apiClient/types.ts
```diff
@@ -0,0 +1,49 @@
+import { ethers } from "ethers";
+
+export type CoingeckoDataReturnType = { price: ethers.BigNumber };
+
+type Fee = {
```
daywiss, line 5: might want to export this jsut in case

src/apiClient/abstractClient.ts
```diff
@@ -0,0 +1,96 @@
...
+  CoingeckoDataReturnType,
+  SuggestedFeeReturnType,
+  BridgeLimitsReturnType,
+  AcrossBridgeStatisticsType,
+} from "./types";
+
+/**
+ * AbstractApiClient is an abstract class that defines the interface for the API client.
+ */
+export default abstract class AbstractApiClient {
```
daywiss, line 13: im not sure the intent of this, but i assume its so that you can create a mock with the same interface as the production class. i thought maybe there was a way you could avoid this by just extending and overriding the production class itself in the mock, but id have to play around with it to be sure.  it just seems like theres a lot of redundant code. not a criticism just looking for a way to simplify!

daywiss (approved):
main feedback id give here is just with the naming of this library, api client is so generic, would prefer something more descriptive, but i see that this is kind of just a pass through to various apis, so maybe its appropriate.

src/apiClient/abstractClient.ts
mrice32: I totally hear this argument. There's a bit of stylistic preference here.

Simple interfaces can be helpful as a clean way to document the expected behavior of the class. However, they definitely add boilerplate, especially in cases where there's only one implementation of the interface used in production (like this).

Either way y'all decide to go, LGTM!

mrice32 (approved):
This looks great!

daywiss (approved):

amateima (approved):
Looks very good. I'll add an issue in the Linear backlog to integrate this into the Scraper API source code and replace the duplicated logic for calling the suggested fees endpoint

PR #152, opened 2023-02-05
feat: :sparkles: add override to relayer fee

james-a-morris:
### Motivation
To stay competitive in the bridging space, we need to reduce fees to zero (both lp and relayer fee to zero) for specific routes. 

### PR Summary
This PR adds an optional override structure that a calling code can use to include a per-route specification for capital cost configs.

### Results
The `RelayerFeeCalculator` has now been extended to include a per-route capital cost configuration. This logic is optional, and previous usage will not be affected.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -2,7 +2,6 @@ import assert from "assert";
 import * as uma from "@uma/sdk";
 import { BigNumber } from "ethers";
 import { BigNumberish, toBNWei, nativeToToken, toBN, min, max, MAX_BIG_INT } from "../utils";
-import { ChainId } from "../transfers-history/adapters/web3/model";
 const { percent, fixedPointAdjustment } = uma.across.utils;
 
 // This needs to be implemented for every chain and passed into RelayFeeCalculator
@@ -198,8 +197,8 @@ export class RelayFeeCalculator {
   async capitalFeePercent(
     _amountToRelay: BigNumberish,
     _tokenSymbol: string,
-    _originRoute?: ChainId,
-    _destinationRoute?: ChainId
+    _originRoute?: ChainIdAsString,
+    _destinationRoute?: ChainIdAsString
   ): Promise<BigNumber> {
     // If amount is 0, then the capital fee % should be the max 100%
     if (toBN(_amountToRelay).eq(toBN(0))) return MAX_BIG_INT;
@@ -252,8 +251,8 @@ export class RelayFeeCalculator {
     amountToRelay: BigNumberish,
     tokenSymbol: string,
     tokenPrice?: number,
-    _originRoute?: ChainId,
-    _destinationRoute?: ChainId
+    _originRoute?: ChainIdAsString,
+    _destinationRoute?: ChainIdAsString
   ): Promise<RelayerFeeDetails> {
     const gasFeePercent = await this.gasFeePercent(amountToRelay, tokenSymbol, tokenPrice);
     const gasFeeTotal = gasFeePercent.mul(amountToRelay).div(fixedPointAdjustment);


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -18,13 +18,20 @@ export interface CapitalCostConfig {
...
 export interface RelayFeeCalculatorConfig {
   nativeTokenDecimals?: number;
   gasDiscountPercent?: number;
   capitalDiscountPercent?: number;
   feeLimitPercent?: number;
   capitalCostsPercent?: number;
-  capitalCostsConfig?: { [token: string]: CapitalCostConfig };
+  capitalCostsConfig?: {
+    [token: string]: CapitalCostConfig | CapitalCostConfigOverride;
+  };
```
daywiss, line 34: personally would have made another property here called `capitalCostsConfigOverride` 
 which is the same shape as `capitalCostsConfig` and prioritized using that before falling back to `capitalCostsConfig`. i think this would have simplified some of the code as you wouldnt need special checks, just a fallback if a value didnt exist.

daywiss (approved):
comment on design, but otherwise it lgtm

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -99,15 +106,63 @@ export class RelayFeeCalculator {
...
+   * @private
+   * @dev This is a type guard that is used to check if a config is a CapitalCostConfigOverride or a CapitalCostConfig.
+   * This is needed because the config can be either a CapitalCostConfig or a CapitalCostConfigOverride. If it's a
+   * CapitalCostConfig, then we need to convert it to a CapitalCostConfigOverride with the default config set with no route
+   * overrides.
+   */
+  private static capitalCostConfigIsOverride(
+    config: CapitalCostConfig | CapitalCostConfigOverride
+  ): config is CapitalCostConfigOverride {
+    return (config as CapitalCostConfigOverride).default !== undefined;
```
mrice32, line 132: Note: (stylistic, you can do it either way) I think typescript will let you do this without casting if you do this: `"default" in config`.

src/relayFeeCalculator/relayFeeCalculator.ts
mrice32: I'm not strongly opinionated, as this is pretty tricky. If we didn't have to worry about backcompat, my preference would be a config that looks like this:
```ts
export interface RelayFeeCalculatorConfig {
  ...
  defaultCapitalCostsPercent: number;
  tokenCapitalCosts?: { [token: string]: number };
  // Note: routes are encoded as strings 1-10 would be the route from chain ID 1 to chain ID 10.
  routeCapitalCosts?: { [token: string]: { [route: string]: number } };
}
```

In that world, you could just store these and do a series of fallbacks that's pretty easy to understand:
```ts
const routeCost = routeCapitalCosts?.[token]?.[routeString];
if (routeCost) return routeCost;
const tokenCost = tokenCapitalCosts?.[token];
if (tokenCost) return tokenCost;
return defaultCost;
```

I think this strategy would get rid of a lot of the pre-parsing and runtime checks that we are doing in this PR and reduce the complexity of the nesting, at the cost of having more config variables.

However, one problem is that we _may_ have to deal with capitalCostsConfig for backwards compatibility reasons. Although, we could publish a new breaking change where we immediately drop this config variable. Then in other applications, when we update to that version, we have to update our config passing simultaneously.

Thoughts? Personally, I think the existing approach is reasonable, just wanted to mention a possible alternative (that may or may not be better).

mrice32 (commented):
Looks great! Only minor comments

src/relayFeeCalculator/relayFeeCalculator.ts
daywiss: @mrice32 @james-a-morris yes perfectly reasonable, i was probably missing some context, but either approach seems fine to me

daywiss (approved):

src/relayFeeCalculator/relayFeeCalculator.ts
mrice32: Agreed, either way seems okay. I would personally prefer less tech debt and complexity over backwards compatibility since we have explicit releases to protect end-users.

Ultimately up to you @james-a-morris!

mrice32 (commented):
https://github.com/across-protocol/sdk-v2/pull/152#discussion_r1100978479

PR #156, opened 2023-03-01
improve(PriceClient): Inherit base HTTP adapter args

pxrl:
Simplifies usage of the base HTTP class in advance of adding support for retries.

### DIFF
--- a/src/priceClient/adapters/baseAdapter.ts
+++ b/src/priceClient/adapters/baseAdapter.ts
@@ -1,8 +1,12 @@
 import assert from "assert";
 import axios, { AxiosError } from "axios";
 
+export type BaseHTTPAdapterArgs = {
+  timeout?: number;
+};
+
 export class BaseHTTPAdapter {
-  private _timeout: number;
+  private _timeout = 0;
 
   get timeout(): number {
     return this._timeout;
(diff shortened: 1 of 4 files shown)


src/priceClient/adapters/baseAdapter.ts
```diff
@@ -13,8 +17,8 @@ export class BaseHTTPAdapter {
     this._timeout = timeout;
   }
 
-  constructor(public readonly name: string, public readonly host: string, { timeout }: { timeout?: number }) {
-    this._timeout = timeout ?? 1000; // ms
+  constructor(public readonly name: string, public readonly host: string, { timeout = 1000 }: BaseHTTPAdapterArgs) {
+    this.timeout = timeout; // ms
```
james-a-morris, line 21: Should this be `this._timeout`?

mrice32 (approved):
LGTM!

src/priceClient/adapters/baseAdapter.ts
pxrl: > I assume it's supposed to go through the setter that has the automatic check?

Yep, correct. It was actually an error that it didn't do this beforehand, because it meant that the initial values passed to the constructor were never validated.

PR #163, opened 2023-04-05
feat: add error interface to common interfaces

james-a-morris:
This PR adds the new relayer Error interface into the SDK

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.4.3",
+  "version": "0.4.4",
   "license": "AGPL-3.0",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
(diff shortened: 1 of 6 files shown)


pxrl (commented):
Thoughts on also adding the isEthersError predicate from TransactionUtils.ts? It's a handy companion utility.

```
const isEthersError = (error?: unknown): error is EthersError =>
    (error as EthersError)?.code in ethers.utils.Logger.errors;
```

james-a-morris:
> d30daf6

Fair point! Added it here: https://github.com/across-protocol/sdk-v2/pull/163/commits/d30daf6611945df0a02b5a592acc67224ca094c6

james-a-morris:
> Thoughts on also adding the isEthersError predicate from TransactionUtils.ts? It's a handy companion utility.
> 
> ```
> const isEthersError = (error?: unknown): error is EthersError =>
>     (error as EthersError)?.code in ethers.utils.Logger.errors;
> ```
> 
> Not sure exactly where it should go; I guess src/interfaces seems to be mostly reserved for type and interface definitions, rather than code. wdyt?

Fair point! Added it here: https://github.com/across-protocol/sdk-v2/pull/163/commits/d30daf6611945df0a02b5a592acc67224ca094c6

pxrl (approved):
Great!

PR #166, opened 2023-04-12
feat: update deposit typed data utility

dohaki:
This is part of ACX-643. Adding this to the SDK makes sense to me in case other projects/components than our FE needs to speed up deposits.

### DIFF
--- a/src/interfaces/TypedData.ts
+++ b/src/interfaces/TypedData.ts
@@ -1,0 +1,21 @@
+export interface MessageTypeProperty {
+  name: string;
+  type: string;
+}
+
+export interface MessageTypes {
+  [additionalProperties: string]: MessageTypeProperty[];
+}
+
+export interface TypedMessage<T extends MessageTypes> {
+  types: T;
+  primaryType: keyof T;
+  domain: {
+    name?: string;
+    version?: string;
+    chainId?: number;
+    verifyingContract?: string;
+    salt?: ArrayBuffer;
+  };
+  message: Record<string, unknown>;
+}
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,6 +5,7 @@ import { isL2Provider as isOptimismL2Provider, L2Provider } from "@eth-optimism/
 import { SpokePool } from "@across-protocol/contracts-v2";
 import assert from "assert";
 import { GasPriceEstimate, getGasPriceEstimate } from "./gasPriceOracle";
+import { TypedMessage } from "./interfaces/TypedData";
 
 export type BigNumberish = string | number | BigNumber;
 export type BN = BigNumber;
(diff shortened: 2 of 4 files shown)


src/utils.ts
```diff
@@ -370,3 +370,46 @@ export async function findBlockAtOrOlder(provider: providers.Provider, desiredLo
...
+        { name: "originChainId", type: "uint256" },
+        { name: "updatedRelayerFeePct", type: "int64" },
+        { name: "updatedRecipient", type: "address" },
+        { name: "updatedMessage", type: "bytes" },
+      ],
+    },
+    domain: {
+      name: "ACROSS-V2",
+      version: "1.0.0",
+      chainId: Number(originChainId),
```
pxrl, line 405: Is the cast necessary? originChainId is already specified as a number in the function declaration.

dohaki: yes good point. removed it here https://github.com/across-protocol/sdk-v2/pull/166/commits/250c572371211142102eab2f23288a1404a0e10d

src/utils.ts
```diff
@@ -370,3 +370,46 @@ export async function findBlockAtOrOlder(provider: providers.Provider, desiredLo
...
+ * @param updatedMessage The new message that should be provided to the recipient.
+ * @return EIP-712 compliant typed data.
+ */
+export function getUpdateDepositTypedData(
+  depositId: number,
+  originChainId: number,
+  updatedRelayerFeePct: BigNumber,
+  updatedRecipient: string,
+  updatedMessage: string
+) {
```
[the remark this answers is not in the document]
dohaki: added here https://github.com/across-protocol/sdk-v2/pull/166/commits/250c572371211142102eab2f23288a1404a0e10d

src/utils.ts
```diff
@@ -370,3 +371,60 @@ export async function findBlockAtOrOlder(provider: providers.Provider, desiredLo
...
+export type UpdateDepositDetailsMessageType = {
+  UpdateDepositDetails: [
+    {
+      name: "depositId";
+      type: "uint32";
+    },
+    { name: "originChainId"; type: "uint256" },
+    { name: "updatedRelayerFeePct"; type: "int64" },
+    { name: "updatedRecipient"; type: "address" },
+    { name: "updatedMessage"; type: "bytes" }
```
james-a-morris, line 384: Is this part of the ABI or something that would be exposed by TypeChain?

src/interfaces/TypedData.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface TypedMessage<T extends MessageTypes> {
+  types: T;
+  primaryType: keyof T;
+  domain: {
+    name?: string;
+    version?: string;
+    chainId?: number;
+    verifyingContract?: string;
+    salt?: ArrayBuffer;
```
mrice32, line 18: nit: why are all these fields optional? Wouldn't we either always set them or never set them for our implementation? Or am I missing something?

mrice32 (approved):
Looks awesome! Very clean! Only one minor comment

src/utils.ts
dohaki: Nope, AFAICT this is not exposed by TypeChain because this struct is not part of the ABI... There is a constant in the `SpokePool.sol` contract
```solidity
    bytes32 public constant UPDATE_DEPOSIT_DETAILS_HASH =
        keccak256(
            "UpdateDepositDetails(uint32 depositId,uint256 originChainId,int64 updatedRelayerFeePct,address updatedRecipient,bytes updatedMessage)"
        );
```
but TypeChain can't extract types for EIP-712 compliant data it seems from that alone.

src/interfaces/TypedData.ts
dohaki: Yea, in our implementation, we only set `name`, `version`, and `chainId`, but according to the spec, all these fields are optional. See https://eips.ethereum.org/EIPS/eip-712.

My initial goal was to make this specific type generic so that other typed data definitions could be implemented by it in the future.

pxrl (approved):
OK from me. Nice work with all the extra annotation. James might have some additional comments.

PR #169, opened 2023-04-13
feat: verify types are supported with spoke pool upgrade

james-a-morris:
This PR aims to bring the SDK in a compilable state with the new spoke pool upgrades added.

### DIFF
--- a/src/transfers-history/client.ts
+++ b/src/transfers-history/client.ts
@@ -206,14 +206,14 @@ export class TransfersHistoryClient {
 
   private insertFilledRelayEvent(event: FilledRelayEvent) {
     const { args, transactionHash } = event;
-    const { totalFilledAmount, depositor, depositId, originChainId, appliedRelayerFeePct } = args;
+    const { totalFilledAmount, depositor, depositId, originChainId, relayerFeePct } = args;
     this.transfersRepository.updateFilledAmount(
       originChainId.toNumber(),
       depositor,
       depositId,
       totalFilledAmount,
       transactionHash,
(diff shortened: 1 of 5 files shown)


nicholaspai (approved):

src/transfers-history/client.ts
```diff
@@ -206,14 +206,14 @@ export class TransfersHistoryClient {
...
-    const { totalFilledAmount, depositor, depositId, originChainId, appliedRelayerFeePct } = args;
+    const { totalFilledAmount, depositor, depositId, originChainId, relayerFeePct } = args;
     this.transfersRepository.updateFilledAmount(
       originChainId.toNumber(),
       depositor,
       depositId,
       totalFilledAmount,
       transactionHash,
-      appliedRelayerFeePct
+      relayerFeePct
```
nicholaspai, line 216: Does this properly handle speed up events?

dohaki: There is a separate handler for that here https://github.com/across-protocol/sdk-v2/blob/1dcf0a1adad7b07b1cf1ef1d38db57ddf1e544ac/src/transfers-history/client.ts#L220

dohaki: AFAICT the new ABIs shouldn't break anything for the speed-ups, but we should in a separate PR adjust the `Transfer` model and the respective handlers. So that they support the new fields `updatedMessage` and `updatedRecipient`.

dohaki (approved):
LGTM 👍

PR #170, opened 2023-04-14
fix(UBA): Tweak predicates and add test

pxrl:
The existing predicates didn't quite work as expected because the
depositId field exists on all of the UbaFlow union types. This change
changes the isUbaInflow() predicate to look at the deposit-exclusive
quoteTimestamp field, and adds some additional predicates to support
sub-classifying a UbaOutflow. While here, add some tests.

### DIFF
--- a/src/interfaces/UBA.ts
+++ b/src/interfaces/UBA.ts
@@ -11,9 +11,17 @@ export type UbaRunningRequest = {
 };
 
 export const isUbaInflow = (flow: UbaFlow): flow is UbaInflow => {
-  return (flow as UbaInflow).depositId !== undefined;
+  return (flow as UbaInflow).quoteTimestamp !== undefined;
 };
 
 export const isUbaOutflow = (flow: UbaFlow): flow is UbaOutflow => {
-  return !isUbaInflow(flow);
+  return !isUbaInflow(flow) && (outflowIsFill(flow) || outflowIsRefund(flow));
+};
+
+export const outflowIsFill = (outflow: UbaOutflow): outflow is FillWithBlock => {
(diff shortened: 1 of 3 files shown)


src/interfaces/UBA.ts
```diff
@@ -11,9 +11,17 @@ export type UbaRunningRequest = {
 };
 
 export const isUbaInflow = (flow: UbaFlow): flow is UbaInflow => {
-  return (flow as UbaInflow).depositId !== undefined;
+  return (flow as UbaInflow).quoteTimestamp !== undefined;
```
james-a-morris, line 14: Good catch, @pxrl !

src/interfaces/UBA.ts
```diff
@@ -11,9 +11,17 @@ export type UbaRunningRequest = {
...
+  return !isUbaInflow(flow) && (outflowIsFill(flow) || outflowIsRefund(flow));
+};
+
+export const outflowIsFill = (outflow: UbaOutflow): outflow is FillWithBlock => {
+  return (outflow as FillWithBlock).isSlowRelay !== undefined;
+};
+
+export const outflowIsRefund = (outflow: UbaOutflow): outflow is RefundRequestWithBlock => {
+  return (outflow as RefundRequestWithBlock).fillBlock !== undefined;
 };
```
james-a-morris, line 27: Nit: now that we have a few of these predicates, would it make sense to place these in their own file?

james-a-morris (commented):
Left a few comments 🙂

james-a-morris (approved):
Approved so that @pxrl can merge #174

PR #174, opened 2023-04-17
refactor: Relocate UBA type predicates + test

pxrl:
Continuing on from discussion here: https://github.com/across-protocol/sdk-v2/pull/170#pullrequestreview-1385628719

My gut feel is that it's probably premature to split these and that it's more convenient to keep them in the same file.

### DIFF
--- a/src/typeguards/UBA.ts
+++ b/src/typeguards/UBA.ts
@@ -1,0 +1,17 @@
+import { FillWithBlock, RefundRequestWithBlock, UbaFlow, UbaInflow, UbaOutflow } from "../interfaces";
+
+export const isUbaInflow = (flow: UbaFlow): flow is UbaInflow => {
+  return (flow as UbaInflow).quoteTimestamp !== undefined;
+};
+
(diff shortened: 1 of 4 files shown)


src/typeguards/UBA.ts
```diff
@@ -0,0 +1,17 @@
...
+export const isUbaInflow = (flow: UbaFlow): flow is UbaInflow => {
+  return (flow as UbaInflow).quoteTimestamp !== undefined;
+};
+
+export const isUbaOutflow = (flow: UbaFlow): flow is UbaOutflow => {
+  return !isUbaInflow(flow) && (outflowIsFill(flow) || outflowIsRefund(flow));
+};
+
+export const outflowIsFill = (outflow: UbaOutflow): outflow is FillWithBlock => {
+  return (outflow as FillWithBlock).isSlowRelay !== undefined;
```
nicholaspai, line 12: This should be `updatableRelayData.isSlowRelay`, have you tried rebasing off my across-2.5 branch?

PR #180, opened 2023-04-20
chore(UBA): Tweak outflowIsFill predicate

pxrl:
Incorporate some fixes from Nick's across-2.5 branch, whilst also adjusting the outflowIsFill predicate to avoid the use of isSlowRelay.

Evaluating isSlowRelay was an unfortunate choice, because it gets relocated as part of the new SpokePool migration. Instead, use repaymentChainId because it's unique to a fill and is not impacted by the new ABI.

### DIFF
--- a/src/interfaces/UBA.ts
+++ b/src/interfaces/UBA.ts
@@ -11,7 +11,7 @@ export type UbaRunningRequest = {
 };
 
 export const isUbaInflow = (flow: UbaFlow): flow is UbaInflow => {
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

nicholaspai:
How is this different from https://github.com/across-protocol/sdk-v2/pull/179/files?

pxrl:
> How is this different from https://github.com/across-protocol/sdk-v2/pull/179/files?

It's effectively the same; it just aligns master with across-2.5, so it'll disappear from your main PR.

PR #183, opened 2023-04-24
feat: migrate across config store & hub pool clients

james-a-morris:
This PR migrates the across config store and the hub pool client from the Across V2 Relayer.


evaldofelipe (approved):
LGMT!

src/clients/index.ts
```diff
@@ -0,0 +1,2 @@
+export * as AcrossConfigStoreClient from "./AcrossConfigStoreClient";
```
nicholaspai, line 1: I would not export everything as `AcrossConfigStoreClient`, otherwise you'll have to instantiate as a new client via `AcrossConfigStoreClient.AcrossConfigStoreClient`, right?

Maybe just `export { GLOBAL_CONFIG_STORE_KEYS, AcrossConfigStoreClient } from "./AcrossConfigStoreClient"`

nicholaspai (commented):
LGTM just one comment on exporting https://github.com/across-protocol/sdk-v2/pull/183/files#r1178278760

src/clients/index.ts
james-a-morris: Good call - I've updated it in a later commit below

src/utils/EventUtils.ts
```diff
@@ -0,0 +1,231 @@
...
+import { SortableEvent } from "../interfaces";
+
+const maxRetries = 3;
+const retrySleepTime = 10;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function spreadEvent(args: Result | undefined): any {
+  if (!args) {
+    throw new Error("No args provided");
+  }
```
pxrl, line 13: Not sure this is a good idea - Event.args is generated by ethers and this is the first place that we inspect it. It's not required to be defined in Ethers, so I imagine there could be theoretical scenarios where we arrive here and throw an error on a legitimate Event (though not one emitted by our contracts).

pxrl (commented):
General question to this; this PR obviously relocates a lot of code, but are there also additional modifications being made? If so (and something goes wrong), that might be hard to debug.

pxrl:
Some follow-up thoughts:
 - There are differences between this code and what we currently have in relayer-v2. For example, ConfigStoreClient in relayer-v2 depends on Redis, but here that dependency has been removed. Given that we need to factor out the necessary changes in relayer-v2 in order to do the migration over here, is it perhaps better to have that change implemented in relayer-v2 _before_ we merge this? I think it'd make it easier to reason about this migration if it was only migrating existing code (with some minor necessary adjustments), rather than refactoring _and_ migrating in the same change.
 - We don't pull across any tests from relayer-v2. This is presumably because they're all too tightly coupled to permit that. Any ideas on how we manage that, moving forward?

src/clients/AcrossHubPoolClient.ts
```diff
@@ -0,0 +1,629 @@
...
+    this.firstBlockToSearch = searchConfig.toBlock + 1; // Next iteration should start off from where this one ended.
+
+    this.logger.debug({ at: "HubPoolClient", message: "HubPool client updated!" });
+  }
+
+  private async fetchTokenInfoFromContract(address: string): Promise<L1Token> {
+    const token = new Contract(address, ERC20__factory.abi, this.hubPool.signer);
+    const [symbol, decimals] = await Promise.all([token.symbol(), token.decimals()]);
+    return { address, symbol, decimals };
+  }
```
pxrl, line 605: IMO it'd be nice if we could factor generic methods like this one out of the HubPoolClient before migration. In this case it _probably_ belongs in the SDK, but for any instances where it belongs in relayer-v2, we'd have to migrate it back over the SDK-v2/relayer-v2 boundary.

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -0,0 +1,446 @@
+import winston from "winston";
```
mrice32, line 1: These are really big files to review start to finish. Can you point out any a) areas you think deserve scrutiny or b) lines that were changed (assuming most of this was copied?)?

src/clients/AcrossHubPoolClient.ts
```diff
@@ -0,0 +1,629 @@
+import { assign, EventSearchConfig, MakeOptional } from "../utils";
```
mrice32, line 1: Same here.

src/clients/AcrossHubPoolClient.ts
pxrl, line 1: Small nit on naming: the entire context of the sdk-v2 repository is Across; what's the reasoning for prepending "Across" to the existing name "HubPoolClient" ?

PR #187, opened 2023-04-26
feat: Update to latest contracts-v2 release

nicholaspai:


hardhat.config.ts
```diff
@@ -0,0 +1,39 @@
+import * as dotenv from "dotenv";
+
+import { HardhatUserConfig } from "hardhat/config";
```
nicholaspai, line 3: I had trouble importing the `hardhat.config` from the `contracts-v2/dist` folder as it couldn't import the files imported by that file. So I copied it directly into here and removed all nonessential parts

src/contracts/acrossConfigStore.test.ts
```diff
@@ -1,4 +1,5 @@
 import { Client } from "./acrossConfigStore";
+jest.useFakeTimers();
```
nicholaspai, line 2: These fix some warnings described by this issue: https://stackoverflow.com/questions/50793885/referenceerror-you-are-trying-to-import-a-file-after-the-jest-environment-has

src/constants.ts
james-a-morris, line 1: This is a good change +1

james-a-morris (approved):
LGTM - I don't see anything that sticks out to me. Otherwise, this seems like a good refactor to upgrade `contracts-v2` to a stable version

PR #193, opened 2023-04-28
refactor: change variable reference terminology

james-a-morris:


src/UBAFeeCalculator/UBAFeeCalculator.ts
```diff
@@ -20,7 +20,7 @@ export default class UBAFeeCalculator {
     private readonly config: UBAConfig,
     private readonly logger: Logger,
     protected readonly originSpoke: UBASpokeBalanceType,
-    protected readonly destinationSpoke: UBASpokeBalanceType
+    protected readonly refundSpoke: UBASpokeBalanceType
```
nicholaspai, line 23: I thought we didn't want to store the refund spoke in the calculator?

I was thinking instead each calculator had a single chain ID for which it provides functions to compute deposit and refund fees for?

nicholaspai (approved):
Looks good. Is this just a PR to rename and other refactors are coming after? https://github.com/across-protocol/sdk-v2/pull/193#discussion_r1180830454

src/UBAFeeCalculator/UBAFeeCalculator.ts
james-a-morris: @nicholaspai Yes, this is just the refactor - the next PR will be for moving this outside of the class

PR #195, opened 2023-05-02
fix: browser env bundle

dohaki:
The latest stable version 0.5.3 of the SDK fails to compile on the FE or in a browser environment with 
```
Failed to compile.

./node_modules/@across-protocol/sdk-v2/node_modules/@uma/common/dist/HardhatConfig.js
Cannot find module: 'hardhat-tracer'. Make sure this package is installed.

You can install this package by running: yarn add hardhat-tracer.
```
Or
```
Creating an optimized production build...
Failed to compile.

./node_modules/@across-protocol/contracts-v2/node_modules/hardhat/internal/util/fs-utils.js
Cannot find module: 'fs/promises'. Make sure this package is installed.

You can install this package by running: yarn add fs/promises.
```

AFAICT this is due to browser-incompatible packages (e.g. `@uma/common`, `@across-protocol/contracts-v2@^2.2.0`) that are not optimized for tree-shaking. Therefore the bundler tries to pack the whole package which contains server-only dependencies.

This PR fixes this by explicitly importing required modules so the bundler can tree-shake and ignore server-only modules. I guess at some point the upstream package should be optimized for tree-shaking and browser environments.

### DIFF
--- a/src/typechain.ts
+++ b/src/typechain.ts
@@ -1,0 +1,42 @@
+/**
+ * This file re-exports some of the typechain bindings so that they can be tree-shaken in the final frontend bundle.
+ * Currently, the packages `@across-protocol/contracts-v2` and `@across-protocol/across-token` are not optimized for tree-shaking
+ * and contain modules that are not compatible in a browser environment. This is a temporary solution until we can fix the issue upstream.
+ */
+export { AcrossMerkleDistributor__factory } from "@across-protocol/contracts-v2/dist/typechain/factories/AcrossMerkleDistributor__factory";
+export { AcrossConfigStore__factory } from "@across-protocol/contracts-v2/dist/typechain/factories/AcrossConfigStore__factory";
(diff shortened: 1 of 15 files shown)


nicholaspai (approved):

src/typechain.ts
james-a-morris, line 1: Are these auto-generated?

james-a-morris (commented):
This is a good refactor - I've added some comments offline

src/typechain.ts
dohaki: No, curated manually 😞 As described in the PR description, the contracts-v2 package doesn't allow for tree-shaking for some reason. I guess in an ideal world we would just re-export all from the SDK like
```typescript
// typechain.ts
export {
  // list of typechain re-exports 
} from "@across-protocol/contracts-v2"
```
And the FE bundler would tree-shake and only bundle the used ones. Unfortunately, with the current setup package, it doesn't work. So I needed to manually to that work for now.

james-a-morris (approved):
Resolved offline. Thanks for following up 🔥

PR #196, opened 2023-05-02
feat(UBA): :sparkles: add threshold trigger in historical balance calculations

james-a-morris:
This change makes the running balance aware of the concept of the trigger balance threshold. To cap the organic growth of a given spoke pool's running balance, a mechanism for preventing increasing balances has been proposed. As the running balance is calculated, if the balance exceeds a certain threshold dictated by the on-chain AcrossConfigStore, then this balance will be artificially lowered by the ACX data worker. As such, this non-event-based event must be captured in the running balance to maintain a uniform understanding of the running balances across all ecosystem relayers/bots/UIs


src/UBAFeeCalculator/UBAFeeConfig.ts
```diff
@@ -67,6 +86,17 @@ class UBAConfig {
...
     return this.balancingFee.override?.[chainId] ?? this.balancingFee.default;
   }
+
+  /**
+   * @description Get the balance trigger threshold for a given chain and token
+   * @param chainId The chain id
+   * @param tokenSymbol The token address
+   * @returns The balance trigger threshold if it exists
+   */
+  public getBalanceTriggerThreshold(chainId: number, tokenSymbol: string): TupleParameter | undefined {
```
mrice32, line 96: nit: don't we normally use mainnet token address as the unique identifier?

Can't the trigger threshold change over time, so don't we need to specify the block number for which we want to get the threshold?

src/UBAFeeCalculator/UBAFeeConfig.ts
```diff
@@ -29,14 +30,32 @@ class UBAConfig {
...
+   * @param utilizationFee A record of piecewise functions for each chain that define the utilization fee to ensure that the bridge responds to periods of high utilization
+   * @param balancingFee A record of piecewise functions for each chain and token that define the balancing fee to ensure either a positive or negative penalty to bridging a token to a chain that is either under or over utilized
+   * @param balanceTriggerThreshold A record of boundry values for each chain and token that define the threshold for when the running balance should be balanced back to a fixed amount. Due to the fact that this type of operation is based on a heuristic and is considered a non-event transient property of the protocol, the threshold must be computed based on the current running balance of the spoke from the last validated running balance.
+   */
   constructor(
     baselineFee: DefaultOverrideStructure<BigNumber, RouteCombination>,
     utilizationFee: BigNumber,
-    balancingFee: DefaultOverrideStructure<TupleParameter[], ChainId>
+    balancingFee: DefaultOverrideStructure<TupleParameter[], ChainId>,
+    balanceTriggerThreshold: Record<ChainTokenCombination, TupleParameter>
```
mrice32, line 53: Just to make sure I understand, this trigger threshold is defined in the mainnet ConfigStore contract, right?

Do you think it would be better to just query the ConfigStoreClient directly instead of trying to track that state here?

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -60,17 +62,30 @@ export default class UBAFeeSpokeCalculator {
...
-        return acc;
+        resultantValue = acc.sub(toBN(flow.amount));
+      }
+
+      // If the trigger hurdle is surpassed, we need to return the trigger hurdle value
+      // as the running balance. This is because the trigger hurdle is the maximum value we'd like to
+      // organically grow the running balance to. If the running balance exceeds the trigger hurdle,
+      // we need to return the trigger hurdle as the running balance because at this point the dataworker
+      // will be triggered to rebalance the running balance.
+      if (triggerHurdle !== undefined && resultantValue.gt(triggerHurdle[0])) {
```
mrice32, line 85: nit: unnamed tuples are a little hard to interpret -- have you considered using a named tuple or an object to represent these two values so it's clearer what's going on? 0 and 1 are not very self-documenting.

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -60,17 +62,30 @@ export default class UBAFeeSpokeCalculator {
...
-        return acc.add(toBN(flow.amount));
+        resultantValue = acc.add(toBN(flow.amount));
       } else if (isUbaOutflow(flow)) {
-        return acc.sub(toBN(flow.amount));
-      } else {
-        return acc;
+        resultantValue = acc.sub(toBN(flow.amount));
+      }
+
+      // If the trigger hurdle is surpassed, we need to return the trigger hurdle value
```
mrice32, line 80: What about a bottom threshold whereby we go below that running balance and need the hub pool to transfer funds into the spoke pool? Do we handle that case?

mrice32 (commented):
Looks great! Just a few questions

src/UBAFeeCalculator/UBAFeeConfig.ts
james-a-morris: I've tried to leave this code in a state where it's as abstracted away from RPC calls as possible. Since this is purely a library wrapper over mathematical operations, I wanted to leave the RPC calls explicitly to the user/calling code.

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
james-a-morris: Good call! https://github.com/across-protocol/sdk-v2/pull/196/commits/a6042c9971bc9aae70498c97ba80a6efe420518c

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
james-a-morris: I don't believe we have this case anymore as of today. I was told by @nicholaspai that our primary thresholds are when the running balance exceeds a target.

mrice32: I'm not sure I agree. Mechanically, imagine all spoke pools start with 0 balance. I make a single transfer from Optimism to Arbitrum for 100 USDC. Let's assume, for simplicity that all fees are 0 (including balancing fee -- we can choose to have the curve be flat).

My deposit makes the running balance on Optimism 100 USDC. The relayer relays my transaction on arbitrum, sending me 100 USDC. They now need to be refunded 100 USDC on arbitrum in the next bundle, but the arbitrum spoke pool needs to be sent 100 USDC from the HubPool to fulfil that liability. How would that happen if no lower threshold exists?

mrice32 (commented):
Looks great -- only one comment!

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
james-a-morris: Fair point! Resolved here: https://github.com/across-protocol/sdk-v2/pull/196/commits/1b468d46e6f0ed4a2f0d91e6cdcb7fb3e6c70f4c

mrice32 (approved):
LGTM!!

mrice32 (approved):
LGTM!!

PR #198, opened 2023-05-04
feat: migrate across config store & hub pool clients

james-a-morris:
This PR migrates the across config store and the hub pool client from the Across V2 Relayer.


src/utils/EventUtils.ts
```diff
@@ -0,0 +1,227 @@
+import { Result } from "@ethersproject/abi";
+import { delay } from "@uma/financial-templates-lib";
+import { SortableEvent } from "../interfaces";
+import { Contract, Event, EventFilter } from "ethers";
```
james-a-morris, line 4: Modified imports from `relayer-v2`

src/utils/EventUtils.ts
```diff
@@ -0,0 +1,227 @@
...
+
+const maxRetries = 3;
+const retrySleepTime = 10;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function spreadEvent(args: Result = {} as Result): any {
```
james-a-morris, line 10: Set a default argument for `args: Result = {} as Result`

src/utils/EventUtils.ts
```diff
@@ -0,0 +1,227 @@
...
+  // rounded up. This gives us the number of queries we need to execute to traverse the whole block range.
+  const paginatedRanges = getPaginatedBlockRanges(searchConfig);
+
+  try {
+    return (
+      (
+        await Promise.all(
+          paginatedRanges.map(([fromBlock, toBlock]) => contract.queryFilter(filter, fromBlock, toBlock))
+        )
+      )
```
james-a-morris, line 92: Changed Promise.map from bluebird into `promise.all(...)`

src/utils/EventUtils.ts
```diff
@@ -0,0 +1,227 @@
...
+    return ex.blockNumber < ey.blockNumber;
+  }
+  if (ex.transactionIndex !== ey.transactionIndex) {
+    return ex.transactionIndex < ey.transactionIndex;
+  }
+  return ex.logIndex < ey.logIndex;
+}
+
+export function getTransactionHashes(events: SortableEvent[]): string[] {
+  return [...Array.from(new Set(events.map((e) => e.transactionHash)))];
```
james-a-morris, line 226: Had to include `Array.from( )` to surround the `new Set(...)`

src/utils/common.ts
```diff
@@ -229,10 +229,30 @@ export const getSamplesBetween = (min: number, max: number, size: number) => {
...
 }
 
+/**
+ * Converts a string to a hex string
+ * @param input The string to convert
+ * @returns The hex string
+ */
+export function utf8ToHex(input: string) {
+  return ethers.utils.formatBytes32String(input);
+}
```
james-a-morris, line 247: Moved `utf8ToHex(...)` to `utils.ts`

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -0,0 +1,436 @@
...
+
+  filterDisabledChains(disabledChains: number[]): number[] {
+    // If any chain ID's are not integers then ignore. UMIP-157 requires that this key cannot include
+    // the chain ID 1.
+    return disabledChains.filter((chainId: number) => !isNaN(chainId) && Number.isInteger(chainId) && chainId !== 1);
+  }
+
+  private async getBlockNumber(timestamp: number): Promise<number | undefined> {
+    return (await this.blockFinder.getBlockForTimestamp(timestamp)).number;
+  }
```
james-a-morris, line 425: Replaced the code with the below. In the `relayer-v2` repository, I plan to overload this function in the extended child class.

```
  private async getBlockNumber(timestamp: number): Promise<number | undefined> {
    return await getBlockForTimestamp(
      this.hubPoolClient.chainId,
      this.hubPoolClient.chainId,
      timestamp,
      getCurrentTime(),
      this.blockFinder
    );
  }
```

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -0,0 +1,439 @@
...
+
+  async computeRealizedLpFeePct(
+    deposit: { quoteTimestamp: number; amount: BigNumber; destinationChainId: number; originChainId: number },
+    l1Token: string
+  ): Promise<{ realizedLpFeePct: BigNumber; quoteBlock: number }> {
+    let quoteBlock = await this.getBlockNumber(deposit.quoteTimestamp);
+
+    if (!quoteBlock) {
+      throw new Error(`Could not find block for timestamp ${deposit.quoteTimestamp}`);
+    }
```
james-a-morris, line 81: Added a throw to guarantee that this quoteBlock is defined in the logic below.

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -0,0 +1,440 @@
...
+  constructor(
+    readonly logger: winston.Logger,
+    readonly configStore: Contract,
+    readonly hubPoolClient: HubPoolClient,
+    readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
+    protected readonly configOverride: {
+      enabledChainIds: number[];
+      defaultConfigStoreVersion: number;
+    } = { enabledChainIds: [], defaultConfigStoreVersion: 0 }
+  ) {
```
james-a-morris, line 68: Added an override here so that we don't need to reference constant variables from the relayer. Throughout the rest of the code, I've replaced calls like `DEFAULT_CONFIG_STORE_VERSION` to `this.configOverride.defaultConfigStoreVersion`;

src/clients/AcrossHubPoolClient.ts
```diff
@@ -0,0 +1,659 @@
...
+  public firstBlockToSearch: number;
+  public latestBlockNumber: number | undefined;
+  public currentTime: number | undefined;
+
+  constructor(
+    readonly logger: winston.Logger,
+    readonly hubPool: Contract,
+    readonly chainId: number = 1,
+    readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 }
+  ) {
```
james-a-morris, line 66: Update the reference to the constant `IGNORED_HUB_PROPOSED_BUNDLES` and `IGNORED_HUB_EXECUTED_BUNDLES` with an overridable instance config variable within the class.

src/clients/AcrossHubPoolClient.ts
```diff
@@ -0,0 +1,667 @@
...
+    }
+
+    // For each enabled Lp token fetch the token symbol and decimals from the token contract. Note this logic will
+    // only run iff a new token has been enabled. Will only append iff the info is not there already.
+    // Filter out any duplicate addresses. This might happen due to enabling, disabling and re-enabling a token.
+    const uniqueL1Tokens = [
+      ...Array.from(
+        new Set(events["L1TokenEnabledForLiquidityProvision"].map((event) => spreadEvent(event.args).l1Token))
+      ),
+    ];
```
james-a-morris, line 565: Had to include Array.from( ) to surround the new Set(...)

src/clients/AcrossHubPoolClient.ts
```diff
@@ -0,0 +1,667 @@
...
+    return sortEventsDescending(this.disputedRootBundles).filter(
+      (bundle: DisputedRootBundle) => bundle.blockNumber >= startingBlock && bundle.blockNumber <= endingBlock
+    );
+  }
+
+  getLatestProposedRootBundle(): ProposedRootBundle {
+    return this.proposedRootBundles[this.proposedRootBundles.length - 1] as ProposedRootBundle;
+  }
+
+  getFollowingRootBundle(currentRootBundle: ProposedRootBundle): ProposedRootBundle | undefined {
```
james-a-morris, line 318: Set the return to include undefined to remove the error on the return undefined this this function

src/utils/TokenUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+type SignerOrProvider = providers.Provider | Signer;
+
+export async function fetchTokenInfo(address: string, signerOrProvider: SignerOrProvider): Promise<L1Token> {
+  const token = new Contract(address, ERC20__factory.abi, signerOrProvider);
+  const [symbol, decimals] = await Promise.all([token.symbol(), token.decimals()]);
+  return { address, symbol, decimals };
+}
+
+export const getL2TokenAddresses = (l1TokenAddress: string): { [chainId: number]: string } | undefined => {
```
james-a-morris, line 15: Add an `undefined` return type because this could potentially return undefined

src/clients/AcrossHubPoolClient.ts
```diff
@@ -511,7 +511,7 @@ export class HubPoolClient {
   }
 
   async update(eventsToQuery?: HubPoolEvent[]): Promise<void> {
-    eventsToQuery ??= Object.keys(this.hubPoolEventFilters()) as HubPoolEvent[]; // Query all events by default.
+    eventsToQuery = eventsToQuery ?? (Object.keys(this.hubPoolEventFilters()) as HubPoolEvent[]); // Query all events by default.
```
james-a-morris, line 514: Change `??=` to ```eventsToQuery = eventsToQuery ?? ...```;

pxrl: Is this needed to avoid some tsc or lint issue as a result of the type assertion on the ??= case?

pxrl:
> This change makes the running balance aware of the concept of the trigger balance threshold. To cap the organic growth of a given spoke pool's running balance, a mechanism for preventing increasing balances has been proposed. As the running balance is calculated, if the balance exceeds a certain threshold dictated by the on-chain AcrossConfigStore, then this balance will be artificially lowered by the ACX data worker. As such, this non-event-based event must be captured in the running balance to maintain a uniform understanding of the running balances across all ecosystem relayers/bots/UIs

This description doesn't seem to match the PR.

src/clients/AcrossHubPoolClient.ts
pxrl, line 1: Can we retain the old names of HubPoolClient and ConfigStoreClient? Prepending "Across" seems redundant.

src/utils/EventUtils.ts
pxrl: For potential follow-up, we can probably refine the return type annotation. I believe this function should return a Record<string, any>. I guess that's an improvement on just returning any.

src/utils/EventUtils.ts
pxrl: We should circle back and see whether we still need Bluebird in relayer-v2 after this migration.

src/utils/common.ts
pxrl: Returns type `string<DataHexString<32>>` according to [ethers](https://docs.ethers.org/v5/api/utils/strings/#utils-formatBytes32). Can we get away with a `string` return type annotation?

james-a-morris:
> > This change makes the running balance aware of the concept of the trigger balance threshold. To cap the organic growth of a given spoke pool's running balance, a mechanism for preventing increasing balances has been proposed. As the running balance is calculated, if the balance exceeds a certain threshold dictated by the on-chain AcrossConfigStore, then this balance will be artificially lowered by the ACX data worker. As such, this non-event-based event must be captured in the running balance to maintain a uniform understanding of the running balances across all ecosystem relayers/bots/UIs
> 
> 
> 
> This description doesn't seem to match the PR.

Whoops! Updated thanks for the callout

src/clients/AcrossHubPoolClient.ts
james-a-morris: Sure! The actual class names are preserved, but I modified the file names for the event that the SDK needs to have non across related client files.

I'll change it back

src/utils/EventUtils.ts
james-a-morris: I personally think we should leave these refactorings to after the relayer migration, but this is a great point to further move toward using a fully typed relayer

src/clients/AcrossConfigStoreClient.ts
nicholaspai: +1

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -0,0 +1,441 @@
...
+    l1Token: string
+  ): Promise<{ realizedLpFeePct: BigNumber; quoteBlock: number }> {
+    let quoteBlock = await this.getBlockNumber(deposit.quoteTimestamp);
+
+    if (!quoteBlock) {
+      throw new Error(`Could not find block for timestamp ${deposit.quoteTimestamp}`);
+    }
+
+    // There is one deposit on optimism for DAI that is right before the DAI rate model was added.
+    if (quoteBlock === 14830339) {
```
nicholaspai, line 86: Just realized that we can remove these hardcoded catches from line 85-97 now that we have new spoke pools.

For Goerli, we'll just redeploy the spoke pool

src/utils/EventUtils.ts
nicholaspai: Yea lets refactor post migration so its easier to review!

nicholaspai (approved):
LGTM thanks for doing this. Let's do further refactors post migration

PR #202, opened 2023-05-08
chore: Sync yarn.lock

pxrl:
I think we somehow got out of sync as part of commit https://github.com/across-protocol/sdk-v2/commit/4d34289f6ae8876a318f8dc94e11b0de216504e3, but I don't understand how that happened.


james-a-morris (approved):
LGTM - all tests pass. We should keep this one on our radar in case any reversions need to be made

PR #204, opened 2023-05-08
improve(HubPoolClient): Track deploymentBlock

pxrl:
Also, set latestBlockNumber in the constructor. If no deploymentBlock is supplied, default to 0.

Once this is pulled into relayer-v2, we can make some follow-on adjustments there and then redefine latestBlockNumber as a number only, so callers don't need to consider the `undefined` scenario.

Fixes ACX-1048


james-a-morris (approved):
LGTM! Is this a lead-up to make an additional feature?

PR #205, opened 2023-05-08
refactor: :fire: remove hardcoded catches for various pools

james-a-morris:
Implementing a change from an offline discussion. "We can remove these hardcoded catches from lines 85-97 now that we have new spoke pools." ~ @nicholaspai


pxrl (approved):
LGTM.

PR #209, opened 2023-05-09
improve: change accessor visibility in clients

james-a-morris:


pxrl (approved):
LGTM.

I noticed that dependabot tried to auto-bump SDK-v2 in the relayer-v2 repo; should we consider bumping the minor version instead of the patch version? i.e. 0.6.0 instead of 0.5.10.

james-a-morris:
> LGTM.
> 
> I noticed that dependabot tried to auto-bump SDK-v2 in the relayer-v2 repo and it failed for some reason; should we consider bumping the minor version instead of the patch version? i.e. 0.6.0 instead of 0.5.10.

SGTM! I've bumped to 0.6.0

PR #211, opened 2023-05-11
chore: downgrade ethers

james-a-morris:


james-a-morris:
Removing this because it's now unneeded.

PR #213, opened 2023-05-17
feat(UBA): Support overloaded ExecutedRootBundle runningBalances

pxrl:
The UBA needs to keep track of the size of the incentive pool for each token on each chain. Per the pending updates to UMIP-157, this will be done by appending the running incentive balances to the ExecutedRootBundle runningBalances array.

This change permits the HubPoolClient to infer the presence of incentive balances in the runningBalances array based on the applicable ConfigStore version at the time.

Fixes ACX-1134

### DIFF
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -1,4 +1,6 @@
+import { isError } from "../typeguards";
 import {
+  isDefined,
   spreadEvent,
   sortEventsDescending,
   spreadEventWithBlockNumber,
@@ -10,7 +12,6 @@ import {
   max,
   sortEventsAscending,
 } from "../utils";
-
 import { Contract, BigNumber } from "ethers";
 import winston from "winston";
 
@@ -25,8 +26,11 @@ import {
   ConfigStoreVersionUpdate,
   DisabledChainsUpdate,
 } from "../interfaces";
-
 import { across } from "@uma/sdk";
+
+// Version 0 is the implicit ConfigStore version from before the version attribute was introduced.
+// @dev Do not change this value.
+export const DEFAULT_CONFIG_STORE_VERSION = 0;
 
 export const GLOBAL_CONFIG_STORE_KEYS = {
   MAX_RELAYER_REPAYMENT_LEAF_SIZE: "MAX_RELAYER_REPAYMENT_LEAF_SIZE",
@@ -56,11 +60,8 @@ export class AcrossConfigStoreClient {
     readonly logger: winston.Logger,
     readonly configStore: Contract,
     readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
-    protected readonly configOverride: {
-      enabledChainIds: number[];
-      defaultConfigStoreVersion: number;
-      configStoreVersion: number;
-    }
+    readonly configStoreVersion: number,
+    readonly enabledChainIds: number[]
   ) {
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
     this.rateModelDictionary = new across.rateModel.RateModelDictionary();
@@ -148,7 +149,7 @@ export class AcrossConfigStoreClient {
   getEnabledChainsInBlockRange(
     fromBlock: number,
     toBlock = Number.MAX_SAFE_INTEGER,
-    allPossibleChains = this.configOverride.enabledChainIds
+    allPossibleChains = this.enabledChainIds
   ): number[] {
     if (toBlock < fromBlock) {
       throw new Error(`Invalid block range: fromBlock ${fromBlock} > toBlock ${toBlock}`);
@@ -178,7 +179,7 @@ export class AcrossConfigStoreClient {
       .sort((a, b) => a - b);
   }
 
-  getEnabledChains(block = Number.MAX_SAFE_INTEGER, allPossibleChains = this.configOverride.enabledChainIds): number[] {
+  getEnabledChains(block = Number.MAX_SAFE_INTEGER, allPossibleChains = this.enabledChainIds): number[] {
     // Get most recent disabled chain list before the block specified.
     const currentlyDisabledChains = this.getDisabledChainsForBlock(block);
     return allPossibleChains.filter((chainId) => !currentlyDisabledChains.includes(chainId));
@@ -196,10 +197,7 @@ export class AcrossConfigStoreClient {
     const config = (sortEventsDescending(this.cumulativeConfigStoreVersionUpdates) as ConfigStoreVersionUpdate[]).find(
       (config) => config.timestamp <= timestamp
     );
-    if (!config) {
-      return this.configOverride.defaultConfigStoreVersion;
-    }
-    return Number(config.value);
+    return isDefined(config) ? Number(config.value) : DEFAULT_CONFIG_STORE_VERSION;
   }
 
   hasValidConfigStoreVersionForTimestamp(timestamp: number = Number.MAX_SAFE_INTEGER): boolean {
@@ -208,7 +206,7 @@ export class AcrossConfigStoreClient {
   }
 
   isValidConfigStoreVersion(version: number): boolean {
-    return this.configOverride.configStoreVersion >= version;
+    return this.configStoreVersion >= version;
   }
 
   async update(): Promise<void> {
@@ -217,14 +215,13 @@ export class AcrossConfigStoreClient {
       toBlock: this.eventSearchConfig.toBlock || (await this.configStore.provider.getBlockNumber()),
       maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
     };
-    if (searchConfig.fromBlock > searchConfig.toBlock) {
-      return;
-    } // If the starting block is greater than
 
     this.logger.debug({ at: "ConfigStore", message: "Updating ConfigStore client", searchConfig });
     if (searchConfig.fromBlock > searchConfig.toBlock) {
+      this.logger.warn({ at: "ConfigStore", message: "Invalid search config.", searchConfig });
       return;
-    } // If the starting block is greater than the ending block return.
+    }
+
     const [updatedTokenConfigEvents, updatedGlobalConfigEvents] = await Promise.all([
       paginatedEventQuery(this.configStore, this.configStore.filters.UpdatedTokenConfig(), searchConfig),
       paginatedEventQuery(this.configStore, this.configStore.filters.UpdatedGlobalConfig(), searchConfig),
@@ -246,6 +243,12 @@ export class AcrossConfigStoreClient {
 
         // If Token config doesn't contain all expected properties, skip it.
         if (!(rateModelForToken && transferThresholdForToken)) {
+          this.logger.warn({
+            at: "ConfigStore",
+            message: "Ignoring invalid rate model update.",
+            update: args,
+            transferThresholdForToken,
+          });
           continue;
         }
 
@@ -299,6 +302,9 @@ export class AcrossConfigStoreClient {
           this.cumulativeRouteRateModelUpdates.push({ ...passedArgs, routeRateModel: {}, l1Token });
         }
       } catch (err) {
+        if (isError(err)) {
+          this.logger.warn({ at: "ConfigStore", message: "Caught error during update.", error: err.message });
+        }
         continue;
       }
     }
@@ -336,7 +342,7 @@ export class AcrossConfigStoreClient {
         // Extract last version
         const lastValue =
           this.cumulativeConfigStoreVersionUpdates.length === 0
-            ? this.configOverride.defaultConfigStoreVersion
+            ? DEFAULT_CONFIG_STORE_VERSION
             : Number(
                 this.cumulativeConfigStoreVersionUpdates[this.cumulativeConfigStoreVersionUpdates.length - 1].value
               );
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -1,3 +1,5 @@
+import assert from "assert";
+import { utils } from "@uma/sdk";
 import { isError } from "../typeguards";
 import {
   isDefined,
@@ -12,7 +14,7 @@ import {
   max,
   sortEventsAscending,
 } from "../utils";
-import { Contract, BigNumber } from "ethers";
+import { Contract, BigNumber, Event } from "ethers";
 import winston from "winston";
 
 import {
@@ -27,6 +29,18 @@ import {
   DisabledChainsUpdate,
 } from "../interfaces";
 import { across } from "@uma/sdk";
+
+type _ConfigStoreUpdate = {
+  success: true;
+  latestBlockNumber: number;
+  searchEndBlock: number;
+  events: {
+    updatedTokenConfigEvents: Event[];
+    updatedGlobalConfigEvents: Event[];
+    globalConfigUpdateTimes: number[];
+  };
+};
+export type ConfigStoreUpdate = { success: false } | _ConfigStoreUpdate;
 
 // Version 0 is the implicit ConfigStore version from before the version attribute was introduced.
 // @dev Do not change this value.
@@ -51,6 +65,7 @@ export class AcrossConfigStoreClient {
 
   protected rateModelDictionary: across.rateModel.RateModelDictionary;
   public firstBlockToSearch: number;
+  public latestBlockNumber = 0;
 
   public hasLatestConfigStoreVersion = false;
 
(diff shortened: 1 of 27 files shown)


pxrl:
We have an unfortunate situation where all of the tests for the {Hub,Spoke}PoolClient still live in the relayer-v2 repository. I've got a draft PR where I am updating the necessary tests here: https://github.com/across-protocol/relayer-v2/pull/685.

src/interfaces/HubPool.ts
```diff
@@ -50,6 +50,11 @@ export interface ExecutedRootBundle extends SortableEvent {
   proof: string[];
 }
 
+export type TokenRunningBalance = {
+  runningBalance: BigNumber;
+  incentiveBalance: BigNumber;
+};
+
```
james-a-morris, line 57: Was this exposed in a previous PR?

src/clients/HubPoolClient.ts
```diff
@@ -452,21 +454,41 @@ export class HubPoolClient {
...
+        // Pre-UBA model (no incentives exist).
+        incentiveBalances = Array.from(runningBalances, () => toBN(0));
+      } else if (runningBalances.length === 2 * l1Tokens.length) {
+        // UBA model: runningBalances array is a concatenation of runningBalance and incentiveBalance.
+
+        runningBalances = runningBalances.slice(0, l1Tokens.length);
+        incentiveBalances = runningBalances.slice(l1Tokens.length);
+        this.logger.debug({
+          at: "HubPoolClient#getRunningBalanceBeforeBlockForChain",
+          message: "",
```
james-a-morris, line 473: Will this mean that the message is empty in the error output?

src/clients/HubPoolClient.ts
```diff
@@ -452,21 +454,41 @@ export class HubPoolClient {
...
-        throw new Error("runningBalances and L1 token of ExecutedRootBundle event are not same length");
+
+    if (executedRootBundle) {
+      // runningBalances length must be a 1x or 2x multiple of l1Tokens length.
+      const { l1Tokens } = executedRootBundle;
+      let { runningBalances } = executedRootBundle;
+      let incentiveBalances: BigNumber[] = [];
+      if (runningBalances.length === l1Tokens.length) {
+        // Pre-UBA model (no incentives exist).
+        incentiveBalances = Array.from(runningBalances, () => toBN(0));
```
james-a-morris, line 465: Should we use `.map()` instead? I think it's a more common usage.
```suggestion
        incentiveBalances = runningBalances.map(() => toBN(0));
```

james-a-morris (commented):
Left a few nits

src/interfaces/HubPool.ts
pxrl: Kinda. I initially implemented it here, then realised you were also defining something similar and that we should share the definition.

This PR is now dependent on the PR where this actually will be defined, so it should disappear from here. (I think github is being weird but not realising that it's the same change).

src/clients/HubPoolClient.ts
pxrl: That's a mistake :) Thanks for spotting it.

src/clients/HubPoolClient.ts
pxrl: Yeah this is fine. I was using map initially but I got complaints about an unused variable (i.e. the variable available in each iteration). Not defining anything is also a good way of avoiding that!

src/clients/HubPoolClient.ts
pxrl: Fixed now.

src/clients/HubPoolClient.ts
```diff
@@ -452,21 +454,42 @@ export class HubPoolClient {
...
-      ) {
-        throw new Error("runningBalances and L1 token of ExecutedRootBundle event are not same length");
+
+    if (executedRootBundle) {
+      // runningBalances length must be a 1x or 2x multiple of l1Tokens length.
+      const { l1Tokens } = executedRootBundle;
+      let { runningBalances } = executedRootBundle;
+      let incentiveBalances: BigNumber[] = [];
+      if (runningBalances.length === l1Tokens.length) {
+        // Pre-UBA model (no incentives exist).
```
nicholaspai, line 464: I think this check should be a bit different:
- If `ConfigStore.Version < 2` (for example), then `assert(runningBalances.length == l1Tokens.length)`
- Else `assert(runningBalances.length == l1Tokens.length * 2)`

nicholaspai (commented):
I think this code rightfully belongs in the `HubPoolClient`, but I do think the checks on the runningBalances have to incorporate the ConfigStore. I'll start with this comment: https://github.com/across-protocol/sdk-v2/pull/213/files#r1197043597 and then take it from there

src/clients/HubPoolClient.ts
pxrl: This makes sense; have implemented it here: https://github.com/across-protocol/sdk-v2/pull/213/commits/c0bafa5372d181fcbab8b12d1143ff429b856231

It now handles this even earlier in the process - i.e. when each ExecutedRootBundle event is parsed in the update() method. It does as such involve extending the ExecutedRootBundle event to have an incentiveBalances array; this should be backwards compatible with existing uses of ExecutedRootBundle though.

pxrl:
Status: am following up with tests in relayer-v2.

src/clients/HubPoolClient.ts
```diff
@@ -646,11 +651,45 @@ export class HubPoolClient {
...
+      }
+      const version = configStoreVersions[event.blockNumber];
+
+      const executedRootBundle = spreadEventWithBlockNumber(event) as ExecutedRootBundle;
+      const { l1Tokens, runningBalances } = executedRootBundle;
+      let expectLength: number = l1Tokens.length;
+
+      if (version < UBA_MIN_CONFIG_STORE_VERSION) {
+        // Pre-UBA model: runningBalances length is 1:1 with l1Tokens length.
+        assert([0, 1].includes(version), `Invalid ConfigStore version: ${version}`);
```
nicholaspai, line 674: Nit: prefer reusing some of the config store client logic to validate the version. Like isValidVersion

src/clients/HubPoolClient.ts
```diff
@@ -498,21 +499,18 @@ export class HubPoolClient {
         );
       }
     ) as ExecutedRootBundle;
-    if (mostRecentExecutedRootBundleEvent) {
-      // Arguably we don't need to even check these array lengths since we should assume that any proposed root bundle
-      // meets this condition.
-      if (
-        mostRecentExecutedRootBundleEvent.l1Tokens.length !== mostRecentExecutedRootBundleEvent.runningBalances.length
-      ) {
-        throw new Error("runningBalances and L1 token of ExecutedRootBundle event are not same length");
```
pxrl, line 507: FWIW, this is now performed in HubPoolClient.update().

src/clients/HubPoolClient.ts
pxrl: I ended up removing the assert since it was mostly useful during development and test.

src/utils/common.ts
```diff
@@ -14,6 +14,9 @@ export type Decimalish = string | number | Decimal;
 export const AddressZero = ethers.constants.AddressZero;
 export const MAX_BIG_INT = BigNumber.from(Number.MAX_SAFE_INTEGER.toString());
 
+// Lowest ConfigStore version where the UBA model is in effect.
+export const UBA_MIN_CONFIG_STORE_VERSION = 2;
```
james-a-morris, line 18: We may want to comment this wish `/* */` so that IDE's like VSCode can get context
```suggestion
/** Lowest ConfigStore version where the UBA model is in effect. */
export const UBA_MIN_CONFIG_STORE_VERSION = 2;
```

src/clients/HubPoolClient.ts
```diff
@@ -691,7 +720,7 @@ export class HubPoolClient {
     this.firstBlockToSearch = update.searchEndBlock + 1; // Next iteration should start off from where this one ended.
 
     this.isUpdated = true;
-    this.logger.debug({ at: "HubPoolClient", message: "HubPool client updated!" });
+    this.logger.debug({ at: "HubPoolClient::update", message: "HubPool client updated!", endBlock: latestBlockNumber });
```
james-a-morris, line 723: Nit: I think @nicholaspai used `#` instead of `::` in the relayer code for specifying where a log was coming from. Should we keep that format?

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -192,6 +192,7 @@ export class AcrossConfigStoreClient {
     );
   }
 
+  // @dev This method assumes a block with a version change must _not_ coincide with an executed root bundle.
```
james-a-morris, line 195: Nit: are we checking for this? I couldn't find it in the code below, but I probably fully missed it 😅

james-a-morris (commented):
Left a few comments. This code is great

src/utils/common.ts
pxrl: `/* */` is actually my preferred style, but I recall getting some feedback early on that the internal standard is to use `//` except for when the comment has special formatting. @mrice32 wdyt?

src/clients/HubPoolClient.ts
pxrl: Yeah, I've seen both in use in the codebase. I think @mrice32 tends for `::`, but we definitely have `#` in places. I'm not strongly opinionated, but I do find it helpful when the method name is included in the `at` field.

src/clients/AcrossConfigStoreClient.ts
pxrl: I don't think we have any specific checks against it, though it's unlikely to happen in practice because we're usually highly conscious of what the proposer is doing when making config changes. It just stuck out to me that it would be worth documenting the implicit requirement.

src/utils/common.ts
mrice32: Sorry, I'm confused, why are /* comments more parseable than //?

src/clients/HubPoolClient.ts
mrice32: Yeah, this is just an inconsistency between myself and Nick. I think either way is fine unless we're planning to parse these logs programmatically, since we all know what they mean. Also happy to just go with whichever y'all prefer going forward!

src/clients/AcrossConfigStoreClient.ts
mrice32: Do we really need this? This seems like an unnecessary fragility? Can we make a rule where the new rules either apply in the block they're added or begin applying in the block after?

src/clients/HubPoolClient.ts
```diff
@@ -646,11 +643,43 @@ export class HubPoolClient {
...
+      if (!isDefined(configStoreVersions[event.blockNumber])) {
+        const timestamp = (await event.getBlock()).timestamp;
+        const version = this.configStoreClient.getConfigStoreVersionForTimestamp(timestamp);
+        configStoreVersions[event.blockNumber] = version;
+      }
+      const version = configStoreVersions[event.blockNumber];
+
+      const executedRootBundle = spreadEventWithBlockNumber(event) as ExecutedRootBundle;
+      const { l1Tokens, runningBalances } = executedRootBundle;
+      const nTokens = l1Tokens.length;
```
mrice32, line 662: We're paying a _fairly_ high costs (extra RPCs in the hot path) for padding the incentiveBalances in the case of pre-UBA versions.

Since we already know the nTokens, we should always be able to correctly slice the array, regardless. What value are we getting out of this padding that we couldn't get just by _always padding_ if there aren't enough values? Note: these are executed, so they technically should already be validated, right?

mrice32 (commented):
Looks great! Just a few comments.

src/clients/AcrossConfigStoreClient.ts
pxrl: The comment is misplaced and doesn't actually relate to the function. It's moreso relevant to the code that uses it, which turns out to be mostly in log messages in the relayer. I'll remove the comment from here.

pxrl:
```suggestion
```

src/clients/HubPoolClient.ts
pxrl: Not sure whether I'm understanding your comment correctly, but the challenge is in knowing when the `incentiveBalances` should be zeroes, and when it should be a slice of `runningBalances`. This PR started out by inferring the correct shape of `runningBalances` and `incentiveBalances` based solely on the length of `runningBalances` in the on-chain event. Nick had a good point that we should only make this evaluation based on the version number, which is how it ended here.

Fortunately, based on your other feedback about the cost of querying the blocks for each update, I realised that we already have all of the necessary information, and that we only lacked a helper method in the `ConfigStoreClient`. See here for a PR to add it in: https://github.com/across-protocol/sdk-v2/pull/230.

src/clients/HubPoolClient.ts
```diff
@@ -646,11 +643,43 @@ export class HubPoolClient {
...
-    );
+
+    const configStoreVersions: { [blockNumber: number]: number } = {};
+    for (const event of events["RootBundleExecuted"]) {
+      if (this.configOverride.ignoredHubExecutedBundles.includes(event.blockNumber)) {
+        continue;
+      }
+
+      if (!isDefined(configStoreVersions[event.blockNumber])) {
+        const timestamp = (await event.getBlock()).timestamp;
```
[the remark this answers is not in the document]
pxrl: These are both really good points, and I'd overlooked how inefficient this is.

I re-evaluated and realised that we only lack a helper method in the `ConfigStoreClient` to be able to resolve the version based on the block number instead of the timestamp. I've proposed that [here](https://github.com/across-protocol/sdk-v2/pull/230), and have made the corresponding change in this PR [here](https://github.com/across-protocol/sdk-v2/pull/213/commits/c7860aba405043cd21f329a7d116c8869989e053).

pxrl:
The corresponding relayer-v2 change can be found here: https://github.com/across-protocol/relayer-v2/pull/685.

src/clients/HubPoolClient.ts
```diff
@@ -646,11 +647,42 @@ export class HubPoolClient {
...
+    for (const event of events["RootBundleExecuted"]) {
+      if (this.configOverride.ignoredHubExecutedBundles.includes(event.blockNumber)) {
+        continue;
+      }
+
+      if (!isDefined(configStoreVersions[event.blockNumber])) {
+        const version = this.configStoreClient.getConfigStoreVersionForBlock(event.blockNumber);
+        configStoreVersions[event.blockNumber] = version;
+      }
+      const version = configStoreVersions[event.blockNumber];
```
pxrl, line 661: Good observation from @nicholaspai here: https://github.com/across-protocol/sdk-v2/pull/230#discussion_r1223131959. This PR is using the wrong block number to resolve the version. The correct block number is that of the corresponding root bundle proposal.

src/clients/HubPoolClient.ts
```diff
@@ -650,11 +647,60 @@ export class HubPoolClient {
...
-        .map((event) => spreadEventWithBlockNumber(event) as ExecutedRootBundle)
-    );
+
+    const configStoreVersions: { [blockNumber: number]: number } = {};
+    for (const event of events["RootBundleExecuted"]) {
+      if (this.configOverride.ignoredHubExecutedBundles.includes(event.blockNumber)) {
+        continue;
+      }
+
+      // The applicable version is determined by the block number of the corresponding proposal.
```
pxrl, line 657: Some thoughts on this:
 - The HubPool does not permit new proposals whilst there are existing un-executed leaves, so it effectively enforces that the proposal immediately before this executed leaf is the right one.
 - We assume that the RPC provider supplies events chronologically and completely.

I'd feel better if we verified the link between the root bundle proposal and the executed leaf with something stronger than the relationship between their blocks, but it's quite difficult to do that this low in the stack. Trying to do so may introduce chicken and egg problems.

src/clients/HubPoolClient.ts
pxrl: Should be resolved here: https://github.com/across-protocol/sdk-v2/pull/213/commits/e5023490f9bc62716ccce8b57c9fae9988cdef78.

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
mrice32: Awesome, thanks!

mrice32 (approved):
LGTM!

src/clients/HubPoolClient.ts
mrice32: This is great! +1. I was mostly just worried about the RPC calls. If we don't make them, then it's no problem!

src/utils/common.ts
james-a-morris: I'm just referring to valid JSDoc formatting. This formatting is interpreted by VSCode and several other IDEs to provide additional context about functions & variables.

https://jsdoc.app/about-getting-started.html

james-a-morris (commented):
Left a comment here: https://github.com/across-protocol/sdk-v2/pull/213#discussion_r1228705307

pxrl:
> Left a comment here: [#213 (comment)](https://github.com/across-protocol/sdk-v2/pull/213#discussion_r1228705307)

Do we want to take this in a separate thread? It's moreso a policy discussion than something related specifically to this PR.

PR #215, opened 2023-05-17
refactor(utils): Migrate relayer-v2 typeguards

pxrl:
Can reasonably foresee a need for these in the sdk-v2 repository, given some of the migrations that are occurring. Will subsequently update the relayer-v2 repository to import these after bumping the SDK version.

Copied from https://github.com/across-protocol/relayer-v2/blob/master/src/utils/TypeGuards.ts


james-a-morris (approved):
Will there be a separate PR opened in the relayer to remove these?

PR #217, opened 2023-05-19
improve: upgrade computation of fees

james-a-morris:
This PR modifies how we are computing the deposit and refund fees (balancing + LP component)


src/UBAFeeCalculator/UBAFeeConfig.ts
```diff
@@ -11,6 +11,7 @@ type DefaultOverrideStructure<PrimaryValue, OverrideKeyType extends string | num
   default: PrimaryValue;
   override?: Record<OverrideKeyType, PrimaryValue>;
 };
+type FlowTupleParameters = { inflow: TupleParameter[]; outflow: TupleParameter[] };
```
nicholaspai, line 14: This probably makes sense as a good way to organize the curve parameters.

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -8,6 +8,11 @@ type TokenRunningBalanceWithNetSend = TokenRunningBalance & {
   netRunningBalanceAdjustment: BigNumber;
 };
 
+type FlowFee = {
+  lpComponent: BigNumber;
```
nicholaspai, line 12: nit maybe name this `lpFee` to be more consistent with naming with `balancingFee`

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -167,61 +172,82 @@ export default class UBAFeeSpokeCalculator {
...
   }
 
+  /**
+   * A convenience method for resolving the event fee for the spoke and the given symbol and a given flow type
+   * @param amount The amount of tokens to simulate
+   * @param flowType The flow type to simulate
+   * @param flowRange The range of the flow to simulate the event for. Defaults to undefined to simulate the event for the entire flow.
+   * @returns The event fee for the spoke and the given symbol and a given flow type
+   */
+  protected getEventFee(amount: BigNumber, flowType: "inflow" | "outflow", flowRange?: UBAFlowRange): FlowFee {
```
nicholaspai, line 182: Can you use the sign of `amount` to infer the `flowType`? Just seeing if there's a way to remove the second param

nicholaspai (approved):
This structure makes sense, have just left comments suggesting to reconsider certain variable names

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
james-a-morris: That should be fine. But would it ever be possible (even in a contrived case to rescue funds or something similar) to want to encode the ability to handle an inverse deposit/refund event?

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
james-a-morris: Completed here: https://github.com/across-protocol/sdk-v2/pull/217/commits/ff3de28daf7fab5f87b37ea10be4b69748fa89bd

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
nicholaspai: hmmm thats an interestingpoint. I can't imagine this case, wouldn't you just call `getEventFee(-amount)` (and negate the sign) to compute the hypothetical reverse event?

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -167,61 +172,82 @@ export default class UBAFeeSpokeCalculator {
...
+
+    // We'll need to now compute the concept of the running balance of the spoke
+    const { runningBalance } = this.calculateHistoricalRunningBalance(flowRange?.startIndex, flowRange?.endIndex);
+
+    // We first need to resolve the inflow/outflow curves for the deposit and refund spoke
+    const flowCurves = this.config.getBalancingFeeTuples(this.chainId);
+
+    // We'll need to determine which flow curve to use based on the flow type
+    const primaryFlowCurve = flowCurves[flowType];
+    const secondaryFlowCurve = flowCurves[flowType === "inflow" ? "outflow" : "inflow"];
```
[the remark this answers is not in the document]
james-a-morris: Updated here: https://github.com/across-protocol/sdk-v2/pull/217/commits/f8e6555c7321949a130a0eafa7efaac0faf4d278

I kept the concept of `mainFlow` instead of just `flow` because flow seemed like not enough info (both the flow & reverse flow are both technically `flows`)

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -167,61 +172,82 @@ export default class UBAFeeSpokeCalculator {
...
+   * @returns The event fee for the spoke and the given symbol and a given flow type
+   */
+  protected getEventFee(amount: BigNumber, flowType: "inflow" | "outflow", flowRange?: UBAFlowRange): FlowFee {
+    // The rough psuedocode for this function is as follows:
+    // We'll need two inflow/outflow curves
+    // We need to determine which flow curve to use based on the flow type
+    // Compute first balancing fee <- f(x, x+amnt)
+    // Compute second balance fee (oportunity cost) <- g(x+amnt, x)
+    // Incentive Fee (LP Fee Component): first balancing fee - second balance fee
+    // Return (LP Fee + Balancing Fee)
```
mrice32, line 189: Does this mean we have an LP fee on the relay side as well? Today we only charge this on the deposit side.

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -167,61 +172,82 @@ export default class UBAFeeSpokeCalculator {
...
+   * @param flowType The flow type to simulate
+   * @param flowRange The range of the flow to simulate the event for. Defaults to undefined to simulate the event for the entire flow.
+   * @returns The event fee for the spoke and the given symbol and a given flow type
+   */
+  protected getEventFee(amount: BigNumber, flowType: "inflow" | "outflow", flowRange?: UBAFlowRange): FlowFee {
+    // The rough psuedocode for this function is as follows:
+    // We'll need two inflow/outflow curves
+    // We need to determine which flow curve to use based on the flow type
+    // Compute first balancing fee <- f(x, x+amnt)
+    // Compute second balance fee (oportunity cost) <- g(x+amnt, x)
```
mrice32, line 187: Just a minor terminology nit: we're overloading the term balancing fee here a little. We're using it for both the component of the deposit/relay fee that does not go to LPs, but we're also using it for the full fee (the value of the curve).

Put differently, we're describing it as:
balancing_fee = balancing_fee - lp_fee

But these two balancing fees refer to different numbers, which makes it somewhat confusing.

mrice32 (commented):
Looks good! I have a few questions to start!

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -167,61 +172,82 @@ export default class UBAFeeSpokeCalculator {
...
+    // Incentive Fee (LP Fee Component): first balancing fee - second balance fee
```
pxrl, line 188: I'd understood `Incentive (fee)` to be the non-LP component - i.e. the amount the actor pays to (or receives from) the penalty pot based on the action that they're taking and the balance of the SpokePool relative to its target balance.

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -167,61 +172,82 @@ export default class UBAFeeSpokeCalculator {
...
+    // the LP fee holders to compute the opposite flow and to essentially reverse the flow
+    const reverseFee = reverseIntegrandFeeFunction(reverseFlowCurve, reverseIntegrandStart, amount);
+
+    // We can now compute the LP fee component of the fee. This is also considered the incentive fee
+    const lpFee = mainFee.sub(reverseFee);
+
+    // We can now return the fee
+    return {
+      lpFee,
+      balancingFee: mainFee.sub(lpFee),
```
pxrl, line 230: As above, would we consider `incentive` rather than `balancingFee` ?

(I'm less thrilled about `incentiveFee` because it seems like the conjunction of two opposites)

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
pxrl: One option would be to pass in the full `Flow` object, and then use the `isInflow` and `isOutflow` predicates to determine which path to take. Presumably that's happening outside of this scope, per the fact that there are separate methods for inflow and outflow.

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
james-a-morris: This line of logic was defined during a call with @nicholaspai - I'll reach out to him to verify. Alternatively, @mrice32 or @cc7768 may be able to weigh in

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -168,60 +173,71 @@ export default class UBAFeeSpokeCalculator {
...
+      flowCurve,
+      runningBalance,
+      amount.mul(flowType === "inflow" ? 1 : -1)
+    );
 
-    // Resolve the historical running balance of the spoke
-    const depositRunningBalance = this.calculateHistoricalRunningBalance(flowRange?.startIndex, flowRange?.endIndex);
+    // Next, we'll need to compute the reverse balancing fee. This is the opportunity cost of
+    // the LP fee holders to compute the opposite flow and to essentially reverse the flow
+    const reverseFee = computePiecewiseLinearFunction(
```
nicholaspai, line 208: I don't think we need this `reverseFee` anymore and therefore its no longer affecting the `lpFee`. See this [commit](https://github.com/UMAprotocol/UMIPs/pull/581/commits/7152e17bc9a192adc16af2961a91d293be05f5ef) which implements the simplified variables that Matt proposed offline

nicholaspai (commented):
Just wanted to make sure you saw latest changes to the UMIP here: https://github.com/UMAprotocol/UMIPs/pull/581/commits/7152e17bc9a192adc16af2961a91d293be05f5ef

This overall should simplify the implementation but we might have to go backwards first... sorry about that. One major change is that `lpFee` is no longer equal to `reverseBalancingFee - balancingFee + utilizationFee` and is now just equal to `utilizationFee`

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
james-a-morris: Updated: https://github.com/across-protocol/sdk-v2/pull/217/commits/c31c655fc37d20307e203edb6da5f5206d937514

nicholaspai (approved):
LGTM from an overall structure perspective. I'm not 100% sure whether `flowCurve` is still relevant. We might only be using one curve per chain per token as opposed to having one inflow/outflow curve per chain per token

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
```diff
@@ -167,61 +171,56 @@ export default class UBAFeeSpokeCalculator {
...
+    // We first need to resolve the inflow/outflow curves for the deposit and refund spoke
+    const flowCurve = this.config.getBalancingFeeTuples(this.chainId);
+
+    // Next, we'll need to compute the first balancing fee from the running balance of the spoke
+    // to the running balance of the spoke + the amount
+    const lpFee = computePiecewiseLinearFunction(flowCurve, runningBalance, amount.mul(flowType === "inflow" ? 1 : -1));
+
+    // We can now return the fee
+    return {
+      lpFee,
```
mrice32, line 203: nit: shouldn't this be the balancing fee portion of the LP fee? Then the LP fee would be computed in a separate module?

mrice32 (commented):
Looks good! Just one question/comment

src/UBAFeeCalculator/UBAFeeSpokeCalculator.ts
james-a-morris: Fair! Updated here: https://github.com/across-protocol/sdk-v2/pull/217/commits/7621170ce3786d57d86e238f862a9cbc56489f75

PR #218, opened 2023-05-19
refactor(HubPoolClient): Reverse ConfigStoreClient dependencies

pxrl:
The HubPoolClient will need to read from the ConfigStoreClient, and so far there don't seem to be any hard reasons for why the ConfigStoreClient needs a HubPoolClient instance. Therefore, migrate the necessary parts of the ConfigStoreClient to the HubPoolClient, and supply the HubPoolClient with a ConfigStoreClient during instantiation.

### DIFF
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -13,7 +13,6 @@ import {
 
 import { Contract, BigNumber } from "ethers";
 import winston from "winston";
-import { Block } from "@ethersproject/abstract-provider";
 
 import {
   L1TokenTransferThreshold,
@@ -27,9 +26,7 @@ import {
   DisabledChainsUpdate,
 } from "../interfaces";
 
-import * as lpFeeCalculator from "../lpFeeCalculator";
-import { BlockFinder, across } from "@uma/sdk";
-import { HubPoolClient } from "./HubPoolClient";
+import { across } from "@uma/sdk";
 
 export const GLOBAL_CONFIG_STORE_KEYS = {
   MAX_RELAYER_REPAYMENT_LEAF_SIZE: "MAX_RELAYER_REPAYMENT_LEAF_SIZE",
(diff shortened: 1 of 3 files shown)


pxrl:
Note: I'm currently finalising the necessary adjustments for relayer-v2. It will be necessary to evaluate those when considering this PR.

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -27,9 +26,7 @@ import {
   DisabledChainsUpdate,
 } from "../interfaces";
 
-import * as lpFeeCalculator from "../lpFeeCalculator";
-import { BlockFinder, across } from "@uma/sdk";
-import { HubPoolClient } from "./HubPoolClient";
+import { across } from "@uma/sdk";
```
nicholaspai, line 29: Why do you import this?

src/clients/HubPoolClient.ts
```diff
@@ -1,3 +1,8 @@
+import { Contract, BigNumber, Event, EventFilter } from "ethers";
```
nicholaspai, line 1: Can you get these from `../utils`?

nicholaspai (approved):
Conceptually this makes sense to me that the ConfigStoreClient is upstream of the HubPoolClient.

Especially the function `getUtilization` belongs in the `HubPoolClient`

I like having the ConfigStoreClient just read ConfigStore variables and convert them into readable TS objects for the HubPoolClient to grab as needed

src/clients/AcrossConfigStoreClient.ts
pxrl: It's already imported (see line 31 above).

src/clients/HubPoolClient.ts
pxrl: I checked a few other subdirectories in sdk-v2 and they all seem to be importing objects directly from ethers, rather than via utils. I'll leave it as-is for now, but we should discuss further to determine whether we want to adjust that strategy.

PR #221, opened 2023-05-23
chore: Migrate mocked {Hub,Spoke}PoolClient from relayer-v2

pxrl:
Will be removed from relayer-v2 in a follow-up commit.

### DIFF
--- a/src/clients/mocks/MockEvents.ts
+++ b/src/clients/mocks/MockEvents.ts
@@ -1,4 +1,3 @@
-import { Result } from "@ethersproject/abi";
 import { utils as ethersUtils, Event, providers } from "ethers";
 import { random } from "lodash";
 
@@ -12,7 +11,8 @@ type EthersEventTemplate = {
   address: string;
   event: string;
   topics: string[];
-  args: Result;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  args: Record<string, any>;
   data?: string;
   blockNumber?: number;
   transactionIndex?: number;
--- a/src/clients/mocks/MockEvents.ts
+++ b/src/clients/mocks/MockEvents.ts
@@ -1,0 +1,76 @@
+import { utils as ethersUtils, Event, providers } from "ethers";
+import { random } from "lodash";
+
+const { id, keccak256, toUtf8Bytes } = ethersUtils;
+
+type Block = providers.Block;
+type TransactionResponse = providers.TransactionResponse;
+type TransactionReceipt = providers.TransactionReceipt;
+
+type EthersEventTemplate = {
+  address: string;
+  event: string;
+  topics: string[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  args: Record<string, any>;
+  data?: string;
+  blockNumber?: number;
+  transactionIndex?: number;
+};
+
+const getBlock = async (): Promise<Block> => {
+  throw new Error("getBlock() not supported");
+};
+const getTransaction = async (): Promise<TransactionResponse> => {
+  throw new Error("getTransaction() not supported");
+};
+const getTransactionReceipt = async (): Promise<TransactionReceipt> => {
+  throw new Error("getTransactionReceipt() not supported");
+};
+const removeListener = (): void => {
+  throw new Error("removeListener not supported");
+};
+
+export class EventManager {
+  private logIndexes: Record<string, number> = {};
+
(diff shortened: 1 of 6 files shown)


pxrl:
@james-a-morris Have confirmed that this works OK with the corresponding change on the relayer-v2 side :+1: Thanks for your help to resolve that!

pxrl:
Note: One thing we might consider here is the placement of the mock files.

In relayer-v2 we use test/mocks, but in sdk-v2 we tend to park all of the tests in a subdirectory of the components they are testing.

I'm not sure which is preferred. Does anyone have any thoughts?

james-a-morris (commented):
Is this a complete copy/paste migration or have there been refactoring changes?

src/clients/mocks/MockEvents.ts
```diff
@@ -0,0 +1,76 @@
+import { utils as ethersUtils, Event, providers } from "ethers";
+import { random } from "lodash";
+
+const { id, keccak256, toUtf8Bytes } = ethersUtils;
+
+type Block = providers.Block;
+type TransactionResponse = providers.TransactionResponse;
+type TransactionReceipt = providers.TransactionReceipt;
```
pxrl, line 8: Additional change: imports were adjusted to be a little bit simpler (i.e. import parts of ethers, rather than the entire package).

src/clients/mocks/MockEvents.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type EthersEventTemplate = {
+  address: string;
+  event: string;
+  topics: string[];
+  args: Result;
```
pxrl, line 15: Additional change: Swap the type annotation of args for the actual ethers `Result` type (weird readonly hybrid Object/Array thing).

At this stage I haven't seen any errors or warnings from doing this, so I think we should persist with this definition unless/until something breaks.

src/clients/mocks/MockHubPoolClient.ts
```diff
@@ -0,0 +1,210 @@
...
+export class MockHubPoolClient extends HubPoolClient {
+  public readonly minBlockRange = 10;
+  public rootBundleProposal = emptyRootBundle;
+
+  private events: Event[] = [];
+
+  private l1TokensMock: L1Token[] = []; // L1Tokens and their associated info.
+  private tokenInfoToReturn: L1Token = { address: "", decimals: 0, symbol: "" };
+  private l1TokensToDestinationTokensMock: { [l1Token: string]: { [destinationChainId: number]: string } } = {};
+  private returnedL1TokenForDeposit = "";
```
pxrl, line 30: Had to add initialisers for some of these, since we're more strict in sdk-v2.

src/clients/mocks/MockHubPoolClient.ts
```diff
@@ -0,0 +1,210 @@
...
+    // Ensure an array for every requested event exists, in the requested order.
+    // All requested event types must be populated in the array (even if empty).
+    const _events: Event[][] = eventNames.map(() => []);
+    this.events.flat().forEach((event) => {
+      const idx = eventNames.indexOf(event.event as string);
+      if (idx !== -1) {
+        _events[idx].push(event);
+      }
+    });
+    this.events = [];
```
pxrl, line 98: This functionality (and everything following) was originally part of [this](https://github.com/across-protocol/relayer-v2/pull/646/files#diff-5f71e436c0671139f600446c187299f6940429e44531633234838dffbc7e4e20) PR in relayer-v2. That PR is held up on something unrelated to the test changes introduced to the HubPoolClient, so I'm taking the opportunity to carve those changes out and get them over here instead. It will reduce the scope of the residual work to be done in relayer-v2 (and it would have had to be migrated anyway).

pxrl:
> Is this a complete copy/paste migration or have there been refactoring changes?

There were some changes that were necessary due to the slightly different (more strict) environment of sdk-v2.

This change does however also introduce new test functionality that is pending in a PR in relayer-v2. My intention is to get it in here instead, and then cut it out of that PR.

I've left annotations for the modifications that I made.

src/clients/mocks/MockSpokePoolClient.ts
pxrl, line 1: This file is taken as-is, with the only changes being adjustments to the import statements to fir the sdk-v2 repo, as well as swapping out nullish coalescing assignment as follows `??= ...` `= ... ?? ...`

james-a-morris (approved):
LGTM! Great refactor, @pxrl 🎉

PR #222, opened 2023-05-23
chore: Bump typescript ^4.6.2 => ^5.0.4

pxrl:
Aligns with relayer-v2.


pxrl:
Not sure why this is failing; seems related tsdx. I feel like we might need to do away with tsdx because it seems unmaintained.

PR #223, opened 2023-05-24
refactor(ConfigStoreClient): Backport relayer-v2 modifications

pxrl:
This change backports some relayer-v2 ConfigStoreClient customisations to the sdk-v2 repository. No behaviour is changed, though there is a small tweak in the way the ConfigStoreClient is instantiated. As part of this, also migrate some constants from relayer-v2, which are anyway more at home in sdk-v2.

### DIFF
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -1,3 +1,4 @@
+import { CHAIN_ID_LIST_INDICES } from "../constants";
 import {
   spreadEvent,
   sortEventsDescending,
@@ -9,6 +10,8 @@ import {
   toBN,
   max,
   sortEventsAscending,
+  CONFIG_STORE_VERSION,
+  DEFAULT_CONFIG_STORE_VERSION,
 } from "../utils";
 
 import { Contract, BigNumber } from "ethers";
@@ -35,6 +38,12 @@ export const GLOBAL_CONFIG_STORE_KEYS = {
   DISABLED_CHAINS: "DISABLED_CHAINS",
 };
 
+type ConfigStoreOverride = {
+  enabledChainIds: number[];
+  defaultConfigStoreVersion: number;
+  configStoreVersion: number;
+};
+
 export class AcrossConfigStoreClient {
   public cumulativeRateModelUpdates: across.rateModel.RateModelEvent[] = [];
   public cumulativeRouteRateModelUpdates: RouteRateModelUpdate[] = [];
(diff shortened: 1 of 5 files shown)


src/clients/AcrossConfigStoreClient.ts
```diff
@@ -56,14 +66,20 @@ export class AcrossConfigStoreClient {
...
-    protected readonly configOverride: {
-      enabledChainIds: number[];
-      defaultConfigStoreVersion: number;
-      configStoreVersion: number;
-    }
+    {
+      enabledChainIds = CHAIN_ID_LIST_INDICES,
+      configStoreVersion = CONFIG_STORE_VERSION,
+      defaultConfigStoreVersion = DEFAULT_CONFIG_STORE_VERSION,
+    }: { enabledChainIds?: number[]; configStoreVersion?: number; defaultConfigStoreVersion?: number } = {}
```
pxrl, line 73: The syntax here is pretty obtuse, but it permits the caller to independently set any of the set of { enabledChainIds, configStoreVersion, defaultConfigStoreVersion } as they please.

src/clients/mocks/MockConfigStoreClient.ts
pxrl, line 1: Copied directly, with only changes made to the import statements.

src/clients/AcrossConfigStoreClient.ts
james-a-morris: My only qualm with this is that it permits a set of opinionated values to this SDK. I'm not sure we should set constants inside the SDK itself but add additional documentation that mentions how to use it best. 

I'm happy to approve this pending @mrice32 sign-off.

james-a-morris (commented):
Left a comment: https://github.com/across-protocol/sdk-v2/pull/223/files#r1205791042

src/clients/AcrossConfigStoreClient.ts
pxrl: It's necessary for the code instantiating the class to specify which config version that are using, so I don't think we can do away with that. I am less confident about overriding the default config store value though - we have strong comments saying that it should only ever be set to 0, so I wonder if it needs to be externally configurable, or whether it can just be internal/private to the ConfigStoreClient.

pxrl: @james-a-morris Have updated after your comments. The user is now required to be explicit about their maximum supported config store version and supported chainIds. I've reinstated the corresponding changes over in the draft PR for relayer-v2: https://github.com/across-protocol/relayer-v2/pull/703.

PR #226, opened 2023-06-01
feat(UBA): :sparkles: add utilization calculation in the uba utils

james-a-morris:
This chaange adds a utility function to the `UBAFeeUtility` file. This simulates calculating the utilization as described by the Across Data team.

### DIFF
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -1,7 +1,6 @@
 import { BigNumber } from "ethers";
 import { MAX_SAFE_JS_INT } from "@uma/common/dist/Constants";
 import { toBN } from "../utils";
-import { parseEther } from "ethers/lib/utils";
 
 /**
  * Computes a linear integral over a piecewise function
@@ -210,6 +209,6 @@ export function calculateUtilization(
 ) {
   const numerator = hubBalance.add(ethSpokeBalance).add(targetSpoke.reduce((a, b) => a.add(b), BigNumber.from(0)));
   const denominator = hubEquity;
-  const result = numerator.mul(parseEther("1")).div(denominator).div(parseEther("1.0"));
+  const result = numerator.div(denominator);
   return BigNumber.from(10).pow(decimals).sub(result);
 }
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -210,6 +210,6 @@ export function calculateUtilization(
 ) {
   const numerator = hubBalance.add(ethSpokeBalance).add(targetSpoke.reduce((a, b) => a.add(b), BigNumber.from(0)));
   const denominator = hubEquity;
-  const dividend = numerator.mul(parseEther("1")).div(denominator).div(parseEther("1.0"));
-  return BigNumber.from(10).pow(decimals).sub(dividend);
+  const result = numerator.mul(parseEther("1")).div(denominator).div(parseEther("1.0"));
+  return BigNumber.from(10).pow(decimals).sub(result);
 }
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -206,11 +206,11 @@ export function calculateUtilization(
   hubBalance: BigNumber,
   hubEquity: BigNumber,
   ethSpokeBalance: BigNumber,
-  targetSpoke: { target: BigNumber; spokeChainId: number }[]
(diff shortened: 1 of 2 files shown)


src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -189,3 +190,26 @@ export function getDepositBalancingFee(
...
+export function calculateUtilization(
+  decimals: number,
+  hubBalance: BigNumber,
+  hubEquity: BigNumber,
+  ethSpokeBalance: BigNumber,
+  targetSpoke: BigNumber[]
+) {
+  const numerator = hubBalance.add(ethSpokeBalance).add(targetSpoke.reduce((a, b) => a.add(b), BigNumber.from(0)));
+  const denominator = hubEquity;
+  const dividend = numerator.mul(parseEther("1")).div(denominator).div(parseEther("1.0"));
```
mrice32, line 213: Two questions:
1. Is this the correct use of the term dividend? It seems like it's being used as the the result of the division, but it's generally synonymous with numerator, no? You could call this unutilizedPct or similar in this case, since I think that's what is represented here.
2. Why are we multiplying _and dividing by_ 1e18? Shouldn't we only be multiplying? My rough understanding is that 18 decimal fixed point divisions of should roughly look like this: `x * 1e18 / y` and you only divide by 1e18 in multiplications, which would look like this: `x * y / 1e18`. This is just my intuition, though, so I could be wrong. It might help to toss in a few quick tests to make sure this math works on a few examples.

mrice32 (commented):
Looks good! Just one comment.

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -189,3 +189,26 @@ export function getDepositBalancingFee(
...
+ * @returns The utilization of the hub pool
+ */
+export function calculateUtilization(
+  decimals: number,
+  hubBalance: BigNumber,
+  hubEquity: BigNumber,
+  ethSpokeBalance: BigNumber,
+  targetSpoke: BigNumber[]
+) {
+  const numerator = hubBalance.add(ethSpokeBalance).add(targetSpoke.reduce((a, b) => a.add(b), BigNumber.from(0)));
```
nicholaspai, line 210: Does `targetSpoke` remove the ETH target spoke? Just want to confirm it does.

Also, nit: should it be named `spokeTargets` or `targetSpokes`

nicholaspai (approved):
One question and variable naming suggestion!

src/UBAFeeCalculator/UBAFeeUtility.ts
mrice32: I think we might have a slight miscommunication: I think the `.mul(parseEther("1"))` is required. I was only suggesting that you remove the `.div(parseEther("1.0"))`. In the original calculation, I think the result would be too small by a factor of 1e18. Now I think you will have rounding issues and the result will be too small by a factor of 1e18.

**Put differently, I think the right way to do this fixed point division (unless I'm missing something!) is `numerator.mul(parseEther("1")).div(denominator)`.**

Here's a quick example:
```
// In 18 decimal precision fixed-point, the raw number 1e18 represents 1.0 with 18 decimals of precision.

// Likewise, a raw number 5e18 represents 5.0.
x = 5e18

// Raw number 3e18 represents 3.0.
y = 3e18

// We want to multiply x, which is representing 5.0, by y, which is representing 3.0.
// This should equal 15.0.

// If we just multiply naively, we get 15e36. This raw number translates to 15e18, which is _not_ what we want.
z_incorrect = x * y

// To correct for this, all multiplications must divide by the fixed point factor.
// This gives us 15e18, which is what we want.
z_correct = x * y / 1e18

// For division, you need a different type of correction.
// This division doesn't work. 15e18 / 3e18 = 5.
// The raw number 5 translates to the semantic value of 5e-18, which is off by a factor of 1e18.
x_recomputed_incorrect = z_correct / y. 

// To correct for this, we multiply by 1e18 before doing the division.
// 15e18 * 1e18 / 3e18 = 5e18, which is equivalent to 5.0. This is what we want!
x_recomputed_correct = z_correct * 1e18 / y.
```

mrice32 (commented):
One comment [here](https://github.com/across-protocol/sdk-v2/pull/226#discussion_r1218665036). I'm not sure the math works. Sorry for the slightly tedious example if you're already familiar with this style of math :). Let me know if you think I'm missing some core assumption of this function!

james-a-morris:
> 7f0648f

No this was perfect. Thanks for the example - I had tunnel vision on the computation. https://github.com/across-protocol/sdk-v2/pull/226/commits/7f0648f6c6bd0a9979883dd775206e5e9db3f452

mrice32 (approved):
LGTM!

PR #227, opened 2023-06-02
improve: Split ConfigStoreClient update()

pxrl:
This change provides an option to override ConfigStoreClient on-chain queries
via _update(). The pattern is pretty consistent with what we have in the
HubPoolClient and SpokePoolClient. For the MockConfigStoreClient it was
necessary to inherit super._update() by default, for compatibility with existing
uses of the MockConfigStoreClient in relayer-v2 tests. Ultimately it would be
nice to remove that.

Addtionally, support mocked getBlock() for HubPoolClient test. This is needed
for the pending runningBalances UBA update.

### DIFF
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -1,3 +1,4 @@
+import assert from "assert";
 import { isError } from "../typeguards";
 import {
   isDefined,
@@ -12,7 +13,7 @@ import {
   max,
   sortEventsAscending,
 } from "../utils";
-import { Contract, BigNumber } from "ethers";
+import { Contract, BigNumber, Event } from "ethers";
 import winston from "winston";
 
 import {
@@ -27,6 +28,18 @@ import {
   DisabledChainsUpdate,
 } from "../interfaces";
 import { across } from "@uma/sdk";
+
+type _ConfigStoreUpdate = {
+  success: true;
+  latestBlockNumber: number;
+  searchEndBlock: number;
+  events: {
+    updatedTokenConfigEvents: Event[];
+    updatedGlobalConfigEvents: Event[];
+    globalConfigUpdateTimes: number[];
+  };
+};
+export type ConfigStoreUpdate = { success: false } | _ConfigStoreUpdate;
 
 // Version 0 is the implicit ConfigStore version from before the version attribute was introduced.
 // @dev Do not change this value.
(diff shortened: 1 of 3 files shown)


src/clients/AcrossConfigStoreClient.ts
```diff
@@ -208,27 +222,53 @@ export class AcrossConfigStoreClient {
     return this.configStoreVersion >= version;
   }
 
-  async update(): Promise<void> {
+  async _update(): Promise<ConfigStoreUpdate> {
+    const latestBlockNumber = await this.configStore.provider.getBlockNumber();
     const searchConfig = {
       fromBlock: this.firstBlockToSearch,
-      toBlock: this.eventSearchConfig.toBlock || (await this.configStore.provider.getBlockNumber()),
+      toBlock: this.eventSearchConfig.toBlock || latestBlockNumber,
```
james-a-morris, line 229: Is there a reason for this modification?

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -208,27 +222,53 @@ export class AcrossConfigStoreClient {
...
+
+    const result = await this._update();
+    if (!result.success) {
+      return;
+    }
+    const { updatedTokenConfigEvents, updatedGlobalConfigEvents, globalConfigUpdateTimes } = result.events;
+    assert(
+      updatedGlobalConfigEvents.length === globalConfigUpdateTimes.length,
+      `GlobalConfigUpdate events mismatch (${updatedGlobalConfigEvents.length} != ${globalConfigUpdateTimes.length})`
+    );
```
james-a-morris, line 270: Should we assert to unwind the stack?

james-a-morris (commented):
Left some comments - this looks great so far

src/clients/AcrossConfigStoreClient.ts
pxrl: Recording the `latestBlockNumber` is a common pattern from the HubPoolClient and SpokePoolClient classes. I had been using it during test, but ultimately found that it wasn't required. I've backed it out from this PR now, since there's not strictly any need for it.

james-a-morris (approved):
LGTM - finalized my questions async. Thanks @pxrl

src/clients/AcrossConfigStoreClient.ts
pxrl: Actually, it's needed for the Mocked variant, where the latest block number needs to be retained between updates. I've restored it.

PR #229, opened 2023-06-05
improve: Reduce noise in ConfigStoreClient

pxrl:
It's annoying to be warned about invalid updates from > 1 year ago.

### DIFF
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -338,9 +338,6 @@ export class AcrossConfigStoreClient {
       } catch (err) {
         // @dev averageBlockTimeSeconds does not actually block.
         const maxWarnAge = (24 * 60 * 60) / (await utils.averageBlockTimeSeconds());
-        console.log(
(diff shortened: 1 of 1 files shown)


src/clients/AcrossConfigStoreClient.ts
```diff
@@ -282,12 +282,15 @@ export class AcrossConfigStoreClient {
 
         // If Token config doesn't contain all expected properties, skip it.
         if (!(rateModelForToken && transferThresholdForToken)) {
-          this.logger.warn({
-            at: "ConfigStore",
-            message: "Ignoring invalid rate model update.",
-            update: args,
-            transferThresholdForToken,
-          });
+          const maxWarnAge = 21_600; // ~3 days @ 12s/block.
```
nicholaspai, line 285: nit: Can you grab the 12s/block hardcoded value from anywhere else so we don't have multiple assumptions in the code?

pxrl: Implemented here: https://github.com/across-protocol/sdk-v2/pull/229/commits/6d6218a891dcffc9a03a151915e4f3f4d112d7d1

PR #230, opened 2023-06-06
improve(ConfigStoreClient): Additional version helper

pxrl:
Allow the applicable version to be searched by block number. This is useful for the pending UBA version bump. Additionally, build the array of cumulative version updates in descending order, so that it doesn't have to be re-ordered on each use.

### DIFF
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -4,6 +4,7 @@ import { isError } from "../typeguards";
 import {
   isDefined,
   spreadEvent,
+  sortEventsAscendingInPlace,
   sortEventsDescending,
   spreadEventWithBlockNumber,
   paginatedEventQuery,
@@ -208,9 +209,13 @@ export class AcrossConfigStoreClient {
   }
 
   getConfigStoreVersionForTimestamp(timestamp: number = Number.MAX_SAFE_INTEGER): number {
-    const config = (sortEventsDescending(this.cumulativeConfigStoreVersionUpdates) as ConfigStoreVersionUpdate[]).find(
-      (config) => config.timestamp <= timestamp
-    );
+    const config = this.cumulativeConfigStoreVersionUpdates.find((config) => config.timestamp <= timestamp);
+    return isDefined(config) ? Number(config.value) : DEFAULT_CONFIG_STORE_VERSION;
+  }
+
+  // @dev This method assumes a block with a version change must _not_ coincide with an executed root bundle.
+  getConfigStoreVersionForBlock(blockNumber: number): number {
+    const config = this.cumulativeConfigStoreVersionUpdates.find((config) => config.blockNumber <= blockNumber);
     return isDefined(config) ? Number(config.value) : DEFAULT_CONFIG_STORE_VERSION;
   }
 
@@ -240,6 +245,9 @@ export class AcrossConfigStoreClient {
       paginatedEventQuery(this.configStore, this.configStore.filters.UpdatedTokenConfig(), searchConfig),
       paginatedEventQuery(this.configStore, this.configStore.filters.UpdatedGlobalConfig(), searchConfig),
     ]);
+
+    // Events *should* normally be received in ascending order, but explicitly enforce the ordering.
+    [updatedTokenConfigEvents, updatedGlobalConfigEvents].forEach((events) => sortEventsAscendingInPlace(events));
 
     const globalConfigUpdateTimes = (
       await Promise.all(updatedGlobalConfigEvents.map((event) => this.configStore.provider.getBlock(event.blockNumber)))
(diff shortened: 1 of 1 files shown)


src/clients/AcrossConfigStoreClient.ts
```diff
@@ -385,23 +392,16 @@ export class AcrossConfigStoreClient {
...
-        const lastValue =
-          this.cumulativeConfigStoreVersionUpdates.length === 0
-            ? DEFAULT_CONFIG_STORE_VERSION
-            : Number(
-                this.cumulativeConfigStoreVersionUpdates[this.cumulativeConfigStoreVersionUpdates.length - 1].value
-              );
-
-        // If version is not > last version, skip.
+        // Extract the current highest version. Require that the version always increments, otherwise skip the update.
+        const lastValue = Number(this.cumulativeConfigStoreVersionUpdates[0]?.value || DEFAULT_CONFIG_STORE_VERSION);
```
pxrl, line 396:
```suggestion
        const lastValue = Number(this.cumulativeConfigStoreVersionUpdates[0]?.value ?? DEFAULT_CONFIG_STORE_VERSION);
```

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -241,6 +245,9 @@ export class AcrossConfigStoreClient {
       paginatedEventQuery(this.configStore, this.configStore.filters.UpdatedGlobalConfig(), searchConfig),
     ]);
 
+    // Events *should* normally be received in ascending order, but explicitly enforce the ordering.
+    [updatedTokenConfigEvents, updatedGlobalConfigEvents].forEach((events) => sortEventsAscendingInPlace(events));
+
```
james-a-morris, line 250: QQ: Will sorting this in place cause any external issues regarding how we treat these variables?

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -385,23 +392,16 @@ export class AcrossConfigStoreClient {
...
         if (value <= lastValue) {
           continue;
         }
 
-        this.cumulativeConfigStoreVersionUpdates.push({
-          ...args,
-          timestamp: globalConfigUpdateTimes[i],
-        });
+        // Prepend the update to impose descending ordering for version updates.
+        this.cumulativeConfigStoreVersionUpdates = [{ ...args, timestamp: globalConfigUpdateTimes[i] }].concat(
```
james-a-morris, line 402: I think we've been using the spread operator `[...a, ...b]` instead of `concat` - should we retain that usage?

james-a-morris (commented):
Looks great, @pxrl ! Left a few nits

src/clients/AcrossConfigStoreClient.ts
pxrl: The objects have a relatively short lifespan - they're populated in `_update()`, and are subsequently iterated over in order to transform them into Across-specific objects. There is an implicit assumption in the code that the events are received in ascending order, so I don't see any possible negative side-effects from enforcing the ordering here.

src/clients/AcrossConfigStoreClient.ts
pxrl: The differences between `.concat()` and the spread operator otherwise don't seem to be very important for our use cases, so I'd probably prefer to use the one that looks the least ugly. Prettier was actually butchering the code with `.concat()` and the spread operator avoids that, so that's an improvement!

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -208,9 +209,12 @@ export class AcrossConfigStoreClient {
...
 
   getConfigStoreVersionForTimestamp(timestamp: number = Number.MAX_SAFE_INTEGER): number {
-    const config = (sortEventsDescending(this.cumulativeConfigStoreVersionUpdates) as ConfigStoreVersionUpdate[]).find(
-      (config) => config.timestamp <= timestamp
-    );
+    const config = this.cumulativeConfigStoreVersionUpdates.find((config) => config.timestamp <= timestamp);
+    return isDefined(config) ? Number(config.value) : DEFAULT_CONFIG_STORE_VERSION;
+  }
+
+  getConfigStoreVersionForBlock(blockNumber: number): number {
```
nicholaspai, line 216: Why would you need these two functions? IIUC you would need to use this function to figure out which config store parameters to apply to a deposit/fill. These events all have `quoteTImestamps` attached to them so you should always be able to compare by timestamp, which is how the client is currently set up. When would you ever need to use block? And in those cases, wouldn't you always be able to grab the timestamp for that block?

nicholaspai (commented):
Why would you need these two functions? IIUC you would need to use this function to figure out which config store parameters to apply to a deposit/fill. These events all have `quoteTImestamps` attached to them so you should always be able to compare by timestamp, which is how the client is currently set up. When would you ever need to use block? And in those cases, wouldn't you always be able to grab the timestamp for that block?

nicholaspai (approved):

src/clients/AcrossConfigStoreClient.ts
pxrl: Great point...have left that as a comment here: https://github.com/across-protocol/sdk-v2/pull/213/files#r1223234563.

PR #235, opened 2023-06-09
fix(mock): Don't override generated fill destinationToken

pxrl:
The SpokePoolClient tries to resolve the destinationToken against the set of known tokens in the HubPoolClient, so it's important to propagate the value supplied by the caller. If no value has been supplied, sub in the ZERO_ADDRESS, because this is what the SpokePoolClient resolves when it has no reference for the HubPoolClient.

In the mean time, there are workarounds in some relayer-v2 tests to plaster over the issue. They will be fixed once this change migrates back to relayer-v2 via an sdk-v2 bump.


james-a-morris (approved):
Good call

PR #236, opened 2023-06-09
improve: perform additional UBA Client calculations

james-a-morris:
Closes ACX-1206

### DIFF
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -3,6 +3,7 @@ import { MAX_SAFE_JS_INT } from "@uma/common/dist/Constants";
 import { toBN } from "../utils";
 import { HUBPOOL_CHAIN_ID } from "../constants";
 import { parseEther } from "ethers/lib/utils";
+import { UBAActionType } from ".";
 
 /**
  * Computes a linear integral over a piecewise function
@@ -306,7 +307,7 @@ export function calculateUtilization(
 
 export function calculateUtilizationBoundaries(
   action: {
-    actionType: "deposit" | "refund";
+    actionType: UBAActionType;
     amount: BigNumber;
     chainId: number;
   },
@@ -319,7 +320,7 @@ export function calculateUtilizationBoundaries(
 ): { utilizationPostTx: BigNumber; utilizationPreTx: BigNumber } {
   let newEthSpokeBalance = ethSpokeBalance;
   if (action.chainId === hubPoolChainId) {
-    if (action.actionType === "deposit") {
+    if (action.actionType === UBAActionType.Deposit) {
       newEthSpokeBalance = newEthSpokeBalance.add(action.amount);
     } else {
       newEthSpokeBalance = newEthSpokeBalance.sub(action.amount);
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -303,6 +303,32 @@ export function calculateUtilization(
   return BigNumber.from(10).pow(decimals).sub(result);
 }
 
+export function calculationUtilizationBoundaries(
+  action: {
+    actionType: "deposit" | "refund";
+    amount: BigNumber;
+    chainId: number;
+  },
+  decimals: number,
+  hubBalance: BigNumber,
+  hubEquity: BigNumber,
+  ethSpokeBalance: BigNumber,
+  spokeTargets: { target: BigNumber; spokeChainId: number }[]
+): { utilizationPostTx: BigNumber; utilizationPreTx: BigNumber } {
+  let newEthSpokeBalance = ethSpokeBalance;
+  if (action.chainId === HUBPOOL_CHAIN_ID) {
+    if (action.actionType === "deposit") {
+      newEthSpokeBalance = newEthSpokeBalance.add(action.amount);
+    } else {
+      newEthSpokeBalance = newEthSpokeBalance.sub(action.amount);
+    }
+  }
+  return {
+    utilizationPreTx: calculateUtilization(decimals, hubBalance, hubEquity, ethSpokeBalance, spokeTargets),
+    utilizationPostTx: calculateUtilization(decimals, hubBalance, hubEquity, newEthSpokeBalance, spokeTargets),
+  };
+}
+
 /**
  * A mapping of the balancing fee functions to the inflow/outflow types. This is used to
  * as a convenience to avoid having to do multiple if/else statements in the UBAFeeCalculator
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -293,11 +293,12 @@ export function calculateUtilization(
   hubBalance: BigNumber,
   hubEquity: BigNumber,
   ethSpokeBalance: BigNumber,
-  spokeTargets: { target: BigNumber; spokeChainId: number }[]
+  spokeTargets: { target: BigNumber; spokeChainId: number }[],
+  hubPoolChainId = HUBPOOL_CHAIN_ID
 ) {
   const numerator = hubBalance
     .add(ethSpokeBalance)
-    .add(spokeTargets.reduce((a, b) => (b.spokeChainId !== HUBPOOL_CHAIN_ID ? a.add(b.target) : a), BigNumber.from(0)));
+    .add(spokeTargets.reduce((a, b) => (b.spokeChainId !== hubPoolChainId ? a.add(b.target) : a), BigNumber.from(0)));
   const denominator = hubEquity;
   const result = numerator.mul(parseEther("1.0")).div(denominator); // We need to multiply by 1e18 to get the correct precision for the result
   return BigNumber.from(10).pow(decimals).sub(result);
@@ -313,10 +314,11 @@ export function calculateUtilizationBoundaries(
   hubBalance: BigNumber,
   hubEquity: BigNumber,
   ethSpokeBalance: BigNumber,
-  spokeTargets: { target: BigNumber; spokeChainId: number }[]
+  spokeTargets: { target: BigNumber; spokeChainId: number }[],
+  hubPoolChainId = HUBPOOL_CHAIN_ID
 ): { utilizationPostTx: BigNumber; utilizationPreTx: BigNumber } {
   let newEthSpokeBalance = ethSpokeBalance;
-  if (action.chainId === HUBPOOL_CHAIN_ID) {
+  if (action.chainId === hubPoolChainId) {
     if (action.actionType === "deposit") {
       newEthSpokeBalance = newEthSpokeBalance.add(action.amount);
     } else {
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -303,7 +303,7 @@ export function calculateUtilization(
   return BigNumber.from(10).pow(decimals).sub(result);
 }
 
-export function calculationUtilizationBoundaries(
+export function calculateUtilizationBoundaries(
   action: {
     actionType: "deposit" | "refund";
     amount: BigNumber;
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -432,6 +432,17 @@ export class AcrossConfigStoreClient {
     return disabledChains.filter((chainId: number) => !isNaN(chainId) && Number.isInteger(chainId) && chainId !== 1);
   }
 
+  async getUBATargetSpokeBalances(
+    chainIds: number[],
+    l1TokenAddress: string,
+    blockNumber?: number
+  ): Promise<{ spokeChainId: number; target: BigNumber }[]> {
+    return chainIds.map((chainId) => {
+      const target = this.getSpokeTargetBalancesForBlock(l1TokenAddress, chainId, blockNumber).target;
+      return { spokeChainId: chainId, target };
+    });
+  }
+
   // THIS IS A STUB FOR NOW
   async getUBAFeeConfig(
     chainId: number,
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -30,6 +30,7 @@ import {
   DisabledChainsUpdate,
 } from "../interfaces";
 import { across } from "@uma/sdk";
+import { UBAFeeConfig } from "../UBAFeeCalculator";
 
 type _ConfigStoreUpdate = {
   success: true;
@@ -429,5 +430,25 @@ export class AcrossConfigStoreClient {
     // If any chain ID's are not integers then ignore. UMIP-157 requires that this key cannot include
     // the chain ID 1.
     return disabledChains.filter((chainId: number) => !isNaN(chainId) && Number.isInteger(chainId) && chainId !== 1);
+  }
+
+  async getUBAFeeConfig(
+    chainId: number,
+    token: string,
+    blockNumber: number | "latest" = "latest"
+  ): Promise<UBAFeeConfig> {
+    chainId;
+    token;
+    blockNumber;
+    return new UBAFeeConfig(
+      {
+        default: toBN(0),
+      },
+      toBN(0),
+      {
+        default: [],
+      },
+      {}
+    );
   }
 }
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -448,7 +448,10 @@ export class AcrossConfigStoreClient {
       {
         default: [],
       },
-      {}
+      {},
+      {
+        default: [],
+      }
     );
   }
 }
--- a/src/clients/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient.ts
@@ -432,6 +432,7 @@ export class AcrossConfigStoreClient {
     return disabledChains.filter((chainId: number) => !isNaN(chainId) && Number.isInteger(chainId) && chainId !== 1);
   }
 
+  // THIS IS A STUB FOR NOW
   async getUBAFeeConfig(
     chainId: number,
     token: string,
--- a/src/clients/UBAClient/UBAClientAbstract.ts
+++ b/src/clients/UBAClient/UBAClientAbstract.ts
@@ -1,16 +1,22 @@
 import winston from "winston";
 import { RefundRequestWithBlock, UbaFlow } from "../../interfaces";
 import { BigNumber } from "ethers";
+import { UBAFeeSpokeCalculator } from "../../UBAFeeCalculator";
 
 export type RequestValidReturnType = { valid: false; reason: string } | { valid: true };
 export type OpeningBalanceReturnType = { blockNumber: number; spokePoolBalance: BigNumber };
+export type BalancingFeeReturnType = { depositBalancingFee: BigNumber; refundBalancingFee: BigNumber };
 
 /**
  * UBAClient is a base class for UBA functionality. It provides a common interface for UBA functionality to be implemented on top of or extended.
  * This class is not intended to be used directly, but rather extended by other classes that implement the abstract methods.
  */
 export abstract class BaseUBAClient {
-  protected constructor(protected readonly chainIdIndices: number[], protected readonly logger?: winston.Logger) {}
+  protected spokeUBAFeeCalculators: { [chainId: number]: { [token: string]: UBAFeeSpokeCalculator } };
+
+  protected constructor(protected readonly chainIdIndices: number[], protected readonly logger?: winston.Logger) {
+    this.spokeUBAFeeCalculators = {};
+  }
 
   /**
    * Retrieves the opening balance for a given token on a given chainId at a given block number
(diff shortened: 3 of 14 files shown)


src/lpFeeCalculator/lpFeeCalculator.ts
james-a-morris, line 1: Removed the `depcrecated` error because honestly when I initially put it there I didn't realize that it could be just as useful as dealing with the class itself.

src/clients/UBAClient/UBAClientWithManual.ts
james-a-morris, line 1: This is a WiP - it doesn't have any useful purpose right now.

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris, line 1: Creating this file for now to earmark it for additional refactoring

src/clients/UBAClient/UBAClientAbstract.ts
james-a-morris, line 1: This is an abstract class that we can use for both the RPC & manual variant of the UBA Client. It contains several stubs that need to be implemented as well as central logic that builds off of these functions.

src/clients/AcrossConfigStoreClient.ts
```diff
@@ -430,4 +431,27 @@ export class AcrossConfigStoreClient {
...
+      toBN(0),
+      {
+        default: [],
+      },
+      {},
+      {
+        default: [],
+      }
+    );
+  }
```
james-a-morris, line 457: This is a placeholder to appease TS

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -303,6 +303,32 @@ export function calculateUtilization(
...
+    } else {
+      newEthSpokeBalance = newEthSpokeBalance.sub(action.amount);
+    }
+  }
+  return {
+    utilizationPreTx: calculateUtilization(decimals, hubBalance, hubEquity, ethSpokeBalance, spokeTargets),
+    utilizationPostTx: calculateUtilization(decimals, hubBalance, hubEquity, newEthSpokeBalance, spokeTargets),
+  };
+}
+
```
james-a-morris, line 331: This generates the initial bounds in order to calculate the system fee.

This is the `u_pre` and `u_post` defined [here](https://github.com/UMAprotocol/UMIPs/blob/680523cbbcb7a896abbe3d60504861ceeabeef29/UMIPs/umip-157.md#computing-the-lp-fee)

src/clients/AcrossConfigStoreClient.ts
pxrl: Just to clarify, is it necessary to introduce UBA terminology into the ConfigStoreClient? I'm not up to speed exactly on how the config records will be set for UBA, but I'd anticipated that we'd be accessing UBA config via the generic token and global config stores.

Fundamentally I think the lower layers in the stack should avoid  depending on upper layers, but I concede that I don't have the full picture on the requirements for it.

src/clients/UBAClient/UBAClientAbstract.ts
```diff
@@ -0,0 +1,139 @@
+import winston from "winston";
+import { RefundRequestWithBlock, UbaFlow } from "../../interfaces";
+import { BigNumber } from "ethers";
+import { UBAFeeSpokeCalculator } from "../../UBAFeeCalculator";
+
+export type RequestValidReturnType = { valid: false; reason: string } | { valid: true };
+export type OpeningBalanceReturnType = { blockNumber: number; spokePoolBalance: BigNumber };
```
pxrl, line 7: Minor: FWIW after naming it `spokePoolBalance` I was wondering if `tokenBalance` would be better.

src/clients/UBAClient/UBAClientAbstract.ts
```diff
@@ -0,0 +1,139 @@
...
+    hubPoolBlockNumber?: number
+  ): OpeningBalanceReturnType;
+
+  /**
+   * Gets the latest block number for a given chainId in the state of the lastest closing block
+   * @param chainId The chainId to get the latest block number for
+   * @returns The latest block number for the given chainId
+   * @note Assumes that the `spoke[...].bundleEndBlocks` are sorted in ascending order
+   */
+  protected abstract resolveClosingBlockNumber(chainId: number, blockNumber: number): number;
```
pxrl, line 42: Is the blockNumber argument intended for use in resolving a closing block number in some previous period?

Also wondering if there should be some consideration for a caller-specified follow distance. i.e. if we're trying to resolve a proposed closing block for the current UBA period, we obviously don't want to take the head of a chain that has re-org risk. It's hard to know exactly where in the stack that should be placed, though.

src/clients/UBAClient/UBAClientAbstract.ts
```diff
@@ -0,0 +1,139 @@
...
+   * @dev  Callers should evaluate 'valid' before 'reason' in the return object.
+   * @dev  The following RefundRequest attributes are not evaluated for validity and should be checked separately:
+   * @dev  - previousIdenticalRequests
+   * @dev  - Age of blockNumber (i.e. according to SpokePool finality)
+   * @param chainId       ChainId of SpokePool where refundRequest originated.
+   * @param refundRequest RefundRequest object to be evaluated for validity.
+   */
+  public abstract refundRequestIsValid(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _chainId: number,
```
pxrl, line 72: FWIW the reason for including the chainId here is because the RefundRequest[WithBlock] type doesn't actually identify which chain the transaction occurred on. This is more or less unique amongst our SpokePool events; we might want to consider adding something like `refundChainId` to the RefundRequest type.

src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -1,29 +1,32 @@
...
+import { UBAFeeSpokeCalculator } from "../../UBAFeeCalculator";
 
-export class UBAClient {
+export class UBAClientWithRefresh extends BaseUBAClient {
   // @dev chainIdIndices supports indexing members of root bundle proposals submitted to the HubPool.
   //      It must include the complete set of chain IDs ever supported by the HubPool.
   // @dev SpokePoolClients may be a subset of the SpokePools that have been deployed.
   constructor(
-    private readonly chainIdIndices: number[],
+    chainIdIndices: number[],
```
pxrl, line 15: Should these not remain at least readonly? I can't see any reason we'd want them to change after instantiation.

james-a-morris: Fair point! I didn't realize you could make any variable readonly. You learn something new every day

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -303,6 +303,32 @@ export function calculateUtilization(
...
+    chainId: number;
+  },
+  decimals: number,
+  hubBalance: BigNumber,
+  hubEquity: BigNumber,
+  ethSpokeBalance: BigNumber,
+  spokeTargets: { target: BigNumber; spokeChainId: number }[]
+): { utilizationPostTx: BigNumber; utilizationPreTx: BigNumber } {
+  let newEthSpokeBalance = ethSpokeBalance;
+  if (action.chainId === HUBPOOL_CHAIN_ID) {
```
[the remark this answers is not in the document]
james-a-morris: Good call - I've used the `hubPoolClient.chainId` to dynamically set the chain id of the hub pool. https://github.com/across-protocol/sdk-v2/pull/236/commits/7be467a9822c43ceb261b16b8c23c654d019a01f

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -293,16 +293,44 @@ export function calculateUtilization(
...
-    .add(spokeTargets.reduce((a, b) => (b.spokeChainId !== HUBPOOL_CHAIN_ID ? a.add(b.target) : a), BigNumber.from(0)));
+    .add(spokeTargets.reduce((a, b) => (b.spokeChainId !== hubPoolChainId ? a.add(b.target) : a), BigNumber.from(0)));
   const denominator = hubEquity;
   const result = numerator.mul(parseEther("1.0")).div(denominator); // We need to multiply by 1e18 to get the correct precision for the result
   return BigNumber.from(10).pow(decimals).sub(result);
 }
 
+export function calculateUtilizationBoundaries(
+  action: {
+    actionType: "deposit" | "refund";
```
nicholaspai, line 309: should "deposit" and "refund" be added to some `enum` if we use them elsewhere?

src/clients/UBAClient/UBAClientAbstract.ts
nicholaspai: isn't it always going to refer to the balance of the spoke pool? If so, I would find `tokenBalance` confusing sometimes since it sounds like the wallet balance of an EOA, such as the relayer

src/clients/UBAClient/UBAClientAbstract.ts
```diff
@@ -0,0 +1,224 @@
...
+    refundChainId: number
+  ): Promise<BalancingFeeReturnType> {
+    // Verify that the spoke clients are instantiated.
+    await Promise.all([
+      this.instantiateUBAFeeCalculator(depositChainId, spokePoolToken, hubPoolBlockNumber),
+      this.instantiateUBAFeeCalculator(refundChainId, spokePoolToken, hubPoolBlockNumber),
+    ]);
+    // Get the balancing fees.
+    const [{ balancingFee: depositBalancingFee }, { balancingFee: refundBalancingFee }] = await Promise.all([
+      this.spokeUBAFeeCalculators[depositChainId][spokePoolToken].getDepositFee(amount),
```
nicholaspai, line 111: Do you always need to compute both the deposit and refund balancing fees for any given `amount`? I thought you'd only need to compute one, based on which direction the flow is

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function computeRealizedLpFeeForRefresh(
+  l1TokenAddress: string,
+  depositChainId: number,
+  refundChainId: number,
+  amount: BigNumber,
+  hubPoolClient: HubPoolClient,
+  spokePoolClients: { [chainId: number]: SpokePoolClient },
+  baselineFee: BigNumber,
+  gammaCutoff: FlowTupleParameters,
+  blockNumber?: number
```
nicholaspai, line 30: Why make `blockNumber` optional? Isn't this always required?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @param depositChainId The chainId of the deposit
+ * @param refundChainId The chainId of the refund
+ * @param amount The amount that is being deposited
+ * @param hubPoolClient A hubpool client instance to query the hubpool
+ * @param spokePoolClients A mapping of spoke chainIds to spoke pool clients
+ * @param baselineFee The baseline fee to use for this given token
+ * @param gammaCutoff The gamma cutoff to use for this given token - used in the piecewise linear function calculation
+ * @returns The realized LP fee for the given token on the given chainId at the given block number
+ */
+export async function computeRealizedLpFeeForRefresh(
```
nicholaspai, line 21:
```suggestion
export async function computeLpFeeForRefresh(
```

Should we rename this to `LpFee` instead of `realizedLpFee`? 

IIUC:
- "LpFee" = component of total `SystemFee` based on integral of utilization function
- "DepositBalancingFee" = component of total `SystemFee` based on balancing fee charged to inflow on origin chain
- "SystemFee" = LPFee + DepositBalancingFee

And ultimately the relayer should set `realizedLpFee = SystemFee` when filling any deposit

src/clients/UBAClient/UBAClientWithManual.ts
```diff
@@ -0,0 +1,119 @@
...
+    protected readonly state: UBAClientState,
+    readonly logger?: winston.Logger
+  ) {
+    super(chainIdIndices, logger);
+    assert(chainIdIndices.length > 0, "No chainIds provided");
+    assert(Object.values(state.spoke).length > 0, "No SpokePools provided");
+  }
+
+  protected resolveClosingBlockNumber(chainId: number, blockNumber: number): number {
+    return this.state.spoke[chainId].bundleEndBlocks.find((bundleEndBlock) => bundleEndBlock <= blockNumber) ?? -1;
```
nicholaspai, line 53: Not sure how you wanna reconcile this code with the very similar function from `relayer-v2` [here](https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/DataworkerUtils.ts#L51)

I think your code is a lot cleaner FWIW.

src/clients/UBAClient/UBAClientWithManual.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  public getFlows(chainId: number, fromBlock?: number, toBlock?: number): UbaFlow[] {
+    const { flows } = this.state.spoke[chainId];
+    if (flows.length === 0) {
+      return [];
+    }
+    return flows.filter((flow) =>
+      isDefined(fromBlock)
+        ? flow.blockNumber >= fromBlock
+        : true && isDefined(toBlock)
```
nicholaspai, line 85:
```suggestion
        : isDefined(toBlock)
```

why is that `true` there?

src/utils/common.ts
```diff
@@ -78,6 +78,16 @@ export function max(a: BigNumberish, b: BigNumberish): BN {
   return bna.gte(bnb) ? bna : bnb;
 }
 
+/**
+ * Compute the absolute value of a BigNumber
+ * @param _x The BigNumber to compute the absolute value of
+ * @returns The absolute value of the given BigNumber
+ */
+export function abs(_x: BigNumberish): BN {
+  const x = toBN(_x);
```
nicholaspai, line 87: Can't you just `return x.abs()`?

nicholaspai (approved):
This is a really great structure. I left comments on specific function implementations but I think +1 to structure!

src/utils/common.ts
james-a-morris: Yep! Resolved. https://github.com/across-protocol/sdk-v2/pull/236/commits/4ea323da458aab0851327499ce6276e83674aa49

src/clients/UBAClient/UBAClientWithManual.ts
james-a-morris: I think we can leave this as is because the relayer-v2 code has logging baked into it

src/clients/UBAClient/UBAClientAbstract.ts
james-a-morris: Good call - updated: https://github.com/across-protocol/sdk-v2/pull/236/commits/65cb9a0ebe0e99e47ff2d71bee03fa486ab54fc1

src/UBAFeeCalculator/UBAFeeUtility.ts
james-a-morris: Good call - updated: https://github.com/across-protocol/sdk-v2/pull/236/commits/65cb9a0ebe0e99e47ff2d71bee03fa486ab54fc1

src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -182,4 +163,55 @@ export class UBAClient {
...
+  }
+
+  protected async computeRelayerFees(
+    tokenSymbol: string,
+    amount: BigNumber,
+    depositChainId: number,
+    refundChainId: number,
+    tokenPrice?: number
+  ): Promise<RelayerFeeDetails> {
+    const relayCalculator = new RelayFeeCalculator(this.relayerConfiguration);
```
pxrl, line 208: Out of curiosity, is there a reason for the RelayFeeCalculator class to be instantiated each time this method is called? I expect it can be called quite a bit by the relayer, so it seems somewhat expensive to be treating it as a disposable object.

src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -182,4 +163,55 @@ export class UBAClient {
...
+      );
+      this.spokeUBAFeeCalculators[chainId] = this.spokeUBAFeeCalculators[chainId] ?? {};
+      this.spokeUBAFeeCalculators[chainId][token] = spokeFeeCalculator;
+    }
+  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected async computeRealizedLpFee(
+    l1TokenAddress: string,
+    depositChainId: number,
+    refundChainId: number,
```
pxrl, line 185: Is it confirmed that the refundChain is relevant here? Per the UMIP, the `refund balancing fee` is handled as part of the relayerFeePct: https://github.com/UMAprotocol/UMIPs/blob/uba/UMIPs/umip-157.md#across-fees.

src/clients/UBAClient/UBAClientAbstract.ts
```diff
@@ -0,0 +1,236 @@
...
+  protected spokeUBAFeeCalculators: { [chainId: number]: { [token: string]: UBAFeeSpokeCalculator } };
+
+  protected constructor(protected readonly chainIdIndices: number[], protected readonly logger?: winston.Logger) {
+    this.spokeUBAFeeCalculators = {};
+  }
+
+  /**
+   * Retrieves the opening balance for a given token on a given chainId at a given block number
+   * @param chainId The chainId to get the opening balance for
+   * @param spokePoolToken The token to get the opening balance for
```
pxrl, line 33: It'd be helpful to clarify that this is the token address, rather than the symbol :+1:

pxrl, line 33:
```suggestion
   * @param spokePoolToken Address of the token to get the opening balance for
```

src/clients/index.ts
```diff
@@ -5,5 +5,5 @@ export {
 } from "./AcrossConfigStoreClient";
 export { HubPoolClient } from "./HubPoolClient";
 export { SpokePoolClient } from "./SpokePoolClient";
-export { UBAClient } from "./UBAClient";
+export * from "./UBAClient";
```
pxrl, line 8: FWIW I used the web interface to propose this change. I'd inteded to propose this as a PR to this branch, but github pushed it directly instead. Would be good to confirm OK @james-a-morris.

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: The default undefined block number polls the RPC with the `"latest"` block number

src/clients/UBAClient/UBAClientAbstract.ts
james-a-morris: This is a fair point. I believe this was part of the previous implementation, and I just re-arranged the structure of this function. It's possible we'd want to include this as a follow-up PR.

src/clients/UBAClient/UBAClientAbstract.ts
james-a-morris: Since this is just an abstract fn, the naming of the variables is mostly optional. However, the `_` was a carryover from my initial implementations. Thanks for catching this 🙂 https://github.com/across-protocol/sdk-v2/pull/236/commits/c881c1bb60c36b62cc9b342dc718b52e4541bd6c

src/clients/index.ts
james-a-morris: Works for me!

src/clients/UBAClient/UBAClientWithRefresh.ts
james-a-morris: Yes, we need it in the calculation of the utilization since we need both the spoke target balances for the deposit & refund chain.

src/clients/UBAClient/UBAClientAbstract.ts
```diff
@@ -0,0 +1,236 @@
...
+      feeType === UBAActionType.Deposit ? "getDepositFee" : "getRefundFee"
+    ](amount);
+    return {
+      balancingFee,
+      actionType: feeType,
+    };
+  }
+
+  protected abstract computeRealizedLpFee(
+    l1TokenAddress: string,
```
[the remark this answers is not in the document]
james-a-morris: Good call: updated here: https://github.com/across-protocol/sdk-v2/pull/236/commits/9ee97e3cdae18f8ff36f318b945ea55d29ea56d1

src/clients/UBAClient/UBAClientAbstract.ts
```diff
@@ -0,0 +1,236 @@
...
+    depositChainId: number,
+    refundChainId: number,
+    tokenPrice?: number
+  ): Promise<RelayerFeeDetails>;
+
+  /**
+   * Compute the entire Relayer fee in the context of the UBA system for a given amount. The relayer fee is the sum of the gas fee, the capital fee, and the balancing fee.
+   * @param depositChain The chainId of the deposit
+   * @param refundChain The chainId of the refund
+   * @param spokePoolToken The token to get the relayer fee for
```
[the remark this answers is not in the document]
james-a-morris: Good call: updated here: https://github.com/across-protocol/sdk-v2/pull/236/commits/9ee97e3cdae18f8ff36f318b945ea55d29ea56d1

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: oh ok

PR #237, opened 2023-06-12
feat(gas-price-oracle): Change Optimism fee model to EIP1559

nicholaspai:
Bedrock implements EIP1559 https://community.optimism.io/docs/developers/bedrock/how-is-bedrock-different/#eip-1559

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.8.1",
+  "version": "0.8.2",
   "license": "AGPL-3.0",
   "main": "dist/index.js",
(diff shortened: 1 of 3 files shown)


pxrl (approved):
Nice - I've also been keen to see this go in since Bedrock went live. Did you test it locally? I expect it should "just work", but it'd be nice to understand how much gas pricing has changed as a result of the upgrade.

nicholaspai:
> Nice - I've also been keen to see this go in since Bedrock went live. Did you test it locally? I expect it should "just work", but it'd be nice to understand how much gas pricing has changed as a result of the upgrade.

I did not but I'll test in the `relayer-v2` repo when opening a PR to use this new version

PR #240, opened 2023-06-14
feat: add parallel determination of UBA fees

james-a-morris:


pxrl:
@james-a-morris Any thoughts on making this public so that I can just compute the balancing fee directly?

https://github.com/across-protocol/sdk-v2/blob/e056dfd3c24df37e4bd2267bc93daf9da96b7643/src/clients/UBAClient/UBAClientAbstract.ts#L92

pxrl (approved):
LGTM!

src/clients/UBAClient/UBAClientAbstract.ts
```diff
@@ -89,7 +90,7 @@ export abstract class BaseUBAClient {
    * @param refundChainId The chainId of the refund
    * @returns The balancing fee for the given token on the given chainId at the given block number
    */
-  protected async computeBalancingFee(
+  public async computeBalancingFee(
```
nicholaspai, line 93: It seems wasteful to pass in both deposit and refund chain id. Why not require user to pass in a single chain, and we assume that that is the deposit chain for a deposit and a refund for a refund?

nicholaspai (approved):
Just one suggestion!

PR #241, opened 2023-06-15
feat(gasPriceOracle): Support Polygon Gas Station

pxrl:
Prompted by the migration from matic.network to polygon.technology today, add support for the Gas Station API and add a fallback strategy in case the gas station fails to resolve.


src/gasPriceOracle/adapters/polygon.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+export async function polygonGasStation(provider: Provider, chainId: number): Promise<GasPriceEstimate> {
+  const gasStation = new PolygonGasStation({ chainId: chainId });
+  try {
+    return await gasStation.getFeeData();
+  } catch (err) {
+    // Fall back to the RPC provider. May be less accurate.
+    return await eip1559(provider, chainId);
```
pxrl, line 79: I'm in two minds about this. I think it's important to have a fallback strategy, but per current implementation, it's a silent failover. If the prices are too low, we might end up with stuck transactions, which is especially bad on Polygon because Polygonscan tends not to display pending transactions.

src/gasPriceOracle/adapters/eip1559.ts
```diff
@@ -0,0 +1,20 @@
+import { BigNumber, providers } from "ethers";
+import { GasPriceEstimate, gasPriceError } from "../oracle";
+
+// @todo: Update to ethers 5.7.x to access FeeData.lastBaseFeePerGas. Use feeData.gasPrice until then.
```
pxrl, line 4: This can be easily handled in a follow-up commit.

pxrl: Addressed here: https://github.com/across-protocol/sdk-v2/pull/242

james-a-morris (approved):
LGTM 😎 Left some nits if you feel they're worth following up on.

src/gasPriceOracle/adapters/polygon.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    return { maxPriorityFeePerGas, maxFeePerGas };
+  }
+
+  protected isPolygon1559GasPrice(gasPrice: unknown): gasPrice is Polygon1559GasPrice {
+    if (!isDefined(gasPrice)) {
+      return false;
+    }
+    const _gasPrice = gasPrice as Polygon1559GasPrice;
+
```
pxrl, line 64:
```suggestion
```

src/gasPriceOracle/adapters/polygon.ts
```diff
@@ -0,0 +1,80 @@
...
+  blockTime: number | string;
+  blockNumber: number | string;
+};
+
+type GasStationArgs = BaseHTTPAdapterArgs & {
+  chainId?: number;
+  host?: string;
+};
+
+type Provider = providers.Provider;
```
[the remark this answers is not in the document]
pxrl: Yeah - it is annoying to do it in multiple places. I don't really have bandwidth for it now so I've just removed the type, but we should look to consolidate some things like this once we hit the UBA milestone.

PR #244, opened 2023-06-16
feat: add update orchestration

james-a-morris:


src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -30,7 +30,33 @@ export class UBAClientWithRefresh extends BaseUBAClient {
...
 
+  public async update(): Promise<void> {
+    // Update the Across config store
+    await this.hubPoolClient.configStoreClient.update();
+    // Update the HubPool
+    await this.hubPoolClient.update();
+    // Update the SpokePools
+    await Promise.all(Object.values(this.spokePoolClients).map(async (spokePoolClient) => spokePoolClient.update()));
+    // Update the UBAFeeCalculators
+    await Promise.all(
```
nicholaspai, line 41: We might want to add some logs here for the future when running in production so we can easily see if any of these clients are taking a long time to update and point to some easy optimizations

src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -30,7 +30,33 @@ export class UBAClientWithRefresh extends BaseUBAClient {
...
+  protected assertNecessaryClientsUpdated(): void {
+    const isUpdated =
+      this.hubPoolClient.configStoreClient.isUpdated &&
+      this.hubPoolClient.isUpdated &&
+      Object.values(this.spokePoolClients).every((spokePoolClient) => spokePoolClient.isUpdated);
+    assert(isUpdated, "UBAClientWithRefresh: Clients not updated");
+  }
+
   protected resolveClosingBlockNumber(chainId: number, blockNumber: number): number {
+    this.assertNecessaryClientsUpdated();
```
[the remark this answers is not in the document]
james-a-morris: It's possible via function decorator - it's the perfect use-case. However, we're restricted by TSDX 😢

src/clients/UBAClient/UBAClientWithRefresh.ts
james-a-morris: I was hoping to make this a second pass when we're more cemented on the components. Should we add the logging now? Totally happy with it either way

PR #246, opened 2023-06-17
fix(gasPriceOracle): Assorted fixes

pxrl:
I found several issues in the gasPriceOracle:
- Polygon gas station lookups are failing because the request resolves to <domain>//v2 instead of <domain>/v2. Polygon don't automatically squash the additional / characters, and instead respond with a 404. This hasn't impacted service because the adapter falls back to an on- chain price lookup, but it delays gas price resolution and probably gives less accurate results.

- After resolving the gas station issue, the oracle.e2e test needed fixing because the Polygon adapter no longer returns mocked gas prices.

- As part of the updates to the oracle.e2e test, I realised that the use of assert.ok() is node-specific and seems irrelevant to jest. So, use expect() instead. This makes the tests a bit more compact and easier to read.

Additionally:
- An Arbitrum-specific adapter has been added that effectively nulifieies the priority fee. It's set to 1 Wei because a priority fee of 0 Wei is used to signal a legacy (type 0) transaction.

- As part of this change I have also updated the RPC list used when local RPCs are not defined. These are the current high-performing publicly availably RPCs, as reported by chainlist.org.

### DIFF
--- a/src/gasPriceOracle/adapters/arbitrum.ts
+++ b/src/gasPriceOracle/adapters/arbitrum.ts
@@ -1,0 +1,21 @@
+import { BigNumber, providers, utils as ethersUtils } from "ethers";
+import { GasPriceEstimate } from "../types";
+import { eip1559 } from "./ethereum";
+
+// Arbitrum Nitro implements EIP-1559 pricing, but the priority fee is always refunded to the caller. Further,
+// ethers typically hardcodes the priority fee to 1.5 Gwei. So, confirm that the priority fee supplied was 1.5
+// Gwei, and then drop it to 1 Wei. Reference: https://developer.arbitrum.io/faqs/gas-faqs#q-priority
+export async function eip1559_arbitrum(provider: providers.Provider, chainId: number): Promise<GasPriceEstimate> {
+  const { maxFeePerGas: _maxFeePerGas, maxPriorityFeePerGas } = await eip1559(provider, chainId);
+
+  // If this throws, ethers default behaviour has changed, or Arbitrum RPCs are returning something more sensible.
+  if (!maxPriorityFeePerGas.eq(ethersUtils.parseUnits("1.5", 9))) {
+    throw new Error(`Expected hardcoded 1.5 Gwei priority fee on Arbitrum, got ${maxPriorityFeePerGas}`);
+  }
+
+  // eip1559() sets maxFeePerGas = lastBaseFeePerGas + maxPriorityFeePerGas, so revert that.
+  // The caller may apply scaling as they wish afterwards.
+  const maxFeePerGas = _maxFeePerGas.sub(maxPriorityFeePerGas).add(1);
+
+  return { maxPriorityFeePerGas: BigNumber.from(1), maxFeePerGas };
+}
--- a/src/gasPriceOracle/adapters/ethereum.ts
+++ b/src/gasPriceOracle/adapters/ethereum.ts
@@ -3,7 +3,7 @@ import { GasPriceEstimate } from "../types";
 import { gasPriceError } from "../util";
 
 export async function eip1559(provider: providers.Provider, chainId: number): Promise<GasPriceEstimate> {
-  const feeData: providers.FeeData = await provider.getFeeData();
+  const feeData = await provider.getFeeData();
 
   [feeData.lastBaseFeePerGas, feeData.maxPriorityFeePerGas].forEach((field: BigNumber | null) => {
     if (!BigNumber.isBigNumber(field) || field.lt(0)) gasPriceError("getFeeData()", chainId, feeData);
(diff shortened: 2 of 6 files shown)


src/gasPriceOracle/adapters/arbitrum.ts
pxrl, line 1: I'm OK with adding a special Arbitrum adapter to nullify the priority fee since Arbitrum themselves describe this as Nitro-specific behaviour. I've added a safeguard to alert us to any changes in the priority fee, so we should be aware of any changes very quickly.

src/gasPriceOracle/adapters/ethereum.ts
pxrl, line 1: Only cosmetic changes applied here.

src/gasPriceOracle/adapters/polygon.ts
```diff
@@ -29,15 +29,15 @@ type GasStationArgs = BaseHTTPAdapterArgs & {
 class PolygonGasStation extends BaseHTTPAdapter {
   readonly chainId: number;
 
-  constructor({ chainId = 137, host, timeout = 3000, retries = 2 }: GasStationArgs = {}) {
+  constructor({ chainId = 137, host, timeout = 1500, retries = 1 }: GasStationArgs = {}) {
```
pxrl, line 32: Dropped the timeouts; 3 seconds is a long time - especially on a total of 3 attempts!

src/gasPriceOracle/oracle.e2e.ts
```diff
@@ -12,7 +11,7 @@ const dummyLogger = winston.createLogger({
 
 type FeeData = providers.FeeData;
 
-class MockedProvider extends providers.JsonRpcProvider {
+class MockedProvider extends providers.StaticJsonRpcProvider {
```
pxrl, line 14: Modified to suppress submitting chainId lookups to the RPC provider.

src/gasPriceOracle/adapters/arbitrum.ts
```diff
@@ -0,0 +1,21 @@
+import { BigNumber, providers, utils as ethersUtils } from "ethers";
+import { GasPriceEstimate } from "../types";
+import { eip1559 } from "./ethereum";
+
+// Arbitrum Nitro implements EIP-1559 pricing, but the priority fee is always refunded to the caller. Further,
+// ethers typically hardcodes the priority fee to 1.5 Gwei. So, confirm that the priority fee supplied was 1.5
+// Gwei, and then drop it to 1 Wei. Reference: https://developer.arbitrum.io/faqs/gas-faqs#q-priority
+export async function eip1559_arbitrum(provider: providers.Provider, chainId: number): Promise<GasPriceEstimate> {
```
nicholaspai, line 8: Have you tested that 1 wei proirity fee works?  This is a really strange gas fee structure...

nicholaspai (approved):
Love these changes, just a bit uncertain about whether arbitrum eip1559 will work but as long as you've tested locally lgtm

src/gasPriceOracle/oracle.e2e.ts
```diff
@@ -12,7 +12,7 @@ const dummyLogger = winston.createLogger({
...
```
james-a-morris, line 15: Will the `chainId`/`network` ever change? IIRC the `StaticJsonRpcProvider` doesn't expect any changing networks

james-a-morris (commented):
Left a comment

src/gasPriceOracle/adapters/arbitrum.ts
pxrl: Just circling back to this one - here's a sample txn where I used this updated gasPriceOracle in the relayer to fetch the gas price: https://arbiscan.io/tx/0x50fe76875656fc58526861e9f20a78a8f95876f6706ea02c69fd4a26b9c79194

The fees supplied by the gasPriceOracle were 0.1 Gwei and 1 Wei respectively, and the maxFeePerGas in the fill transaction was 0.36 Gwei (after scaling in the relayer-v2 getGasPrice() function). Ultimately, a price of 0.1 Gwei was charged, and the rest was refunded to the relayer. So the on-chain behaviour seems consistent with the description in the docs :+1:

src/gasPriceOracle/oracle.e2e.ts
pxrl: Nope, there's a 1:1 relationship between a Provider object and a remote RPC provider, and they never change in the bots. Note however that this is in the e2e test file, so it's not even part of the production code. I switched to StaticJsonRpcProvider to reduce the chance of rate-limiting (requesting chainId can be very spammy), because sometimes these tests are run against public RPC endpoints.

pxrl:
Tested OK w/ Polygon here: https://polygonscan.com/tx/0x5fff2a70acb5356aebade6ac7be9c271944e240e20c8e4def6a23ea06044491a

PR #249, opened 2023-06-21
fix: browser compatibility

dohaki:
Fixes browser compatibility of sdk by:
- bumping contracts-v2 with compatible typechain bindings
- removing browser incompatible `@uma/financial-templates-lib` dep
- using tree-shakeable imports for `@across-protocol/contracts-v2`


src/utils/common.test.ts
```diff
@@ -30,6 +30,8 @@ describe("Utils test", () => {
   });
 
   it("apply gas multiplier", async () => {
+    jest.setTimeout(50_000);
+
```
dohaki, line 34: Not 100% sure why I had to change this here... When I ran the tests it errored with
```
 FAIL  src/utils/common.test.ts
  ● Test suite failed to run

    src/utils/common.test.ts:64:6 - error TS2554: Expected 1-2 arguments, but got 3.

    64   }, 50000);
            ~~~~~
```
Weird that it seems to correctly run previously

pxrl: Hm that's odd. One side-effect of this - jset.setTimeout() [affects all subsequent tests in the file](https://jestjs.io/docs/jest-object#jestsettimeouttimeout).

yarn.lock
pxrl, line 1: Cuts almost 2k lines in yarn.lock - that's nice!

src/utils/common.test.ts
pxrl: I reverted this change and tested it locally, and it works for me:

```
riskybiz$ git diff origin/fix-browser-compatibility
diff --git a/src/utils/common.test.ts b/src/utils/common.test.ts
index 229554f..8bbadbb 100644
--- a/src/utils/common.test.ts
+++ b/src/utils/common.test.ts
@@ -30,8 +30,6 @@ describe("Utils test", () => {
   });
 
   it("apply gas multiplier", async () => {
-    jest.setTimeout(50_000);
-
     const spokePoolAddress = "0xB88690461dDbaB6f04Dfad7df66B7725942FEb9C"; // mainnet
     const usdcAddress = "0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8";
     const relayerAddress = "0x893d0d70ad97717052e3aa8903d9615804167759";
@@ -63,5 +61,5 @@ describe("Utils test", () => {
       const gasMultiplier = toBNWei(1.0 + gasMarkup);
       expect(toBN(gasEstimate).eq(toBN(refGasEstimate).mul(gasMultiplier).div(toBNWei(1)))).toBe(true);
     }
-  });
+  }, 50000);
 });
riskybiz$ yarn test src/utils/common.test.ts       
yarn run v1.22.19
$ tsdx test src/utils/common.test.ts
(node:468268) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at /home/paul/dev/across-protocol/sdk-v2/node_modules/tsdx/node_modules/tslib/package.json.
Update this package.json to use a subpath pattern like "./*".
(Use `node --trace-deprecation ...` to show where the warning was created)
(node:468268) V8: /home/paul/dev/across-protocol/sdk-v2/node_modules/rustbn.js/lib/index.asm.js:2 Linking failure in asm.js: Unexpected stdlib member
 PASS  src/utils/common.test.ts (11.877s)
  Utils test
    ✓ retry (2006ms)
    ✓ apply gas multiplier (4316ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        11.895s, estimated 17s
Ran all test suites matching /src\/utils\/common.test.ts/i.
Done in 12.87s.
```

dohaki: Hmm ok thanks for trying that out. I will try with a cleaned state again

dohaki: Yea you were right https://github.com/across-protocol/sdk-v2/actions/runs/5332713044/jobs/9662260149. Seems to be a local env issue on my side... Thanks for flagging!

pxrl (approved):
This tested fine for me in relayer-v2 via unit test, relayer and validateRootBundle script.

PR #251, opened 2023-06-21
fix(UBAClient): Override refresh computeSystemFee

pxrl:
The base computeSystemFee() method receives a SpokePool token address and passes it in to computeLpFee(), which seems to expect a HubPool token address. Since the abstract base class doesn't have any reference to the HubPoolClient, it's not possible to handle the translation directly. Therefore, override it for the time being.

Note that this doesn't completely address the issue - it seems like there will be an issue in the base class in any case. We might need to consider removing that base implementation.


james-a-morris (commented):
@pxrl do we still need this with the new change to the SDK?

pxrl:
> @pxrl do we still need this with the new change to the SDK?

Nope - your change addresses it :+1:

PR #255, opened 2023-06-22
style(UBA): modify uba fee interface

james-a-morris:
This change allows for additional information to be passed to the UBA fee calculations which are needed. The `spokePoolToken` variable has been changed to include the more reasonable collection of variables `tokenSymbol` and `hubPoolTokenAddress` which represent the symbol of the token & its address on the hubpool chain.

Breaking Changes: This will change the interface


pxrl (approved):
LGTM

PR #257, opened 2023-06-26
test: add test configuration to UBA Client

james-a-morris:
This series of tests aims to verify the code correctness of the various UBA-specific calculations. Checked against data provided from @cc7768

### DIFF
--- a/src/UBAFeeCalculator/UBAFeeCalculator.test.ts
+++ b/src/UBAFeeCalculator/UBAFeeCalculator.test.ts
@@ -63,46 +63,46 @@ describe("UBA Fee Calculations", () => {
 
   it("should integrate the correct value: test #1", () => {
     const result = performLinearIntegration(tuples, 0, toBN(0), toBN(100_000));
-    expect(result.div(parseEther("1")).toString()).to.eq("-40000");
+    expect(result.toString()).to.eq("-40000");
   });
 
   it("should integrate the correct value: test #2", () => {
     const result = performLinearIntegration(tuples, 0, toBN(100_000), toBN(0));
-    expect(result.div(parseEther("1")).toString()).to.eq("40000");
+    expect(result.toString()).to.eq("40000");
   });
 
   it("should integrate the correct value: test #3", () => {
     const result = performLinearIntegration(tuples, 1, toBN(100_000), toBN(250_000));
-    expect(result.div(parseEther("1")).toString()).to.eq("-48750");
+    expect(result.toString()).to.eq("-48750");
   });
 
   it("should integrate the correct value: test #4", () => {
     const result = performLinearIntegration(tuples, 1, toBN(250_000), toBN(100_000));
-    expect(result.div(parseEther("1")).toString()).to.eq("48750");
+    expect(result.toString()).to.eq("48750");
   });
 
   it("should compute the correct deposit fee #1", () => {
     const result = getDepositBalancingFee(tuples, toBN(300_000), toBN(50_000));
-    expect(result.div(parseEther("1")).toString()).to.eq("-10250");
+    expect(result.toString()).to.eq("-10250");
   });
 
   it("should compute the correct deposit fee #2", () => {
     const result = getDepositBalancingFee(tuples, toBN(300_000), toBN(100_000));
-    expect(result.div(parseEther("1")).toString()).to.eq("-19000");
+    expect(result.toString()).to.eq("-19000");
   });
 
   it("should compute the correct refund fee #1", () => {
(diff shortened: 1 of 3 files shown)


src/UBAFeeCalculator/UBAFeeCalculator.test.ts
```diff
@@ -56,46 +58,117 @@ describe("UBA Fee Calculations", () => {
...
+    expect(result.toString()).to.eq("25500");
+  });
+});
+
+describe("UBA Fee Calculations from Data", () => {
+  let gammaCutoffArray: FlowTupleParameters;
+  let omegaCutoffArray: FlowTupleParameters;
+
+  beforeEach(() => {
+    gammaCutoffArray = [
```
nicholaspai, line 110: Why set these arrays in the `beforeEach` and why not in a single `before()`?

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -19,14 +19,17 @@ export function performLinearIntegration(
   integralStart: BigNumber,
   integralEnd: BigNumber
 ): BigNumber {
+  const scaler = parseEther("1.0");
   const lengthUnderCurve = integralEnd.sub(integralStart);
   const resolveValue = (index: number): BigNumber => cutoffArray[index][1];
-  let feeIntegral = resolveValue(Math.min(index, cutoffArray.length - 1)).mul(lengthUnderCurve);
+  let feeIntegral = resolveValue(index === 0 ? 0 : index === cutoffArray.length ? cutoffArray.length - 1 : index - 1)
```
nicholaspai, line 25: This is a bit hard to read, can you add a comment or refactor with more intermediate variables to understand how to change this in the future?

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -72,7 +78,7 @@ export function getInterval(
 ): [number, [BigNumber, BigNumber]] {
   let result: [number, [BigNumber, BigNumber]] = [
     -1,
-    [BigNumber.from(-MAX_SAFE_JS_INT), BigNumber.from(MAX_SAFE_JS_INT)],
+    [BigNumber.from(-MAX_SAFE_JS_INT).mul(parseEther("1.0")), BigNumber.from(MAX_SAFE_JS_INT).mul(parseEther("1.0"))],
```
nicholaspai, line 81: nit: do we store `parseEther("1.0")` anywhere as a constant? Feel like its a very commonly used value. But i can't recall if its in `relayer-v2`/`sdk-v2`/`contracts-v2`. I feel like @pxrl has put this somewhere nice

src/UBAFeeCalculator/UBAFeeUtility.ts
james-a-morris: Good call - I've done both (simplified & added documentation). In the end, this could be reduced to a simple

`MAX(0, MIN(INDEX - 1, ARR_LENGTH))`

Updated: https://github.com/across-protocol/sdk-v2/pull/257/commits/f606f38fdfdff158c6a1aeb98f1ed14d11b725b8

src/UBAFeeCalculator/UBAFeeCalculator.test.ts
james-a-morris: Good call - I've updated to a single `beforeAll` instead of the `beforeEach`.

src/UBAFeeCalculator/UBAFeeUtility.ts
james-a-morris: It's defined in the sdk - good call.

nicholaspai (approved):

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -51,12 +62,14 @@ export function performLinearIntegration(
  * @returns The upper and lower bounds of the interval
  */
 export function getBounds(cutoffArray: [BigNumber, BigNumber][], index: number): [BigNumber, BigNumber] {
+  const largestBound = BigNumber.from(parseUnits((-MAX_SAFE_JS_INT).toString(), 18)).mul(-1);
```
pxrl, line 65: I'm curious about the motivation for instantiating a large negative number and then multiplying it by -1. I tested with both and it seems to produce the same result, and the MAX_SAFE_JS_INT already seems to be 2^32 - 1, so -MAX_SAFE_JS_INT is probably not the same as "MIN_SAFE_JS_INT" (if there were one).

Worth a comment to clarify?

james-a-morris: Updated here: #259

PR #266, opened 2023-06-28
fix: refactor UBAClient to achieve similar use as other clients

james-a-morris:
This PR aims to upgrade the UBAClient such that it works in a similar way/fashion to the other clients in the SDK. Primarily, this entails the UBAClient having a single asynchronous update function and all logic related to computations being synchronous.


src/clients/UBAClient/UBAClientTypes.ts
```diff
@@ -0,0 +1,41 @@
...
+  openingBalance: BigNumber;
+  blockNumber: number;
+  config: {
+    ubaConfig: UBAConfig;
+  };
+  flows: {
+    flow: UbaFlow;
+    systemFee: SystemFeeResult;
+    relayerFee: RelayerFeeResult;
+    runningBalance: BigNumber;
```
nicholaspai, line 39: nit: can you add `netRunningBalanceAdjustment` here too? We'll need this to construct `netSendAmounts` in the Dataworker and you should take into account when computing this `runningBalance` whether we need to make an adjustment due to thresholds

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -77,3 +83,296 @@ export async function getUBAFeeConfig(
...
+    .filter(
+      (deposit: DepositWithBlock) =>
+        deposit.blockNumber >= (fromBlock as number) && deposit.blockNumber <= (toBlock as number)
+    );
+
+  // Filter out:
+  // - Fills that request refunds on a different chain.
+  // - Subsequent fills after an initial partial fill.
+  // - Slow fills.
+  const fills: UbaFlow[] = spokePoolClient.getFills().filter((fill: FillWithBlock) => {
```
nicholaspai, line 287: nit: will need to validate these fills used the correct `realizedLpFeePct` and also match perfectly with a deposit. See [here](https://github.com/across-protocol/sdk-v2/blob/master/src/utils/FillUtils.ts#L30) how we currently validate deposits. This assumes that the `SpokePoolClient` has mapped a `realizedLpFeePct` to each `DepositWithBlock` in its `update` call. This `realizedLpFeePct` obviously needs to be computed differently in the UBA but the same idea applies for validating a fill: it must match a deposit and have set the expected `realizedLpFeePct`.

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -77,3 +83,296 @@ export async function getUBAFeeConfig(
...
+        && fill.amount.eq(amount)
+        && fill.realizedLpFeePct.eq(realizedLpFeePct)
+        && fill.blockNumber === fillBlock.toNumber()
+      );
+  });
+  if (!isDefined(fill)) {
+    return { valid: false, reason: "Unable to find matching fill" };
+  }
+
+  const deposit = originSpoke.getDepositForFill(fill);
```
nicholaspai, line 362: See [here](https://github.com/across-protocol/sdk-v2/pull/266/files#r1246726140) for note on validating the fill

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -77,3 +83,296 @@ export async function getUBAFeeConfig(
...
+  const balances = hubPoolClient.getRunningBalanceBeforeBlockForChain(hubPoolBlockNumber, chainId, hubPoolToken);
+  const endBlock = hubPoolClient.getLatestBundleEndBlockForChain([chainId], hubPoolBlockNumber, chainId);
+  return {
+    blockNumber: endBlock,
+    spokePoolBalance: balances.runningBalance,
+    incentiveBalance: balances.incentiveBalance,
+  };
+}
+
+function getFlows(
```
nicholaspai, line 261: I think this function will be worth unit testing in depth with mocked data

src/clients/UBAClient/UBAClientTypes.ts
james-a-morris: Added

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -0,0 +1,346 @@
...
+    tokenSymbol: string,
+    amount: BigNumber,
+    balancingActionBlockNumber: number,
+    chainId: number,
+    feeType: UBAActionType
+  ): BalancingFeeReturnType {
+    // Opening balance for the balancing action blockNumber.
+    const relevantBundleStates = this.retrieveBundleStates(chainId, tokenSymbol);
+    const specificBundleState = relevantBundleStates.findLast(
+      (bundleState) => bundleState.blockNumber <= balancingActionBlockNumber
```
nicholaspai, line 162: I keep getting a bit confused about which `chainId` the `bundleState.blockNumber` is referring to: is it the Ethereum block at which the proposal happened, or the `bundleEndBlock` for the `chainId` that was proposed in the bundle? I assume the latter, so maybe can we rename the param to `endBlockForChain`?

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -0,0 +1,346 @@
...
+    );
+    if (!specificBundleState) {
+      throw new Error(`No bundle states found for token ${tokenSymbol} on chain ${chainId}`);
+    }
+    /** @TODO ADD TX INDEX COMPARISON */
+    const flows = (specificBundleState?.flows ?? []).filter(
+      (flow) => flow.flow.blockNumber <= balancingActionBlockNumber
+    );
+    const calculator = new UBAFeeSpokeCalculator(
```
nicholaspai, line 171: Just a suggestion but isn't it kind of strange to instantiate a new calculator each time we call `computeBalancingFee`? Should we instead treat the calculator as a library with a bunch of static functions and make calls like:

```ts
const { balancingFee } = UBAFeeSpokeCalculator[feeType === UBAActionType.Deposit ? "getDepositFee" : "getRefundFee"](
    chainId, 
    tokenSymbol, 
    flows.map(({ flow }) => flow), 
    specificBundleState.openingBalance,
    specificBundleState.openingIncentiveBalance,
    specificBundleState.config.ubaConfig,
    amount
)
```

src/clients/UBAClient/UBAClientTypes.ts
```diff
@@ -0,0 +1,65 @@
...
+  };
+  bundles: {
+    [tokenSymbol: string]: UBABundleState[];
+  };
+};
+
+export type UBABundleState = {
+  openingBalance: BigNumber;
+  openingIncentiveBalance: BigNumber;
+  blockNumber: number;
```
nicholaspai, line 32: Should this be renamed `bundleEndBlockNumber`? https://github.com/across-protocol/sdk-v2/pull/266/files#r1247925856

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -77,3 +127,336 @@ export async function getUBAFeeConfig(
...
+        ),
+        latestBlockNumber: spokePoolClient.latestBlockNumber,
+      },
+    };
+
+    for (const tokenSymbol of relevantTokenSymbols) {
+      // Get the bundles for this token
+      const availableBundles = chainState.bundles[tokenSymbol] ?? [];
+      // Get the block number and opening balance for this token
+      const { blockNumber, spokePoolBalance, incentiveBalance } = getOpeningTokenBalances(
```
nicholaspai, line 171: This `blockNumber` is the `bundleEndBlock` for the `chainId` right? I think it might be helpful to rename so as not to confuse yourself about which chain you're comparing blocks on

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -77,3 +127,336 @@ export async function getUBAFeeConfig(
...
+        availableBundles[referenceBundleIndex] = constructedBundle;
+      }
+      // Otherwise, add it to the list of bundles
+      else {
+        availableBundles.push(constructedBundle);
+      }
+      // Find the last flow avaialble in the bundle
+      const lastFlow = constructedBundle.flows[constructedBundle.flows.length - 1];
+      // If the last flow exists, call getFlows with the last flow's block number as the fromBlock
+      const recentFlows = getFlows(
```
nicholaspai, line 219: Wouldn't you want to load all flows from the beginning of the bundle to the end, so wouldn't you want to set the `fromBlock` equal to the blockNumber of the first flow?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -77,3 +127,336 @@ export async function getUBAFeeConfig(
...
+      // Instantiate a UBAFeeSpokeCalculator
+      const calculator = new UBAFeeSpokeCalculator(
+        chainId,
+        tokenSymbol,
+        [...constructedBundle.flows.map(({ flow }) => flow), ...recentFlows],
+        spokePoolBalance,
+        incentiveBalance,
+        constructedBundle.config.ubaConfig
+      );
+      for (const flow of recentFlows) {
```
nicholaspai, line 235: nit: I assume `recentFlows` is returned in ascending order, right? Just want to make sure you've seen this helpful functions to deterministically sort events on the same chain: https://github.com/across-protocol/relayer-v2/blob/master/src/utils/EventUtils.ts#L181

nicholaspai (commented):
This is really really good work I think we're making great progress. I mainly looked at `updateUBAClient` and `computeBalancingFee` and I think they're pretty much spot on! My biggest question marks are regarding [variable naming](https://github.com/across-protocol/sdk-v2/pull/266/files#r1247931534) and [how the dataworker will use `updateUBAClient`](https://github.com/across-protocol/sdk-v2/pull/266/files#r1247939648)

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: Yep! The end of `getFlows()` returns the flows in ascending order.

james-a-morris: Added docs to make this more explicit: https://github.com/across-protocol/sdk-v2/pull/266/commits/a03635d2b538c3ebc003b2e240857b44d4d17b0f

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: Good point - answered below

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: This new PR regenerates all the bundles on update. https://github.com/across-protocol/sdk-v2/pull/266/commits/617c110861852a75cfe6c538ac56c5e4e0901f03

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -77,3 +127,336 @@ export async function getUBAFeeConfig(
     }
   );
 }
+
+export async function updateUBAClient(
```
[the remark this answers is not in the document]
james-a-morris: Yep! I've reworked it so that the update now regenerates up to the `MAX_BUNDLE_LOOKBACK_LENGTH` that's currently stored as a constant. (We could always make this an ENV var or pass it as an optional param).

src/clients/UBAClient/UBAClientTypes.ts
james-a-morris: This is more apt as `openingBlockNumberForSpokeChain` because it denotes the `start` of the current bundle (since the last verified block bundle): https://github.com/across-protocol/sdk-v2/pull/266/commits/5334cac2b391dc7f07f65c48c186bb67fe1fcdc9

src/clients/UBAClient/UBAClientBase.ts
james-a-morris: I think this is contained in: https://github.com/across-protocol/sdk-v2/pull/266#discussion_r1248156998

src/clients/UBAClient/UBAClientBase.ts
james-a-morris: Sounds good - I've created analog versions of all the UBA Spoke Fee calculators so you can run it without instantiating a class: https://github.com/across-protocol/sdk-v2/pull/266/commits/6cd922196f94c49119642da9940dc501814c0ece

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts
```diff
@@ -0,0 +1,202 @@
...
+  // This is the case when the UBA Fee Calculator is first initialized or run on a range
+  // that we haven't computed the running balance for yet
+  const historicalResult: TokenRunningBalanceWithNetSend = flows.reduce(
+    (acc, flow) => {
+      // If the flow is an inflow, we need to add the amount to the running balance
+      // If the flow is an outflow, we need to subtract the amount from the running balance
+      // This is reflected in the incentive balance as well
+      const resultant: TokenRunningBalanceWithNetSend = {
+        netRunningBalanceAdjustment: toBN(acc.netRunningBalanceAdjustment.toString()), // Deep copy via string conversion
+        runningBalance: acc.runningBalance[isUbaInflow(flow) ? "add" : "sub"](flow.amount),
```
nicholaspai, line 43: Where should we compute the incentive fee and subtract it from this line and add it in next line?

src/clients/UBAClient/UBAClientBase.ts
nicholaspai: I don't see us using the analog version in This function. Shouldn't we?

nicholaspai (approved):
This looks good thanks for addressing my comments. It looks like in the update function you're traversing the bundles backwards which is interesting. I guess there is no extra benefit to loading them forwards or backwards since each bundle state is kind of independent. 

I want to resurface this comment about validating fills https://github.com/across-protocol/sdk-v2/pull/266#discussion_r1246726872

Otherwise this LGTM and I look foreword to reviewing the unit tests as that will be the real test !!

james-a-morris:
@nicholaspai Resolved in the commit above: https://github.com/across-protocol/sdk-v2/pull/266#discussion_r1248207001

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts
james-a-morris: We are now accounting for the incentive fee.

Commit: https://github.com/across-protocol/sdk-v2/pull/266/commits/340e5d554ccdd39c0b0a4abd29c0466f0e36c864

james-a-morris:
@pxrl @nicholaspai This commit now fully parallelizes the update function. I was able to do this because the UBA client effectively works on independent bundles/chains/tokens.

Commit: https://github.com/across-protocol/sdk-v2/pull/266/commits/fd2d16be7d6c3e5909221e97ccae0897dbc4fa7b

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: Sounds good! Updated

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts
```diff
@@ -0,0 +1,229 @@
...
+  lastValidatedRunningBalance: BigNumber,
+  lastValidatedIncentiveRunningBalance: BigNumber,
+  chainId: number,
+  tokenSymbol: string,
+  config: UBAConfig
+): TokenRunningBalanceWithNetSend {
+  // Attempt to resolve the trigger hurdle to include in the running
+  // balance calculation
+  const { upperBound: upperBoundTriggerHurdle, lowerBound: lowerBoundTriggerHurdle } =
+    config.getBalanceTriggerThreshold(chainId, tokenSymbol);
```
nicholaspai, line 32: Wouldn't you need to pass in `blockNumber` into this call to figure out which trigger threshold to pull from the config store? The threshold can be updated between flows so you want the correct one

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts
```diff
@@ -0,0 +1,229 @@
...
+    previousFlows,
+    lastValidatedRunningBalance,
+    lastValidatedIncentiveRunningBalance,
+    chainId,
+    tokenSymbol,
+    config
+  );
+
+  // We first need to resolve the inflow/outflow curves for the deposit and refund spoke
+  const flowCurve = config.getBalancingFeeTuples(chainId);
```
nicholaspai, line 138: Same as this: https://github.com/across-protocol/sdk-v2/pull/266/files#r1253650456

nicholaspai (approved):
This looks amazing overall, huge feature!

Sending some comments out that might cause you to think about other changes but these stood out to me
- https://github.com/across-protocol/sdk-v2/pull/266/files#r1253655193
- https://github.com/across-protocol/sdk-v2/pull/266#discussion_r1253650456

src/utils/TypeUtils.ts
```diff
@@ -1,3 +1,7 @@
+import { SpokePoolClient } from "../clients";
+
 export type MakeOptional<Type, Key extends keyof Type> = Omit<Type, Key> & Partial<Pick<Type, Key>>;
 
 export type AnyObject = Record<string, unknown>;
+
+export type SpokePoolClients = Record<string, SpokePoolClient>;
```
pxrl, line 7: This is normally specified as `{ [chainId: number]: SpokePoolClient }`, but afaik object keys are coaxed into being strings anyway. Does it make a difference whether it's `Record<string, ...>` or `Record<number, ...>` ?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts
james-a-morris: That's fair - my understanding with this function is that it this function computes the running balances of a current bundle. For efficiency, the assumption is made that the config does not change during intra-bundle calculations - I believe we discussed this and my understanding is that this is an acceptable assumption.

Because of that, the `UBAConfig` is fixed for the calculations of each flow.

src/utils/TypeUtils.ts
james-a-morris: I totally agree. This is something that I keep seeing and is why I decided to just go ahead and collapse it. Let me run through the code and verify that I've replaced all the references to spokeChainClients.

As a general rule, it's not that painful to switch between each, but it will require a loop so we should be uniform. Should we do `Record<number, SpokePoolClient>` everywhere?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts
james-a-morris: Resolved offline

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts
james-a-morris: Resolved offline 👀

src/utils/TypeUtils.ts
james-a-morris: Changed all of the outstanding `{ [chainId: string | number]: SpokePoolClient }` into `SpokePoolClients`

PR #267, opened 2023-07-04
docs(UBA): :memo: include JSDocs for spoke client

james-a-morris:
This change includes valid JSDoc comments for all the functions within the SpokePool Client.

As the relayer clients are now within the SDK, it may be effective to provide proper doc comments at all levels for developers to use / better understand the gist of the code.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -42,6 +42,10 @@ type _SpokePoolUpdate = {
 };
 export type SpokePoolUpdate = { success: false } | _SpokePoolUpdate;
 
+/**
+ * SpokePoolClient is a client for the SpokePool contract. It is responsible for querying the SpokePool contract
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -663,7 +827,21 @@ export class SpokePoolClient {
...
     this.logger[level]({ at: "SpokePoolClient", chainId: this.chainId, message, ...data });
   }
+
+  /**
+   * Retrieves the current time from the SpokePool contract.
+   * @returns The current time.
+   */
+  public getCurrentTime(): number {
+    return this.currentTime;
+  }
```
james-a-morris, line 846: @pxrl I've included this for convenience. This satisfies the comment left in: https://github.com/across-protocol/relayer-v2/pull/787/files

nicholaspai (approved):

PR #269, opened 2023-07-05
feat: include dataworker client utils

james-a-morris:


src/clients/bridges/BundleDataClient.ts
```diff
@@ -0,0 +1,363 @@
...
+  // - Bundles that passed liveness but have not had all of their pool rebalance leaves executed.
+  // - Bundles that are pending liveness
+  // - Not yet proposed bundles
+  async getNextBundleRefunds(): Promise<FillsToRefund> {
+    const futureBundleEvaluationBlockRanges = getWidestPossibleExpectedBlockRange(
+      this.chainIdListForBundleEvaluationBlockNumbers,
+      this.spokePoolClients,
+      getEndBlockBuffers(this.chainIdListForBundleEvaluationBlockNumbers, this.blockRangeEndBlockBuffer),
+      this.clients,
+      this.clients.hubPoolClient.latestBlockNumber ?? 0,
```
james-a-morris, line 100: Included this to escape the fact that `latestBlockNumber` can potentially be undefined. I feel comfortable adding `?? 0` since we don't do any checks currently.

PR #270, opened 2023-07-06
improve: Return a copy of RefundRequestWithBlock objects

pxrl:
Removes a potential sharp edge if the recipient of this array happens to call Array.pop().

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -198,9 +198,10 @@ export class SpokePoolClient {
    * @returns A list of refund requests.
    */
   public getRefundRequests(fromBlock?: number, toBlock?: number): RefundRequestWithBlock[] {
-    return fromBlock === undefined || toBlock === undefined || isNaN(fromBlock) || isNaN(toBlock)
+    const refundRequests = fromBlock === undefined || toBlock === undefined || isNaN(fromBlock) || isNaN(toBlock)
       ? this.refundRequests
       : this.refundRequests.filter((request) => request.blockNumber >= fromBlock && request.blockNumber <= toBlock);
+    return [ ...refundRequests ];
   }
 
   /**


pxrl (commented):
IMO we probably want to apply this to all of the `get...()` methods that return event data.

src/clients/SpokePoolClient.ts
```diff
@@ -198,9 +198,10 @@ export class SpokePoolClient {
    * @returns A list of refund requests.
    */
   public getRefundRequests(fromBlock?: number, toBlock?: number): RefundRequestWithBlock[] {
-    return fromBlock === undefined || toBlock === undefined || isNaN(fromBlock) || isNaN(toBlock)
+    const refundRequests = fromBlock === undefined || toBlock === undefined || isNaN(fromBlock) || isNaN(toBlock)
       ? this.refundRequests
       : this.refundRequests.filter((request) => request.blockNumber >= fromBlock && request.blockNumber <= toBlock);
+    return [ ...refundRequests ];
```
james-a-morris, line 204: If we're looking to deep-copy a variable, should we use `deepCopy`?
```suggestion
    return deepCopy(refundRequests);
```

james-a-morris, line 204: Formatting per our linter file
```suggestion
    const refundRequests =
      fromBlock === undefined || toBlock === undefined || isNaN(fromBlock) || isNaN(toBlock)
        ? this.refundRequests
        : this.refundRequests.filter((request) => request.blockNumber >= fromBlock && request.blockNumber <= toBlock);
    return deepCopy(refundRequests);
```

james-a-morris (commented):
LGTM - left a few nits

src/clients/SpokePoolClient.ts
nicholaspai: what's the difference? I think using native TS syntax might be better here

james-a-morris: Sorry for the delay on this - I wasn't re-requested / notified of this. The `deepCopy` is going to deep copy the `refundRequest` elements whereas `[...refundRequests]` are just a shallow copy of the elements in a new array instance.

james-a-morris (commented):
Left another comment

PR #272, opened 2023-07-06
fix(UBA): Fix RefundRequest validation

pxrl:
Refund Request validation has been wrong since it was first added, because it relied on the deposits having previously been validated, and that wasn't the case. It also needs to be async because it must be able to search for old fills and deposits via an RPC provider.

While here, remove refundRequestIsValid() from UBAClientBase because it's unlikely to be used in its current form, and refund request validation belongs between the UBAClient and SpokePoolClients.

### DIFF
--- a/src/clients/UBAClient/UBAClientUtilities.ts
+++ b/src/clients/UBAClient/UBAClientUtilities.ts
@@ -401,9 +401,9 @@ async function getFlows(
     return validWithinBounds || hasMatchingDeposit;
   });
 
-  const refundRequests: UbaFlow[] = spokePoolClient
-    .getRefundRequests(fromBlock, toBlock)
-    .filter(async (refundRequest) => {
+  const refundRequests: UbaFlow[] = await filterAsync(
+    spokePoolClient.getRefundRequests(fromBlock, toBlock),
+    async (refundRequest) => {
       const result = await refundRequestIsValid(chainIdIndices, spokePoolClients, hubPoolClient, refundRequest);
       if (!result.valid && logger !== undefined) {
         logger.info({
--- a/src/clients/UBAClient/UBAClientUtilities.ts
+++ b/src/clients/UBAClient/UBAClientUtilities.ts
@@ -401,19 +401,21 @@ async function getFlows(
     return validWithinBounds || hasMatchingDeposit;
   });
 
-  const refundRequests: UbaFlow[] = spokePoolClient.getRefundRequests(fromBlock, toBlock).filter((refundRequest) => {
-    const result = refundRequestIsValid(chainId, chainIdIndices, spokePoolClients, hubPoolClient, refundRequest);
-    if (!result.valid && logger !== undefined) {
-      logger.info({
-        at: "UBAClient::getFlows",
-        message: `Excluding RefundRequest on chain ${chainId}`,
-        reason: result.reason,
-        refundRequest,
-      });
-    }
+  const refundRequests: UbaFlow[] = spokePoolClient
+    .getRefundRequests(fromBlock, toBlock)
+    .filter(async (refundRequest) => {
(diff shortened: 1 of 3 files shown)


src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -401,19 +401,21 @@ async function getFlows(
...
-    const result = refundRequestIsValid(chainId, chainIdIndices, spokePoolClients, hubPoolClient, refundRequest);
-    if (!result.valid && logger !== undefined) {
-      logger.info({
-        at: "UBAClient::getFlows",
-        message: `Excluding RefundRequest on chain ${chainId}`,
-        reason: result.reason,
-        refundRequest,
-      });
-    }
+  const refundRequests: UbaFlow[] = spokePoolClient
```
nicholaspai, line 404: Should we `Promise.all()` these?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -401,19 +401,21 @@ async function getFlows(
...
+    .getRefundRequests(fromBlock, toBlock)
+    .filter(async (refundRequest) => {
```
james-a-morris, line 406: You may want to use `filterAsync` in `ArrayUtils.ts`

james-a-morris (approved):
Good changes - one nit

src/clients/UBAClient/UBAClientUtilities.ts
pxrl: Implemented: https://github.com/across-protocol/sdk-v2/pull/272/commits/61e56bd591669ae429af2d179f7dd589b0c9b025

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: The `filterAsync` function in utils internally uses this. We will need to rewrite this though b/c this will return a promise.

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -401,19 +401,21 @@ async function getFlows(
...
+          message: `Excluding RefundRequest on chain ${chainId}`,
+          reason: result.reason,
+          refundRequest,
+        });
+      }
 
-    return result.valid;
-  });
+      return result.valid;
+    });
```
james-a-morris: We're going to need to modify this to get the output the way we want.

```
  const refundRequests: UbaFlow[] = spokePoolClient
    .getRefundRequests(fromBlock, toBlock)
    .filter(async (refundRequest) => {
      const result = await refundRequestIsValid(chainIdIndices, spokePoolClients, hubPoolClient, refundRequest);
      if (!result.valid && logger !== undefined) {
        logger.info({
          at: "UBAClient::getFlows",
          message: `Excluding RefundRequest on chain ${chainId}`,
          reason: result.reason,
          refundRequest,
        });
      }

      return result.valid;
    });
```

should probably become

```
  const refundRequests: UbaFlow[] = await filterAsync(
    spokePoolClient.getRefundRequests(fromBlock, toBlock),
    async (refundRequest) => {
      const result = await refundRequestIsValid(chainIdIndices, spokePoolClients, hubPoolClient, refundRequest);
      if (!result.valid && logger !== undefined) {
        logger.info({
          at: "UBAClient::getFlows",
          message: `Excluding RefundRequest on chain ${chainId}`,
          reason: result.reason,
          refundRequest,
        });
      }
      return result.valid;
    });
```

src/clients/UBAClient/UBAClientUtilities.ts
pxrl: Handled via James' `filterAsync()` suggestion :+1:

PR #274, opened 2023-07-06
improve: refactor spokepool tests to SDK-V2

james-a-morris:


james-a-morris:
This is on hold until we change the SDK-V2 to using mocha/chai

PR #275, opened 2023-07-06
improve: Additional exports

pxrl:
For use in relayer-v2.

### DIFF
--- a/src/clients/UBAClient/index.ts
+++ b/src/clients/UBAClient/index.ts
@@ -1,3 +1,5 @@
+export * from "./UBAClientUtilities";
+
 export { UBAActionType } from "../../UBAFeeCalculator/UBAFeeTypes";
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):

src/clients/UBAClient/index.ts
```diff
@@ -1,4 +1,5 @@
 export { UBAActionType } from "../../UBAFeeCalculator/UBAFeeTypes";
 export { BalancingFeeReturnType as UBABalancingFee, SystemFeeResult as UBASystemFee } from "./UBAClientTypes";
+export { refundRequestIsValid } from "./UBAClientUtilities";
```
james-a-morris, line 3: We may want to export everything on the off-chance it's needed

james-a-morris (approved):

src/clients/UBAClient/index.ts
pxrl:
```suggestion
export * from "./UBAClientUtilities";
```

PR #277, opened 2023-07-10
feat: add base-client extension to all clients

james-a-morris:
This PR essentially adds a base-client that is extended by all of our bot clients. The reason being: we have common `isUpdated` logic in all clients and that should be strictly enforced.

### DIFF
--- a/src/clients/BaseAbstractClient.ts
+++ b/src/clients/BaseAbstractClient.ts
@@ -21,7 +21,7 @@ export abstract class BaseAbstractClient {
    * @param value Whether the client has been updated since it was created.
    * @throws Error if the client has already been updated and the value is false.
    */
-  protected set isUpdated(value: boolean) {
+  public set isUpdated(value: boolean) {
     if (this._isUpdated === true && !value) {
       throw new Error("Cannot set isUpdated to false once it is true");
     }
--- a/src/clients/BaseAbstractClient.ts
+++ b/src/clients/BaseAbstractClient.ts
@@ -1,0 +1,39 @@
+/**
+ * Base class for all clients to extend.
+ */
+export abstract class BaseAbstractClient {
+  private _isUpdated: boolean;
+
+  constructor() {
+    this._isUpdated = false;
+  }
+
+  /**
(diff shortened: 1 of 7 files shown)


src/clients/BaseAbstractClient.ts
```diff
@@ -0,0 +1,39 @@
...
+      throw new Error("Cannot set isUpdated to false once it is true");
+    }
+    this._isUpdated = value;
+  }
+
+  /**
+   * Asserts that the client has been updated.
+   */
+  protected assertUpdated(): void {
+    if (!this.isUpdated) {
```
nicholaspai, line 35: should this be `_isUpdated`?

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -358,6 +360,7 @@ export abstract class BaseUBAClient {
     if (state) {
       this.bundleStates = state;
     }
+    this.isUpdated = true;
```
nicholaspai, line 363: Should this be `_isUpdated`? 

Also, I'm a bit confused how this extension works. The other clients IIUC all set `isUpdated = true` at the end of their `update()` calls. Does this mean we need to change those calls to set `_isUpdated = true` instead?

src/clients/BaseAbstractClient.ts
james-a-morris: We could do either-or. In terms of the `get` `set` functions, these are added to the object as a class variable so we could reference `obj.isUpdated` or `obj.isUpdated = ...` and it will trigger these getter/setter functions

src/clients/UBAClient/UBAClientBase.ts
james-a-morris: https://github.com/across-protocol/sdk-v2/pull/277#discussion_r1258332373

The `get` `set` definitions in TS allow us to enforce/dynamically generate values from accessing the variables directly.

We can use `isUpdated` and `isUpdated = ...` anywhere in the code now and TS will implicitly route the logic back to the functions created in the base abstract client.

PR #280, opened 2023-07-10
feat: parse UBA config data

james-a-morris:


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -299,6 +304,16 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
         // class itself to work with the generalized ConfigStore.
         const l1Token = args.key;
 
+        // For now, we'll need to check if this value is undefined. In the
+        // future, we should make this a required field.
+        if (parsedValue.uba !== undefined) {
```
nicholaspai, line 309: Can we instead set default values? We can add default values to the UMIP so this is in-protocol

src/clients/AcrossConfigStoreClient/ConfigStoreParsingUtilities/ConfigStoreParsingUtilities.ts
```diff
@@ -0,0 +1,60 @@
...
+import { UBAOnChainConfigType, UBAParsedConfigType } from "../../../interfaces";
+import { UBA_CONFIG_ONCHAIN_SCHEMA } from "./superstructs";
+
+/**
+ * Type guard for UBAOnChainConfigType
+ * @param input An unknown value that may or may not be a UBAOnChainConfigType
+ * @returns True if the input is a UBAOnChainConfigType, false otherwise
+ */
+export function isUBAOnChainConfig(input: unknown): input is UBAOnChainConfigType {
+  return UBA_CONFIG_ONCHAIN_SCHEMA.is(input);
```
nicholaspai, line 11: typescript for the win!

nicholaspai (commented):
Questions on:
- default values [here](https://github.com/across-protocol/sdk-v2/pull/280/files#r1259823595)

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Yep! These will be needed to be supplied by Data. I'll create a ticket for it

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -441,4 +456,18 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+
+  /**
+   * Retrieves a UBA config for a given L1 token address and block number.
+   * @param l1TokenAddress The L1 token address to retrieve the config for
+   * @param blockNumber The block number to retrieve the config for. If not provided, the latest config will be returned.
+   * @returns The UBA config for the given L1 token address and block number, or undefined if no config exists.
+   */
+  getUBAConfig(l1TokenAddress: string, blockNumber?: number): UBAParsedConfigType | undefined {
+    const config = findLast(
+      sortEventsAscending(this.ubaConfigUpdates),
```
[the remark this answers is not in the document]
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/280/commits/822abf0269f68b5342632a198ef274b0f66e9a80

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -299,6 +304,17 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
         // class itself to work with the generalized ConfigStore.
         const l1Token = args.key;
 
+        // For now, we'll need to check if this value is undefined. In the
+        // future, we should make this a required field.
+        if (parsedValue.uba !== undefined) {
```
nicholaspai, line 309: Can we skip this parsing step in prod while continuing to be able to unit test it? I ask because we should push the config store changes on chain without fearing the existing bots crashing as soon as it tries to parse the `uba` string

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -299,6 +304,17 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+          // Parse and store UBA config
+          const ubaConfig = parseUBAConfigFromOnChain(parsedValue?.uba);
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          const { value: _value, key: _key, ...passedArgs } = args;
+          this.ubaConfigUpdates.push({ ...passedArgs, config: ubaConfig, l1Token });
+          sortEventsAscendingInPlace(this.ubaConfigUpdates);
```
nicholaspai, line 315: Why would you sort after each `.push`? Why not sort after going through all `ubaConfigUpdates`?

src/clients/AcrossConfigStoreClient/ConfigStoreParsingUtilities/ConfigStoreParsingUtilities.test.ts
```diff
@@ -0,0 +1,233 @@
...
+        },
+      },
+    },
+  },
+  {
+    reason: "Invalid config with rebalance having a key that isn't '{number}-{number}'",
+
+    value: {
+      alpha: {
+        default: "23",
```
nicholaspai, line 118: What if `default` is empty, is that a case we should handle?

nicholaspai (approved):
LGTM, one suggestion to reduce chance of existing bots crashing once we push config changes on chain

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/280/commits/c0664473b6fa3fb8f9a34cb0bb2d57aaab087244

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Do we have an ENV variable currently in place that can tell if the system is in prod or not?

src/clients/AcrossConfigStoreClient/ConfigStoreParsingUtilities/ConfigStoreParsingUtilities.test.ts
james-a-morris: Happy to enforce it! do you think this case will appear often?

james-a-morris: Alternatively, since we have a fixed default that will be hardcoded, it shouldn't necessarily be required. Following that logic, we may not need default at all since we have a protocol default.

nicholaspai: I would agree it shouldn't be required if we have a fixed default

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
nicholaspai: we do not but I would opt to hardcode the skip so that we can test locally and open a PR to remove the hardcoded skip

nicholaspai (approved):
https://github.com/across-protocol/sdk-v2/pull/280#discussion_r1260285845

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -299,6 +304,16 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
         // class itself to work with the generalized ConfigStore.
         const l1Token = args.key;
 
+        // For now, we'll need to check if this value is undefined. In the
+        // future, we should make this a required field.
+        if (parsedValue.uba !== undefined) {
+          // Parse and store UBA config
+          const ubaConfig = parseUBAConfigFromOnChain(parsedValue?.uba);
```
pxrl, line 311:
```suggestion
        if (parsedValue?.uba !== undefined) {
          // Parse and store UBA config
          const ubaConfig = parseUBAConfigFromOnChain(parsedValue.uba);
```

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
pxrl, line 1: I see this is being moved to a new directory - is that because additional files will be added?

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -28,9 +29,12 @@ import {
   RouteRateModelUpdate,
   ConfigStoreVersionUpdate,
   DisabledChainsUpdate,
-} from "../interfaces";
+  UBAConfigUpdates,
+  UBAParsedConfigType,
+} from "../../interfaces";
```
pxrl, line 34: This sets off some alarm bells for me - does the ConfigStoreClient class need to know about UBA concepts and types? It seems like an inversion to me.

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -119,27 +120,80 @@ export function computeLpFeeStateful(
...
-  token;
-  blockNumber;
+  // If the config client has not been updated at least
+  // once, throw
+  if (!configClient.isUpdated) {
+    throw new Error("Config client not updated");
+  }
+  const ubaConfig = configClient.getUBAConfig(token, blockNumber === "latest" ? undefined : blockNumber);
+  if (ubaConfig === undefined) {
+    throw new Error("UBA config not found");
```
pxrl, line 158:
```suggestion
    throw new Error(`UBA config for blockTag ${blockNumber} not found");
```

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Yep! I thought it was more effective to colocate the files

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/280/commits/8b5f6e9280008df22c2ab4b2e35fc3b53ead276f

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: These types are specifically defined within the `./interfaces/ConfigStore.ts` file. They represent the on-chain definition of the UBA Configuration (similar to the rate model). 

I think the ConfigStore will need to know about the uba concepts insofar as to parse out the config from our configstore contract

cc @nicholaspai

pxrl: On second review I can see that there are already other upper-layer primitives that we depend on in here - i.e. `L1TokenTransferThreshold` and `SpokePoolTargetBalance`.

It still generally seems circular to me, because the ConfigStoreClient otherwise sits at the very low layers of our part of the stack, but it ends up knowing about high level concepts. I may just have the wrong perspective on it though, and (in light of the above) the changes being introduced are at least consistent with the existing usage :shrug:.

nicholaspai: I think where you put the `UBAConfigUpdates` and `UBAParsedConfigType` type, in the `interfaces/ConfigStore` makes sense to me. The ConfigStoreClient definitely needs to know the shape of the UBA config in order to parse the on chain `uba` key into the correct curves

pxrl (approved):
OK from me :+1:

PR #281, opened 2023-07-10
fix: Various fixes

nicholaspai:
- getUBAFeeConfig should probably be a `sync` method
- expose `getOpeningTokenBalances` utility method
- `getOpeningTokenBalances` is called incorrectly by `UBAClientBase.update` by passing in token symbol instead of address. I simplify this interface and just pass in `l1TokenAddress` instead of the `l2TokenAddress`


james-a-morris (approved):
LGTM - good changes

PR #282, opened 2023-07-11
improve(test): Singleton event manager in mocked clients

pxrl:
This change migrates to the use of a singleton for event managers, enabling multiple "client" classes to see consistent block number increments. This addresses a race condition in the relayer-v2 runningBalances test where the ConfigStoreClient and HubPoolClient can get out of sync and disagree about the applicable version.

While here, add an additional check to ensure that FundsDeposited events are generated with sequential (or at least incrementing) depositIds.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -695,7 +695,7 @@ export class HubPoolClient extends BaseAbstractClient {
       // Safeguard
       if (executedRootBundle.runningBalances.length !== nTokens) {
         throw new Error(
-          `Invalid runningBalances length (${executedRootBundle.runningBalances.length} != ${nTokens})` +
+          `Invalid runningBalances length (${executedRootBundle.runningBalances.length} !== ${nTokens})` +
             ` for ConfigStore version ${version} in chain ${this.chainId} transaction ${event.transactionHash}`
         );
       }
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -694,7 +694,7 @@ export class HubPoolClient {
       // Safeguard
       if (executedRootBundle.runningBalances.length !== nTokens) {
         throw new Error(
-          `Invalid runningBalances length (${executedRootBundle.runningBalances.length})` +
+          `Invalid runningBalances length (${executedRootBundle.runningBalances.length} != ${nTokens})` +
             ` for ConfigStore version ${version} in chain ${this.chainId} transaction ${event.transactionHash}`
         );
       }
--- a/src/clients/mocks/MockEvents.ts
+++ b/src/clients/mocks/MockEvents.ts
@@ -1,6 +1,7 @@
+import assert from "assert";
 import { utils as ethersUtils, Event, providers } from "ethers";
 import { random } from "lodash";
-import { toBN, randomAddress } from "../../utils";
+import { isDefined, randomAddress, toBN } from "../../utils";
 
 const { id, keccak256, toUtf8Bytes } = ethersUtils;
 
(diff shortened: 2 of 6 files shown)


nicholaspai (approved):

src/clients/HubPoolClient.ts
```diff
@@ -695,7 +695,7 @@ export class HubPoolClient extends BaseAbstractClient {
       // Safeguard
       if (executedRootBundle.runningBalances.length !== nTokens) {
         throw new Error(
-          `Invalid runningBalances length (${executedRootBundle.runningBalances.length})` +
+          `Invalid runningBalances length (${executedRootBundle.runningBalances.length} != ${nTokens})` +
```
james-a-morris, line 698: Should we use the strict equality operator here? 
```suggestion
          `Invalid runningBalances length (${executedRootBundle.runningBalances.length} !== ${nTokens})` +
```

src/clients/mocks/MockEvents.ts
```diff
@@ -95,3 +114,22 @@ export class EventManager {
     } as Event;
   }
 }
+
+export function getEventManager(
```
james-a-morris, line 118: Should we leave JSDocs here?

src/clients/mocks/MockEvents.ts
```diff
@@ -34,8 +37,19 @@ const removeListener = (): void => {
 
 export class EventManager {
   private logIndexes: Record<string, number> = {};
+  public readonly minBlockRange = 10;
+  public readonly eventSignatures: Record<string, string> = {};
+
+  constructor(public blockNumber = 0) {}
 
-  constructor(public readonly eventSignatures: Record<string, string>) {}
+  addEventSignatures(eventSignatures: Record<string, string>): void {
```
james-a-morris, line 45: Should we apply an explicit visibility operator here? `public` `protected` `private` etc

src/clients/mocks/MockEvents.ts
```diff
@@ -44,7 +58,12 @@ export class EventManager {
 
     let { blockNumber, transactionIndex } = inputs;
 
-    blockNumber = blockNumber ?? random(1, 100_000, false);
+    // Increment the block number by at least 1, by default. The caller may override
+    // to force the same block number to be used, but never a previous block number.
+    blockNumber = blockNumber ?? random(this.blockNumber + 1, this.blockNumber + this.minBlockRange, false);
+    assert(blockNumber >= this.blockNumber, `${blockNumber} < ${this.blockNumber}`);
```
james-a-morris, line 64: Should we leave a more explicit error? 
```suggestion
    assert(blockNumber >= this.blockNumber, `Inequality failed to match ${blockNumber} < ${this.blockNumber}`);
```

src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -94,8 +95,12 @@ export class MockSpokePoolClient extends SpokePoolClient {
   generateDeposit(deposit: DepositWithBlock): Event {
     const event = "FundsDeposited";
 
-    const { depositId, blockNumber, transactionIndex } = deposit;
-    let { depositor, destinationChainId } = deposit;
+    const { blockNumber, transactionIndex } = deposit;
+    let { depositId, depositor, destinationChainId } = deposit;
+    depositId = depositId ?? this.numberOfDeposits;
+    assert(depositId >= this.numberOfDeposits, `${depositId} < ${this.numberOfDeposits}`);
+    this.numberOfDeposits = ++depositId;
```
james-a-morris, line 102: Is `depositId` saved outside of this function? If not, we may want to use `depositId + 1` for readability
```suggestion
    this.numberOfDeposits = depositId + 1;
```

james-a-morris (commented):
LGTM! I left a few nits 🙂 Great work on this

src/clients/mocks/MockEvents.ts
pxrl: `Public` is the default so there's no change in behaviour as a result...is there any advantage to explicitly specifying it?

src/clients/mocks/MockEvents.ts
pxrl: I'd prefer to leave it as is - `Inequality failed to match ...` isn't adding any information that isn't already there, and when these errors throw there tends to be a wall of error text output, so it can be hard to spot the actual error amongst all the text.

PR #285, opened 2023-07-11
chore: bump package version

nicholaspai:


nicholaspai:
> @nicholaspai What's the trigger for building this new patch version?

It will include all the small bug fixes that we've pushed in the last 12 hours

PR #287, opened 2023-07-12
improve(UBAClient): Retain HubPool chain ID statefully

pxrl:
The HubPool chain ID is key information for the UBAClient, so it makes sense to retain it as a state variable, rather than to require it to be passed in as an argument to various methods.

### DIFF
--- a/src/clients/UBAClient/UBAClientBase.ts
+++ b/src/clients/UBAClient/UBAClientBase.ts
@@ -36,6 +36,7 @@ export abstract class BaseUBAClient extends BaseAbstractClient {
    */
   constructor(
     protected readonly chainIdIndices: number[],
+    protected readonly hubPoolChainId: number,
     protected readonly tokens: string[],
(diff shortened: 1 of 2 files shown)


src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -194,15 +194,14 @@ export abstract class BaseUBAClient extends BaseAbstractClient {
   protected computeLpFee(
     amount: BigNumber,
     depositChainId: number,
-    hubPoolChainId: number,
```
pxrl, line 197: I can see that `computeLpFee()` is called by `computeSystemFee()`, except in that case the chainId might _not_ be `hubPoolChainId`. The caller may instead pass in any destination chain ID. If so, is it correct to name this `hubPoolChainId`?

pxrl:
Looks like the name hubPoolChainId is incorrect, so will close this now.

PR #291, opened 2023-07-12
fix(UBAFeeCalculator): Refine `calculateHistoricalRunningBalance`

nicholaspai:
- We need to make sure we're only passing the snapshotted running balances back and forth between `getEventFee` and `calculateHistoricalRunningBalance`
- Add clarifying comments as to the usage of these functions


james-a-morris (approved):
LGTM

PR #295, opened 2023-07-13
feat: allow passing `queriesMap` to relay fee calc config

dohaki:
While working on the `GET /uba-bundle-state` endpoint for the quote engine, I noticed that we pass the wrong (?) queries here https://github.com/across-protocol/sdk-v2/blob/use-correct-queries-relay-fee-calculator/src/clients/UBAClient/UBAClientUtilities.ts#L351. The old `RelayFeeCalculatorConfig` expects as a property an instance of `QueryInterface` which implements the respective `gasCosts`, `price`, `decimals` queries for a single chain. But AFAICT we need the queries of the respective destination chain.

This PR tries to fix this by allowing to instantiate an instance of `RelayerFeeCalculator` using a `queriesMap` and an additional `destinationChainId` parameter. I chose this approach in order to not break other code that uses the calculator.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -39,7 +39,6 @@ export interface RelayFeeCalculatorConfigWithQueries extends BaseRelayFeeCalcula
 }
 export interface RelayFeeCalculatorConfigWithMap extends BaseRelayFeeCalculatorConfig {
   queriesMap: Record<number, QueryInterface>;
-  destinationChainId: number;
 }
 export type RelayFeeCalculatorConfig = RelayFeeCalculatorConfigWithQueries | RelayFeeCalculatorConfigWithMap;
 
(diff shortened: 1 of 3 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -33,8 +33,14 @@ export interface RelayFeeCalculatorConfig {
   capitalCostsConfig?: {
     [token: string]: CapitalCostConfig | CapitalCostConfigOverride;
   };
+}
+export interface RelayFeeCalculatorConfigWithQueries extends BaseRelayFeeCalculatorConfig {
   queries: QueryInterface;
 }
+export interface RelayFeeCalculatorConfigWithMap extends BaseRelayFeeCalculatorConfig {
+  queriesMap: Record<number, QueryInterface>;
```
nicholaspai, line 41: +1

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -84,8 +90,19 @@ export class RelayFeeCalculator {
...
+  constructor(config: RelayFeeCalculatorConfigWithQueries, logger?: Logger);
+  constructor(config: RelayFeeCalculatorConfigWithMap, logger?: Logger, destinationChainId?: number);
+  constructor(config?: RelayFeeCalculatorConfig, logger?: Logger, destinationChainId?: number) {
+    assert(config, "config must be provided");
+
+    if ("queries" in config) {
+      this.queries = config.queries;
+    } else {
+      assert(destinationChainId !== undefined, "destinationChainId must be provided if queriesMap is provided");
+      assert(config.queriesMap[destinationChainId], "No queries provided for destination chain");
```
nicholaspai, line 102: +1 for safety

nicholaspai (approved):

PR #299, opened 2023-07-13
improve(UBAClient): Make spokePoolClients public

nicholaspai:


src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -3,30 +3,24 @@ import winston from "winston";
...
-
   // @dev chainIdIndices supports indexing members of root bundle proposals submitted to the HubPool.
   //      It must include the complete set of chain IDs ever supported by the HubPool.
   // @dev SpokePoolClients may be a subset of the SpokePools that have been deployed.
   constructor(
     readonly chainIdIndices: number[],
     readonly tokens: string[],
     protected readonly hubPoolClient: HubPoolClient,
-    protected readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
+    readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
```
james-a-morris, line 16: I think we should be explicit with the typing. 
```suggestion
    public readonly spokePoolClients: { [chainId: number]: SpokePoolClient },
```

james-a-morris (approved):
LGTM

PR #300, opened 2023-07-13
fix(UBAClient): Fix various bugs in refund request validation

nicholaspai:
For reference, [this PR](https://github.com/across-protocol/contracts-v2/pull/264/files) implemented most of the refund validation features we take advantage of in this PR.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -175,7 +175,7 @@ export class HubPoolClient extends BaseAbstractClient {
     return l1Token;
   }
 
-  getL1TokenCounterpartAtBlock(l2ChainId: number, l2Token: string, l1Block: number): string {
+  getL1TokenCounterpartAtBlock(l2ChainId: number, l2Token: string, hubPoolBlock: number): string {
     const l1Token = Object.keys(this.l1TokensToDestinationTokensWithBlock).find((_l1Token) => {
       // If this token doesn't exist on this L2, return false.
       if (this.l1TokensToDestinationTokensWithBlock[_l1Token][l2ChainId] === undefined) {
@@ -184,12 +184,12 @@ export class HubPoolClient extends BaseAbstractClient {
 
       // Find the last mapping published before the target block.
       return sortEventsDescending(this.l1TokensToDestinationTokensWithBlock[_l1Token][l2ChainId]).find(
-        (mapping: DestinationTokenWithBlock) => mapping.l2Token === l2Token && mapping.blockNumber <= l1Block
+        (mapping: DestinationTokenWithBlock) => mapping.l2Token === l2Token && mapping.blockNumber <= hubPoolBlock
       );
     });
     if (!l1Token) {
       throw new Error(
-        `Could not find L1 token mapping for chain ${l2ChainId} and L2 token ${l2Token} equal to or earlier than block ${l1Block}!`
+        `Could not find L1 token mapping for chain ${l2ChainId} and L2 token ${l2Token} equal to or earlier than block ${hubPoolBlock}!`
       );
     }
     return l1Token;
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -175,7 +175,7 @@ export class HubPoolClient extends BaseAbstractClient {
     return l1Token;
   }
 
-  getL1TokenCounterpartAtBlock(l2ChainId: number, l2Token: string, block: number): string {
+  getL1TokenCounterpartAtBlock(l2ChainId: number, l2Token: string, l1Block: number): string {
     const l1Token = Object.keys(this.l1TokensToDestinationTokensWithBlock).find((_l1Token) => {
       // If this token doesn't exist on this L2, return false.
       if (this.l1TokensToDestinationTokensWithBlock[_l1Token][l2ChainId] === undefined) {
@@ -184,12 +184,12 @@ export class HubPoolClient extends BaseAbstractClient {
 
       // Find the last mapping published before the target block.
       return sortEventsDescending(this.l1TokensToDestinationTokensWithBlock[_l1Token][l2ChainId]).find(
-        (mapping: DestinationTokenWithBlock) => mapping.l2Token === l2Token && mapping.blockNumber <= block
+        (mapping: DestinationTokenWithBlock) => mapping.l2Token === l2Token && mapping.blockNumber <= l1Block
       );
     });
     if (!l1Token) {
       throw new Error(
-        `Could not find L1 token mapping for chain ${l2ChainId} and L2 token ${l2Token} equal to or earlier than block ${block}!`
+        `Could not find L1 token mapping for chain ${l2ChainId} and L2 token ${l2Token} equal to or earlier than block ${l1Block}!`
       );
     }
     return l1Token;
--- a/src/clients/UBAClient/UBAClientTypes.ts
+++ b/src/clients/UBAClient/UBAClientTypes.ts
@@ -1,9 +1,9 @@
 import { BigNumber } from "ethers";
-import { UbaFlow } from "../../interfaces";
+import { FillWithBlock, UbaFlow } from "../../interfaces";
 import { UBAActionType, FlowTupleParameters } from "../../UBAFeeCalculator/UBAFeeTypes";
 import UBAConfig from "../../UBAFeeCalculator/UBAFeeConfig";
 
-export type RequestValidReturnType = { valid: false; reason: string } | { valid: true };
+export type RequestValidReturnType = { valid: boolean; reason?: string; matchingFill?: FillWithBlock };
 export type BalancingFeeReturnType = { balancingFee: BigNumber; actionType: UBAActionType };
 export type SystemFeeResult = { lpFee: BigNumber; depositBalancingFee: BigNumber; systemFee: BigNumber };
 export type RelayerFeeResult = {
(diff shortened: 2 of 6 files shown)


src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -541,14 +542,23 @@ export async function refundRequestIsValid(
     };
   }
 
-  // Validate relayer and depositId.
+  if (previousIdenticalRequests.gt(0)) {
```
nicholaspai, line 545: Is this true @mrice32 ? We should only handle refunds where `previousIdenticalRequests == 0`?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -541,14 +542,23 @@ export async function refundRequestIsValid(
...
     // prettier-ignore
     return (
         fill.depositId === depositId
         && fill.originChainId === originChainId
         && fill.destinationChainId === destinationChainId
+        // Must have requested refund on chain that refund was sent on.
+        && fill.repaymentChainId === repaymentChainId
+        // Must be a full fill to qualify for a refund.
         && fill.amount.eq(amount)
+        && fill.fillAmount.eq(amount)
```
nicholaspai, line 561: This check isn't necessarily required since partial fills are forced by the contracts to take repayment on destination chain but its a safety check?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -562,17 +572,27 @@ export async function refundRequestIsValid(
...
 
-  // Verify that the refundToken maps to a known HubPool token.
+  // Verify that the refundToken maps to a known HubPool token and is the correct
+  // token for the chain where the refund was sent from.
   // Note: the refundToken must be valid at the time of the Fill *and* the RefundRequest.
-  // @todo: Resolve to the HubPool block number at the time of the RefundRequest ?
-  const hubPoolBlockNumber = hubPoolClient.latestBlockNumber ?? hubPoolClient.deploymentBlock - 1;
+  // @todo: This really should be using the hub pool block equivalent of the fillBlock. Its unlikely to be a problem
+  // since L1 tokens are rarely expected to be remapped but its worth solving.
+  const hubPoolBlockNumber = hubPoolClient.latestBlockNumber!;
```
nicholaspai, line 580: This is an annoying artifact. We really should be passing in a mainnet bundle end block, but that would require adding an optional param to this call (and all the way up the stack).

Its pretty ugly but we might have to do it to be correct

src/clients/HubPoolClient.ts
```diff
@@ -175,7 +175,7 @@ export class HubPoolClient extends BaseAbstractClient {
     return l1Token;
   }
 
-  getL1TokenCounterpartAtBlock(l2ChainId: number, l2Token: string, block: number): string {
+  getL1TokenCounterpartAtBlock(l2ChainId: number, l2Token: string, l1Block: number): string {
```
pxrl, line 178: Thoughts on `hubPoolBlock` instead of `l1Block` ?

We will soon support SpokePool deployments to other L1s, so the existing naming conventions of `l1`and `l2` terms won't make as much sense and IMO should be phased out.

src/clients/UBAClient/UBAClientTypes.ts
```diff
@@ -1,9 +1,9 @@
 import { BigNumber } from "ethers";
-import { UbaFlow } from "../../interfaces";
+import { FillWithBlock, UbaFlow } from "../../interfaces";
 import { UBAActionType, FlowTupleParameters } from "../../UBAFeeCalculator/UBAFeeTypes";
 import UBAConfig from "../../UBAFeeCalculator/UBAFeeConfig";
 
-export type RequestValidReturnType = { valid: false; reason: string } | { valid: true };
+export type RequestValidReturnType = { valid: boolean; reason?: string; matchingFill?: FillWithBlock };
```
pxrl, line 6: Agree this is good, and might also be worth returning the corresponding deposit as well?  Given that we already do the work to look it up...

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -562,17 +574,24 @@ export async function refundRequestIsValid(
...
-  const hubPoolBlockNumber = hubPoolClient.latestBlockNumber ?? hubPoolClient.deploymentBlock - 1;
+  // Verify that the refundToken maps to a known HubPool token and is the correct
+  // token for the chain where the refund was sent from.
+  // Note: the refundToken must be valid at the time the deposit was sent.
   try {
-    hubPoolClient.getL1TokenCounterpartAtBlock(repaymentChainId, refundToken, hubPoolBlockNumber);
+    const l1TokenForFill = hubPoolClient.getL1TokenCounterpartAtBlock(
+      fill.destinationChainId,
+      fill.destinationToken,
+      deposit.quoteBlockNumber
```
pxrl, line 584: Nice, the quote block number is a good fit for this.

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -562,17 +574,24 @@ export async function refundRequestIsValid(
...
+    );
+    const expectedRefundToken = hubPoolClient.getDestinationTokenForL1Token(l1TokenForFill, repaymentChainId);
+    if (expectedRefundToken !== refundToken) {
+      return { valid: false, reason: `Refund token does not map to expected refund token ${refundToken}` };
+    }
   } catch {
-    return { valid: false, reason: `Refund token unknown at HubPool block ${hubPoolBlockNumber}` };
+    return { valid: false, reason: `Refund token unknown at deposit quote block ${deposit.quoteBlockNumber}` };
```
pxrl, line 591: Thoughts on leaving this as is? `HubPool block` is more obvious than `deposit quote block` IMO.

pxrl, line 591:
```suggestion
    return { valid: false, reason: `Refund token unknown at HubPool block ${deposit.quoteBlockNumber}` };
```

Thoughts on the message as is? `HubPool block` is more obvious than `deposit quote block` IMO.

src/clients/UBAClient/UBAClientTypes.ts
pxrl, line 6:
```suggestion
// @todo: Revert to this after bumping typescript: { valid: true, fill: FillWithBlock } | { valid: false, reason: string } ;
export type RequestValidReturnType = { valid: boolean; reason?: string; matchingFill?: FillWithBlock };
```

src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -12,7 +12,7 @@ export class UBAClientWithRefresh extends BaseUBAClient {
   constructor(
     readonly chainIdIndices: number[],
     readonly tokens: string[],
-    protected readonly hubPoolClient: HubPoolClient,
+    public readonly hubPoolClient: HubPoolClient,
```
pxrl, line 15: Curious about this change - it doesn't seem to be used elsewhere in this commit.

src/clients/UBAClient/UBAClientTypes.ts
nicholaspai: I want to be practical. We don't need it as far as I can tell in the dataworker. Does the Relayer needd it from this function?

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: It's not only a good fit it's the only correct block to insert here. The previous block passed was incorrect

src/clients/UBAClient/UBAClientWithRefresh.ts
nicholaspai: I use it in the dataworker actually. I like enforcing that the dataworker is using the same clients that the uba client is using so it's events line up

src/clients/HubPoolClient.ts
nicholaspai: yes good point!

src/clients/UBAClient/UBAClientWithRefresh.ts
pxrl: Yeah - if we are instantiating multiple HubPoolClient instances then we're doing something wrong. But the HubPoolClient is part of the CommonClients type, so doesn't the Dataworker already have a direct reference for the HubPoolClient?

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: This makes sense to me

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -562,17 +574,24 @@ export async function refundRequestIsValid(
...
+    if (expectedRefundToken !== refundToken) {
+      return { valid: false, reason: `Refund token does not map to expected refund token ${refundToken}` };
+    }
   } catch {
-    return { valid: false, reason: `Refund token unknown at HubPool block ${hubPoolBlockNumber}` };
+    return { valid: false, reason: `Refund token unknown at HubPool block ${deposit.quoteBlockNumber}` };
   }
 
-  return { valid: true };
+  return { valid: true, matchingFill: fill };
```
james-a-morris, line 594: Good addition here

james-a-morris (approved):

src/clients/UBAClient/UBAClientUtilities.ts
mrice32: I think so, at least to start. We should probably log an error (page) in this case so we can figure out why it was double requested and if the initial one was really thrown out.

The main issue this tries to handle is timing issues where the initial request was too early and was thrown out due to a lack of a match.

How do you think we should handle?

nicholaspai: I agree with you about logging a warning but assuming that only the first one is valid for now; that's the solution I've implemented.

PR #307, opened 2023-07-14
improve(ConfigStoreClient): Refine logs

nicholaspai:
- much easier to read a JSON rather than stringified version
- remove debug log with no information


james-a-morris (approved):
LGTM

PR #308, opened 2023-07-15
fix(UBAClient): Add various fixes to updateUBAClients

nicholaspai:
This PR overall tries to make `updateUBAClients` re-use existing functions that have been battle tested in production code.

- `updateUBAClient`'s first step is to construct bundle block ranges for a specified number of the most recently validated bundles. The first fix refactors the block range construction logic into a function `getMostRecentBundles` which has unit tests accompanying it [here](https://github.com/across-protocol/relayer-v2/pull/815/files).
- `calculateUtilizationBoundaries` doesn't correctly take into account both the `refundChain` AND the `originChain` when computing utilization. The UMIP states that utilization calculation should take into account flows being added to the ETH spoke pool from a deposit and subtracted when ETH is the destination spoke pool.
- `getUBATargetSpokeBalances` is used to get spoke target balances set in the config store that we can use to compute the LP fee. The LP fee is a utilization based fee that is based on the cumulative spoke target across all chains. Therefore it is dangerous to allow the caller to specify which `chainIds` to grab spoke targets for. Instead, hardcode the list of spoke targets to ALL chains supported by Across.
    - Moreover, fixed the `getUBATargetSpokeBalances` call. Currently its fetching the wrong configuration. It should be fetching `target` values within the `uba` key, not the legacy `spokeTargetBalance`
- `computeLpFee` is supposed to be used to grab the latest LP fee. This is useful for the `Relayer` and also the Fee Quoting API. Therefore this function should be grabbing fresh HubPool balance, SpokePool balance, and spoke target information.
- Removes the  HubPool balance, SpokePool balance information from the `UBABundleState`. 
- Removes the relayer gas and capital fee information from a flow. Similar to the LP fee calculation inputs above, its not really useful to know these for each flow. The only information that is really needed to save with each flow is the running balance, incentive balance, and balancing fee at the time of the flow. These are saved so that we can easily recompute any historical flow's balancing fees, which are dependent only on the preceding flows' running balances, incentive balances, and balancing fees. The relayer gas + capital fee and the LP fees are not needed to reconstruct running balances.
    - I could see one use case where someone wants to reconstruct the LP fees and relayer gas + capital fees at some historical point in time to figure out what these values would have been at the time of some historical flow. However, this information is grabbed in O(1) time, by querying block chain state at that flow's block time. Unlike the balancing fees, these fees are not dependent on preceding historical information, and therefore derive a lot less benefit from caching the data.

### DIFF
--- a/src/UBAFeeCalculator/UBAFeeConfig.ts
+++ b/src/UBAFeeCalculator/UBAFeeConfig.ts
@@ -1,5 +1,6 @@
 import { BigNumber } from "ethers";
 import { ThresholdBoundType, FlowTupleParameters } from "./UBAFeeTypes";
+import { CHAIN_ID_LIST_INDICES, HUBPOOL_CHAIN_ID } from "../constants";
 
 type ChainId = number;
 type RouteCombination = string;
@@ -107,12 +108,34 @@ class UBAConfig {
   /**
    * @description Get the balance trigger threshold for a given chain and token
    * @param chainId The chain id
-   * @param tokenSymbol The token address
+   * @param l1TokenAddress The token address
    * @returns The balance trigger threshold if it exists
    */
-  public getBalanceTriggerThreshold(chainId: number, tokenSymbol: string): ThresholdBoundType {
-    const chainTokenCombination = `${chainId}-${tokenSymbol}`;
+  public getBalanceTriggerThreshold(chainId: number, l1TokenAddress: string): ThresholdBoundType {
+    const chainTokenCombination = `${chainId}-${l1TokenAddress}`;
     return this.balanceTriggerThreshold.override?.[chainTokenCombination] ?? this.balanceTriggerThreshold.default;
+  }
+
+  /**
+   * Arbitrarily return upper bound target. This could be the average of the upper and lower bound targets but
+   * for now return upper bound target.
+   * @param chainId
+   * @param l1TokenAddress
+   * @returns
+   */
+  public getTargetBalance(chainId: number, l1TokenAddress: string): BigNumber {
+    const thresholdConfig = this.getBalanceTriggerThreshold(chainId, l1TokenAddress);
+    return thresholdConfig?.upperBound?.target ?? BigNumber.from(0);
+  }
+
+  /**
+   * Get sum of all spoke target balances for all chains besides hub pool chain for l1TokenAddress.
+   * This output should be used to compute LP fee based on total spoke target
+   */
+  public getTotalSpokeTargetBalanceForComputingLpFee(l1TokenAddress: string): BigNumber {
+    return CHAIN_ID_LIST_INDICES.filter((chainId) => chainId !== HUBPOOL_CHAIN_ID).reduce((sum, chainId) => {
+      return sum.add(this.getTargetBalance(chainId, l1TokenAddress));
+    }, BigNumber.from(0));
   }
 
   /**
--- a/src/UBAFeeCalculator/UBAFeeConfig.ts
+++ b/src/UBAFeeCalculator/UBAFeeConfig.ts
@@ -108,11 +108,11 @@ class UBAConfig {
   /**
    * @description Get the balance trigger threshold for a given chain and token
    * @param chainId The chain id
-   * @param l1TokenAddress The token address
+   * @param tokenSymbol The token
    * @returns The balance trigger threshold if it exists
    */
-  public getBalanceTriggerThreshold(chainId: number, l1TokenAddress: string): ThresholdBoundType {
-    const chainTokenCombination = `${chainId}-${l1TokenAddress}`;
+  public getBalanceTriggerThreshold(chainId: number, tokenSymbol: string): ThresholdBoundType {
+    const chainTokenCombination = `${chainId}-${tokenSymbol}`;
     return this.balanceTriggerThreshold.override?.[chainTokenCombination] ?? this.balanceTriggerThreshold.default;
   }
 
@@ -123,8 +123,8 @@ class UBAConfig {
    * @param l1TokenAddress
    * @returns
    */
-  public getTargetBalance(chainId: number, l1TokenAddress: string): BigNumber {
-    const thresholdConfig = this.getBalanceTriggerThreshold(chainId, l1TokenAddress);
+  public getTargetBalance(chainId: number, tokenSymbol: string): BigNumber {
+    const thresholdConfig = this.getBalanceTriggerThreshold(chainId, tokenSymbol);
     return thresholdConfig?.upperBound?.target ?? BigNumber.from(0);
   }
 
@@ -132,9 +132,9 @@ class UBAConfig {
    * Get sum of all spoke target balances for all chains besides hub pool chain for l1TokenAddress.
    * This output should be used to compute LP fee based on total spoke target
    */
-  public getTotalSpokeTargetBalanceForComputingLpFee(l1TokenAddress: string): BigNumber {
+  public getTotalSpokeTargetBalanceForComputingLpFee(tokenSymbol: string): BigNumber {
     return CHAIN_ID_LIST_INDICES.filter((chainId) => chainId !== HUBPOOL_CHAIN_ID).reduce((sum, chainId) => {
-      return sum.add(this.getTargetBalance(chainId, l1TokenAddress));
+      return sum.add(this.getTargetBalance(chainId, tokenSymbol));
     }, BigNumber.from(0));
   }
 
--- a/src/UBAFeeCalculator/UBAFeeConfig.ts
+++ b/src/UBAFeeCalculator/UBAFeeConfig.ts
@@ -1,6 +1,6 @@
 import { BigNumber, ethers } from "ethers";
 import { ThresholdBoundType, FlowTupleParameters } from "./UBAFeeTypes";
-import { CHAIN_ID_LIST_INDICES, HUBPOOL_CHAIN_ID } from "../constants";
+import { CHAIN_ID_LIST_INDICES } from "../constants";
 
 type ChainId = number;
 type RouteCombination = string;
@@ -133,7 +133,7 @@ class UBAConfig {
    * This output should be used to compute LP fee based on total spoke target
    */
   public getTotalSpokeTargetBalanceForComputingLpFee(tokenSymbol: string): BigNumber {
-    return CHAIN_ID_LIST_INDICES.filter((chainId) => chainId !== HUBPOOL_CHAIN_ID).reduce((sum, chainId) => {
+    return CHAIN_ID_LIST_INDICES.reduce((sum, chainId) => {
       return sum.add(this.getTargetBalance(chainId, tokenSymbol));
     }, ethers.constants.Zero);
   }
--- a/src/UBAFeeCalculator/UBAFeeConfig.ts
+++ b/src/UBAFeeCalculator/UBAFeeConfig.ts
@@ -131,6 +131,8 @@ class UBAConfig {
   /**
    * Get sum of all spoke target balances for all chains besides hub pool chain for l1TokenAddress.
    * This output should be used to compute LP fee based on total spoke target
+   * @param tokenSymbol The token to find target balances for
+   * @returns The sum of token balances for a specific token symbol
    */
   public getTotalSpokeTargetBalanceForComputingLpFee(tokenSymbol: string): BigNumber {
     return CHAIN_ID_LIST_INDICES.reduce((sum, chainId) => {
--- a/src/UBAFeeCalculator/UBAFeeConfig.ts
+++ b/src/UBAFeeCalculator/UBAFeeConfig.ts
@@ -108,7 +108,7 @@ class UBAConfig {
   /**
    * @description Get the balance trigger threshold for a given chain and token
    * @param chainId The chain id
-   * @param tokenSymbol The token
+   * @param tokenSymbol The symbol of the token which will be resolved
    * @returns The balance trigger threshold if it exists
    */
   public getBalanceTriggerThreshold(chainId: number, tokenSymbol: string): ThresholdBoundType {
--- a/src/UBAFeeCalculator/UBAFeeConfig.ts
+++ b/src/UBAFeeCalculator/UBAFeeConfig.ts
@@ -1,4 +1,4 @@
-import { BigNumber } from "ethers";
+import { BigNumber, ethers } from "ethers";
 import { ThresholdBoundType, FlowTupleParameters } from "./UBAFeeTypes";
 import { CHAIN_ID_LIST_INDICES, HUBPOOL_CHAIN_ID } from "../constants";
 
@@ -125,7 +125,7 @@ class UBAConfig {
    */
   public getTargetBalance(chainId: number, tokenSymbol: string): BigNumber {
     const thresholdConfig = this.getBalanceTriggerThreshold(chainId, tokenSymbol);
-    return thresholdConfig?.upperBound?.target ?? BigNumber.from(0);
+    return thresholdConfig?.upperBound?.target ?? ethers.constants.Zero;
   }
 
   /**
@@ -135,7 +135,7 @@ class UBAConfig {
   public getTotalSpokeTargetBalanceForComputingLpFee(tokenSymbol: string): BigNumber {
     return CHAIN_ID_LIST_INDICES.filter((chainId) => chainId !== HUBPOOL_CHAIN_ID).reduce((sum, chainId) => {
       return sum.add(this.getTargetBalance(chainId, tokenSymbol));
-    }, BigNumber.from(0));
+    }, ethers.constants.Zero);
   }
 
   /**
@@ -144,7 +144,7 @@ class UBAConfig {
    * @returns The incentive pool adjustment. Defaults to 0 if not set
    */
   public getIncentivePoolAdjustment(chainId: string): BigNumber {
-    return this.incentivePoolAdjustment?.[chainId] ?? BigNumber.from(0); // Default to 0 if not set
+    return this.incentivePoolAdjustment?.[chainId] ?? ethers.constants.Zero; // Default to 0 if not set
   }
 
   /**
@@ -153,7 +153,7 @@ class UBAConfig {
    * @returns The UBA reward multiplier. Defaults to 1 if not set
    */
   public getUbaRewardMultiplier(chainId: string): BigNumber {
-    return this.ubaRewardMultiplier?.[chainId] ?? BigNumber.from(1); // Default to 1 if not set
+    return this.ubaRewardMultiplier?.[chainId] ?? ethers.constants.One; // Default to 1 if not set
   }
 }
 
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -2,7 +2,6 @@ import { BigNumber } from "ethers";
 import { MAX_SAFE_JS_INT } from "@uma/common/dist/Constants";
 import { fixedPointAdjustment, toBN } from "../utils";
 import { HUBPOOL_CHAIN_ID } from "../constants";
-import { UBAActionType } from "./UBAFeeTypes";
 
 /**
  * Computes a linear integral over a piecewise function
@@ -319,26 +318,13 @@ export function calculateUtilization(
 }
 
 export function calculateUtilizationBoundaries(
-  action: {
-    actionType: UBAActionType;
-    amount: BigNumber;
-    chainId: number;
-  },
   decimals: number,
   hubBalance: BigNumber,
   hubEquity: BigNumber,
   ethSpokeBalance: BigNumber,
-  spokeTargets: { target: BigNumber; spokeChainId: number }[],
-  hubPoolChainId = HUBPOOL_CHAIN_ID
+  newEthSpokeBalance: BigNumber,
+  spokeTargets: { target: BigNumber; spokeChainId: number }[]
 ): { utilizationPostTx: BigNumber; utilizationPreTx: BigNumber } {
-  let newEthSpokeBalance = ethSpokeBalance;
-  if (action.chainId === hubPoolChainId) {
-    if (action.actionType === UBAActionType.Deposit) {
-      newEthSpokeBalance = newEthSpokeBalance.add(action.amount);
-    } else {
-      newEthSpokeBalance = newEthSpokeBalance.sub(action.amount);
-    }
-  }
   return {
     utilizationPreTx: calculateUtilization(decimals, hubBalance, hubEquity, ethSpokeBalance, spokeTargets),
     utilizationPostTx: calculateUtilization(decimals, hubBalance, hubEquity, newEthSpokeBalance, spokeTargets),
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -1,7 +1,6 @@
 import { BigNumber } from "ethers";
 import { MAX_SAFE_JS_INT } from "@uma/common/dist/Constants";
 import { fixedPointAdjustment, toBN } from "../utils";
-import { HUBPOOL_CHAIN_ID } from "../constants";
 
 /**
  * Computes a linear integral over a piecewise function
@@ -306,12 +305,9 @@ export function calculateUtilization(
   hubBalance: BigNumber,
   hubEquity: BigNumber,
   ethSpokeBalance: BigNumber,
-  spokeTargets: { target: BigNumber; spokeChainId: number }[],
-  hubPoolChainId = HUBPOOL_CHAIN_ID
+  cumulativeSpokeTargets: BigNumber
 ) {
-  const numerator = hubBalance
-    .add(ethSpokeBalance)
-    .add(spokeTargets.reduce((a, b) => (b.spokeChainId !== hubPoolChainId ? a.add(b.target) : a), BigNumber.from(0)));
+  const numerator = hubBalance.add(ethSpokeBalance).add(cumulativeSpokeTargets);
   const denominator = hubEquity;
   const result = numerator.mul(fixedPointAdjustment).div(denominator); // We need to multiply by 1e18 to get the correct precision for the result
   return BigNumber.from(10).pow(decimals).sub(result);
@@ -323,11 +319,17 @@ export function calculateUtilizationBoundaries(
   hubEquity: BigNumber,
   ethSpokeBalance: BigNumber,
   newEthSpokeBalance: BigNumber,
-  spokeTargets: { target: BigNumber; spokeChainId: number }[]
+  cumulativeSpokeTargets: BigNumber
 ): { utilizationPostTx: BigNumber; utilizationPreTx: BigNumber } {
   return {
-    utilizationPreTx: calculateUtilization(decimals, hubBalance, hubEquity, ethSpokeBalance, spokeTargets),
-    utilizationPostTx: calculateUtilization(decimals, hubBalance, hubEquity, newEthSpokeBalance, spokeTargets),
+    utilizationPreTx: calculateUtilization(decimals, hubBalance, hubEquity, ethSpokeBalance, cumulativeSpokeTargets),
+    utilizationPostTx: calculateUtilization(
+      decimals,
+      hubBalance,
+      hubEquity,
+      newEthSpokeBalance,
+      cumulativeSpokeTargets
+    ),
   };
 }
 
--- a/src/UBAFeeCalculator/UBAFeeUtility.ts
+++ b/src/UBAFeeCalculator/UBAFeeUtility.ts
@@ -290,44 +290,6 @@ export function computePiecewiseLinearFunction(
   return integral.mul(scale);
 }
 
-export function calculateUtilization(
-  decimals: number,
-  hubBalance: BigNumber,
-  hubLiquidReserves: BigNumber,
-  ethSpokeDelta: BigNumber,
-  cumulativeSpokeTargets: BigNumber
-) {
-  const numerator = hubLiquidReserves.add(ethSpokeDelta).add(cumulativeSpokeTargets);
-  const denominator = hubBalance;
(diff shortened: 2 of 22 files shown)


src/constants.ts
```diff
@@ -2,3 +2,10 @@ export { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/contracts-v2/dist
 export const ZERO_ADDRESS = "0x0000000000000000000000000000000000000000";
 
 export const HUBPOOL_CHAIN_ID = 1;
+
+// This array contains all chains that Across supports, although some of the chains could be currently disabled.
+// The order of the chains is important to not change, as the dataworker proposes "bundle block numbers" per chain
+// in the same order as the following list. To add a new chain ID, append it to the end of the list. Never delete
+// a chain ID. The on-chain ConfigStore should store a list of enabled/disabled chain ID's that are a subset
+// of this list, so this list is simply the list of all possible Chain ID's that Across could support.
```
james-a-morris, line 10: I think we can use TypeDocs here

```suggestion
/**
 * This array contains all chains that Across supports, although some of the chains could be currently disabled.
 * The order of the chains is important to not change, as the dataworker proposes "bundle block numbers" per chain
 * in the same order as the following list. To add a new chain ID, append it to the end of the list. Never delete
 * a chain ID. The on-chain ConfigStore should store a list of enabled/disabled chain ID's that are a subset
 * of this list, so this list is simply the list of all possible Chain ID's that Across could support.
**/
```

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -235,23 +301,21 @@ export async function updateUBAClient(
...
-          startingBlock,
-          chainId
-        );
-        // Push the structure to the start of the list
-        bundleBounds.unshift({
-          start: startingBlock,
-          end: lastPreviousBlock,
-        });
-        startingBlock = lastPreviousBlock - 1;
-      }
```
james-a-morris, line 254: Nit: this logic trims the bundle blocks so that we only have N of them in memory. Should we refactor the code to include it in a different location

james-a-morris (commented):
Left a few nits. Looks great

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: What do you propose? I thought the goal was to have only `maxBundleStates` bundles total?

I'm also curious why you use `.unshift` instead of `.push`?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -14,21 +14,19 @@ import {
...
-import { RelayFeeCalculator, RelayFeeCalculatorConfigWithMap } from "../../relayFeeCalculator";
-import { TOKEN_SYMBOLS_MAP } from "../../constants";
+import { CHAIN_ID_LIST_INDICES, TOKEN_SYMBOLS_MAP } from "../../constants";
 import { getDepositFee, getRefundFee } from "../../UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog";
 import { AcrossConfigStoreClient } from "../AcrossConfigStoreClient";
+import {
+  blockRangesAreInvalidForSpokeClients,
+  getBlockRangeForChain,
+  getImpliedBundleBlockRanges,
+} from "../../utils/BundleUtils";
```
james-a-morris, line 29: Good refactor

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -53,33 +53,33 @@ export async function computeLpFeeForRefresh(
   gammaCutoff: FlowTupleParameters,
   blockNumber?: number
 ): Promise<BigNumber> {
-  const configStoreClient = hubPoolClient.configStoreClient;
   const erc20 = ERC20__factory.connect(hubPoolTokenAddress, hubPoolClient.hubPool.provider);
-  const [decimals, ethSpokeBalance, hubBalance, hubEquity, spokeTargets] = await Promise.all([
-    erc20.decimals({ blockTag: blockNumber }),
+  // Grab the balances of the spoke pool and hub pool at the given block number.
```
james-a-morris, line 57: This would save us 1 RPC call, but it shouldn't cause that much overhead as the results are being run in parallel. This adds the onus of resolving the decimals on the calling function. Could we instead use a lookup from the  `TOKEN_LOOKUP` class to avoid having to pass decimals into this function?

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -15,6 +14,10 @@ import { computeLpFeeStateful } from "./UBAClientUtilities";
 import { findLast } from "../../utils/ArrayUtils";
 import { analog } from "../../UBAFeeCalculator";
 import { BaseAbstractClient } from "../BaseAbstractClient";
+import { HubPoolClient } from "../HubPoolClient";
+import { SpokePoolClients } from "../../utils";
+import { ERC20__factory } from "../../typechain";
+import { TOKEN_SYMBOLS_MAP } from "@across-protocol/contracts-v2";
```
james-a-morris, line 20: The goal of this class was to not use any RPC endpoint calls and rely entirely on a state object for the Vercel API. Do you recommend that we keep the original method or should we deviate?

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -89,6 +90,26 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+   * @returns The UBA config for the given L1 token address and block number, or undefined if no config exists
+   * before blockNumber.
+   */
+  getUBAConfig(l1TokenAddress: string, blockNumber = Number.MAX_SAFE_INTEGER): UBAParsedConfigType | undefined {
+    const config = findLast(
+      this.ubaConfigUpdates,
+      (config) => config.l1Token === l1TokenAddress && config.blockNumber <= blockNumber
+    );
+    return config?.config;
+  }
```
james-a-morris, line 107: Should the UBA specific functions reside in the UBAClient utilities?

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -246,78 +269,37 @@ export class BaseUBAClient extends BaseAbstractClient {
...
-    overrides?: UBALPFeeOverride
-  ): Record<number, RelayerFeeResult & SystemFeeResult> {
-    const transformation = refundChainCandidates.map((refundChain) => {
-      return [
-        refundChain,
-        this.getUBAFee(depositChain, refundChain, tokenSymbol, amount, hubPoolBlockNumber, overrides),
-      ] as [number, RelayerFeeResult & SystemFeeResult];
-    });
-    return Object.fromEntries(transformation.filter(([, result]) => !result.amountTooLow));
-  }
```
james-a-morris, line 319: Is this no longer needed?

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -246,78 +269,37 @@ export class BaseUBAClient extends BaseAbstractClient {
...
-  ): { relayerBalancingFee: BigNumber } & SystemFeeResult {
-    const [relayerFee, systemFee] = [
-      this.computeBalancingFee(tokenSymbol, amount, hubPoolBlockNumber, refundChain, UBAActionType.Refund),
-      this.computeSystemFee(depositChain, refundChain, tokenSymbol, amount, hubPoolBlockNumber, overrides),
-    ];
-    return {
-      relayerBalancingFee: relayerFee.balancingFee,
-      ...systemFee,
-    };
-  }
```
james-a-morris, line 291: Do we no longer need the Relay Balancing Fee?

james-a-morris (commented):
Initial comments

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
nicholaspai: I don't see this as a UBA specific function, rather I see it as a function to fetch and parse the on chain `TokenConfig` event emitted by the `AcrossConfigStore` and return it as a usable object. the `UBAFeeConfig` then takes the result of this `ConfigStoreClient.getUBAConfig` call and shapes it further as an opinionated `UBAFeeConfig` that can be used by the `UBAClient`

src/clients/UBAClient/UBAClientBase.ts
nicholaspai: I didn't see how this would be used, how would it? I also don't see a point to the `getUBAFee` function. Why not just call `computeBalancingFee ` combined with  `computeSystemFee`?

src/clients/UBAClient/UBAClientBase.ts
nicholaspai: The Relayer doesn't use this. The relayer calls `computeSystemFee` and sets `realizedLpFee` using that value.

The Vercel API will call both `computeSystemFee` and `computeBalancingFee` to quote the `realizedLpFee` and the `relayerFeePct` respectively. Perhaps @dohaki might find this useful but I think the extra function adds more burden on maintainers and I found it difficult keeping this function up to date with all the internal changes I was making to `computeSystemFee` and `computeBalancingFee` so I've removed it from now.

Happy to add it once we have those internal functions more stable and finalized

src/clients/UBAClient/UBAClientBase.ts
nicholaspai: I see what you're saying. So original issue I'm trying to fix with `computeLpFee` is that you're using the incorrect `ethSpokeBalance`, `hubBalance`, and `hubEquity`. Those values need to be linked to the passed in `hubPoolBlockNumber`, not the values associated with the `bundleEndBlock` of the latest bundle state returned by `retrieveLastBundleState`.

So one way to make this stateless is to require the caller to pass in `ethSpokeBalance`, `hubBalance`, and `hubEquity`.

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: This is a good call. Here's a commit where I load the `decimals` directly from a mapping from the `HubPoolClient`. This results in `decimals` being passed around less from function to function. It also means we can make the `UBABundleState` class simpler I think. 68356887bee75beda8a219929db2fe3e3028e089

src/UBAFeeCalculator/UBAFeeConfig.ts
```diff
@@ -107,14 +108,36 @@ class UBAConfig {
   /**
    * @description Get the balance trigger threshold for a given chain and token
    * @param chainId The chain id
-   * @param tokenSymbol The token address
+   * @param l1TokenAddress The token address
```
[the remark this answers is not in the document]
nicholaspai: Good call out, reverted here e3b199d22ee03c7756543b7f50ce099ab7c54fb1

src/UBAFeeCalculator/UBAFeeConfig.ts
```diff
@@ -107,14 +108,36 @@ class UBAConfig {
...
+  /**
+   * Arbitrarily return upper bound target. This could be the average of the upper and lower bound targets but
+   * for now return upper bound target.
+   * @param chainId
+   * @param l1TokenAddress
+   * @returns
+   */
+  public getTargetBalance(chainId: number, l1TokenAddress: string): BigNumber {
+    const thresholdConfig = this.getBalanceTriggerThreshold(chainId, l1TokenAddress);
+    return thresholdConfig?.upperBound?.target ?? BigNumber.from(0);
```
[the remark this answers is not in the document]
nicholaspai: oh nice find!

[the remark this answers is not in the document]
nicholaspai: 8a491a3f766ffe215f0ffd02e1e5e6b06ec670d3

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: SGTM

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: SGTM

src/clients/UBAClient/UBAClientBase.ts
james-a-morris: SGTM!

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -295,7 +321,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
         const transferThresholdForToken = parsedValue.transferThreshold;
 
         // If Token config doesn't contain all expected properties, skip it.
-        if (!(rateModelForToken && transferThresholdForToken)) {
+        if (!this.isUBA() && !(rateModelForToken && transferThresholdForToken)) {
```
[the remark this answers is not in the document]
nicholaspai: hmm good point it should point to version applicable at time of update

src/clients/UBAClient/UBAClientUtilities.ts
pxrl: This could alternatively have been done via the HubPoolClient itself - it has the complete list of L1 tokens so it's possible to find the decimals and address with an `Array.find` that searches on the token symbol. But TOKEN_SYMBOLS_MAP is also fine.

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -197,47 +200,67 @@ export class BaseUBAClient extends BaseAbstractClient {
...
+  protected async computeLpFee(
+    hubPoolBlockNumber: number,
     amount: BigNumber,
     depositChainId: number,
+    refundChainId: number,
     hubPoolChainId: number,
     tokenSymbol: string,
-    refundChainId?: number,
+    hubPoolClient: HubPoolClient,
+    spokePoolClients: SpokePoolClients,
```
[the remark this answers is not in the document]
nicholaspai: Fixed this

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -197,47 +200,67 @@ export class BaseUBAClient extends BaseAbstractClient {
...
+      // TODO: Fix this by looking up the token address from the token symbol at the time of the hubPoolBlockNumber
+      const tokenMappingLookup = (
+        TOKEN_SYMBOLS_MAP as Record<string, { addresses: { [x: number]: string }; decimals: number }>
+      )[tokenSymbol];
+      const hubPoolTokenAddress = tokenMappingLookup.addresses[hubPoolClient.chainId];
+      const erc20 = ERC20__factory.connect(hubPoolTokenAddress, hubPoolClient.hubPool.provider);
+      const [ethSpokeBalance, hubBalance, hubEquity] = await Promise.all([
+        erc20.balanceOf(spokePoolClients[hubPoolClient.chainId].spokePool.address, { blockTag: hubPoolBlockNumber }),
+        erc20.balanceOf(hubPoolClient.hubPool.address, { blockTag: hubPoolBlockNumber }),
+        erc20.balanceOf(hubPoolClient.hubPool.address, { blockTag: hubPoolBlockNumber }),
```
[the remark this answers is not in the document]
nicholaspai: Nope good catch, will fix

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -246,78 +269,37 @@ export class BaseUBAClient extends BaseAbstractClient {
...
-      this.computeLpFee(amount, depositChainId, destinationChainId, tokenSymbol, undefined, overrides),
-      this.computeBalancingFee(tokenSymbol, amount, hubPoolBlockNumber, depositChainId, UBAActionType.Deposit),
-    ];
+  ): Promise<SystemFeeResult> {
+    const lpFee = await this.computeLpFee(
+      hubPoolBlockNumber,
+      amount,
+      depositChainId,
+      destinationChainId,
+      hubPoolClient.chainId,
```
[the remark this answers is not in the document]
nicholaspai: good call

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -292,56 +290,41 @@ export function computePiecewiseLinearFunction(
...
- * Computes the utilization at a given point in time based on the
- * current balances and equity of the hub and spoke pool targets.
- * @param decimals The number of decimals for the token
- * @param hubBalance The current balance of the hub pool for the token
- * @param hubEquity The current equity of the hub pool for the token
- * @param ethSpokeBalance The current balance of the ETH spoke pool for the token
- * @param targetSpoke The current balance of the target spoke pool for the token - this is a list.
- * @returns The utilization of the hub pool
- */
 export function calculateUtilization(
```
james-a-morris, line 293:
```suggestion
/**
 * Computes the utilization at a given point in time based on the
 * current balances and equity of the hub and spoke pool targets.
 * @param decimals The number of decimals for the token
 * @param hubBalance The current balance of the hub pool for the token
 * @param hubLiquidReserves The current equity of the hub pool for the token
 * @param ethSpokeDelta The current balance of the ETH spoke pool for the token
 * @param cumulativeSpokeTargets The sum of all spoke targets for a given token
 * @returns The utilization of the hub pool
 */
export function calculateUtilization(
```

src/UBAFeeCalculator/UBAFeeUtility.ts
```diff
@@ -292,56 +290,41 @@ export function computePiecewiseLinearFunction(
...
+  hubLiquidReserves: BigNumber,
+  ethSpokeDelta: BigNumber,
+  cumulativeSpokeTargets: BigNumber
 ) {
-  const numerator = hubBalance
-    .add(ethSpokeBalance)
-    .add(spokeTargets.reduce((a, b) => (b.spokeChainId !== hubPoolChainId ? a.add(b.target) : a), BigNumber.from(0)));
-  const denominator = hubEquity;
+  const numerator = hubLiquidReserves.add(ethSpokeDelta).add(cumulativeSpokeTargets);
+  const denominator = hubBalance;
```
james-a-morris, line 301: Would we be susceptible to the same vector as we were discussing in the call today? The case where a user sends funds to the hub pool contract directly

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -460,29 +486,4 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
-    });
-  }
-
-  /**
-   * Retrieves a UBA config for a given L1 token address and block number.
-   * @param l1TokenAddress The L1 token address to retrieve the config for
-   * @param blockNumber The block number to retrieve the config for. If not provided, the latest config will be returned.
-   * @returns The UBA config for the given L1 token address and block number, or undefined if no config exists.
-   */
-  getUBAConfig(l1TokenAddress: string, blockNumber?: number): UBAParsedConfigType | undefined {
```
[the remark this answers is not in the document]
james-a-morris: I agree. This is something I've seen as well

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -461,27 +463,23 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
    * @param l1TokenAddress The L1 token address to retrieve the config for
-   * @param blockNumber The block number to retrieve the config for. If not provided, the latest config will be returned.
-   * @returns The UBA config for the given L1 token address and block number, or undefined if no config exists.
+   * @param blockNumber The block number to retrieve the config for. If not specified, sets block to max integer
+   * meaning that this function will return the latest config.
+   * @returns The UBA config for the given L1 token address and block number, or undefined if no config exists
+   * before blockNumber.
    */
-  getUBAConfig(l1TokenAddress: string, blockNumber?: number): UBAParsedConfigType | undefined {
+  getUBAConfig(l1TokenAddress: string, blockNumber = Number.MAX_SAFE_INTEGER): UBAParsedConfigType | undefined {
```
james-a-morris, line 474: Nit should we make this a public function and explicitly set the access modifier?
```suggestion
  public getUBAConfig(l1TokenAddress: string, blockNumber = Number.MAX_SAFE_INTEGER): UBAParsedConfigType | undefined {
```

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -461,27 +463,23 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+    // We only care about searching on the block number and not any events that occurred in the same block
+    // but with a lower transaction index. In other words, if the UBA config was updated as the absolute
+    // last transaction in a block, the update still applies to all preceding UBA events in the same block.
+    // This is a simplifying assumption that we can make because the ConfigStore admin role is whitelisted and assumed
+    // to be acting in the best interest of the protocol.
     const config = findLast(
       this.ubaConfigUpdates,
-      (config) => config.l1Token === l1TokenAddress && (!blockNumber || config.blockNumber <= blockNumber)
+      (config) => config.l1Token === l1TokenAddress && config.blockNumber <= blockNumber
```
james-a-morris, line 482: Fair refactor!

src/clients/HubPoolClient.ts
```diff
@@ -500,6 +500,10 @@ export class HubPoolClient extends BaseAbstractClient {
       }
     ) as ExecutedRootBundle;
 
+    return this.getRunningBalanceForToken(l1Token, executedRootBundle);
+  }
+
+  getRunningBalanceForToken(l1Token: string, executedRootBundle: ExecutedRootBundle): TokenRunningBalance {
```
james-a-morris, line 506: Nit: should we make this public

```suggestion
  public getRunningBalanceForToken(l1Token: string, executedRootBundle: ExecutedRootBundle): TokenRunningBalance {
```

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -64,36 +65,42 @@ export class BaseUBAClient extends BaseAbstractClient {
     return this.retrieveBundleStates(chainId, tokenSymbol).at(-1);
   }
 
+  /**
+   * Returns the most recent bundle state for a chain and token that was created before a given block number.
+   * @param hubPoolBlockNumber The bundle state was proposed at or before this block
+   * @param chainId
+   * @param tokenSymbol
+   * @returns
```
james-a-morris, line 73: Nit: 
```suggestion
   * @returns the most recent bundle state for a given chain and token combination prior to the given block number.
```

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -64,36 +65,42 @@ export class BaseUBAClient extends BaseAbstractClient {
...
+   */
+  public retrieveBundleStateForBlock(
+    hubPoolBlockNumber: number,
+    chainId: number,
+    tokenSymbol: string
+  ): UBABundleState | undefined {
+    const sortedDescendingBundleStates = [
+      ...this.retrieveBundleStates(chainId, tokenSymbol).sort(
```
james-a-morris, line 81: QQ: this won't deep copy the bundle states. Is there a reason we're using the spread operator to re-create a new array of bundle states?

src/clients/UBAClient/UBAClientBase.ts
james-a-morris: Fair enough!

src/clients/UBAClient/UBAClientTypes.ts
```diff
@@ -37,39 +38,14 @@ export type ModifiedUBAFlow = {
...
-    target: BigNumber;
-  }[];
-};
-
 export type UBABundleState = {
   openingBalance: BigNumber;
   openingIncentiveBalance: BigNumber;
   openingBlockNumberForSpokeChain: number;
-  config: UBABundleConfig;
+  config: UBAConfig;
```
james-a-morris, line 45: Good refactor, @nicholaspai

james-a-morris (commented):
Comments on all non-ubaclientutilities functions. Will look into the UBAClientUtilities next.

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -313,114 +385,141 @@ export async function updateUBAClient(
...
+                    // @dev Assume that flow is taking refund on destination chain ID for purposes of computing the LP fee.
+                    // This is encoded in the UMIP as the way to compute utilization.
+                    flow.originChainId,
+                    flow.destinationChainId,
+                    flow.amount,
+                    hubPoolClient,
+                    constructedBundle.config.getBaselineFee(flow.destinationChainId, flow.originChainId),
+                    constructedBundle.config.getLpGammaFunctionTuples(flow.destinationChainId),
+                    flow.quoteBlockNumber
+                  );
```
pxrl, line 486: I'm unclear on why we compute the deposit fee, refund fee and LP fee for each flow. Is it just for simplicity, such that some of them will resolve to 0 because that type of fee is not relevant for the specific flow event?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -461,35 +560,56 @@ async function getFlows(
...
+          const matchingDeposit = await resolveCorrespondingDepositForFill(fill, spokePoolClients);
+          if (matchingDeposit === undefined) {
+            return undefined;
+          } else {
+            return {
+              ...fill,
+              quoteBlockNumber: matchingDeposit.quoteBlockNumber,
+            };
+          }
+        }
```
pxrl, line 576: The ` } else {` statements here are redundant, and the code squashes down nicely if they're removed.

```suggestion
        if (!validWithinBounds) {
          return undefined;
        }

        const matchingDeposit = await resolveCorrespondingDepositForFill(fill, spokePoolClients);
        if (matchingDeposit === undefined) {
          return undefined;
        }

        return { ...fill, quoteBlockNumber: matchingDeposit.quoteBlockNumber };
```

src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -47,14 +46,64 @@ export class UBAClientWithRefresh extends BaseUBAClient {
...
+   * @param amount The amount that is being deposited
+   * @param hubPoolClient A hubpool client instance to query the hubpool
+   * @param spokePoolClients A mapping of spoke chainIds to spoke pool clients
+   * @returns The realized LP fee for the given token on the given chainId at the given block number
+   */
+  public async computeLpFee(
+    hubPoolBlockNumber: number,
+    amount: BigNumber,
+    depositChainId: number,
+    refundChainId: number,
```
pxrl, line 69: Do we want to name it `refundChainId` ? I think it's dangerous for third-party integrators who might end up computing invalid LP fees when filling relays.

src/constants.ts
```diff
@@ -2,3 +2,12 @@ export { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/contracts-v2/dist
...
 export const HUBPOOL_CHAIN_ID = 1;
+
+/**
+ * This array contains all chains that Across supports, although some of the chains could be currently disabled.
+ * The order of the chains is important to not change, as the dataworker proposes "bundle block numbers" per chain
+ * in the same order as the following list. To add a new chain ID, append it to the end of the list. Never delete
+ * a chain ID. The on-chain ConfigStore should store a list of enabled/disabled chain ID's that are a subset
+ * of this list, so this list is simply the list of all possible Chain ID's that Across could support.
+ **/
+export const CHAIN_ID_LIST_INDICES = [1, 10, 137, 288, 42161];
```
pxrl, line 13: For post-UBA launch, should we consider pushing this to the config store?

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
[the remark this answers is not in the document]
nicholaspai: Done

src/UBAFeeCalculator/UBAFeeConfig.ts
```diff
@@ -107,21 +108,43 @@ class UBAConfig {
...
+   * @param l1TokenAddress
+   * @returns
+   */
+  public getTargetBalance(chainId: number, tokenSymbol: string): BigNumber {
+    const thresholdConfig = this.getBalanceTriggerThreshold(chainId, tokenSymbol);
+    return thresholdConfig?.upperBound?.target ?? ethers.constants.Zero;
+  }
+
+  /**
+   * Get sum of all spoke target balances for all chains besides hub pool chain for l1TokenAddress.
```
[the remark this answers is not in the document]
nicholaspai: nah it wasn't at the time, but we're not using this function anymore

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -162,7 +195,7 @@ export function getUBAFeeConfig(
 
   const threshold = ubaConfig.rebalance[String(chainId)];
 
-  const chainTokenCombination = `${chainId}-${token}`;
+  const chainTokenCombination = `${chainId}-${l1TokenAddress}`;
```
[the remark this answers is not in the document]
nicholaspai: good bug catch!

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -205,106 +238,145 @@ export function getUBAFeeConfig(
...
+ * @param spokePoolClients
+ * @param relevantChainIds
+ * @param relevantTokenSymbols
+ * @param latestHubPoolBlockNumber
+ * @param updateInternalClients
+ * @param relayFeeCalculatorConfig
+ * @param maxBundleStates
+ * @returns
+ */
 export async function updateUBAClient(
```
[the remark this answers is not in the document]
nicholaspai: I will leave that for follow up for @james-a-morris !

src/constants.ts
nicholaspai: yea

james-a-morris (approved):

PR #317, opened 2023-07-18
feat: Relocate computeSystemFee and refactor getModifiedFlows

nicholaspai:
- `computeLpFee` shouldn't try to retrieve a bundle config from state since its expected to be used by the relayer. If the relayer is trying to set the LP fee for a very recent deposit, then it should not be using the config included at the time of the last validated bundle, instead it should be using the latest config in the `ConfigStore` contract
- `getModifiedFlows` is refactored and now is used in `resolveHistoricalDepositForFill` to reduce the DOS vector
- The only remaining DOS vector is looking up an old fill for a refund. I've left a TODO

Fixes #ACX1321

### DIFF
--- a/src/clients/UBAClient/UBAClientUtilities.ts
+++ b/src/clients/UBAClient/UBAClientUtilities.ts
@@ -760,6 +760,7 @@ async function getFlows(
   // - Fills that request refunds on a different chain.
   // - Subsequent fills after an initial partial fill.
   // - Slow fills.
+  // - Fills that are not complete fills.
   // - Fills that are considered "invalid" by the spoke pool client.
   const fills: UbaFlow[] = (
     await getValidFillCandidates(
@@ -771,6 +772,7 @@ async function getFlows(
         toBlock,
         repaymentChainId: chainId,
         isSlowRelay: false,
+        isCompleteFill: true,
       },
       ["realizedLpFeePct"]
     )
@@ -913,6 +915,7 @@ export type SpokePoolFillFilter = {
   toBlock?: number;
   repaymentChainId?: number;
   isSlowRelay?: boolean;
+  isCompleteFill?: boolean;
 };
 
 /**
@@ -933,7 +936,7 @@ export async function getValidFillCandidates(
   const spokePoolClient = spokePoolClients[chainId];
   assert(isDefined(spokePoolClient));
 
-  const { repaymentChainId, relayer, isSlowRelay, fromBlock, toBlock } = filter;
+  const { repaymentChainId, relayer, isSlowRelay, isCompleteFill, fromBlock, toBlock } = filter;
 
   const fills = (
     await mapAsync(spokePoolClient.getFills(), async (fill) => {
@@ -950,6 +953,10 @@ export async function getValidFillCandidates(
         (isDefined(relayer) && fill.relayer !== relayer) ||
         (isDefined(isSlowRelay) && fill.updatableRelayData.isSlowRelay !== isSlowRelay)
       ) {
+        return undefined;
+      }
+
+      if (isDefined(isCompleteFill) && isCompleteFill !== fill.fillAmount.eq(fill.totalFilledAmount)) {
         return undefined;
       }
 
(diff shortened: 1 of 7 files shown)


src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -835,39 +834,43 @@ export type SpokePoolFillFilter = SpokePoolEventFilter & {
...
-    if (
-      (isDefined(originChainId) && fill.originChainId !== originChainId) ||
-      (isDefined(repaymentChainId) && fill.repaymentChainId !== repaymentChainId) ||
-      (isDefined(relayer) && fill.relayer !== relayer) ||
-      (isDefined(isSlowRelay) && fill.updatableRelayData.isSlowRelay !== isSlowRelay)
-    ) {
-      return false;
-    }
+  const fills = (
+    await mapAsync(spokePoolClient.getFills(), async (fill) => {
```
pxrl, line 847: What's the reason for switching to use `mapAsync` here insteead of `filterAsync` ? As far as I can see, `filterAsync` does what we want and doesn't require the subsequent loop to filter out all the `undefined` instances.

pxrl: Ah - I see, it's because `quoteBlockNumber`  is being added to each fill.

nicholaspai: yep

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -297,6 +293,27 @@ export function getMostRecentBundleBlockRanges(
...
+export async function resolveCorrespondingDepositForFill(
+  fill: FillWithBlock,
+  spokePoolClients: SpokePoolClients,
+  hubPoolClient: HubPoolClient
+): Promise<DepositWithBlock | undefined> {
+  // Matched deposit for fill was not found in spoke client. This situation should be rare so let's
+  // send some extra RPC requests to blocks older than the spoke client's initial event search config
+  // to find the deposit if it exists.
+  return queryHistoricalDepositForFill(hubPoolClient, spokePoolClients, fill);
+}
```
james-a-morris, line 315: Is this currently just a wrapper over `resolveCorrespondingDepositForFill`

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -672,24 +748,17 @@ async function getFlows(
...
+      (
+        await getFills(chainId, hubPoolClient, spokePoolClients, {
+          fromBlock,
+          toBlock,
+          repaymentChainId: chainId,
+          isSlowRelay: false,
+        })
+      ).filter((fill: FillWithBlock) => {
+        // We only want to include full fills as flows. Partial fills need to request refunds and those refunds
+        // will be included as flows.
```
james-a-morris, line 760: Could we use `filterAsync`

james-a-morris (commented):
Left comments

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: yeah basically

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: this is a `sync` filter actually. The `array` is `await getFills` and the filter function is syncrhonous

james-a-morris (approved):
LGTM - pending @pxrl signoff and the resolution of my outstanding comment

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -951,48 +923,36 @@ export async function getFills(
...
-      (isDefined(destinationChainId) && refundRequest.destinationChainId !== destinationChainId) ||
-      (isDefined(relayer) && refundRequest.relayer !== relayer)
-    ) {
-      return false;
-    }
-
-    const result = await refundRequestIsValid(spokePoolClients, hubPoolClient, refundRequest);
-
-    return result.valid;
-  });
```
pxrl: Can we retain this filtering? It's optional to use, and the relayer relies on filtering by `relayer` to find its existing refund requests.

nicholaspai: sure

PR #318, opened 2023-07-18
fix(updateUBAClient): Add defaults to loading bundle block ranges and opening balances

nicholaspai:
Handles case where there are no validated proposals in memory before a block


james-a-morris (approved):

pxrl (approved):
OK from me :+1:

PR #320, opened 2023-07-18
chore: optimize update in ubaClient

james-a-morris:


src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -25,38 +25,26 @@ export class UBAClientWithRefresh extends BaseUBAClient {
...
-      await updateUBAClient(
-        this.hubPoolClient,
-        this.spokePoolClients,
-        this.chainIdIndices,
-        this.tokens,
-        forceClientRefresh,
-        this.maxBundleStates
-      )
-    );
+  public async update(state: UBAClientState, forceClientRefresh?: boolean): Promise<void> {
```
pxrl, line 28:
```suggestion
  public async update(state?: UBAClientState, forceClientRefresh?: boolean): Promise<void> {
```

pxrl, line 28: It's confusing to only use `forceClientRefresh` if `state` is undefined.

james-a-morris: Done https://github.com/across-protocol/sdk-v2/pull/320/commits/ee401eb44159ae3458ae43dfa986e6a2236a8996

james-a-morris: Followed up offline

pxrl: I'm skeptical about the UX of this, but you're right that it was pre-existing, so fair enough :+1:

pxrl (approved):

PR #324, opened 2023-07-18
fix: account for no bundle state available

james-a-morris:


pxrl:
Works for me!

src/clients/UBAClient/UBAClientBase.ts
```diff
@@ -129,6 +129,17 @@ export class BaseUBAClient extends BaseAbstractClient {
   ): BalancingFeeReturnType {
     // Opening balance for the balancing action blockNumber.
     const relevantBundleStates = this.retrieveBundleStates(chainId, tokenSymbol);
+
+    // If there are no bundle states for this token on this chain, then there is no fee.
+    // This is a case where the token is newly supported and there are both no flows
+    // yet and no root bundles that have been proposed.
+    if (relevantBundleStates.length === 0) {
```
nicholaspai, line 136: I don't think this is a good solution as this only works for the first flow after the bundle start block. (i.e. if we just deployed the spoke pool, only the first flow after that deployment)

I think a better solution is to make sure that the loading of the bundle state block ranges is done to cover any times before we had published bundles. Thoughts on [this](https://github.com/across-protocol/sdk-v2/pull/325/files#diff-23b9ca7566c8f516ceddd46d0cdc63f33562b3f9d55002c60564f691bdcdba82R269) solution?

It does require us to update the `ConfigStore` contract with a UBA version to "activate" the UBA

nicholaspai:
I think this might work if we change the balancing fees to only be based on the opening bundle running balance

james-a-morris:
No longer needed.

PR #325, opened 2023-07-18
feat: match events with correct UBA config based on bundle start block not quote block

nicholaspai:
In the UBA model, events should be matched to UBA config based on the bundle start block containing the event, rather than the event.quoteBlock

### DIFF
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -15,7 +15,6 @@ import {
   max,
   sortEventsAscending,
   findLast,
-  isUBA,
 } from "../../utils";
 import { Contract, BigNumber, Event } from "ethers";
 import winston from "winston";
@@ -234,32 +233,6 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     const version = this.getConfigStoreVersionForTimestamp(timestamp);
     return this.isValidConfigStoreVersion(version);
   }
-
-  /**
-   * Return first block number where UBA was activated by setting "Version" GlobalConfig in ConfigStore
-   * @returns
-   */
-  getUbaActivationBlock(): number {
-    const config = this.cumulativeConfigStoreVersionUpdates.find((config) => {
-      isUBA(Number(config.value));
-    });
-    if (config) {
-      return config.blockNumber;
-    } else {
-      return Number.MAX_SAFE_INTEGER;
-    }
-  }
-
-  public isUbaBlock(block: number): boolean {
-    const versionAppliedToDeposit = this.getConfigStoreVersionForBlock(block);
-    return isUBA(versionAppliedToDeposit);
-  }
-
-  public canSupportUbaBlock(block: number): boolean {
-    const versionAppliedToDeposit = this.getConfigStoreVersionForBlock(block);
-    return this.isValidConfigStoreVersion(versionAppliedToDeposit);
-  }
-
   isValidConfigStoreVersion(version: number): boolean {
     return this.configStoreVersion >= version;
   }
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -15,6 +15,7 @@ import {
   max,
   sortEventsAscending,
   findLast,
+  isUBA,
 } from "../../utils";
 import { Contract, BigNumber, Event } from "ethers";
 import winston from "winston";
@@ -232,6 +233,31 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
   hasValidConfigStoreVersionForTimestamp(timestamp: number = Number.MAX_SAFE_INTEGER): boolean {
     const version = this.getConfigStoreVersionForTimestamp(timestamp);
     return this.isValidConfigStoreVersion(version);
+  }
+
+  /**
+   * Return first block number where UBA was activated by setting "Version" GlobalConfig in ConfigStore
+   * @returns
+   */
+  getUbaActivationBlock(): number {
+    const config = this.cumulativeConfigStoreVersionUpdates.find((config) => {
+      isUBA(Number(config.value));
+    });
+    if (config) {
+      return config.blockNumber;
+    } else {
+      return Number.MAX_SAFE_INTEGER;
+    }
+  }
+
+  public isUbaBlock(block: number): boolean {
+    const versionAppliedToDeposit = this.getConfigStoreVersionForBlock(block);
+    return isUBA(versionAppliedToDeposit);
+  }
+
+  public canSupportUbaBlock(block: number): boolean {
+    const versionAppliedToDeposit = this.getConfigStoreVersionForBlock(block);
+    return this.isValidConfigStoreVersion(versionAppliedToDeposit);
   }
 
   isValidConfigStoreVersion(version: number): boolean {
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -20,7 +20,15 @@ import {
   paginatedEventQuery,
   toBN,
 } from "../utils";
-import { Deposit, L1Token, CancelledRootBundle, DisputedRootBundle, LpToken, TokenRunningBalance } from "../interfaces";
+import {
+  Deposit,
+  L1Token,
+  CancelledRootBundle,
+  DisputedRootBundle,
+  LpToken,
+  TokenRunningBalance,
+  DepositWithBlock,
+} from "../interfaces";
 import { ExecutedRootBundle, PendingRootBundle, ProposedRootBundle } from "../interfaces";
 import { CrossChainContractsSet, DestinationTokenWithBlock, SetPoolRebalanceRoot } from "../interfaces";
 import * as lpFeeCalculator from "../lpFeeCalculator";
@@ -253,13 +261,10 @@ export class HubPoolClient extends BaseAbstractClient {
   }
 
   async computeRealizedLpFeePct(
-    deposit: {
-      quoteTimestamp: number;
-      amount: BigNumber;
-      destinationChainId: number;
-      originChainId: number;
-      blockNumber: number;
-    },
+    deposit: Pick<
+      DepositWithBlock,
+      "quoteTimestamp" | "amount" | "destinationChainId" | "originChainId" | "blockNumber"
+    >,
     l1Token: string
   ): Promise<{ realizedLpFeePct: BigNumber | undefined; quoteBlock: number }> {
     const quoteBlock = await this.getBlockNumber(deposit.quoteTimestamp);
@@ -754,7 +759,7 @@ export class HubPoolClient extends BaseAbstractClient {
       const nTokens = l1Tokens.length;
 
       // Safeguard
-      if (![nTokens, nTokens*2].includes(runningBalances.length)) {
+      if (![nTokens, nTokens * 2].includes(runningBalances.length)) {
         throw new Error(
           `Invalid runningBalances length: ${runningBalances.length}. Expected ${nTokens} or ${nTokens * 2} for chain ${
             this.chainId
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -733,24 +733,6 @@ export class HubPoolClient extends BaseAbstractClient {
         continue;
       }
 
-      // The applicable version is determined by the block number of the corresponding proposal.
-      let proposalBlockNumber = event.blockNumber;
-      for (let idx = this.proposedRootBundles.length - 1; idx >= 0; --idx) {
-        const rootBundleProposal = this.proposedRootBundles[idx];
-        if (event.blockNumber > rootBundleProposal.blockNumber) {
-          proposalBlockNumber = rootBundleProposal.blockNumber;
-          break;
-        }
-      }
-      if (proposalBlockNumber === event.blockNumber) {
-        this.logger.warn({
-          at: "HubPoolClient#update",
-          message: `Unable to find RootBundleProposal before blockNumber ${event.blockNumber}`,
-          executedRootBundle: event.transactionHash,
-        });
-        continue;
-      }
-
       // Set running balances and incentive balances for this bundle.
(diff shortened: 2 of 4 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -244,17 +254,32 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    const bundleStartBlockContainingDeposit = this.getBundleStartBlockContainingBlock(
+      deposit.blockNumber,
+      deposit.originChainId,
+      this.latestBlockNumber
+    );
+    const versionAppliedToDeposit = this.configStoreClient.getConfigStoreVersionForBlock(
+      bundleStartBlockContainingDeposit
+    );
+    if (isUBA(versionAppliedToDeposit)) {
+      if (!this.configStoreClient.isValidConfigStoreVersion(versionAppliedToDeposit)) {
```
james-a-morris, line 282: I see so we're essentially determining this on a per-flow basis? Good catch

src/clients/HubPoolClient.ts
```diff
@@ -306,6 +332,45 @@ export class HubPoolClient extends BaseAbstractClient {
...
+   * bundle contains the event
+   */
+  getBundleStartBlockContainingBlock(eventBlock: number, eventChain: number, hubPoolLatestBlock?: number): number {
+    // First find the latest executed bundle as of `hubPoolLatestBlock`.
+    const latestExecutedBundle = this.getNthFullyExecutedRootBundle(-1, hubPoolLatestBlock);
+
+    // If there is no latest executed bundle, then return the spoke's activation block. This means that there is no
+    // bundle before `hubPoolLatestBlock` containing the event block so the next bundle will start at
+    // the activation block and contain the event.
+    if (!latestExecutedBundle) {
```
james-a-morris, line 349: Should we be more explicit with these if statements? Not sure if it matters in practice, but technically `latestExecutedBundle = 0` will fail this conditional. We do this in other locations as well. We may want to employ the `isDefined(latestExecutedBundle)` type guard or just `latestExecutedBundle === undefined`

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -252,18 +253,23 @@ export function getMostRecentBundleBlockRanges(
...
 
   // Reconstruct bundle ranges based on published end blocks.
+
+  // Bundle states are examined in chronological descending order.
   const bundleData: { start: number; end: number; proposalBlock: number }[] = [];
   for (let i = 0; i < maxBundleStates; i++) {
     // Get the most recent bundle end range for this chain published in a bundle before `toBlock`.
     const latestExecutedRootBundle = hubPoolClient.getNthFullyExecutedRootBundle(-1, toBlock);
     if (!latestExecutedRootBundle) {
+      // TODO: I'm not sure the following logic is correct but I'm trying to handle the case where
```
james-a-morris, line 263: Should we generate a Linear task?

james-a-morris (commented):
Left some comments

src/clients/HubPoolClient.ts
nicholaspai: yea i think that is more correct I realized. If a deposit happened after the UBA activation block but is contained in a bundle who's start block is before the UBA activation block, then the deposit can't have a UBA fee applied to it because of the way we compute running balances

src/clients/HubPoolClient.ts
nicholaspai: `latestExecutedBundle` is a `ProposedRootBundle` object though

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: i removed this i think i fixed it

src/clients/HubPoolClient.ts
```diff
@@ -323,7 +381,7 @@ export class HubPoolClient extends BaseAbstractClient {
     latestMainnetBlock: number,
     block: number,
     chain: number,
-    chainIdList: number[]
+    chainIdList = CHAIN_ID_LIST_INDICES
   ): number | undefined {
     let endingBlockNumber: number | undefined;
```
pxrl, line 386: We should have the same list stored within the configStoreClient instance. If those arrays are not identical then we'd likely end up with undefined behaviour, so it should be both easier and safer to source the array from there.

```suggestion
  ): number | undefined {
    const chainIdList = this.configStoreClient.enabledChainIds;
    let endingBlockNumber: number | undefined;
```

src/clients/HubPoolClient.ts
```diff
@@ -703,7 +776,12 @@ export class HubPoolClient extends BaseAbstractClient {
       if (isUBA(version)) {
         // runningBalances array is a concatenation of pre-UBA runningBalances and incentiveBalances.
         executedRootBundle.runningBalances = runningBalances.slice(0, nTokens);
-        executedRootBundle.incentiveBalances = runningBalances.slice(nTokens);
+        // If bundle hasn't added incentive balance values, then assume they are 0. This is in place
+        // to make sure that the incentiveBalances isn't undefined.
+        executedRootBundle.incentiveBalances =
+          runningBalances.length > nTokens
+            ? runningBalances.slice(nTokens)
+            : Array(nTokens).fill(ethers.constants.Zero);
```
pxrl, line 784: This loosens the validation logic a lot and opens up for non-conformant runningBalances. The UMIP is very specific about runningBalances being 2x the length of l1Tokens, and I like that because it's easy to validate. Is it really necessary to change it?

2023-07-19-152131_1017x300_scrot

nicholaspai:
> Member

Good catch. I think the UMIP is wrong because it violates an assumption each event in a bundle uses the same UBA config: 2882fdcd55bf92923566bb0a88f1154b2c3f5a2a

src/clients/HubPoolClient.ts
nicholaspai: so I added this so that it can still validate pre UBA bundles. I'll remove it for now but I think we might run into problems on prod without this

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -234,6 +235,31 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+  public isUbaBlock(block: number): boolean {
+    const versionAppliedToDeposit = this.getConfigStoreVersionForBlock(block);
+    return isUBA(versionAppliedToDeposit);
+  }
+
+  public canSupportUbaBlock(block: number): boolean {
+    const versionAppliedToDeposit = this.getConfigStoreVersionForBlock(block);
+    return this.isValidConfigStoreVersion(versionAppliedToDeposit);
+  }
+
```
pxrl, line 262: Thoughts on relocating these to `UBAClientUtilities.ts`?

src/clients/HubPoolClient.ts
nicholaspai: Actually I think this can all be simpler. Its not correct as is because we need to compare the bundle start block, not the proposal block, against the config store version. I think its easier to just check here if running balances is 2x l1 tokens or 1x. We don't need to check version https://github.com/across-protocol/sdk-v2/commit/97d0b647fcdb51e82900ccdc79c0330dae318c77

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
nicholaspai: will do good call

src/clients/HubPoolClient.ts
pxrl: > Actually I think this can all be simpler. Its not correct as is because we need to compare the bundle start block, not the proposal block, against the config store version. I think its easier to just check here if running balances is 2x l1 tokens or 1x. We don't need to check version [97d0b64](https://github.com/across-protocol/sdk-v2/commit/97d0b647fcdb51e82900ccdc79c0330dae318c77)

Just to be sure on that, we previously agreed that it was important to verify that the format complies with the version (I agree with this):
https://github.com/across-protocol/sdk-v2/pull/213#discussion_r1197043597

On the applicable block number to determine the version, we also discussed this previously and I'm pretty confident that it will work correctly (forwards and backwards) by evaluating on the proposal block number. There are potentially some TBDs on this though - i.e. it's very much dependent on how the UBA migration is handled.
https://github.com/across-protocol/sdk-v2/pull/230#discussion_r1223131959

src/clients/HubPoolClient.ts
```diff
@@ -244,29 +253,36 @@ export class HubPoolClient extends BaseAbstractClient {
...
 
   async computeRealizedLpFeePct(
-    deposit: { quoteTimestamp: number; amount: BigNumber; destinationChainId: number; originChainId: number },
+    deposit: {
+      quoteTimestamp: number;
+      amount: BigNumber;
+      destinationChainId: number;
+      originChainId: number;
+      blockNumber: number;
+    },
```
james-a-morris, line 262: Should this be made into a type?

src/clients/HubPoolClient.ts
```diff
@@ -690,33 +746,24 @@ export class HubPoolClient extends BaseAbstractClient {
...
-        executedRootBundle.runningBalances = runningBalances.slice(0, nTokens);
-        executedRootBundle.incentiveBalances = runningBalances.slice(nTokens);
-      } else {
-        // Pre-UBA: runningBalances length is 1:1 with l1Tokens length. Pad incentiveBalances with zeroes.
-        executedRootBundle.incentiveBalances = runningBalances.map(() => toBN(0));
-      }
-
       // Safeguard
-      if (executedRootBundle.runningBalances.length !== nTokens) {
+      if (runningBalances.length !== nTokens || runningBalances.length !== nTokens * 2) {
```
james-a-morris, line 757:
```suggestion
      if (![nTokens, nTokens*2].includes(runningBalances.length)) {
```

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -281,10 +292,17 @@ export function getMostRecentBundleBlockRanges(
...
       );
     }
-    const blockRangeForChain = getBlockRangeForChain(rootBundleBlockRanges, chainId);
 
-    // Push the structure to the start of the list
-    bundleData.unshift({
+    // If UBA is not enabled for this bundle, exit early since no subsequent bundles will be enabled
+    // for the UBA, as the UBA was a non-reversible change.
+    const hubPoolStartBlock = getBlockRangeForChain(rootBundleBlockRanges, hubPoolClient.chainId)[0];
+    if (!isUbaBlock(hubPoolStartBlock, hubPoolClient.configStoreClient)) {
```
james-a-morris, line 299: Would we want to be checking for a specific spoke or is the hubpool block the one we're looking to verify?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -513,6 +531,38 @@ export async function updateUBAClient(
...
+ */
+export function getUbaActivationBlock(configStoreClient: AcrossConfigStoreClient): number {
+  const config = configStoreClient.cumulativeConfigStoreVersionUpdates.find((config) => {
+    isUBA(Number(config.value));
+  });
+  if (config) {
+    return config.blockNumber;
+  } else {
+    return Number.MAX_SAFE_INTEGER;
+  }
```
james-a-morris, line 558:
```suggestion
  return configStoreClient.cumulativeConfigStoreVersionUpdates.find((config) => {
    isUBA(Number(config.value));
  })?.blockNumber ?? Number.MAX_SAFE_INTEGER;

```

src/clients/HubPoolClient.ts
nicholaspai: Can we somehow tell Typescript that we only want to pluck these parts out of the `Deposit` type?

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: I think we need to always compare the hub pool start block against the config store blocks since they're both on mainnet

src/clients/HubPoolClient.ts
nicholaspai: what if the UBA VERSION is updated after the first UBA bundle start block but before the proposal block number?

Checking the version at the proposal block  would return 2

Checking the version at the bundle start block would return 1

The correct version for this bundle should be 1 right?

src/clients/HubPoolClient.ts
james-a-morris: Yep! We can use the `Pick<T, k typeof keyof T>`. It would be something like:

```
deposit: Pick<Deposit, "quoteTimestamp" | "amount" | "destinationChainId" | "blockNumber">
```

src/clients/HubPoolClient.ts
james-a-morris: Ah shoot - please disregard then. There's other locations in the code where we check the blocks which could be undefined or zero.

src/clients/HubPoolClient.ts
pxrl: > The correct version for this bundle should be 1 right?

Yeah, but that presupposes a specific upgrade path. My assumption was that we'd need to pause deposits, fill all outstanding deposits and then propose a bundle before bumping the version. If we did that and then re-enabled deposits then we should have a clean upgrade where the same version applies to both the opening block and the proposal block. It does however have the downside of being open to a dispute if the upgrade proceeds within the challenge window, though. So if we can get over the version hump without all that overhead then that's great :+1: 

The removal of the version guard still strikes me as a step backwards - I'd prefer it if we were strict about the length as a function of the applicable version (i.e. I think your feedback on this in May was definitely valid).

pxrl: Update - I've been convinced the other way out of band :+1: It's probably not necessary to validate the length based on the version number because the bundle has already cleared liveness, meaning that consensus is that it was correct.

src/clients/HubPoolClient.ts
```diff
@@ -690,33 +751,24 @@ export class HubPoolClient extends BaseAbstractClient {
         continue;
```
pxrl, line 751: I think you also want to delete the entire block of code above this line.

2023-07-19-201258_676x351_scrot

pxrl (approved):
LGTM

src/clients/HubPoolClient.ts
nicholaspai: I did add some `isDefined` though! good reminder

PR #329, opened 2023-07-19
docs: add typedoc

james-a-morris:


src/clients/UBAClient/UBAClientTypes.ts
```diff
@@ -50,3 +50,19 @@ export type UBABundleState = {
 export type UBAClientState = {
   [chainId: number]: UBAChainState;
 };
+
+export type SpokePoolEventFilter = {
```
nicholaspai, line 54: Not sure I agree with adding these back until we need them. For example, I don't see any of the relayer bots using the `originChainId` or `destinationChainId` filters currently

nicholaspai (approved):
LGTM but I don't see the reason for the change in `UBAClientTypes`

PR #337, opened 2023-07-25
refactor: Simplify MockConfigStoreClient

pxrl:
Instead of tracking two state variables, only instantiate EventManager if mockUpdate is true. This should also make it more difficult to accidentally misuse the class.


james-a-morris (approved):
Good refactor

PR #338, opened 2023-07-25
improve: Append block timestamp to SpokePool events

pxrl:
Add the block timestamp to FundsDeposited, FilledRelay and RefundRequested events. This allows upper layers to perform additional checks based on cross-chain timestamps.

This change is plucked from @nicholaspai's existing draft PR here: https://github.com/across-protocol/sdk-v2/pull/332/files#diff-8aaa64629d1940a0ea618d935e9d93044f09df5bd32d7352ee02c415a763d250

This change is also implicitly dependent on the following change in relayer-v2, because without it the relayer will experience a potentially significant performance decrease (and a potentially large increase in RPC usage): https://github.com/across-protocol/relayer-v2/pull/835

ACX-1328

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -727,11 +727,11 @@ export class SpokePoolClient extends BaseAbstractClient {
         });
       }
       await forEachAsync(fillEvents, async (event) => {
-        const fillData = spreadEventWithBlockNumber(event);
-        const fill = {
-          ...fillData,
-          blockTimestamp: (await this.spokePool.provider.getBlock(fillData.blockNumber)).timestamp,
-        } as FillWithBlock;
+        const rawFill = spreadEventWithBlockNumber(event) as FillWithBlock;
+        const fill: FillWithBlock = {
+          ...rawFill,
+          blockTimestamp: (await this.spokePool.provider.getBlock(rawFill.blockNumber)).timestamp,
+        };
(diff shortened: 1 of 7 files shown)


nicholaspai (approved):

pxrl:
@nicholaspai I had to modify this because it wasn't really compatible with the way the mocks work, and broke a few tests in relayer-v2. But we now get some benefit from filtering out duplicate blocks before we request them in parallel, so it has the advantage of fewer overheads.

nicholaspai (approved):
Nice changes

src/clients/SpokePoolClient.ts
```diff
@@ -751,12 +783,15 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-      // repaymentChainId is not part of the on-chain event, so add it here.
-      for (const refundRequest of refundRequests) {
-        this.refundRequests.push({
-          ...spreadEventWithBlockNumber(refundRequest),
+      for (const event of refundRequests) {
+        const rawRefundRequest = spreadEventWithBlockNumber(event) as RefundRequestWithBlock;
+        const refundRequest: RefundRequestWithBlock = {
+          ...rawRefundRequest,
+          // repaymentChainId is not part of the on-chain event, so add it here.
           repaymentChainId: this.chainId,
```
pxrl, line 791:
```suggestion
          repaymentChainId: this.chainId,  // repaymentChainId is not part of the on-chain event, so add it here.
```

PR #339, opened 2023-07-25
feat(clients): add `toJSON` and `updateFromJSON` methods

dohaki:
In an effort to make the states of `ConfigStoreClient`, `HubPoolClient` and `SpokePoolClient` serializable, i.e. cachable, I added custom `toJSON` methods. The `updateFromJSON` method allows updating the client state from the output of `toJSON`.

I added these because the performances of the `uba-client-state` job and serverless handler of `/suggested-fees` can be increased significantly by utilizing the above methods and caching.

For example, in the `/suggested-fees` handler, we can use cached `HubPoolClient` and `ConfigStoreClient` states in addition to the already cached `UBAClientState`. This way we don't need to call the `update()` methods to save time.

### DIFF
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -21,6 +21,7 @@ import winston from "winston";
 
 import {
   L1TokenTransferThreshold,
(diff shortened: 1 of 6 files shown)


nicholaspai (approved):
This seems like a reasonable solution.

PR #340, opened 2023-07-25
feat: add undefined return type

james-a-morris:


james-a-morris:
Closed per @nicholaspai

PR #341, opened 2023-07-26
feat: add IPFS caching client

james-a-morris:
This PR adds an IPFS client to the SDK. Additionally, this client complies with the `CachingMechanism` interface so that we can use it in arbitrary locations throughout the SDK where caching is concerned.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.14.13",
+  "version": "0.14.14",
   "license": "AGPL-3.0",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "@across-protocol/across-token": "^1.0.0",
     "@across-protocol/contracts-v2": "2.4.0",
     "@eth-optimism/sdk": "^2.1.0",
+    "@pinata/sdk": "^2.1.0",
     "@uma/sdk": "^0.34.1",
     "axios": "^0.27.2",
     "decimal.js": "^10.3.1",
--- a/src/caching/IPFS/PinataIPFSClient.ts
+++ b/src/caching/IPFS/PinataIPFSClient.ts
@@ -87,6 +87,6 @@ export class PinataIPFSClient implements CachingMechanismInterface {
         function: "setWithReturnID",
       },
     });
-    return storeValueInIPFS(JSON.stringify(value, jsonReplacerWithBigNumbers), this.client);
+    return storeValueInIPFS(key, JSON.stringify(value, jsonReplacerWithBigNumbers), this.client);
   }
 }
--- a/src/caching/IPFS/PinataIPFSClient.ts
+++ b/src/caching/IPFS/PinataIPFSClient.ts
@@ -1,16 +1,16 @@
-import { buildIPFSClient, retrieveValueFromIPFS, storeValueInIPFS } from "../utils/IPFSUtils";
-import { CachingMechanismInterface } from "../interfaces";
+import { buildIPFSClient, retrieveValueFromIPFS, storeValueInIPFS } from "../../utils/IPFSUtils";
+import { CachingMechanismInterface } from "../../interfaces";
 import { Struct } from "superstruct";
 import winston from "winston";
 import PinataClient from "@pinata/sdk";
-import { jsonReviverWithBigNumbers, jsonReplacerWithBigNumbers, formattedLog } from "../utils";
+import { jsonReviverWithBigNumbers, jsonReplacerWithBigNumbers, formattedLog } from "../../utils";
 
 /**
  * A client for interacting with the IPFS. This is a wrapper around the IPFS API.
  * This client also is a part of the caching mechanism interface.
  * @see https://docs.ipfs.io/reference/http/api/
  */
-export class IPFSClient implements CachingMechanismInterface {
+export class PinataIPFSClient implements CachingMechanismInterface {
   /**
    * The IPFS client instance
    */
--- a/src/caching/IPFSClient.ts
+++ b/src/caching/IPFSClient.ts
@@ -3,7 +3,7 @@ import { CachingMechanismInterface } from "../interfaces";
 import { Struct } from "superstruct";
 import winston from "winston";
 import PinataClient from "@pinata/sdk";
-import { jsonReviverWithBigNumbers, jsonReplacerWithBigNumbers } from "../utils";
+import { jsonReviverWithBigNumbers, jsonReplacerWithBigNumbers, formattedLog } from "../utils";
 
 /**
  * A client for interacting with the IPFS. This is a wrapper around the IPFS API.
@@ -42,7 +42,15 @@ export class IPFSClient implements CachingMechanismInterface {
     if (!key) {
       return null;
     }
-    this.logger?.info(`Retrieving value from IPFS with key ${key}`);
+    formattedLog(this.logger, {
+      level: "debug",
+      message: `Retrieving value from IPFS with key ${key}`,
+      at: {
+        location: "IPFSClient",
+        function: "get",
+      },
+    });
+
     const arrivedResult = await retrieveValueFromIPFS(key, this.publicGatewayURL);
     if (!arrivedResult) {
       return null;
@@ -50,7 +58,14 @@ export class IPFSClient implements CachingMechanismInterface {
     const revivedResult = JSON.parse(arrivedResult, jsonReviverWithBigNumbers);
 
     if (_structValidator && !_structValidator.is(revivedResult)) {
-      this.logger?.warn(`Retrieved value from IPFS with key ${key} does not match the expected type`);
+      formattedLog(this.logger, {
+        level: "warn",
+        message: `Retrieved value from IPFS with key ${key} does not match the expected type`,
+        at: {
+          location: "IPFSClient",
+          function: "get",
+        },
+      });
       return null;
     }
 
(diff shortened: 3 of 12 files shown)


src/utils/IPFSUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Build an IPFS client for interacting with the IPFS API
+ * @param APIKey The project ID
+ * @param secretAPIKey The project secret
+ * @returns An IPFS client for interacting with Pinata
+ */
+export function buildIPFSClient(APIKey: string, secretAPIKey: string): PinataClient {
+  return new PinataClient(APIKey, secretAPIKey);
+}
```
pxrl, line 13: Curious about this wrapper - it seems unnecessary. Do you anticipate that it could be expanded in future?

src/utils/IPFSUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+  // We need to check for both strings and big numbers, because the JSON parser
+  // is not aware of BigNumbers.
+  if (typeof value === "string" && /^-?\d+$/.test(value)) {
+    const bigNumber = BigNumber.from(value);
+    if (bigNumber.toString() === value) {
+      return bigNumber;
+    }
+  }
+  return value;
+}
```
pxrl, line 84: This isn't really related to IPFS - is there a better home for it in this repo?

src/utils/IPFSUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+ * @param value The value to convert
+ * @returns The converted value
+ */
+export function jsonReplacerWithBigNumbers(_key: string, value: unknown): unknown {
+  // We need to check if this is a big number, because the JSON parser
+  // is not aware of BigNumbers and will convert them to the string representation
+  // of the object itself which is not what we want.
+  if (value instanceof BigNumber) {
+    return value.toString();
+  }
```
pxrl, line 64: There's also `BigNumber.is(value)`. Not sure if either has any advantage over the other though :shrug:

src/utils/IPFSUtils.ts
james-a-morris: Done https://github.com/across-protocol/sdk-v2/pull/341/commits/e8993ab9114d4ea9443b7cca002a17f407c3c755

src/caching/IPFSClient.ts
```diff
@@ -0,0 +1,87 @@
...
+   * Calls to a public IPFS gateway to retrieve a value.
+   * @param key The key to retrieve.
+   * @param _structValidator An optional struct validator to validate the retrieved value. If the value does not match the struct, null is returned.
+   * @returns The value if it exists, otherwise null.
+   */
+  async get<ObjectType>(key?: string, _structValidator?: Struct<unknown, unknown>): Promise<ObjectType | null> {
+    if (!key) {
+      return null;
+    }
+    this.logger?.info(`Retrieving value from IPFS with key ${key}`);
```
nicholaspai, line 45: This logging doesn't comply with our existing log format as it should include the following keys:
- `at`: i.e. "IPFSClient#get"
- `message`: i.e. `Retrieving value from IPFS with key ${key}`. 
- ...other props you want to log

src/utils/IPFSUtils.ts
james-a-morris: Done! https://github.com/across-protocol/sdk-v2/pull/341/commits/57e90679e634103a40c3a441a18c44b4f27e6c7d

src/utils/IPFSUtils.ts
james-a-morris: I was trying to put as much logic in necessary utils & the class itself is for the caching interface.

src/interfaces/CachingMechanism.ts
```diff
@@ -9,7 +11,11 @@ export interface CachingMechanismInterface {
    * @param key The key to retrieve.
    * @returns The value if it exists, otherwise null.
    */
-  get<T>(key: string): Promise<T | null>;
+  get<ObjectType, OverrideType>(
```
nicholaspai, line 14: I'm kind of confused how these generics work. For example with the RedisClient, every time I want to `get` a value do I have to use the syntax `get<ModifiedFlow[]>`?

I'm also confused how this OverrideType works

src/interfaces/CachingMechanism.ts
```diff
@@ -20,5 +26,26 @@ export interface CachingMechanismInterface {
...
+  /**
+   * Identical to set, but returns the ID of the value stored. This is useful for storing values in a cache that
+   * are not known ahead of time. For example, if you want to store a value in a cache that is the result of a
+   * computation, you can use this method to store the value and retrieve the ID of the value stored.
+   * @param key The canonical key to store.
+   * @param value The value to store.
+   * @param ttl The time to live in seconds.
+   * @param overrides Any overrides to the default caching mechanism configuration for the given caching protocol.
+   */
+  setWithReturnID<ObjectType, OverrideType>(
```
nicholaspai, line 45: Do you think this should be in the interface? Or should we have `set` always return a `string` rather than `boolean`? Just curious what you think I'm unopinonated but keeping interfaces slim generally seems like a good idea

src/utils/IPFSUtils.ts
nicholaspai: Agreed

src/utils/IPFSUtils.ts
nicholaspai: How does this handle deeply nested BigNumbers?

src/caching/IPFSClient.ts
```diff
@@ -0,0 +1,82 @@
...
+import winston from "winston";
+import PinataClient from "@pinata/sdk";
+import { jsonReviverWithBigNumbers, jsonReplacerWithBigNumbers } from "../utils";
+
+/**
+ * A client for interacting with the IPFS. This is a wrapper around the IPFS API.
+ * This client also is a part of the caching mechanism interface.
+ * @see https://docs.ipfs.io/reference/http/api/
+ */
+export class IPFSClient implements CachingMechanismInterface {
```
dohaki, line 13: NIT: We are actually pretty opinionated on using the specific Pinata API here. So maybe renaming to `PinataIPFSClient` would be more accurate. The class `IPFSClient` then could be used in the future to implement a platform-agnostic IPFS HTTP Client that, for example, could connect directly to an IPFS node.

src/utils/IPFSUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+      // any anomalies with the response
+      headers: {
+        Accept: "text/plain",
+      },
+      // We want just the raw response, not the parsed response
+      transformResponse: (r) => r,
+    });
+    return result.data;
+  } catch (e) {
+    return undefined;
```
dohaki, line 36: Hmm, any specific reason not to let an error bubble up? The consumer of the libarary should be the one deciding what to do in case of a fetching error

src/caching/IPFSClient.ts
james-a-morris: Done & included a new formatting fn so that the logger is typed. https://github.com/across-protocol/sdk-v2/pull/341/commits/18bc951b56ddad980fd167bf86e29e72806f958e

dohaki (commented):
Nice work! Left one nit and one comment on suppressing errors.

src/interfaces/CachingMechanism.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/341/commits/aa633b002adab888f786e24d1a464b775e28e9c7

src/caching/IPFSClient.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/341/commits/79537dbf655bcac073d3400dca068adb8f25c3f3

src/utils/IPFSUtils.ts
james-a-morris: I'm a bit weary of this because the system will error if a key isn't found. However, that's fair that we want to propagate the error so I've removed it. Good callout! https://github.com/across-protocol/sdk-v2/pull/341/commits/421ea066c570643c6900e4272a688b78c73b8de2

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.14.12",
+  "version": "0.14.13",
```
pxrl, line 4: FWIW I just bumped the version so you probably want to go to 0.14.14: https://github.com/across-protocol/sdk-v2/releases/tag/v0.14.13
```suggestion
  "version": "0.14.14",
```

james-a-morris: Rebased

dohaki (approved):
LGTM

src/caching/IPFS/PinataIPFSClient.ts
```diff
@@ -0,0 +1,92 @@
...
+   * Calls to a public IPFS gateway to retrieve a value.
+   * @param key The key to retrieve.
+   * @param _structValidator An optional struct validator to validate the retrieved value. If the value does not match the struct, null is returned.
+   * @returns The value if it exists, otherwise null.
+   */
+  async get<ObjectType>(key?: string, _structValidator?: Struct<unknown, unknown>): Promise<ObjectType | null> {
+    if (!key) {
+      return null;
+    }
+    formattedLog(this.logger, {
```
nicholaspai, line 45: I don't like this pattern that much, do we have to use `formattedLog` every time we log?

james-a-morris: My understanding is that this is an initial pass of how we can move toward a cohesive formatting tool. I'll definitely workshop it, but I think it's a good medium if you'd like to roll with it for now

james-a-morris:
> I'm curious now that we have the `RedisClient` in relayer-v2, and this client in `sdk-v2` it seems like we ideally should be putting these into the same file. I think adding them both to the sdk-v2 makes more sense but I'm not really sure TBH, wdyt?

I think this may be beneficial, my only reason for keeping redis in the relayer is that the redis client is very large & is only used there where it could be possible that pinata could be used by both the FE & the relayer

PR #342, opened 2023-07-26
tests: getEventFee and adjacent methods

james-a-morris:
This PR aims to add unit tests to the `getEventFee` function for rigorous testing. Additionally, `getDepositFee` and `getRefundFee` are tested to ensure that they exactly follow the outputs of the adjacent `getEventFee` fn call.


src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,182 @@
...
+
+describe("UBAFeeSpokeCalculatorAnalog", () => {
+  const defaultConfig = new UBAConfig(
+    {
+      default: toBN(0),
+    },
+    {
+      default: [
+        [toBN(0), toBN(0)],
+        [toBN(0), parseEther("2")],
```
nicholaspai, line 15: Can you add a comment about this curve? It looks like you want to hardcode 200% as the fee for every `amount`?

nicholaspai: Also, isn't this curve illegal because there are duplicates `x` points with `x=0`?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      const amount = BigNumber.from(10);
+      const lastRunningBalance = BigNumber.from(1000);
+      const lastIncentiveBalance = BigNumber.from(-1000);
+      const chainId = 1;
+      const flowType = "outflow";
+      const fee = getEventFee(amount, flowType, lastRunningBalance, lastIncentiveBalance, chainId, defaultConfig);
+      expect(fee.balancingFee.toString()).toEqual("0");
+    });
+
+    it("should return a balance fee 0 if lastIncentiveBalance is 0", () => {
```
nicholaspai, line 73: Is this a duplicate test as the one above?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      const chainId = 1;
+      const flowType = "outflow";
+      const fee = getEventFee(amount, flowType, lastRunningBalance, lastIncentiveBalance, chainId, defaultConfig);
+      expect(fee.balancingFee.toString()).toEqual("0");
+    });
+
+    describe("Should correctly apply a reward multiplier", () => {
+      for (const fee of [-23, 0, 23]) {
+        const signToZero = fee === 0 ? "==" : fee > 0 ? ">" : "<";
+        it(`should return a balance fee a discounted balance fee if the fee is ${signToZero} 0`, () => {
```
nicholaspai, line 86: Looks like there is a typo in this test name?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      const fee = getEventFee(amount, flowType, lastRunningBalance, lastIncentiveBalance, chainId, defaultConfig);
+      expect(fee.balancingFee.toString()).toEqual("20");
+    });
+  });
+
+  for (const [flowName, flowType] of [
+    ["getDepositFee", "inflow"],
+    ["getRefundFee", "outflow"],
+  ]) {
+    describe(flowName, () => {
```
nicholaspai, line 148: I'm not entireley sure what this test is supposed to achieve can you elaborate?

nicholaspai (commented):
Where do you test the logic about:

1. Applying the hardcoded multiplier:
```ts
    // First, apply hardcoded multiplier if incentive fee is a reward instead of a penalty
    // This should never error. `getUbaRewardMultiplier` should default to 1 and be in 18 decimal precision
    const ubaRewardMultiplier = config.getUbaRewardMultiplier(chainId.toString());
    balancingFee = balancingFee.mul(ubaRewardMultiplier).div(fixedPointAdjustment);
```

2. Apply a discount if the incentive balance is below the `feesToBringFeePctToZero`:
```ts
// Next, compute the amount of fees that would be required to be paid out of the incentive balance
    // to bring the the balance fee to 0. This is found by finding the point on the flow curve
    // where the fee is 0 and integrating from there to the last running balance without including
    // the current action.
    const zeroFeePoint = config.getZeroFeePointOnBalancingFeeCurve(chainId);
    const feesToBringFeePctToZero = computePiecewiseLinearFunction(flowCurve, zeroFeePoint, lastRunningBalance).abs();
    console.log(`- feesToBringFeePctToZero = ${feesToBringFeePctToZero.toString()}`);

    // The discount factor to apply is equal to the balance fee to bring the fee % back to 0 divided
    // by the incentive balance, capped at 100%.
    // @dev balancing fee, amount, and last incentive balance should all be in the same decimals precision.
    if (feesToBringFeePctToZero.gt(lastIncentiveBalance)) {
      console.log(
        `- discounting balancing reward because feesToBringFeePctToZero exceeds incentive balance. Starting balancing fee = ${balancingFee.toString()}, incentive balance = ${lastIncentiveBalance.toString()}`
      );

      // Discount factor should be in 18 decimal precision and should always be <= 1 since
      // feesToBringFeePctToZero > lastIncentiveBalance
      const discountFactor = lastIncentiveBalance.mul(fixedPointAdjustment).div(feesToBringFeePctToZero);
      balancingFee = balancingFee.mul(discountFactor).div(fixedPointAdjustment);
      console.log(`- Discount factor = ${discountFactor}%, resultant balancing fee = ${balancingFee}`);
    }
```

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,178 @@
+import { BigNumber } from "ethers";
+import UBAConfig from "./UBAFeeConfig";
+import { getEventFee, getDepositFee, getRefundFee } from "./UBAFeeSpokeCalculatorAnalog";
+import { toBN } from "../utils";
+import { parseEther } from "ethers/lib/utils";
+
+describe.only("UBAFeeSpokeCalculatorAnalog", () => {
```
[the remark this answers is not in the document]
james-a-morris: I don't believe there are any in the most recent version of this PR - at least in the git CI's most recent commit

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Yep! Good call - also `fee` is replaced with the correct variable `multiplier`

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: These are meant to guarantee that the `getDepositFee`/`getRefundFee` return the same as an adjacent call to `getEventFee`.

Added this comment:

```
  // The following tests are designed around ensuring that our `getEventFee` specific functions (getDepositFee and getRefundFee) are
  // calculating the same fee as the `getEventFee` function. This is to ensure that the `getEventFee` function is working as expected
  // as well as the `getDepositFee` and `getRefundFee` functions.
  // Based on their design, these calculation functions should always return the same fee as the `getEventFee` function.
```

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Resolved these two: https://github.com/across-protocol/sdk-v2/pull/342/commits/2d38b389a3cd4462cdf176a628355226be7b3876

james-a-morris:
> Where do you test the logic about:
> 
> 1. Applying the hardcoded multiplier:
> 
> ```ts
>     // First, apply hardcoded multiplier if incentive fee is a reward instead of a penalty
>     // This should never error. `getUbaRewardMultiplier` should default to 1 and be in 18 decimal precision
>     const ubaRewardMultiplier = config.getUbaRewardMultiplier(chainId.toString());
>     balancingFee = balancingFee.mul(ubaRewardMultiplier).div(fixedPointAdjustment);
> ```
> 
> 2. Apply a discount if the incentive balance is below the `feesToBringFeePctToZero`:
> 
> ```ts
> // Next, compute the amount of fees that would be required to be paid out of the incentive balance
>     // to bring the the balance fee to 0. This is found by finding the point on the flow curve
>     // where the fee is 0 and integrating from there to the last running balance without including
>     // the current action.
>     const zeroFeePoint = config.getZeroFeePointOnBalancingFeeCurve(chainId);
>     const feesToBringFeePctToZero = computePiecewiseLinearFunction(flowCurve, zeroFeePoint, lastRunningBalance).abs();
>     console.log(`- feesToBringFeePctToZero = ${feesToBringFeePctToZero.toString()}`);
> 
>     // The discount factor to apply is equal to the balance fee to bring the fee % back to 0 divided
>     // by the incentive balance, capped at 100%.
>     // @dev balancing fee, amount, and last incentive balance should all be in the same decimals precision.
>     if (feesToBringFeePctToZero.gt(lastIncentiveBalance)) {
>       console.log(
>         `- discounting balancing reward because feesToBringFeePctToZero exceeds incentive balance. Starting balancing fee = ${balancingFee.toString()}, incentive balance = ${lastIncentiveBalance.toString()}`
>       );
> 
>       // Discount factor should be in 18 decimal precision and should always be <= 1 since
>       // feesToBringFeePctToZero > lastIncentiveBalance
>       const discountFactor = lastIncentiveBalance.mul(fixedPointAdjustment).div(feesToBringFeePctToZero);
>       balancingFee = balancingFee.mul(discountFactor).div(fixedPointAdjustment);
>       console.log(`- Discount factor = ${discountFactor}%, resultant balancing fee = ${balancingFee}`);
>     }
> ```

The first has been completed in this test: 
```
getEventFee should correctly apply a reward multiplier
```

I will follow up offline about the second

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
nicholaspai: i see nice

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,188 @@
...
+          const originalFee = getEventFee(
+            amount,
+            flowType,
+            lastRunningBalance,
+            lastIncentiveBalance,
+            chainId,
+            defaultConfig
+          );
+
+          const config = new UBAConfig(
```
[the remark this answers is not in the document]
james-a-morris: Done! Good call

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    it("should apply a discount if the incentive balance is below the feesToBringFeePctToZero", () => {
+      const config = new MockUBAConfig();
+      // These values are set such that our feeToBringFeePctToZero is higher than the balancing fee
+      // within the getEventFee function. This is to ensure that the discount is applied for this test
+      // case.
+      const amount = BigNumber.from(1000);
+      const lastRunningBalance = BigNumber.from(1000);
+      const lastIncentiveBalance = BigNumber.from(1000);
+      const chainId = 1;
+      const flowType = "outflow";
```
nicholaspai, line 108: Can you add an `expect` to log out that the `feeToBringFeePctToZero.gt(lastIncentiveBalance)`? So we don't have a false positive test on some complex logic

Also, why is the `balancingFee` 20? Shouldn't it be discounted?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -92,6 +93,16 @@ describe("UBAFeeSpokeCalculatorAnalog", () => {
...
       const flowType = "inflow";
+
+      const zeroFeePoint = defaultConfig.getZeroFeePointOnBalancingFeeCurve(chainId);
+      const feeToBringFeePctToZero = computePiecewiseLinearFunction(
+        defaultConfig.getBalancingFeeTuples(chainId),
+        zeroFeePoint,
+        lastRunningBalance
+      ).abs();
+
+      expect(feeToBringFeePctToZero.gt(lastIncentiveBalance)).toBeTruthy();
```
nicholaspai, line 104: why not use `expect().to.be.greater.than()`? 

Also, can you print out what you think the non discounted balancing fee and what the discounted one should be? I think we should check that the discount factor is computed and then applied correctly.

For example if we expect the discount factor to be 0.8, we should check that balancingFee.eq(potentialBalancingFee.mul(toBNWei("0.8")).div(fixedPointAdjustment))

james-a-morris: These are `jest` tests & they don't compute the inequalities on BigNumbers out of the box

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,167 @@
...
+      const lastRunningBalance = BigNumber.from(1000);
+      const lastIncentiveBalance = BigNumber.from(0);
+      const chainId = 1;
+      const flowType = "outflow";
+      const fee = getEventFee(amount, flowType, lastRunningBalance, lastIncentiveBalance, chainId, defaultConfig);
+      expect(fee.balancingFee.toString()).toEqual("0");
+    });
+
+    describe("getEventFee should correctly apply a reward multiplier", () => {
+      for (const multiplier of [-23, 0, 23]) {
```
nicholaspai, line 63: These multipliers should be in `wei` so rather than `-23` do `-toBNWei("23")`

nicholaspai: I'd also like to see some test values with other numbers like:
- `toBNWei("5.5")`
- `toBNWei("0.2")`
- `toBNWei("1.2")` etc.

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/342/commits/abbdafb07855a718155ff9306655a93c0c0d2b9b

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Yep! That's now added in: 
```
      for (const rawMultiplier of ["-1", "1.2", "0.2", "1.2", "3.4", "7"]) {
        const multiplier = utils.parseEther(rawMultiplier);
```

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Done - these were from Chase's original truth conditions. I've just tried to port over the code.

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,176 @@
...
+      const lastRunningBalance = BigNumber.from(1000);
+      const lastIncentiveBalance = BigNumber.from(0);
+      const chainId = 1;
+      const flowType = "outflow";
+      const fee = getEventFee(amount, flowType, lastRunningBalance, lastIncentiveBalance, chainId, defaultConfig);
+      expect(fee.balancingFee.toString()).toEqual("0");
+    });
+
+    describe("getEventFee should correctly apply a reward multiplier", () => {
+      for (const rawMultiplier of ["-1", "1.2", "0.2", "1.2", "3.4", "7"]) {
```
nicholaspai, line 63: nit: can you add a "0" multiplier test too?

james-a-morris: Done https://github.com/across-protocol/sdk-v2/pull/342/commits/e71573b51b6703995dbc0ffb6973a873e9588ec7

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
nicholaspai: This shouldn't rely at all on chase's truth conditions we should be reasoning only by looking at the code within `getEventFee`. This test in particular should be testing lines 177-185 here: https://github.com/across-protocol/sdk-v2/blob/master/src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts#L184

nicholaspai:
https://github.com/across-protocol/sdk-v2/pull/342/files#r1278082059

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    it("should calculate the balancing fee for an outflow event", () => {
+      const amount = BigNumber.from(10);
+      const lastRunningBalance = BigNumber.from(1000);
+      const lastIncentiveBalance = BigNumber.from(1000);
+
+      const chainId = 1;
+      const flowType = "outflow";
+      const fee = getEventFee(amount, flowType, lastRunningBalance, lastIncentiveBalance, chainId, defaultConfig);
+      expect(fee.balancingFee.toString()).toEqual("0");
```
nicholaspai, line 29: I don't love using tests where the output is supposed to be "0" because it increases the likelihood of false positive tests because "0" is often a default value.

Can you change the `UBAConfig` so that the curve is more symmetrical and produces negative and positive fees?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,213 @@
+import { BigNumber, utils } from "ethers";
+import { getEventFee, getDepositFee, getRefundFee } from "./UBAFeeSpokeCalculatorAnalog";
+import { fixedPointAdjustment, toBNWei } from "../utils";
+import { MockUBAConfig } from "../clients/mocks";
+import { computePiecewiseLinearFunction } from "./UBAFeeUtility";
+
+describe("UBAFeeSpokeCalculatorAnalog", () => {
+  const defaultConfig = new MockUBAConfig();
```
nicholaspai, line 8: I don't think we should be testing with a default config as the defaults might change. I think we should explicitly set balancing fee tuples for this test

nicholaspai (commented):
I don't think this set of tests actually gives me much confidence that the logic works. First of all the UBA config is not explained in comments or in code what the expected curve looks like, so testing something like "should return a balanceFee of 0 if the amount is 0" isn't helpful for me because I don't know what the curve looks like.

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      const amount = BigNumber.from(0);
+      const lastRunningBalance = BigNumber.from(1000);
+      const lastIncentiveBalance = BigNumber.from(1000);
+      const chainId = 1;
+      const flowType = "outflow";
+      const fee = getEventFee(amount, flowType, lastRunningBalance, lastIncentiveBalance, chainId, defaultConfig);
+      expect(fee.balancingFee.toString()).toEqual("0");
+    });
+
+    it("should return a balance fee 0 if lastIncentiveBalance is negative", () => {
```
nicholaspai, line 42: This is a good test because the logic explicitly returns 0 if `lastIncentiveBalance` is negative.

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        lastRunningBalance,
+        lastIncentiveBalance,
+        chainId,
+        config
+      ).balancingFee;
+
+      const nonDiscountedFee = getEventFee(
+        amount,
+        flowType,
+        lastRunningBalance,
```
nicholaspai, line 87:
```suggestion
        // This fee is "non-discounted" because the `lastIncentiveBalance` is increased such that the balancing fee no longer needs to be discounted
        lastRunningBalance,
```

I think we should not set `lastIncentiveBalance = lastRunningBalance` but should set it explicitly to something like `toBNWei("1000")` for example so that its very clear how this test is working. Also comments should be added here.

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,236 @@
+import { BigNumber, utils } from "ethers";
+import { getEventFee, getDepositFee, getRefundFee } from "./UBAFeeSpokeCalculatorAnalog";
+import { toBNWei } from "../utils";
+import { MockUBAConfig } from "../clients/mocks";
+import { computePiecewiseLinearFunction } from "./UBAFeeUtility";
+
+describe("UBAFeeSpokeCalculatorAnalog", () => {
+  describe.only("getEventFee", () => {
```
nicholaspai, line 8:
```suggestion
  describe("getEventFee", () => {
```

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    // Create a chain Id that we can reference anywhere
+    // without any magic numbers.
+    const chainId = "1";
+
+    // This is the base case for this function. We want to ensure that if we have a
+    // positive balancing fee result, that the result is equal to the integration of
+    // the balancing fee curve without any additional modifiers.
+    it("should expect if our balancing fee is positive, that the result is a single integration", () => {
```
nicholaspai, line 16:
```suggestion
    it("if balancing fee is positive, then the result is a single integration", () => {
```

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      // We start with a basic config that's been instantiated.
+      const config = new MockUBAConfig();
+      // We set the balancing fee curve to be a basic curve that has a positive slope.
+      // We can work in terms of any decimal as long as we're consistent & the second value
+      // in the tuple is in terms of wei or any other 18 decimal token.
+      config.setBalancingFeeCurve(chainId, [
+        [toBNWei(0, 6), toBNWei(0, 0)],
```
nicholaspai, line 23: You might want to pull out the decimals into a separate variable at the top of the tests and perform the full test for two types of decimals, 6 and 18 for example

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      const lastRunningBalance = toBNWei(1000, 6);
+      const lastIncentiveBalance = toBNWei(1000, 6);
+      // We call the getEventFee function with the parameters we've set above.
+      const fee = getEventFee(amount, "inflow", lastRunningBalance, lastIncentiveBalance, 1, config).balancingFee;
+      // We are expecting that the fee will be positive. As a result, let's assert
+      // that the fee is greater than 0.
+      expect(fee.gt(0)).toBeTruthy();
+      // Because the fee is greater than zero, we applied no additional reward multiplier,
+      // or any other factors that would affect the fee, we expect that the fee will be
+      // equal to the integration of the balancing fee curve.
```
nicholaspai, line 41: +1 great comment clearly explaining why we expect the fee to be what it is

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,236 @@
...
+        config.getBalancingFeeTuples(Number(chainId)),
+        lastRunningBalance,
+        lastRunningBalance.add(amount)
+      );
+      // We expect that the fee is equal to the integration of the balancing fee curve.
+      // Note: we need to ensure that we're using the same decimals as the balancing fee curve.
+      // Note: The expectation above is that our fee is positive
+      expect(fee.toString()).toEqual(integration.toString());
+
+      // As an edge case, we can also test that if the balancing fee that is initially computed
```
nicholaspai, line 52: Should we make this edge case a separate unit test in its own `it(...)`?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      lastIncentiveBalance = toBNWei(-10, 6);
+      fee = getEventFee(amount, "outflow", lastRunningBalance, lastIncentiveBalance, 1, config).balancingFee;
+      expect(fee.toString()).toEqual("0");
+    });
+
+    // We now want to test the specific code path where the balancing fee is negative initially
+    // and the incentive balance is positive, but the zero fee threshold is not met. In this case,
+    // we expect that the fee will simply be discounted by the reward multiplier provided in the
+    // config.
+    it("should return a balancing fee that is discounted by the reward multiplier if the balancing fee is initially negative and incentive balance is positive but below the zero fee threshold", () => {
```
nicholaspai, line 129: This is a good test for the UBARewardMultiplier but have we lost the tests that test the discount factor based purely on the incentive balance being greater than the incentives needed to bring the fee back to 0? i.e. when `config.setRewardMultiplier(chainId, utils.parseEther("1"))`

nicholaspai (commented):
+1 really appreciate all the comments. I think I'd still like to see two additions:
- Testing the discount factor logic from lines 170-187 [here](https://github.com/across-protocol/sdk-v2/blob/master/src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.ts#L170)
- Fuzz/stress testing with a lot of randomly generated balancing fee tuple inputs, reward multiplier inputs, incentive, running balance and amount inputs.

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: This is handled now within the code blocks

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Done!

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: I personally think we should keep it as is. The reason being: if we separated it we'd have a lot of code duplication

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,331 @@
...
+            // We start with a basic config that's been instantiated.
+            const config = new MockUBAConfig();
+            // We set the balancing fee curve to be a basic curve that has a positive slope.
+            // We can work in terms of any decimal as long as we're consistent & the second value
+            // in the tuple is in terms of wei or any other 18 decimal token.
+            config.setBalancingFeeCurve(chainId, [
+              [toBNWei(0, decimalCount), toBNWei(0, 0)],
+              [toBNWei(1, decimalCount), positiveSlope],
+            ]);
+            // We set the reward multiplier to be 1. This is to ensure that we don't have any
```
nicholaspai, line 43: Is this set to 1? It looks to be a random number `const rewardMultiplier = utils.parseEther(Math.random().toFixed(18))`

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,331 @@
...
+            // additional reward multiplier that would affect our calculations.
+            config.setRewardMultiplier(chainId, rewardMultiplier);
+            // We call the getEventFee function with the parameters we've set above.
+            const fee = getEventFee(amount, "inflow", lastRunningBalance, lastIncentiveBalance, 1, config).balancingFee;
```
nicholaspai, line 47: is it worth testing outflow too or does that not increase test coverage?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,331 @@
...
+                config
+              ).balancingFee.toString()
+            ).toEqual("0");
+          }
+        });
+
+        // We now want to test the specific code path where the balancing fee is negative initially
+        // but the incentive balance is not positive (i.e. 0 or negative). In this case, we expect
+        // that the fee will be zero.
+        it("should return a balancing fee of 0 if the balancing fee is initially negative and incentive balance is non-positive", () => {
```
nicholaspai, line 100: nit: does this test also the edge case where the balancing fee is 0 and incentive balance is < 0, should also return balancing fee of 0?

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Resolved

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,331 @@
...
+            fee = getEventFee(amount, "outflow", lastRunningBalance, lastIncentiveBalance, 1, config).balancingFee;
+            expect(fee.toString()).toEqual("0");
+          }
+        });
+
+        // We now want to test the specific code path where the balancing fee is negative initially
+        // and the incentive balance is positive, but the zero fee threshold is not met. In this case,
+        // we expect that the fee will simply be discounted by the reward multiplier provided in the
+        // config.
+        it("should return a balancing fee that is discounted by the reward multiplier if the balancing fee is initially negative and incentive balance is positive but below the zero fee threshold", () => {
```
nicholaspai, line 153: Should this read "above the zero fee threshold"? or am I confusing things? If yes, then following test should read "below the zero fee threshold"

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
```diff
@@ -0,0 +1,331 @@
...
+            config
+          ).balancingFee;
+          // We expect that the fee is greater than the baseline fee.
+          expect(fee.gt(baselineFee)).toBeTruthy();
+
+          // From the above data, we know that our zeroPoint is 0. So, in order to compute the
+          // zeroPointFee, we need to compute the integral of the curve from 0 to the lastRunningBalance.
+          // In numerical terms, this is an integral from the range [0, 1000e{decimalCount}].
+          // We can compute this discount factor now.
+          const zeroPointFee = computePiecewiseLinearFunction(
```
nicholaspai, line 263: is this a fuzz test or unit test? If just a unit test can you create a fuzz test version where the `zeroPointFee's` and `lastIncentiveBalance` are changed such that the `multiplier` changes and we can test edge cases where its very close to 0 and 100?

nicholaspai (commented):
Really well commented tests, have left some comments but this LGTM

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: The difference between `inflow` and `outflow` is designed to be the direction of the integral. It shouldn't affect code coverage as it only appears in `getEventFee` as 

```
  let balancingFee = computePiecewiseLinearFunction(
    flowCurve,
    lastRunningBalance,
    lastRunningBalance.add(amount.mul(flowType === "inflow" ? 1 : -1))
  );

```

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Yes it should. We initially resolve a negative balancing fee. Then, we set `lastIncentiveBalance` to be negative and re-confirm that the balance is yielded as zero.

```
            // Finally, we can set the lastIncentiveBalance to be negative to ensure that calling
            // our fee again will result in a zero
            lastIncentiveBalance = toBNWei(Math.floor(Math.random() * -100) + 1, decimalCount);
            fee = getEventFee(amount, "outflow", lastRunningBalance, lastIncentiveBalance, 1, config).balancingFee;
            expect(fee.toString()).toEqual("0");
```

src/UBAFeeCalculator/UBAFeeSpokeCalculatorAnalog.test.ts
james-a-morris: Done

PR #346, opened 2023-08-03
feat: ✨ add global chain index scraping from configstore 

james-a-morris:
Adds indices scraping to the across config store. This change is aimed at providing a dynamic approach to determining the across chain indices across block ranges.

<details><summary>Details</summary>
<p>

[feat: ✨ add global chain index scraping from configstore](https://github.com/across-protocol/sdk-v2/commit/6cbe2c82ddc3448a8844fd77cc3e0521795d4204) 

This change adds global config store updates for the chain ID list. If the user requests chain indices prior to the relayer's knowledge then it will return the hardcoded value per the UMIP

[feat: ✨ add generic to global config updates](https://github.com/across-protocol/sdk-v2/commit/e5d67a929732891a67c41fb42bc563484515982e) 

This change adds generic value types for global config updates. Currently, value updates were only able to be numeric types. However, now there is an optional type to include open ended values. These values default to the `number` type for events where the generic is not included.
</p>
</details>

### DIFF
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -454,7 +454,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
         // indices are [1, 10, 137, 288, 42161] (outlined in UMIP-157)
         const previousUpdate = this.chainIdIndicesUpdates.at(-1)?.value ?? [1, 10, 137, 288, 42161];
         // We should now check that previousUpdate is a subset of chainIndices.
-        if (!previousUpdate.every((chainId) => chainIndices.includes(chainId))) {
+        if (!previousUpdate.every((chainId, idx) => chainIndices[idx] === chainId)) {
           this.logger.warn({
             at: "ConfigStoreClient#update",
             message: `The array ${rawChainIndices} is invalid. It must be a superset of the previous array ${previousUpdate}`,
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -57,12 +57,13 @@ export type ConfigStoreUpdate = { success: false } | _ConfigStoreUpdate;
 // @dev Do not change this value.
 export const DEFAULT_CONFIG_STORE_VERSION = 0;
 
-export const GLOBAL_CONFIG_STORE_KEYS = {
-  MAX_RELAYER_REPAYMENT_LEAF_SIZE: "MAX_RELAYER_REPAYMENT_LEAF_SIZE",
-  MAX_POOL_REBALANCE_LEAF_SIZE: "MAX_POOL_REBALANCE_LEAF_SIZE",
-  VERSION: "VERSION",
-  DISABLED_CHAINS: "DISABLED_CHAINS",
-};
+export enum GLOBAL_CONFIG_STORE_KEYS {
+  MAX_RELAYER_REPAYMENT_LEAF_SIZE = "MAX_RELAYER_REPAYMENT_LEAF_SIZE",
+  MAX_POOL_REBALANCE_LEAF_SIZE = "MAX_POOL_REBALANCE_LEAF_SIZE",
+  VERSION = "VERSION",
+  DISABLED_CHAINS = "DISABLED_CHAINS",
+  CHAIN_ID_INDICES = "CHAIN_ID_INDICES",
+}
 
 export class AcrossConfigStoreClient extends BaseAbstractClient {
   public cumulativeRateModelUpdates: across.rateModel.RateModelEvent[] = [];
@@ -71,6 +72,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
   public cumulativeTokenTransferUpdates: L1TokenTransferThreshold[] = [];
   public cumulativeMaxRefundCountUpdates: GlobalConfigUpdate[] = [];
   public cumulativeMaxL1TokenCountUpdates: GlobalConfigUpdate[] = [];
+  public chainIdIndicesUpdates: GlobalConfigUpdate<number[]>[] = [];
   public cumulativeSpokeTargetBalanceUpdates: SpokeTargetBalanceUpdate[] = [];
   public cumulativeConfigStoreVersionUpdates: ConfigStoreVersionUpdate[] = [];
   public cumulativeDisabledChainUpdates: DisabledChainsUpdate[] = [];
@@ -85,8 +87,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     readonly logger: winston.Logger,
     readonly configStore: Contract,
     readonly eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
-    readonly configStoreVersion: number,
-    readonly enabledChainIds: number[]
+    readonly configStoreVersion: number
   ) {
     super();
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
@@ -121,6 +122,13 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
       return undefined;
     }
     return across.rateModel.parseAndReturnRateModelFromString(config.routeRateModel[route]);
+  }
+
+  getChainIdIndicesForBlock(blockNumber: number = Number.MAX_SAFE_INTEGER): number[] {
+    const config = (sortEventsDescending(this.chainIdIndicesUpdates) as GlobalConfigUpdate<number[]>[]).find(
+      (config) => config.blockNumber <= blockNumber
+    );
+    return config?.value ?? [1, 10, 137, 288, 42161];
   }
 
   getTokenTransferThresholdForBlock(l1Token: string, blockNumber: number = Number.MAX_SAFE_INTEGER): BigNumber {
@@ -177,17 +185,13 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
    * @param allPossibleChains Returned list will be a subset of this list.
    * @returns List of chain IDs that have been enabled at least once in the block range. Sorted from lowest to highest.
    */
-  getEnabledChainsInBlockRange(
-    fromBlock: number,
-    toBlock = Number.MAX_SAFE_INTEGER,
-    allPossibleChains = this.enabledChainIds
-  ): number[] {
+  getEnabledChainsInBlockRange(fromBlock: number, toBlock = Number.MAX_SAFE_INTEGER): number[] {
     if (toBlock < fromBlock) {
       throw new Error(`Invalid block range: fromBlock ${fromBlock} > toBlock ${toBlock}`);
     }
     // Initiate list with all chains enabled at the fromBlock.
     const disabledChainsAtFromBlock = this.getDisabledChainsForBlock(fromBlock);
-    const enabledChainsAtFromBlock = allPossibleChains.filter(
+    const enabledChainsAtFromBlock = this.getChainIdIndicesForBlock(fromBlock).filter(
       (chainId) => !disabledChainsAtFromBlock.includes(chainId)
     );
 
@@ -200,7 +204,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
         }
         // If any of the possible chains are not listed in this disabled chain update and are not already in the
         // enabled chain list, then add them to the list.
-        allPossibleChains.forEach((chainId) => {
+        this.getChainIdIndicesForBlock(disabledChainUpdate.blockNumber).forEach((chainId) => {
           if (!disabledChainUpdate.chainIds.includes(chainId) && !enabledChains.includes(chainId)) {
             enabledChains.push(chainId);
           }
@@ -210,10 +214,10 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
       .sort((a, b) => a - b);
   }
 
-  getEnabledChains(block = Number.MAX_SAFE_INTEGER, allPossibleChains = this.enabledChainIds): number[] {
+  getEnabledChains(block = Number.MAX_SAFE_INTEGER): number[] {
     // Get most recent disabled chain list before the block specified.
     const currentlyDisabledChains = this.getDisabledChainsForBlock(block);
-    return allPossibleChains.filter((chainId) => !currentlyDisabledChains.includes(chainId));
+    return this.getChainIdIndicesForBlock(block).filter((chainId) => !currentlyDisabledChains.includes(chainId));
   }
 
   getDisabledChainsForBlock(blockNumber: number = Number.MAX_SAFE_INTEGER): number[] {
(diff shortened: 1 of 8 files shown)


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -57,12 +57,13 @@ export type ConfigStoreUpdate = { success: false } | _ConfigStoreUpdate;
 // @dev Do not change this value.
 export const DEFAULT_CONFIG_STORE_VERSION = 0;
 
-export const GLOBAL_CONFIG_STORE_KEYS = {
-  MAX_RELAYER_REPAYMENT_LEAF_SIZE: "MAX_RELAYER_REPAYMENT_LEAF_SIZE",
-  MAX_POOL_REBALANCE_LEAF_SIZE: "MAX_POOL_REBALANCE_LEAF_SIZE",
-  VERSION: "VERSION",
-  DISABLED_CHAINS: "DISABLED_CHAINS",
-};
+export enum GLOBAL_CONFIG_STORE_KEYS {
```
nicholaspai, line 60: +1

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -123,6 +124,13 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     return across.rateModel.parseAndReturnRateModelFromString(config.routeRateModel[route]);
   }
 
+  getChainIdIndicesForBlock(blockNumber: number = Number.MAX_SAFE_INTEGER): number[] {
+    const config = (sortEventsDescending(this.chainIdIndicesUpdates) as GlobalConfigUpdate<number[]>[]).find(
+      (config) => config.blockNumber <= blockNumber
+    );
+    return config?.value ?? [1, 10, 137, 288, 42161];
```
nicholaspai, line 131: nit: can you add a comment how you arrive at this default value? i.e. some link to the UMIP ?

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -417,6 +421,33 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
         if (!isNaN(args.value)) {
           this.cumulativeMaxRefundCountUpdates.push(args);
         }
+      } else if (args.key === utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.CHAIN_ID_INDICES)) {
+        // First remove out all the spaces
+        const rawChainIndices = String(args.value).replace(/\s/g, "");
```
nicholaspai, line 426: Instead of removing spaces can't we just `JSON.parse(args.value)` and then check that its a list of positive integers including the chain ID 1? Said another way, why isn't the regex check a superfluous check?

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -417,6 +421,33 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+        // we'll skip this update.
+        if (
+          chainIndices.length !== new Set(chainIndices).size ||
+          chainIndices.some((index) => index < 0 || index >= chainIndices.length)
+        ) {
+          this.logger.warn({ at: "ConfigStore", message: `The array ${chainIndices} is invalid.` });
+          continue;
+        }
+        // If all else passes, we can add this update.
+        this.chainIdIndicesUpdates.push({ ...args, value: chainIndices });
```
nicholaspai, line 450: can we add one more check: the UMIP states that chain ID's can only append chains. Can you add a check that the chain ID list only adds ID's to the previous list?

The default case is [1,10,137,288,42161] which we can store as a variable and we should use this as the "previous" list for the first update

nicholaspai (commented):
This is really good code. I left a few nits!

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Resolved offline

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/346/commits/b2b3a6f73d5b7bf174df83112bb32b0d85b3bae8

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/346/commits/c3af2dff62d51862a10462a45d3696b6ca229b03

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -568,8 +614,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     return {
       eventSearchConfig: this.eventSearchConfig,
       configStoreVersion: this.configStoreVersion,
-      enabledChainIds: this.enabledChainIds,
-
+      enabledChainIds: this.chainIdIndicesUpdates,
```
nicholaspai, line 617:
```suggestion
      chainIdIndicesUpdates: this.chainIdIndicesUpdates,
```

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -193,7 +188,7 @@ export function getMostRecentBundleBlockRanges(
     const hubPoolStartBlock = getBlockRangeForChain(
       rootBundleBlockRanges,
       hubPoolClient.chainId,
-      hubPoolClient.configStoreClient.enabledChainIds
+      hubPoolClient.configStoreClient.getEnabledChains(latestExecutedRootBundle.blockNumber)
```
nicholaspai, line 191:
```suggestion
      hubPoolClient.configStoreClient.getEnabledChains()
```

I don't know how you arrived at passing in `latestExecutedRootBundle.blockNumber` as the block here but we should just use the latest chain ID list because the list can only get longer

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -203,7 +198,7 @@ export function getMostRecentBundleBlockRanges(
     const blockRangeForChain = getBlockRangeForChain(
       rootBundleBlockRanges,
       chainId,
-      hubPoolClient.configStoreClient.enabledChainIds
+      hubPoolClient.configStoreClient.getEnabledChains(latestExecutedRootBundle.blockNumber)
```
nicholaspai, line 201:
```suggestion
      hubPoolClient.configStoreClient.getEnabledChains()
```

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -222,7 +217,9 @@ export function getMostRecentBundleBlockRanges(
     const ubaActivationBundleStartBlockForChain = getBlockForChain(
       ubaActivationBundleStartBlocks,
       chainId,
-      hubPoolClient.configStoreClient.enabledChainIds
+      // We want to pull the most recent enabled chains because we should assume that the
+      // protocol defaults as per the UMIP (https://github.com/UMAprotocol/UMIPs/pull/590) will be triggered.
+      hubPoolClient.configStoreClient.getEnabledChains()
```
nicholaspai, line 222: Correct

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -380,12 +377,11 @@ export async function getMatchedDeposit(
 export function isUBAActivatedAtBlock(hubPoolClient: HubPoolClient, block: number, chain: number): boolean {
   try {
     const ubaActivationBlocks = getUbaActivationBundleStartBlocks(hubPoolClient);
-    const ubaActivationStartBlockForChain = getBlockForChain(
-      ubaActivationBlocks,
-      chain,
-      hubPoolClient.configStoreClient.enabledChainIds
-    );
-    return block >= ubaActivationStartBlockForChain;
+    const enabledChainsIndices = hubPoolClient.configStoreClient.getEnabledChains(ubaActivationBlocks[0]);
```
nicholaspai, line 380:
```suggestion
    const enabledChainsIndices = hubPoolClient.configStoreClient.getEnabledChains();
```

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -417,7 +413,7 @@ export function getUbaActivationBundleStartBlocks(hubPoolClient: HubPoolClient):
       return bundleStartBlocks;
     } else {
       // No validated bundles after UBA activation block, UBA should be activated on next bundle start blocks.
-      const chainIdIndices = hubPoolClient.configStoreClient.enabledChainIds;
+      const chainIdIndices = hubPoolClient.configStoreClient.getEnabledChains(ubaActivationBlock);
```
nicholaspai, line 416:
```suggestion
      const chainIdIndices = hubPoolClient.configStoreClient.getEnabledChains();
```

src/clients/UBAClient/UBAClientWithRefresh.ts
```diff
@@ -924,7 +924,7 @@ export class UBAClientWithRefresh extends BaseAbstractClient {
           const startBlock = getBlockForChain(
             startBlocks,
             chainId,
-            this.hubPoolClient.configStoreClient.enabledChainIds
+            this.hubPoolClient.configStoreClient.getEnabledChains(mainnetStartBlock)
```
nicholaspai, line 927:
```suggestion
          this.chainIdIndices
```

src/clients/mocks/MockConfigStoreClient.ts
```diff
@@ -23,11 +23,10 @@ export class MockConfigStoreClient extends AcrossConfigStoreClient {
     configStore: Contract,
     eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock"> = { fromBlock: 0, maxBlockLookBack: 0 },
     configStoreVersion: number,
-    enabledChainIds: number[],
     chainId = 1,
     mockUpdate = false
   ) {
-    super(logger, configStore, eventSearchConfig, configStoreVersion, enabledChainIds);
+    super(logger, configStore, eventSearchConfig, configStoreVersion);
```
nicholaspai, line 29: I think we're going to actually want to mock the `getEnabledChains()` function and have it return `enabledChainIdsOverride` if there is no GlobalConfig update. does that make sense?

src/utils/BundleUtils.ts
```diff
@@ -69,13 +69,13 @@ export function getImpliedBundleBlockRanges(
...
   // since this function is designed to reconstruct a historical bundle block range.
-  const enabledChains = configStoreClient.getEnabledChains(rootBundle.blockNumber, configStoreClient.enabledChainIds);
+  const enabledChains = configStoreClient.getEnabledChains(rootBundle.blockNumber);
 
   return rootBundle.bundleEvaluationBlockNumbers.map((endBlock, i) => {
-    const chainId = configStoreClient.enabledChainIds[i];
     const fromBlock = prevRootBundle?.bundleEvaluationBlockNumbers?.[i]
       ? prevRootBundle.bundleEvaluationBlockNumbers[i].toNumber() + 1
       : 0;
+    const chainId = configStoreClient.getEnabledChains(fromBlock)[i];
```
nicholaspai, line 78: Should be using the same chain ID list for the full bundle

```suggestion
    const chainId = enabledChains[i];
```

james-a-morris: Ah I see - so we can collapse this

src/clients/UBAClient/UBAClientUtilities.ts
james-a-morris: Right, my understanding was that we wanted to get as close to the start of the bundle as possible. But yes, since we're using an append-only list, then this should be gtg with the additional commit

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -417,6 +432,37 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+
+        // We now need to check that we're only appending positive integers to the
+        // chainIndices array on each update. If this isn't the case, we're going to
+        // need to skip this update & warn.
+        // Resolve the previous update. If there is no previous update, then we can
+        // assume that the default chain indices are being used. These default chain
+        // indices are [1, 10, 137, 288, 42161] (outlined in UMIP-157)
+        const previousUpdate = this.chainIdIndicesUpdates.at(-1)?.value ?? [1, 10, 137, 288, 42161];
+        // We should now check that previousUpdate is a subset of chainIndices.
+        if (!previousUpdate.every((chainId) => chainIndices.includes(chainId))) {
```
nicholaspai, line 457: WDYT about a stricter check where we check `JSON.stringify(chainIndices.slice(0, previousUpdate.length)) === JSON.stringify(previousUpdate)`?

james-a-morris: This would require sorting both the `chainIndices` & `previousUpdate` & converting them both to strings. I believe this would result in the same comparison but with a larger time complexity

nicholaspai: resolved this offline

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -200,7 +215,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
         }
         // If any of the possible chains are not listed in this disabled chain update and are not already in the
         // enabled chain list, then add them to the list.
-        allPossibleChains.forEach((chainId) => {
+        this.getChainIdIndicesForBlock(disabledChainUpdate.blockNumber).forEach((chainId) => {
```
nicholaspai, line 218:
```suggestion
        enabledChainsAtFromBlock.forEach((chainId) => {
```

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -380,12 +377,11 @@ export async function getMatchedDeposit(
...
     const ubaActivationBlocks = getUbaActivationBundleStartBlocks(hubPoolClient);
-    const ubaActivationStartBlockForChain = getBlockForChain(
-      ubaActivationBlocks,
-      chain,
-      hubPoolClient.configStoreClient.enabledChainIds
-    );
-    return block >= ubaActivationStartBlockForChain;
+    const enabledChainsIndices = hubPoolClient.configStoreClient.getEnabledChains();
+    // Find the first activation block where the index matches the chain
+    const activationBlock =
```
nicholaspai, line 382: I'm not sure this is correct, can we just use `getBlockForChain` using `enabledChainsIndices`?

src/clients/HubPoolClient.ts
```diff
@@ -348,7 +348,7 @@ export class HubPoolClient extends BaseAbstractClient {
     chain: number,
     chainIdListOverride?: number[]
   ): number | undefined {
-    const chainIdList = chainIdListOverride ?? this.configStoreClient.enabledChainIds;
+    const chainIdList = chainIdListOverride ?? this.configStoreClient.getEnabledChains();
```
nicholaspai, line 351:
```suggestion
    const chainIdList = chainIdListOverride ?? this.configStoreClient.getEnabledChains(latestMainnetBlock);
```

src/clients/mocks/MockConfigStoreClient.ts
james-a-morris: Follwed up here: https://github.com/across-protocol/sdk-v2/pull/346/commits/9533b39849528443840d6329d3249144017bb2db

src/clients/UBAClient/UBAClientUtilities.ts
nicholaspai: Actually this should be fine

PR #355, opened 2023-08-04
feat: define protocol defaults for chain id indices

james-a-morris:


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -445,13 +446,22 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
           }
+          // Let's also check that the array doesn't contain any duplicates.
+          if (new Set(chainIndices).size !== chainIndices.length) {
+            this.logger.warn({
+              at: "ConfigStore",
+              message: `The array ${chainIndices} contains duplicates making it invalid.`,
+            });
+            // If not a valid array, skip.
+            continue;
+          }
```
james-a-morris, line 457: This adds a check against duplicate values

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -445,13 +446,22 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
             // If not a valid array, skip.
             continue;
           }
+          // Let's also check that the array doesn't contain any duplicates.
+          if (new Set(chainIndices).size !== chainIndices.length) {
```
mrice32, line 450: One other validation we might want to do is to make sure this array always starts with PROTOCOL_DEFAULT_CHAIN_ID_INDICES.

I think the UMIP specifies that this array can _only be added to_, and we might want to validate that so we don't end up inadvertently dropping a chain, for instance.

mrice32 (commented):
Looks good, just one comment!

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: Good call! I've accounted for this here: 

https://github.com/across-protocol/sdk-v2/blob/53d4c110b1b86bf8f06674426cc8d445f16d8393/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts#L458-L472

daywiss (approved):

PR #356, opened 2023-08-05
fix: Pass in chainIdIndices instead of enabledChainIds

nicholaspai:
There is a very important difference between calling `getEnabledChains` and `getChainIdIndices` on the `ConfigStoreClient`. The latter is used to index block numbers to chains while the former is used much more seldomly to limit which chains we care about for whatever operation.

This PR fixes critical bugs in the various clients that create block ranges. I discovered this bug while integration testing the dataworker and UBA client

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -347,7 +347,7 @@ export class HubPoolClient extends BaseAbstractClient {
     chain: number,
     chainIdListOverride?: number[]
   ): number | undefined {
-    const chainIdList = chainIdListOverride ?? this.configStoreClient.getEnabledChains(latestMainnetBlock);
+    const chainIdList = chainIdListOverride ?? this.configStoreClient.getChainIdIndicesForBlock(latestMainnetBlock);
     let endingBlockNumber: number | undefined;
     // Search proposed root bundles in reverse chronological order.
     for (let i = this.proposedRootBundles.length - 1; i >= 0; i--) {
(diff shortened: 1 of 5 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -347,7 +347,7 @@ export class HubPoolClient extends BaseAbstractClient {
     chain: number,
     chainIdListOverride?: number[]
   ): number | undefined {
-    const chainIdList = chainIdListOverride ?? this.configStoreClient.getEnabledChains(latestMainnetBlock);
+    const chainIdList = chainIdListOverride ?? this.configStoreClient.getChainIdIndicesForBlock(latestMainnetBlock);
```
james-a-morris, line 350: Should we remove the `chainIdListOverride` entirely in favor of enforcing the on-chain call?

james-a-morris (commented):
Initial pass - will pull this down now and test in sim mode

src/clients/HubPoolClient.ts
nicholaspai: Yea in prod we'd never override it but it might be useful for tests. Wdyt how do the tests look without this override option?

nicholaspai:
Discussed offline with James about logging clearer messages

src/clients/HubPoolClient.ts
james-a-morris: Yeah we should keep it in the end. A lot of things break and we'd need addtl functionality to get to a place where we are currently.

james-a-morris (approved):
The errors from the relayer are removed when this PR is integrated.

PR #365, opened 2023-08-09
improve(ConfigStoreClient): Add Chain ID awareness

pxrl:
This makes the ConfigStoreClient more compatible with testnet deployments, and helps to ignore obviously incorrect GlobalConfig updates impacting the chain ID indices array.

### DIFF
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -133,7 +133,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
    * @param ConfigStore chain ID.
    * @dev If the resolved chain ID is part of the default set, assume the protocol defaults.
    *      Otherwise, assume this is a test deployment with a lone chain ID.
-   * @dev The protocol defaults are [1, 10, 137, 288, 42161](outlined in UMIP-157).
+   * @dev The protocol defaults are [1, 10, 137, 288, 42161] (outlined in UMIP-157).
    * @dev chainId is marked optional to appease tsc. It must always be passed in.
    */
   protected implicitChainIdIndices(chainId?: number): number[] {
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -45,6 +45,7 @@ import { PROTOCOL_DEFAULT_CHAIN_ID_INDICES } from "../../constants";
 
 type _ConfigStoreUpdate = {
   success: true;
+  chainId: number;
   latestBlockNumber: number;
   searchEndBlock: number;
   events: {
@@ -84,6 +85,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
   public latestBlockNumber = 0;
 
   public hasLatestConfigStoreVersion = false;
+  public chainId: number | undefined;
 
   constructor(
     readonly logger: winston.Logger,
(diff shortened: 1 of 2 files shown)


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -126,6 +128,21 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+   * Resolve the implied set of chain ID indices based on the chain ID of the ConfigStore.
+   * @param ConfigStore chain ID.
+   * @dev If the resolved chain ID is part of the default set, assume the protocol defaults.
+   *      Otherwise, assume this is a test deployment with a lone chain ID.
+   * @dev The protocol defaults are [1, 10, 137, 288, 42161](outlined in UMIP-157).
+   * @dev chainId is marked optional to appease tsc. It must always be passed in.
+   */
+  protected implicitChainIdIndices(chainId?: number): number[] {
+    assert(isDefined(chainId), `ConfigStoreClient used before update`);
+    return PROTOCOL_DEFAULT_CHAIN_ID_INDICES[0] === chainId
```
pxrl, line 141: nb. This imposes an implicit assumption that the ConfigStore contact was deployed to the first chain in the array, which is also implicitly the Hub chain. FWIW I think that's OK.

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -269,6 +287,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
   }
 
   async _update(): Promise<ConfigStoreUpdate> {
+    const chainId = this.chainId ?? (await this.configStore.provider.getNetwork()).chainId;
```
pxrl, line 290: Because we consciously use a StaticJsonRpcProvider, this should not result in an external RPC call.

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -126,6 +128,19 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     return across.rateModel.parseAndReturnRateModelFromString(config.routeRateModel[route]);
   }
 
+  /**
+   * Resolve the implied set of chain ID indices based on the chain ID of the ConfigStore.
+   * @param ConfigStore chain ID.
```
pxrl, line 133:
```suggestion
   * @param chainId Chain ID of the ConfigStore.
```

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
nicholaspai: can we still Promise.all it with the below call?

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -126,6 +128,19 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+   * Resolve the implied set of chain ID indices based on the chain ID of the ConfigStore.
+   * @param chainId Chain ID of the ConfigStore.
+   * @dev If the resolved chain ID is part of the default set, assume the protocol defaults.
+   *      Otherwise, assume this is a test deployment with a lone chain ID.
+   * @dev The protocol defaults are [1, 10, 137, 288, 42161] (outlined in UMIP-157).
+   * @dev chainId is marked optional to appease tsc. It must always be passed in.
+   */
+  protected implicitChainIdIndices(chainId?: number): number[] {
+    assert(isDefined(chainId), "ConfigStoreClient used before update");
+    return PROTOCOL_DEFAULT_CHAIN_ID_INDICES[0] === chainId ? PROTOCOL_DEFAULT_CHAIN_ID_INDICES : [chainId];
```
nicholaspai, line 141: I don't fully understand this logic, why should the first default chain ID index be [chainId]? Does that make sense? maybe it does

nicholaspai (approved):

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
pxrl: As I see it, if we detect that we're on a deployment to anything other than chain ID 1, it means we're in some test/alt environment. In that case, all assumptions are thrown out the window and the ConfigStoreClient has no option but to assume it's the only chain ID in existence.

It might seem counter-intuitive to require that the caller to specify `chainId` as an input argument, rather than using `this.chainId`. The reason I did this is that we update the class state variables at the end of the `ConfigStoreClient.update()` method, but we need to call `implicitChainIdIndices()` in the middle of the `_update()` method. So in the case where we actually need to know the implicit chain IDs, `this.chainId` is actually undefined. So basically, it's done this way to be consistent with the rest of the implementation.

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
pxrl: Yeah, that's also fine and more consistent. I will apply that change.

pxrl: Implemented here: https://github.com/across-protocol/sdk-v2/pull/365/commits/ee39a86be93d0507766ef5c1755a91572ba61e7e

PR #367, opened 2023-08-10
improve(zkSync): Add zkSync-Görli relayFeeCalculator

pxrl:
Useful for testing fills on zkSync-Görli.

### DIFF
--- a/src/relayFeeCalculator/chain-queries/zksync.ts
+++ b/src/relayFeeCalculator/chain-queries/zksync.ts
@@ -26,3 +26,27 @@ export class ZkSyncQueries extends QueryBase {
     );
   }
 }
+
+export class zkSyncGoerliQueries extends QueryBase {
+  constructor(
+    provider: providers.Provider,
+    symbolMapping = TOKEN_SYMBOLS_MAP,
+    spokePoolAddress = "0x863859ef502F0Ee9676626ED5B418037252eFeb2",
+    usdcAddress = "0x0faF6df7054946141266420b43783387A78d82A9",
+    simulatedRelayerAddress = "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D",
+    coingeckoProApiKey?: string,
+    logger: Logger = DEFAULT_LOGGER,
+    gasMarkup = 0
+  ) {
+    super(
+      provider,
+      symbolMapping,
+      spokePoolAddress,
+      usdcAddress,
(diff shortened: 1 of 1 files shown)


src/relayFeeCalculator/chain-queries/zksync.ts
```diff
@@ -26,3 +26,27 @@ export class ZkSyncQueries extends QueryBase {
     );
   }
 }
+
+export class zkSyncGoerliQueries extends QueryBase {
```
james-a-morris, line 30: Should we comment on this class/file?

james-a-morris (approved):

src/relayFeeCalculator/chain-queries/zksync.ts
```diff
@@ -26,3 +26,27 @@ export class ZkSyncQueries extends QueryBase {
...
+  constructor(
+    provider: providers.Provider,
+    symbolMapping = TOKEN_SYMBOLS_MAP,
+    spokePoolAddress = "0x863859ef502F0Ee9676626ED5B418037252eFeb2",
```
nicholaspai, line 34: Can we load any of these variables from existing files? Like TOKEN_SYMBOLS_MAP?

Or at least comment where we find these from so in the future we can more easily modify

pxrl: This is a good call - I'd been meaning to do that later, but it's a really good idea to get it done now. The first part is here: https://github.com/across-protocol/sdk-v2/pull/367/commits/69e8ea01971ec504d9f1ebce3f31206f7ac78209

I'll follow up with a commit to make the SpokePool address resolution dynamic as well.

src/relayFeeCalculator/chain-queries/zksync.ts
pxrl: Done: https://github.com/across-protocol/sdk-v2/pull/367/commits/dcde52c57286e8fdedc8e52bf7a806c99c687a12

james-a-morris (commented):
Looks good, but would it make sense to either remove the external call to `chainId` and instead rely on `provider.chainId`?

pxrl:
> Looks good, but would it make sense to either remove the external call to `chainId` and instead rely on `provider.chainId`?

Ultimately yes, but I think we'd need to change the interface because it's technically a blocking (or blockable) call to retrieve the chainId, so it can't be done in the constructor.

PR #370, opened 2023-08-11
fix(queries): browser compatibility + use chain id enum

dohaki:
The import path
```js
import { getDeployedAddress } from "@across-protocol/contracts-v2";
```
broke the FE because the package is not tree shakable and contains dependencies that are not compatible with a browser env. We fix this by importing the `getDeployedAddress` directly from the source file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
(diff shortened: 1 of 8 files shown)


nicholaspai (approved):

pxrl (approved):
LGTM

mrice32 (approved):
LGTM!

PR #374, opened 2023-08-15
improve: use deployment rather than hardcoded addr

james-a-morris:


nicholaspai (approved):

src/relayFeeCalculator/chain-queries/zksync.ts
```diff
@@ -4,12 +4,13 @@ import { providers } from "ethers";
 import { TOKEN_SYMBOLS_MAP, CHAIN_IDs } from "../../constants";
 import { DEFAULT_SIMULATED_RELAYER_ADDRESS } from "./baseQuery";
 import QueryBase from "./baseQuery";
+import { getDeployedAddress } from "@across-protocol/contracts-v2";
```
dohaki, line 7: This will break the FE unfortunately... Can you import from 
```js
import { getDeployedAddress } from "../../utils/DeploymentUtils";
```
instead? Like for example in https://github.com/across-protocol/sdk-v2/blob/7ef7316168d3e463d7b20a9236c410abfa685e77/src/relayFeeCalculator/chain-queries/arbitrum.ts#L1

dohaki (changes requested):
Left one mandatory change to not break FE

pxrl:
I think we need to bump contracts-v2 before this will work, since the current version of contracts-v2 doesn't contain any metadata for the zkSync deployments. I've been waiting for the Base work to wrap up over there before doing that.

james-a-morris:
@dohaki I resolved the imports to use the local version of `getDeployedAddress`. I also changed the base version which has been in this version already

dohaki (approved):

PR #377, opened 2023-08-22
fix: correct etherscan link & refactoring

james-a-morris:
Correct etherscan link template for Base. Currently, without the trailing `/` after the URL then our assumptions break for logging. Also, `mainnet.basescan.org` is now defunct replaced by just `basescan.org`

This PR now also includes a refactoring to bring `./formattingUtils` from the relayer into the SDK.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.15.18",
+  "version": "0.15.20",
   "license": "AGPL-3.0",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
@@ -23,9 +23,9 @@
     "prepare": "tsdx build && husky install",
     "size": "size-limit",
     "analyze": "size-limit --why",
-    "bump-version:major": "yarn version --major --no-git-tag-version --no-commit-hooks",
-    "bump-version:minor": "yarn version --minor --no-git-tag-version --no-commit-hooks",
-    "bump-version:patch": "yarn version --patch --no-git-tag-version --no-commit-hooks"
+    "bump-version:major": "yarn version --major --no-git-tag-version --no-commit-hooks && git commit -m 'chore: bump version' ./package.json",
+    "bump-version:minor": "yarn version --minor --no-git-tag-version --no-commit-hooks && git commit -m 'chore: bump version' ./package.json",
+    "bump-version:patch": "yarn version --patch --no-git-tag-version --no-commit-hooks && git commit -m 'chore: bump version' ./package.json"
   },
   "lint-staged": {
     "*.ts": "yarn lint"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.15.20",
+  "version": "0.15.19",
   "license": "AGPL-3.0",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -19,18 +19,18 @@ export const PROTOCOL_DEFAULT_CHAIN_ID_INDICES = [1, 10, 137, 288, 42161];
 export const PUBLIC_NETWORKS: { [chainId: number]: { name: string; etherscan: string } } = {
   1: {
     name: "mainnet",
-    etherscan: "https://etherscan.io/",
+    etherscan: "https://etherscan.io",
   },
-  5: { name: "goerli", etherscan: "https://goerli.etherscan.io/" },
-  10: { name: "optimism", etherscan: "https://optimistic.etherscan.io/" },
+  5: { name: "goerli", etherscan: "https://goerli.etherscan.i/" },
+  10: { name: "optimism", etherscan: "https://optimistic.etherscan.io" },
   137: {
     name: "polygon-matic",
-    etherscan: "https://polygonscan.com/",
+    etherscan: "https://polygonscan.com",
   },
-  324: { name: "zksync", etherscan: "https://explorer.zksync.io/" },
+  324: { name: "zksync", etherscan: "https://explorer.zksync.io" },
   8453: { name: "base", etherscan: "https://basescan.org" },
-  42161: { name: "arbitrum", etherscan: "https://arbiscan.io/" },
-  43114: { name: "avalanche", etherscan: "https://snowtrace.io/" },
-  84531: { name: "base-goerli", etherscan: "https://basescan.org/" },
-  421613: { name: "arbitrum-goerli", etherscan: "https://goerli.arbiscan.io/" },
+  42161: { name: "arbitrum", etherscan: "https://arbiscan.io" },
+  43114: { name: "avalanche", etherscan: "https://snowtrace.io" },
+  84531: { name: "base-goerli", etherscan: "https://basescan.org" },
+  421613: { name: "arbitrum-goerli", etherscan: "https://goerli.arbiscan.io" },
 };
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -21,7 +21,7 @@ export const PUBLIC_NETWORKS: { [chainId: number]: { name: string; etherscan: st
     name: "mainnet",
     etherscan: "https://etherscan.io",
   },
-  5: { name: "goerli", etherscan: "https://goerli.etherscan.i/" },
+  5: { name: "goerli", etherscan: "https://goerli.etherscan.io" },
   10: { name: "optimism", etherscan: "https://optimistic.etherscan.io" },
   137: {
     name: "polygon-matic",
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -31,6 +31,6 @@ export const PUBLIC_NETWORKS: { [chainId: number]: { name: string; etherscan: st
   8453: { name: "base", etherscan: "https://mainnet.basescan.org" },
   42161: { name: "arbitrum", etherscan: "https://arbiscan.io/" },
   43114: { name: "avalanche", etherscan: "https://snowtrace.io/" },
-  84531: { name: "base-goerli", etherscan: "https://basescan.org" },
+  84531: { name: "base-goerli", etherscan: "https://basescan.org/" },
   421613: { name: "arbitrum-goerli", etherscan: "https://goerli.arbiscan.io/" },
 };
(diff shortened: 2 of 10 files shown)


src/constants.ts
```diff
@@ -28,9 +28,9 @@ export const PUBLIC_NETWORKS: { [chainId: number]: { name: string; etherscan: st
     etherscan: "https://polygonscan.com/",
   },
   324: { name: "zksync", etherscan: "https://explorer.zksync.io/" },
-  8453: { name: "base", etherscan: "https://mainnet.basescan.org" },
+  8453: { name: "base", etherscan: "https://basescan.org" },
```
pxrl, line 31: There's no trailing `/` here, but fwiw I think it's a mistake to require it here - we should be inserting the `/` when we generate the link.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.15.18",
+  "version": "0.15.20",
```
pxrl, line 4: Intentional to skip 0.15.19?

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,165 @@
...
+  return createFormatFunction(2, 4, false, 16)(toBN(relayerFeePct).toString());
+};
+
+/**
+ * Creates an etherscan link for a transaction or address on a given network.
+ * @param txHashOrAddress The transaction hash or address to link to.
+ * @param chainId The chainId to link to.
+ * @returns The etherscan link. This link will be formatted in markdown.
+ */
+export function etherscanLink(txHashOrAddress: string, chainId: number | string): string {
```
pxrl, line 53: While we're at it, thoughts on this?

```suggestion
export function blockExplorerLink(txHashOrAddress: string, chainId: number | string): string {
```

package.json
james-a-morris: No - I was testing the new package.json `yarn bump-version:patch` which commits & updates the patch version.

james-a-morris: Resolved: https://github.com/across-protocol/sdk-v2/pull/377/commits/dcb7e14f5890ea4f0bd5cb329a761128d64d9838

src/constants.ts
james-a-morris: Resolved

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,165 @@
...
+  return _createEtherscanLinkMarkdown(txHashOrAddress, Number(chainId)) ?? "<>";
+}
+
+/**
+ * Resolves a domain to an etherscan link for a given network.
+ * @param networkId The network to link to.
+ * @returns The etherscan link. If the networkId is not supported, the default etherscan mainnet link will be returned.
+ */
+export function createEtherscanLinkFromTx(networkId: number): string {
```
pxrl, line 62: There's no input transaction here, so the name doesn't quite fit properly.

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,165 @@
...
+    return null;
+  }
+  const shortURLString = createShortHexString(hex);
+  // Transaction hash
+  if (hex.length == 66) {
+    return `<${createEtherscanLinkFromTx(chainId)}tx/${hex}|${shortURLString}>`;
+  }
+  // Account
+  else if (hex.length == 42) {
+    return `<${createEtherscanLinkFromTx(chainId)}address/${hex}|${shortURLString}>`;
```
pxrl, line 93: I'm a firm believer that we should be applying the trailing / here, since the hostname is totally separate from the subsequent path/resource.

```suggestion
    return `<${createEtherscanLinkFromTx(chainId)}/tx/${hex}|${shortURLString}>`;
  }
  // Account
  else if (hex.length == 42) {
    return `<${createEtherscanLinkFromTx(chainId)}/address/${hex}|${shortURLString}>`;
```

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Converts a value from wei to a decimal value with the specified number of decimal places.
+ * @param weiVal - The value in wei to convert.
+ * @param decimals - The number of decimal places to include in the converted value.
+ * @returns The converted value as a string.
+ */
+export function convertFromWei(weiVal: string, decimals: number): string {
+  const formatFunction = createFormatFunction(2, 4, false, decimals);
+  return formatFunction(weiVal);
+}
```
pxrl, line 146: Aren't we conflating the PR by bundling all this in here?

FWIW I'm happy to see it out of `src/utils/common.ts` but the BN stuff doesn't really have much to do with block explorers. In general the concept of `FormattingUtils` is fairly vague to me. I think this fits better in a dedicated BigNumber file instead.

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Shortens a lengthy hexadecimal string to a shorter version with an ellipsis in the middle.
+ * @param hex A hexadecimal string to be shortened.
+ * @returns The shortened hexadecimal string.
+ * @example shortenHexString("0x772871a444c6e4e9903d8533a5a13101b74037158123e6709470f0afbf6e7d94") -> "0x7787...7d94"
+ */
+export function createShortHexString(hex: string): string {
+  return hex.substring(0, 5) + "..." + hex.substring(hex.length - 6, hex.length);
```
pxrl, line 73: Consider allowing the width to be variable, as implemented here? https://github.com/across-protocol/relayer-v2/blob/e38356ed0496f85a44bca0461a34200a0139d388/scripts/spokepool.ts#L23

src/utils/FormattingUtils.ts
james-a-morris: I think we should change the function signature in a different PR.

src/utils/FormattingUtils.ts
james-a-morris: I think we should change the function signature in a different PR.

src/utils/FormattingUtils.ts
james-a-morris: I think we should change the function signature in a different PR.

src/utils/FormattingUtils.ts
james-a-morris: Refactored

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * Shortens a list of addresses to a shorter version with only the first 10 characters.
+ * @param addresses A list of addresses to shorten.
+ * @returns A list of shortened addresses.
+ * @see shortenHexString
+ */
+export function shortenHexStrings(addresses: string[]): string[] {
+  return addresses.map((address) => shortenHexString(address));
```
mrice32, line 102:
```suggestion
  return addresses.map(shortenHexString);
```

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+  return formatFunction(weiVal);
+}
+
+/**
+ * Shortens a list of addresses to a shorter version with only the first 10 characters.
+ * @param addresses A list of addresses to shorten.
+ * @returns A list of shortened addresses.
+ * @see shortenHexString
+ */
+export function shortenHexStrings(addresses: string[]): string[] {
```
mrice32, line 101: This is the only one that feels like it may be a _little_ overkill, since you could do this in a super short one-liner: `addresses.map(shortenHexString);`

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+  return addresses.map((address) => shortenHexString(address));
+}
+
+/**
+ * Shortens a hexadecimal string to a shorter version with only the first 10 characters.
+ * @param hexString A hexadecimal string to shorten.
+ * @returns The shortened hexadecimal string.
+ */
+export function shortenHexString(hexString: string): string {
```
mrice32, line 110: When would/do we use this over `createShortHexString` above?

mrice32 (commented):
Looks good! Just a few minor comments

mrice32 (approved):
@james-a-morris showed me where these were being moved from and that they were copy-pastes. LGTM!

pxrl (approved):
LGTM, subject to a couple of Matt's comments above :+1:

src/utils/FormattingUtils.ts
james-a-morris: Resolved in https://github.com/across-protocol/sdk-v2/pull/378

src/utils/FormattingUtils.ts
james-a-morris: Resolved in #377

src/utils/BlockExplorerUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+export function etherscanLink(txHashOrAddress: string, chainId: number | string): string {
+  return _createEtherscanLinkMarkdown(txHashOrAddress, Number(chainId)) ?? "<>";
+}
+
+/**
+ * Resolves a domain to an etherscan link for a given network.
+ * @param networkId The network to link to.
+ * @returns The etherscan link. If the networkId is not supported, the default etherscan mainnet link will be returned.
+ */
+export function createEtherscanLinkFromTx(networkId: number): string {
```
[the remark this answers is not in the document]
james-a-morris: Pulled from the relayer-v2 🙂

src/utils/FormattingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Converts a stringified number into a BigNumber. If the string version of the num contains a `.` then it is a number which needs to be parsed to a string int.
+ * @param num The number to parse.
+ * @returns The parsed BigNumber.
+ * @note This is a temporary function until we can backport support for decimal points to @across-protocol/sdk-v2.
+ */
+export const toBN = (num: BigNumberish): BN => {
+  // If the string version of the num contains a `.` then it is a number which needs to be parsed to a string int.
+  if (num.toString().includes(".")) {
+    return BigNumber.from(parseInt(num.toString()));
+  }
```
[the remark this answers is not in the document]
james-a-morris: This is an extension of the basecase of `toBN` that exists in the relayer. I've updated it here with more functionality: #378

src/utils/FormattingUtils.ts
james-a-morris: #378

src/utils/FormattingUtils.ts
james-a-morris: #378

src/utils/FormattingUtils.ts
james-a-morris: #378

PR #378, opened 2023-08-23
improve: post refactoring optimizations

james-a-morris:
This PR aims to reconcile the changes requested by both @mrice32 and @pxrl in the previous PR (#377).

These changes include:

- https://github.com/across-protocol/sdk-v2/pull/377#discussion_r1302202280
- https://github.com/across-protocol/sdk-v2/pull/377#discussion_r1302212074
- https://github.com/across-protocol/sdk-v2/pull/377#discussion_r1302235530
- https://github.com/across-protocol/sdk-v2/pull/377#discussion_r1302250483
- https://github.com/across-protocol/sdk-v2/pull/377#discussion_r1302251513
- https://github.com/across-protocol/sdk-v2/pull/377#discussion_r1302207317

### DIFF
--- a/src/utils/BlockExplorerUtils.ts
+++ b/src/utils/BlockExplorerUtils.ts
@@ -1,5 +1,7 @@
+import { ethers } from "ethers";
 import { PUBLIC_NETWORKS } from "../constants";
 import { createShortHexString } from "./FormattingUtils";
+import { isDefined } from "./TypeGuards";
 
 /**
  * Creates a block explorer link for a transaction or address on a given network.
@@ -16,12 +18,29 @@ export function blockExplorerLink(txHashOrAddress: string, chainId: number | str
  * @param networkId The network to link to.
  * @returns The block explorer link. If the networkId is not supported, the default block explorer mainnet link will be returned.
  */
-export function resolveBlockExplorerDomain(networkId: number): string {
-  const explorerLink = PUBLIC_NETWORKS[networkId]?.etherscan;
-  if (!explorerLink) {
-    throw new Error(`Block explorer link does not exist for networkId: ${networkId}`);
-  }
-  return explorerLink;
+export function resolveBlockExplorerDomain(networkId: number): string | undefined {
+  return PUBLIC_NETWORKS[networkId]?.etherscan;
+}
+
+/**
+ * Constructs a URL for a given list of domain parts.
+ * @param domain The domain to construct a URL for.
+ * @param parts The relative path of parts to append to the domain. These parts will be joined with a "/".
+ * @returns The constructed URL.
+ * @see resolveBlockExplorerDomain
+ * @example constructURL("https://example.com", ["path", "to", "resource"]) => "https://example.com/path/to/resource"
+ * @example constructURL("https://example.com", ["path", "to", "resource", ""]) => "https://example.com/path/to/resource"
+ * @example constructURL("https://example.com", ["path", "to//", "resource",]) => "https://example.com/path/to/resource"
+ */
+function constructURL(domain: string, parts: string[]): string {
+  // Further split the parts by "/" to handle any parts that contain multiple "/".
+  parts = parts.flatMap((p) => p.split("/"));
+  // Remove any empty parts.
+  parts = parts.filter((p) => p !== "");
+  // Join the parts with a "/".
+  const path = parts.join("/");
+  // Remove any trailing "/".
+  return `${domain}/${path}`.replace(/\/+$/, "");
 }
 
 /**
@@ -31,21 +50,31 @@ export function resolveBlockExplorerDomain(networkId: number): string {
  * @returns A formatted markdown block explorer link to the given transaction hash or address on the given chainId.
  */
 function _createBlockExplorerLinkMarkdown(hex: string, chainId = 1): string | null {
-  try {
-    if (hex.substring(0, 2) != "0x") {
-      return null;
+  // Attempt to resolve the block explorer domain for the given chainId.
+  const explorerDomain = resolveBlockExplorerDomain(chainId);
+  // If the chainId is not supported, return an unsupported link.
+  if (!isDefined(explorerDomain)) {
+    return `<unsupported chain/hash ${chainId}:${hex}>}`;
+  }
+  // Ensure that the first two characters are "0x". If they are not, append them.
+  if (hex.substring(0, 2) !== "0x") {
+    hex = `0x${hex}`;
+  }
+  // Ensure that the hex string is a valid hexadecimal string.
+  if (!ethers.utils.isHexString(hex)) {
+    return null;
+  }
+  // Resolve the short URL string.
+  const shortURLString = createShortHexString(hex);
+  // Iterate over the two possible hex lengths.
+  for (const [length, route] of [
+    [66, "tx"],
+    [42, "address"],
+  ] as [number, string][]) {
+    // If the hex string is the correct length, return the link.
+    if (hex.length === length) {
+      return `<${constructURL(explorerDomain, [route, hex])}|${shortURLString}>`;
     }
-    const shortURLString = createShortHexString(hex);
-    // Transaction hash
-    if (hex.length == 66) {
-      return `<${resolveBlockExplorerDomain(chainId)}/tx/${hex}|${shortURLString}>`;
-    }
-    // Account
-    else if (hex.length == 42) {
-      return `<${resolveBlockExplorerDomain(chainId)}/address/${hex}|${shortURLString}>`;
-    }
-  } catch (e) {
-    throw new Error(`Could not create block explorer link. ChainId: ${chainId}, Hex: ${hex}`);
   }
   return null;
 }
@@ -57,6 +86,6 @@ function _createBlockExplorerLinkMarkdown(hex: string, chainId = 1): string | nu
  * @returns A list of formatted markdown block explorer links to the given transaction hashes or addresses on the given chainId.
  * @see blockExplorerLink
  */
-export function blockExplorer(txHashesOrAddresses: string[], chainId: number | string): string {
+export function blockExplorerLinks(txHashesOrAddresses: string[], chainId: number | string): string {
   return txHashesOrAddresses.map((hash) => `${blockExplorerLink(hash, chainId)}\n`).join("");
 }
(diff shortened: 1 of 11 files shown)


src/utils/FormattingUtils.ts
```diff
@@ -27,14 +27,19 @@ export function toGWei(num: BigNumberish): BN {
...
  */
-export const toBN = (num: BigNumberish): BN => {
+export const toBN = (num: BigNumberish, rounding: "floor" | "round" | "ceil" = "floor"): BN => {
   // If the string version of the num contains a `.` then it is a number which needs to be parsed to a string int.
   if (num.toString().includes(".")) {
-    return BigNumber.from(parseInt(num.toString()));
+    // Resolve a rounding function from the rounding parameter.
+    const roundingFunction = Math[rounding];
+    // Parse the number to a string float and round it.
+    return BigNumber.from(roundingFunction(parseFloat(num.toString())));
```
mrice32, line 40: If I'm not mistaken, this can lose precision _on the part before the decimal_ if we convert to a javascript number. I would not want to use this function anywhere near sensitive fee calculations that must always be repeatable and correct unless we can be sure that the integer part of the calculation never loses precision.

> Unlike many other programming languages, JavaScript does not define different types of numbers, like integers, short, long, floating-point etc. JavaScript numbers are always stored as double precision floating point numbers, following the international IEEE 754 standard.

> Integers can only be represented without loss of precision in the range -253 + 1 to 253 - 1, inclusive (obtainable via [Number.MIN_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER) and [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)), because the mantissa can only hold 53 bits (including the leading 1).

To avoid precision loss in the integer portion, we can do something like (I haven't checked this):
```ts
const [integer, decimal, ...rest] = num.toString().split(".");
assert(rest.length === 0, `invalid number ${num.toString()}`);
// Resolve a rounding function from the rounding parameter.
const roundingFunction = Math[rounding];
// Parse the decimal to a string float with '0.' in front, round it, and then add it.
// This isolates the precision loss to the part _after_ the decimal.
return BigNumber.from(integer).add(BigNumber.from(roundingFunction(parseFloat(`0.${decimal}`)));
```

Note: this does not eliminate all precision loss, but the worst that precision loss in the decimal could do is cause the output to be rounded the wrong way. Not sure if this is an issue, but it should reduce the likelihood. In general, we still need to be very careful about using this function due to the possible loss of precision and unpredictable rounding.

mrice32 (approved):
LGTM! Just one comment on precision loss

src/utils/FormattingUtils.ts
```diff
@@ -47,10 +52,28 @@ export const formatFeePct = (relayerFeePct: BigNumber): string => {
...
+  // We can simulate rounding by adding 0.5 to the integer. If
+  // we had an odd division, the floor will add one additional
+  // character to the left side.
+  const leftCharacters = Math.floor(maxLength / 2 + 0.5);
+  // A simple floor division between the max character length
+  const rightCharacters = Math.floor(maxLength / 2);
+  // Combine the two sides.
+  const result = `${hex.substring(0, leftCharacters)}...${hex.substring(hex.length - rightCharacters)}`;
+  // Append the prefix if it was originally provided
+  return (hasPrefix ? "0x" : "") + result;
```
pxrl, line 76: As a user of this function, I'd intuitively expect that `maxLength` limits the output length of the resulting string - i.e. including including the `...`separator and optional `0x` prefix. If it does this, then there'd actually be no need to detect and strip off any leading `0x`. I would concede that the name `createShortHexString` seems a bit off though, since I'm describing something that works at the string level, and is basically ignorant of hexadecimal notation...

src/utils/DeploymentUtils.ts
```diff
@@ -1,5 +1,7 @@
-export {
+import * as contracts from "@across-protocol/contracts-v2";
```
dohaki, line 1: This will again break the FE bundle unfortunately...

dohaki: The problem is that the `@across-protocol/contracts-v2` package contains modules that don't work in a browser env. In the [main entry file](https://github.com/across-protocol/contracts-v2/blob/master/index.ts) of the package, we also export everything via `export * from "..."`. This results in the behavior that nothing can be tree-shaken as soon as you import something from the "default" path `@across-protocol/contracts-v2`. That on the other hand, resulted in browser-incompatible modules being bundled into the SDK. But because the SDK can also not be tree-shaken (due to tsdx's default behavior to put everything into one file), building the frontend bundle fails.

So we used the workaround to only import browser-compatible modules via something like `@across-protocol/contracts-v2/dist/<MODULE>`. This would omit the browser-incompatible imports from the SDK bundle.

james-a-morris: Reverted. Thanks for continuing to call this out @dohaki

src/utils/FormattingUtils.ts
james-a-morris: Done - all precision is completed via bignumber comparisons: https://github.com/across-protocol/sdk-v2/pull/378/commits/aa34de8007b4ef122919125aebbd57b2bc4b334f

src/utils/FormattingUtils.ts
james-a-morris: Happy to dive into this if you feel there's anything actionable 🙂

pxrl: I do think we should be enforcing that the output string is no longer than `maxLength` characters, so the resulting string should be accounting for "0x" (if present) and "..." in the middle.

In general I think it'd be simpler if the code was just ignorant of whether the 0x prefix was there or not.

james-a-morris: Good call - resolved

src/utils/BlockExplorerUtils.ts
```diff
@@ -2,53 +2,53 @@ import { PUBLIC_NETWORKS, DEFAULT_BLOCKCHAIN_EXPLORER_DOMAIN } from "../constant
...
- * Resolves a domain to an etherscan link for a given network.
+ * Resolves a domain to an block explorer link for a given network.
  * @param networkId The network to link to.
- * @returns The etherscan link. If the networkId is not supported, the default etherscan mainnet link will be returned.
+ * @returns The block explorer link. If the networkId is not supported, the default block explorer mainnet link will be returned.
  */
-export function createEtherscanLinkFromTx(networkId: number): string {
+export function resolveBlockExplorerDomain(networkId: number): string {
   return PUBLIC_NETWORKS[networkId]?.etherscan ?? DEFAULT_BLOCKCHAIN_EXPLORER_DOMAIN;
 }
```
pxrl, line 21: Minor observation for future consideration (i.e. probably not to be changed in this PR): the existing behaviour of defaulting to etherscan doesn't make sense to me, since if we don't recognise the chainId then any etherscan link will just resolve to an etherscan error page.

I think it'd be fine to throw (or return undefined) and then handle that within the caller to produce some meaningful text (...and include the txn hash).

src/utils/FormattingUtils.ts
```diff
@@ -27,14 +27,25 @@ export function toGWei(num: BigNumberish): BN {
...
+    // can just return the integer. However, if it has a decimal point with decimal
+    // places, then we can automatically round up if ceil is specified or if round is
+    // specified and the first decimal is greater than or equal to 5.
+    const roundUp = decimal.length > 0 && (rounding === "ceil" || (rounding === "round" && parseInt(decimal[0]) >= 5));
+    // If we need to round up, we can just add 1 to the integer.
+    return BigNumber.from(integer).add(roundUp ? 1 : 0);
   }
+  // Otherwise, it is a string int and we can parse it directly.
   return BigNumber.from(num.toString());
 };
```
pxrl, line 50: For follow up; do we have any test cases for this implementation?

pxrl (approved):
LGTM but it'd be safer if we could fall back on some toBN() test cases, given the change in implementation (these could also be split out in order to get the rest in, since they don't relate to the block explorer work).

src/utils/FormattingUtils.ts
james-a-morris: Done https://github.com/across-protocol/sdk-v2/pull/378/commits/20546de00b564e3b3c84d1dd05177278db7a1f88

src/utils/BlockExplorerUtils.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/378/commits/31f986b365bc6f606e71a6d009418841a0d85df5

src/utils/BlockExplorerUtils.ts
```diff
@@ -27,17 +31,21 @@ export function resolveBlockExplorerDomain(networkId: number): string {
...
+      return `<${resolveBlockExplorerDomain(chainId)}/tx/${hex}|${shortURLString}>`;
+    }
+    // Account
+    else if (hex.length == 42) {
+      return `<${resolveBlockExplorerDomain(chainId)}/address/${hex}|${shortURLString}>`;
+    }
+  } catch (e) {
+    throw new Error(`Could not create block explorer link. ChainId: ${chainId}, Hex: ${hex}`);
   }
   return null;
```
pxrl, line 50: Small proposal to avoid it blowing up when we're adding new chains:

```suggestion
  const unsupported = `<unsupported hash (${hex})>`;
  if (hex.substring(0, 2) != "0x") {
    return unsupported;
  }
  
  try {
    const shortURLString = createShortHexString(hex);
    // Transaction hash
    if (hex.length == 66) {
      return `<${resolveBlockExplorerDomain(chainId)}/tx/${hex}|${shortURLString}>`;
    }
    // Account
    else if (hex.length == 42) {
      return `<${resolveBlockExplorerDomain(chainId)}/address/${hex}|${shortURLString}>`;
    }
  } catch (e) {
    return `<unsupported chainId (${chainId}), hash: ${hex}>`;
  }
  return unsupported;
```

pxrl (approved):
This is good to go. It _could_ optionally be a bit less rigid on unrecognised chains, since it might be annoying to be forced to fix this in the middle of adding a new chain, but it's not a deal-breaker.

PR #379, opened 2023-08-24
improve: reintroduce hub pool caching

james-a-morris:
This PR re-introduces the hub pool caching mechanism for storing utilization data in an external cache (i.e. redis). A subsequent PR will need to be introduced in relayer-v2 to make use of this functionality. Otherwise, it is completely pass through in the current state.

This functionality relies on the agnostic caching interface that the relayer can use to provide its own interpretation of a cache that exposes get and set functions.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -39,6 +39,7 @@ import * as lpFeeCalculator from "../lpFeeCalculator";
 import { AcrossConfigStoreClient as ConfigStoreClient } from "./AcrossConfigStoreClient/AcrossConfigStoreClient";
 import { BaseAbstractClient } from "./BaseAbstractClient";
 import { isUBAActivatedAtBlock } from "./UBAClient/UBAClientUtilities";
+import { DEFAULT_CACHING_TTL } from "../constants";
 
 type _HubPoolUpdate = {
   success: true;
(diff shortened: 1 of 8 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -254,10 +258,39 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    if (isDefined(result)) {
+      const [current, post] = result.split(",").map(BigNumber.from);
+      return { current, post };
+    }
+    // We were not able to find a valid result, so let's resolve the function
+    // and store the result in the cache
+    else {
+      const { current, post } = await resolver();
+      // First determine if we should cache the result
+      if (shouldCache(getCurrentTime(), timestamp, 60 * 60 * 24 * 90)) {
```
nicholaspai, line 287: So just to confirm, you are saying we should only cache utilization data that is at least 90 days old?

james-a-morris: Correct - I've modified it to at least one day old. This should cache more items while reducing reorg risk

src/clients/HubPoolClient.ts
```diff
@@ -254,10 +258,39 @@ export class HubPoolClient extends BaseAbstractClient {
...
+      return { current, post };
+    }
+    // We were not able to find a valid result, so let's resolve the function
+    // and store the result in the cache
+    else {
+      const { current, post } = await resolver();
+      // First determine if we should cache the result
+      if (shouldCache(getCurrentTime(), timestamp, 60 * 60 * 24)) {
+        // If we should cache the result, then let's store it
+        await cache.set(key, `${current.toString()},${post.toString()}`, 60 * 60 * 24);
```
nicholaspai, line 289: As discussed offline, setting the TTL to 1 day doesn't seem right here. I think the TTL should be as old as the expected bot's lookback period, so up to 14 days

james-a-morris: Great call - this is now resolved.

nicholaspai (approved):

PR #380, opened 2023-08-30
improve: Export useful ethers constants

pxrl:
Using these can deliver a small benefit over spinning up new BNs for common numbers like 0 and 1.


james-a-morris (approved):
This is quite helpful so as to rely on existing constants.

PR #382, opened 2023-09-08
chore: migrate from tsdx

james-a-morris:
This PR removes TSDX from the SDK-V2


package.json
```diff
@@ -13,14 +13,14 @@
     "node": ">=16.18.0"
   },
   "scripts": {
-    "start": "tsdx watch",
-    "build": "tsdx build",
-    "test": "tsdx test",
-    "test:watch": "tsdx test --watch",
-    "test:e2e": "tsdx test --config jest-e2e-config.json",
+    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build --incremental'",
+    "build": "tsc --project ./tsconfig.build.json",
```
dohaki, line 17: Currently, this will only build a `CommonJS` module. But we should also target an `ESM` module. We could use a similar approach to here https://github.com/wagmi-dev/viem/blob/main/package.json#L9-L11 and add additional build commands. These would make use of the CLI flags `--module` and `--outDir`

dohaki (commented):
I think we still need to add some more tweaks to fully support ESM and CommonJS with the new setup. The wagmi/viem people are doing a lot of things right IMO with the packaging approach, so we could be inspired by their setup:
- [ ] Separate build step for commonjs -> outputs to `./dist/cjs`
- [ ] Separate build step for esm -> outputs to `./dist/esm`
- [ ] Separate build step of types -> outputs to `./dist/types`
- [ ] Specify correct entrypoints in `package.json` file:
```
  "type": "module",
  "main": "./dist/cjs/index.js",
  "module": "./dist/esm/index.js",
  "types": "./dist/types/index.d.ts",
  "typings": "./dist/types/index.d.ts",
```
- [ ] (Optional) Specify alternative entrypoints to submodules like:
```
    ".": {
      "types": "./dist/types/index.d.ts",
      "import": "./dist/esm/index.js",
      "default": "./dist/cjs/index.js"
    },
    "./abi": {
      "types": "./dist/types/abi.d.ts",
      "import": "./dist/esm/abi.js",
      "default": "./dist/cjs/abi.js"
    },
    "./accounts": {
      "types": "./dist/types/accounts/index.d.ts",
      "import": "./dist/esm/accounts/index.js",
      "default": "./dist/cjs/accounts/index.js"
    },
```

dohaki:
@james-a-morris If the requested changes in this [comment](https://github.com/across-protocol/sdk-v2/pull/382#pullrequestreview-1619126783) are too much in your opinion, then we can also try to add these changes successively. But not sure if dropping ESM support in this PR will break other integrators or the frontend

package.json
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/382/commits/f8f57164632dc2b38ae659018942bef31458b0d8

dohaki (commented):
Nice looks good now! I am trying to test it locally though and build the frontend with a linked package but failing to do so... Did you have any success?

package.json
```diff
@@ -13,16 +11,20 @@
...
+    "test": "jest",
+    "test:watch": "jest --watch",
+    "test:e2e": "jest --config jest-e2e-config.json",
     "lint": "eslint --fix src",
     "lint-check": "eslint src",
-    "prepare": "tsdx build && husky install",
+    "prepare": "tsc --project ./tsconfig.build.json && husky install",
     "size": "size-limit",
     "analyze": "size-limit --why",
+    "clean": "rimraf dist",
```
pxrl, line 27: Are there advantages to using rimraf? I don't see it as a direct dependency, so we're inheriting it from some other dependency.

FWIW, chaining `mv ... && rm -rf ... &` is usually pretty quick because it cleans up with having to wait on the deletion to finish.

```suggestion
    "clean": "mv dist .dist && rm -rf .dist &",
```

package.json
```diff
@@ -13,16 +11,20 @@
     "node": ">=16.18.0"
   },
   "scripts": {
-    "start": "tsdx watch",
-    "build": "tsdx build",
-    "test": "tsdx test",
-    "test:watch": "tsdx test --watch",
-    "test:e2e": "tsdx test --config jest-e2e-config.json",
+    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build --incremental'",
```
pxrl, line 14: typescript seems to support this natively; is nodemon doing something that tsc doesn'tdo (or doesn't do _well_) ?

https://www.typescriptlang.org/docs/handbook/configuring-watch.html

package.json
james-a-morris: The primary advantage is that the module is agnostic of OS. The command effectively has fallbacks for POSIX & Windows

package.json
james-a-morris: Normally this would be acceptable but we'd need to repeat multiple commands to build all with `--watch` whereas `nodemon` sits outside of the compiler and can issue arbitrary commands on file change.

package.json
pxrl: We don't target Windows as a platform and rm is POSIX, so is it worth depending on a package just for this?

package.json
pxrl: Ah, so this means that we'll be using it in more places than just `yarn start` in future?

james-a-morris: Yep!

package.json
james-a-morris: Done https://github.com/across-protocol/sdk-v2/pull/382/commits/80599f4aa5d76861e754bc034267dfdaf79edeff

I've kept this as a synchronous call since we'll need to ensure dist is deleted before rebuilding

james-a-morris:
@dohaki The FE should now compile after this PR: https://github.com/across-protocol/frontend-v2/pull/845

.gitignore
```diff
@@ -22,3 +22,5 @@ dist
 .env.development.local
 .env.test.local
 .env.production.local
+
+tsconfig.build.tsbuildinfo
```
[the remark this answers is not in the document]
james-a-morris: Resolved: https://github.com/across-protocol/sdk-v2/pull/382/commits/01b75cb22a30c326168c7faf4faaff52197efb2b

dohaki (approved):
Nice tested on the FE with this branch https://github.com/across-protocol/frontend-v2/pull/847#pullrequestreview-1636316699 and works!

PR #383, opened 2023-09-08
fix: resolve bug in spoke pool deposit search

james-a-morris:
There is a bug in the binary search function of the SpokePoolClient where, if the deposit ID is outside the range, then it still returns a non-failing value. This PR changes that by adding an additional safety check to ensure that the largest deposit ID is greater than or equal to the target search.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,53 +1,53 @@
 import { Contract, providers } from "ethers";
 import { groupBy } from "lodash";
+import winston from "winston";
 import {
-  assign,
-  EventSearchConfig,
-  DefaultLogLevels,
-  MakeOptional,
-  mapAsync,
   AnyObject,
+  DefaultLogLevels,
+  EventSearchConfig,
   MAX_BIG_INT,
+  MakeOptional,
+  assign,
+  mapAsync,
   stringifyJSONWithNumericString,
   toBN,
 } from "../utils";
-import { validateFillForDeposit, filledSameDeposit } from "../utils/FlowUtils";
 import {
-  spreadEvent,
   paginatedEventQuery,
-  spreadEventWithBlockNumber,
   sortEventsAscending,
   sortEventsAscendingInPlace,
+  spreadEvent,
+  spreadEventWithBlockNumber,
 } from "../utils/EventUtils";
-import winston from "winston";
+import { filledSameDeposit, validateFillForDeposit } from "../utils/FlowUtils";
 
 import { BigNumber, Event, EventFilter, ethers } from "ethers";
 
+import { ZERO_ADDRESS } from "../constants";
 import {
   Deposit,
   DepositWithBlock,
+  DepositWithBlockStringified,
   Fill,
   FillWithBlock,
+  FillWithBlockStringified,
+  FundsDepositedEvent,
+  FundsDepositedEventStringified,
   RefundRequestWithBlock,
+  RefundRequestWithBlockStringified,
   RelayerRefundExecutionWithBlock,
+  RelayerRefundExecutionWithBlockStringified,
   RootBundleRelayWithBlock,
   SpeedUp,
-  TokensBridged,
-  FundsDepositedEvent,
-  DepositWithBlockStringified,
-  FillWithBlockStringified,
   SpeedUpStringified,
+  TokensBridged,
   TokensBridgedStringified,
-  RelayerRefundExecutionWithBlockStringified,
-  FundsDepositedEventStringified,
-  RefundRequestWithBlockStringified,
 } from "../interfaces";
-import { HubPoolClient } from "./HubPoolClient";
-import { ZERO_ADDRESS } from "../constants";
-import { getNetworkName } from "../utils/NetworkUtils";
-import { BaseAbstractClient } from "./BaseAbstractClient";
-import { getBlockRangeForDepositId, getDepositIdAtBlock } from "../utils/SpokeUtils";
 import { SpokePool } from "../typechain";
+import { getNetworkName } from "../utils/NetworkUtils";
+import { getBlockRangeForDepositId, getDepositIdAtBlock } from "../utils/SpokeUtils";
+import { BaseAbstractClient } from "./BaseAbstractClient";
+import { HubPoolClient } from "./HubPoolClient";
 
 type Block = providers.Block;
 
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -438,11 +438,22 @@ export class SpokePoolClient extends BaseAbstractClient {
     low: number;
     high: number;
   }> {
+    console.log("NEXT");
+
     if (initLow > initHigh) {
       throw new Error("Binary search failed because low > high");
     }
+    if (initLow < 0 || initHigh < 0) {
+      throw new Error("Binary search failed because low or high < 0");
+    }
     if (maxSearches <= 0) {
       throw new Error("maxSearches must be > 0");
+    }
+
+    // If the deposit ID at the initial high block is less than the target deposit ID, then we know that
+    // the target deposit ID must be greater than the initial high block, so we can throw an error.
+    if ((await this._getDepositIdAtBlock(initHigh)) < targetDepositId) {
+      throw new Error("Failed to find deposit ID");
     }
 
     let low = initLow;
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,4 +1,4 @@
-import { providers } from "ethers";
+import { Contract, providers } from "ethers";
 import { groupBy } from "lodash";
 import {
   assign,
@@ -99,7 +99,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    */
   constructor(
     readonly logger: winston.Logger,
-    readonly spokePool: SpokePool,
+    readonly spokePool: Contract,
     // Can be excluded. This disables some deposit validation.
     readonly hubPoolClient: HubPoolClient | null,
     readonly chainId: number,
@@ -440,7 +440,7 @@ export class SpokePoolClient extends BaseAbstractClient {
     low: number;
     high: number;
   }> {
-    return getBlockRangeForDepositId(targetDepositId, initLow, initHigh, maxSearches, this.spokePool);
+    return getBlockRangeForDepositId(targetDepositId, initLow, initHigh, maxSearches, this.spokePool as SpokePool);
   }
 
   /**
@@ -449,7 +449,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @returns The deposit ID.
    */
   public async _getDepositIdAtBlock(blockTag: number): Promise<number> {
-    return getDepositIdAtBlock(this.spokePool, blockTag);
+    return getDepositIdAtBlock(this.spokePool as SpokePool, blockTag);
   }
 
   /**
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -438,7 +438,6 @@ export class SpokePoolClient extends BaseAbstractClient {
     low: number;
     high: number;
   }> {
-    console.log("NEXT");
 
     if (initLow > initHigh) {
       throw new Error("Binary search failed because low > high");
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -440,7 +440,14 @@ export class SpokePoolClient extends BaseAbstractClient {
     low: number;
     high: number;
   }> {
-    return getBlockRangeForDepositId(targetDepositId, initLow, initHigh, maxSearches, this.spokePool as SpokePool);
+    return getBlockRangeForDepositId(
+      targetDepositId,
+      initLow,
+      initHigh,
+      maxSearches,
+      this.spokePool as SpokePool,
+      this.deploymentBlock
+    );
   }
 
   /**
@@ -901,15 +908,16 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @note This method is used to find deposits that are outside of the search range of this client.
    */
   async findDeposit(depositId: number, destinationChainId: number, depositor: string): Promise<DepositWithBlock> {
-    // Binary search for block where SpokePool.numberOfDeposits incremented to fill.depositId + 1.
-    // This way we can get the blocks before and after the deposit with deposit ID = fill.depositId
-    // and use those blocks to optimize the search for that deposit. Stop searches after a maximum
-    // # of searches to limit number of eth_call requests. Make an eth_getLogs call on the remaining block range
-    // (i.e. the [low, high] remaining from the binary search) to find the target deposit ID.
+    // Binary search for block. This way we can get the blocks before and after the deposit with
+    // deposit ID = fill.depositId and use those blocks to optimize the search for that deposit.
+    // Stop searches after a maximum # of searches to limit number of eth_call requests. Make an
+    // eth_getLogs call on the remaining block range (i.e. the [low, high] remaining from the binary
+    // search) to find the target deposit ID.
+    //
     // @dev Limiting between 5-10 searches empirically performs best when there are ~300,000 deposits
     // for a spoke pool and we're looking for a deposit <5 days older than HEAD.
     const searchBounds = await this._getBlockRangeForDepositId(
-      depositId + 1,
+      depositId,
       this.deploymentBlock,
       this.latestBlockNumber,
       7
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,4 +1,4 @@
-import { Contract, providers } from "ethers";
+import { BigNumber, Contract, Event, EventFilter, ethers, providers } from "ethers";
 import { groupBy } from "lodash";
 import winston from "winston";
 import {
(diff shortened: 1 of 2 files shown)


james-a-morris:
Needed to resolve: https://github.com/across-protocol/relayer-v2/pull/930

src/clients/SpokePoolClient.ts
```diff
@@ -438,13 +438,24 @@ export class SpokePoolClient extends BaseAbstractClient {
     low: number;
     high: number;
   }> {
+    console.log("NEXT");
```
pxrl, line 441:
```suggestion
```

src/utils/SpokeUtils.ts
```diff
@@ -42,13 +41,15 @@ export async function getBlockRangeForDepositId(
   // in the queriedIds cache.
   const _getDepositIdAtBlock = async (blockNumber: number): Promise<number> => {
     if (queriedIds[blockNumber] === undefined) {
-      queriedIds[blockNumber] = await getDepositIdAtBlock(spokePool, blockNumber);
+      const resultId = await getDepositIdAtBlock(spokePool, blockNumber);
+      if (!Number.isInteger(resultId)) {
+        throw new Error("Invalid deposit count");
+      }
```
pxrl, line 47: Would it be better to locate this validation centrally in `getDepositIdAtBlock()`, so callers don't have to check this every time?

src/utils/SpokeUtils.ts
```diff
@@ -76,7 +77,7 @@ export async function getBlockRangeForDepositId(
     //                                     // targetId = 11  <- fail (triggers this error)          // 10 <= 11
     //                                     // targetId = 10  <- fail (triggers this error)          // 10 <= 10
     //                                     // targetId = 09  <- pass (does not trigger this error)  // 10 <= 09
-    throw new Error("Failed to find deposit ID");
+    throw new Error("Failed to find deposit ID: Target Id is greater than the initial high block");
```
pxrl, line 80:
```suggestion
    throw new Error(`Target depositId is greater than the initial high block (${targetDepositId} > ${initHigh})`);
```

src/utils/SpokeUtils.ts
```diff
@@ -87,7 +88,7 @@ export async function getBlockRangeForDepositId(
     //                                     // targetId = 1 <- fail (triggers this error)
     //                                     // targetId = 2 <- pass (does not trigger this error)
     //                                     // targetId = 3 <- pass (does not trigger this error)
-    throw new Error("Failed to find deposit ID");
+    throw new Error("Failed to find deposit ID: Target Id is less than the initial low block");
```
pxrl, line 91:
```suggestion
    throw new Error(`Target depositId is less than the initial low block (${targetDepositId} > ${initLow})`);
```

src/utils/SpokeUtils.ts
```diff
@@ -104,34 +105,27 @@ export async function getBlockRangeForDepositId(
...
-      high = mid;
+      low = mid + 1;
     }
-  } while (++searches < maxSearches && low < high);
+  } while (++searches <= maxSearches && low < high);
+
+  // Sanity check to ensure that our low was not greater than our high.
+  if (low > high) {
+    throw new Error("Binary search failed. (low > high) SHOULD NEVER HAPPEN (but here we are)");
+  }
```
pxrl, line 128:
```suggestion
  if (low > high) {
    throw new Error(`Binary search failed (${low} > ${high}). SHOULD NEVER HAPPEN (but here we are)`);
  }
```

pxrl (commented):
@james-a-morris The following request timestamps might be good candidates for testing this with the `validateRootBundle.ts` script.

```
1690202003
1690210931
```

src/clients/SpokePoolClient.ts
```diff
@@ -21,7 +21,7 @@ import {
 } from "../utils/EventUtils";
 import winston from "winston";
 
-import { Contract, BigNumber, Event, EventFilter, ethers } from "ethers";
+import { BigNumber, Event, EventFilter, ethers } from "ethers";
```
pxrl, line 24: This is the 2nd import from ethers - candidate for consolidation? (also per your recent lint change in relayer-v2).

src/clients/SpokePoolClient.ts
```diff
@@ -481,7 +449,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @returns The deposit ID.
    */
   public async _getDepositIdAtBlock(blockTag: number): Promise<number> {
```
pxrl, line 451: Curious that we have a public method that's prefixed with an underscore :eyes:

src/utils/SpokeUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+ *        // This way the caller can search for a FundsDeposited event between [low, high] that will always
+ *        // contain the event emitted when deposit ID was incremented to targetDepositId + 1. This is the same transaction
+ *        // where the deposit with deposit ID = targetDepositId was created.
+ */
+export async function getBlockRangeForDepositId(
+  targetDepositId: number,
+  initLow: number,
+  initHigh: number,
+  maxSearches: number,
+  spokePool: SpokePool,
```
pxrl, line 23: Does this need to be a `SpokePool` type, or could we live with a regular ethers `Contract` definition?

src/utils/SpokeUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  // Sanity check to ensure that deploymentBlock is greater than or equal to zero.
+  if (deploymentBlock < 0) {
+    throw new Error("deploymentBlock must be >= 0");
+  }
+
+  // If the deposit ID at the initial high block is less than the target deposit ID, then we know that
+  // the target deposit ID must be greater than the initial high block, so we can throw an error.
+  if ((await _getDepositIdAtBlock(initHigh)) <= targetDepositId) {
```
pxrl, line 75: Consider combining this with the call to `getBlockNumber()` (at the top) so they can be executed in parallel?

src/utils/SpokeUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+    // initLow   = 5: Deposits Num: 10
+    //                                     // targetId = 11  <- fail (triggers this error)          // 10 <= 11
+    //                                     // targetId = 10  <- fail (triggers this error)          // 10 <= 10
+    //                                     // targetId = 09  <- pass (does not trigger this error)  // 10 <= 09
+    throw new Error("Failed to find deposit ID: Target Id is greater than the initial high block");
+  }
+
+  // If the deposit ID at the initial low block is greater than the target deposit ID, then we know that
+  // the target deposit ID must be less than the initial low block, so we can throw an error.
+  if ((await _getDepositIdAtBlock(Math.max(deploymentBlock, initLow - 1))) > targetDepositId) {
```
pxrl, line 85: Consider combining this with the call to `getBlockNumber()` (at the top) so they can be executed in parallel?

src/utils/SpokeUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+ * @param maxSearches The maximum number of searches to perform. This is used to prevent infinite loops.
+ * @returns The block range that contains the deposit ID.
+ * @note  // We want to find the block range that satisfies these conditions:
+ *        // - the low block has deposit count <= targetDepositId
+ *        // - the high block has a deposit count > targetDepositId.
+ *        // This way the caller can search for a FundsDeposited event between [low, high] that will always
+ *        // contain the event emitted when deposit ID was incremented to targetDepositId + 1. This is the same transaction
+ *        // where the deposit with deposit ID = targetDepositId was created.
+ */
+export async function getBlockRangeForDepositId(
```
pxrl, line 18: One consideration - it might be nice to wrap this method in another that accepts an array of deposits for the origin SpokePool. The reason for this is that we're likely to call `getBlockRangeForDepositId()` in parallel, and so there will be many concurrent calls to `getBlockNumber()` (etc etc). When these are executed in parallel for the first time, they're unlikely to be served out of the local redis cache, so we'll probably hit the RPC provider with a bunch of identical requests.

It might therefore be nice if we resolve that once centrally, and supply the values in as optional arguments. If the caller has access to a `SpokePoolClient` instance then they could then supply `SpokePoolClient.latestBlockNumber`, and that would save a lookup.

src/utils/SpokeUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+  low: number;
+  high: number;
+}> {
+  // Get the most recent block from the spoke pool contract.
+  const mostRecentBlockNumber = await spokePool.provider.getBlockNumber();
+  // Set the initial high block to the most recent block number or the initial high block, whichever is smaller.
+  initHigh = Math.min(initHigh, mostRecentBlockNumber);
+
+  // Define a mapping of block numbers to deposit IDs. This is used to cache the deposit ID at a block number
+  // so we don't need to make an eth_call request to get the deposit ID at a block number more than once.
```
pxrl, line 35:
```suggestion
  // Define a mapping of block numbers to number of deposits at that block. This saves repeated lookups.
```

src/utils/SpokeUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+  const queriedIds: Record<number, number> = {};
+
+  // Define a llambda function to get the deposit ID at a block number. This function will first check the
+  // queriedIds cache to see if the deposit ID at the block number has already been queried. If not, it will
+  // make an eth_call request to get the deposit ID at the block number. It will then cache the deposit ID
+  // in the queriedIds cache.
```
pxrl, line 41: Just my 2c - I think the function is pretty self-explanatory, so 4 lines of comments to a 6 line function might be a bit verbose.

src/utils/SpokeUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  // Sanity check to ensure that maxSearches is greater than zero.
+  if (maxSearches <= 0) {
+    throw new Error("maxSearches must be > 0");
+  }
+
+  // Sanity check to ensure that deploymentBlock is greater than or equal to zero.
+  if (deploymentBlock < 0) {
+    throw new Error("deploymentBlock must be >= 0");
+  }
```
pxrl, line 71: Basic input validation takes up a lot of space in this function, and it's a series of checks that we probably never expect to trigger. Thoughts on squashing it?

```suggestion
  const errors: [boolean, string][] = [ 
    [initLow < deploymentBlock, "Binary search failed because low must be >= 0"],
    [initLow > initHigh, "Binary search failed because low > high"],
    [maxSearches <= 0, "maxSearches must be > 0"],
    [deploymentBlock < 0, "deploymentBlock must be >= 0"],
  ];  

  errors.forEach(([condition, error]) => {
    if (!condition) {
      throw new Error(error);
    }   
  });
```

src/clients/SpokePoolClient.ts
```diff
@@ -929,15 +904,16 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    // Binary search for block. This way we can get the blocks before and after the deposit with
+    // deposit ID = fill.depositId and use those blocks to optimize the search for that deposit.
+    // Stop searches after a maximum # of searches to limit number of eth_call requests. Make an
+    // eth_getLogs call on the remaining block range (i.e. the [low, high] remaining from the binary
+    // search) to find the target deposit ID.
+    //
     // @dev Limiting between 5-10 searches empirically performs best when there are ~300,000 deposits
     // for a spoke pool and we're looking for a deposit <5 days older than HEAD.
     const searchBounds = await this._getBlockRangeForDepositId(
-      depositId + 1,
```
nicholaspai, line 940: so assuming that your implementation of `_getBlockRangeForDepositId` is logically equivalent to the current implementation, why do you not  opt to search for `depositId + 1`? The reason we search for the +1 number currently is that we are interested ultimately in finding the block containing the deposit with `depositId`, which is emitted in the same block where `SpokePool.depositId` gets incremented +1. So by searching for the block where `SpokePool.depositId  = depositId + 1` we find the same block where the deposit with `depositId` is mined.

What do you think?

src/clients/SpokePoolClient.ts
james-a-morris: Absolutely! This is a good call

src/utils/SpokeUtils.ts
james-a-morris: I personally prefer to be as explicit with the typecasting as possible. Do you think it negatively affects runtime? Happy to switch it back

src/utils/SpokeUtils.ts
james-a-morris: Yep!

src/utils/SpokeUtils.ts
james-a-morris: I agree. In this scenario, do you envision the multi-id function to be the main & the current existing process to be the one-off or vice versa?

src/clients/SpokePoolClient.ts
james-a-morris: `_getBlockRangeForDepositId` is now slightly modified in that we account for the deposit ID at a particular block as essentially `[depositCount[block-1], depositCount[block] - 1]`. As a result, the `+1` addition to account for this is no longer needed since we will get to the block that contains the deposit count without needing to modify the deposit we're looking for.

src/utils/SpokeUtils.ts
pxrl: I was thinking that we'd collate a list of deposit Ids that we're interested in, and then query them all at once. It might require further refactoring upstream in the calling code though, so perhaps it's not practical given the current code structure.

src/utils/SpokeUtils.ts
pxrl: I guess I don't really see any benefits from it, vs. a standard Ethers `Contract` type. A standard Ethers `Contract` seems preferable to me because it's more portable and imposes the least on the caller. I'm pretty sure you could swap the type back to `Contract` and it'd work without any further modifications at all.

src/utils/SpokeUtils.ts
pxrl: I'd still argue that these specific checks are just simple and basic input validation...I don't see any risk of them hiding any complexity or creating confusion. The error messages that are thrown are also matched almost 1:1 by the comments, so I don't really see how this kind of redundancy helps the reader. To me, it just seems like this validation just occupies vertical real estate and draws attention away from where the real complexity is (which is now very clear and well-documented, btw).

src/clients/SpokePoolClient.ts
mrice32: @nicholaspai

@james-a-morris and I went through the old implementation and the new one around this specific question. It is logically equivalent, but the + 1 is taken care of inside the function. I thought this would probably be more intuitive to the caller who may not understand the precise way that deposit ids are incremented or how the contract variable is set. It seems more logical to ask "what was the block that this deposit was in" not "what block was the deposit number variable set to this value".

mrice32 (approved):
LGTM! (But we went through this together IRL, so I clearly agree with the implementation :))

src/utils/SpokeUtils.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/383/commits/08e9de5d6fedb88a3384d25720ce712b2be04fd8

src/utils/SpokeUtils.ts
james-a-morris: I want to push on this and keep the documentation specifically because it's an anonymous function that uses the scope of the variable above it.

src/utils/SpokeUtils.ts
james-a-morris: Fair enough - I've modified the tests here: https://github.com/across-protocol/sdk-v2/pull/383/commits/0134433543e75a83862c2abedc2084585a7d0ec4

src/utils/SpokeUtils.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/383/commits/f62f826707b56ee6930189853dbcc863667fec52

src/clients/SpokePoolClient.ts
```diff
@@ -482,7 +457,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @returns The deposit ID.
    */
   public async _getDepositIdAtBlock(blockTag: number): Promise<number> {
```
nicholaspai, line 458: Does this function stil need to be a sync?

src/clients/SpokePoolClient.ts
```diff
@@ -439,41 +441,14 @@ export class SpokePoolClient extends BaseAbstractClient {
     low: number;
     high: number;
   }> {
-    if (initLow > initHigh) {
```
nicholaspai, line 441: Is this function still an async one?

nicholaspai (commented):
Left some questions on first pass. I trust the algorithm is good as I've reviewed it and Matt has worked on it

src/clients/SpokePoolClient.ts
james-a-morris: The function itself returns a promise. But no, it doesn't need the `async` modifier, which won't change the execution logic.  

I'd like to handle this in a secondary change that applies the same relayer eslint file to the SDK. To do this, it would be an easier lift to do so after this & the other SDK change is merged.

src/clients/SpokePoolClient.ts
james-a-morris: It is currently not needed, but I do think we should do this in a separate PR.

nicholaspai (approved):

PR #385, opened 2023-09-10
fix: disallow computing lp fees for future timestamps

mrice32:
This should not happen, in general, but this adds a check to throw an error if a future timestamp is ever requested.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -270,6 +270,13 @@ export class HubPoolClient extends BaseAbstractClient {
     if (!isDefined(this.currentTime)) {
       throw new Error("HubPoolClient has not set a currentTime");
     }
+
+    if (deposit.quoteTimestamp > this.currentTime) {
+      throw new Error(
+        `Cannot compute lp fee percent for quote timestamp ${deposit.quoteTimestamp} in the future. Current time: ${this.currentTime}.`
+      );
+    }
+
     const quoteBlock = await this.getBlockNumber(deposit.quoteTimestamp);
     if (!isDefined(quoteBlock)) {
       throw new Error(`Could not find block for timestamp ${deposit.quoteTimestamp}`);
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -10,6 +10,7 @@ import {
   MAX_BIG_INT,
   stringifyJSONWithNumericString,
   toBN,
+  isDefined,
 } from "../utils";
 import { validateFillForDeposit, filledSameDeposit } from "../utils/FlowUtils";
 import {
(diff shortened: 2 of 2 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -270,6 +270,13 @@ export class HubPoolClient extends BaseAbstractClient {
     if (!isDefined(this.currentTime)) {
       throw new Error("HubPoolClient has not set a currentTime");
     }
+
+    if (deposit.quoteTimestamp > this.currentTime) {
```
james-a-morris, line 274: Good call - it would need to be specifically greater than.

src/clients/HubPoolClient.ts
```diff
@@ -270,6 +270,13 @@ export class HubPoolClient extends BaseAbstractClient {
...
+      throw new Error(
+        `Cannot compute lp fee percent for quote timestamp ${deposit.quoteTimestamp} in the future. Current time: ${this.currentTime}.`
+      );
+    }
```
james-a-morris, line 278: Should we also log here?

src/clients/SpokePoolClient.ts
```diff
@@ -672,6 +673,11 @@ export class SpokePoolClient extends BaseAbstractClient {
       }
     }
 
+    const hubCurrentTime = this.hubPoolClient?.currentTime;
+    if (!isDefined(hubCurrentTime)) {
```
james-a-morris, line 677: Could we short circuit this and directly check `!isDefined(hubCurrentTime)` or is this needed to appease TS?

james-a-morris (approved):
LGTM - left some nits

pxrl (approved):
Seems fine to me; my approach in https://github.com/across-protocol/sdk-v2/pull/384 was to compute an impossible LP fee in order to try to handle this more gracefully, but this is nice in that it will loudly alert us to any latent issues here.

src/clients/HubPoolClient.ts
mrice32: Wouldn't the error show up in logging when it gets handled?

src/clients/SpokePoolClient.ts
mrice32: This is mainly to convince ts that hubCurrentTime is a valid number if this if statement isn't hit.

src/clients/HubPoolClient.ts
james-a-morris: True, fair point!

src/clients/SpokePoolClient.ts
james-a-morris: SGTM! Just wanted to double check that this was the reason.

james-a-morris (approved):
LGTM - thanks for addressing my concerns

PR #386, opened 2023-09-11
chore: 🚨 linter warnings 

james-a-morris:
The new upgrade to eslint requires new linting

### DIFF
--- a/src/pool/poolClient.ts
+++ b/src/pool/poolClient.ts
@@ -112,7 +112,10 @@ export type PooledToken = {
 };
 
 class PoolState {
-  constructor(private contract: hubPool.Instance, private address: string) {}
+  constructor(
+    private contract: hubPool.Instance,
+    private address: string
+  ) {}
   public async read(l1Token: string, latestBlock: number, previousBlock?: number) {
     // typechain does not have complete types for call options, so we have to cast blockTag to any
     const exchangeRatePrevious = await this.exchangeRateAtBlock(l1Token, previousBlock || latestBlock - 1);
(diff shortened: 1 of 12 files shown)


mrice32 (commented):
Are these auto-formatted? Are these only lint warnings?

In general, I think lint warnings are of little to no use, since they don't block merges. I think we should only fix lint errors. If these are only lint warnings, IMO, we should either convert them to lint errors or not fix them.

src/pool/poolClient.ts
```diff
@@ -144,7 +147,10 @@ export class PoolEventState {
   private iface: ethers.utils.Interface;
   // maintain ordered unique list of events so we can calculate state
   private events: uma.SerializableEvent[] = [];
-  constructor(private contract: hubPool.Instance, private startBlock = 0) {
+  constructor(
+    private contract: hubPool.Instance,
+    private startBlock = 0
+  ) {
```
pxrl, line 153: This rule seems unnecessarily strict to me - isn't it fine to have the constructor arguments on a single line as long as they don't exceed the line length limit? I'd propose to tweak the eslint config for this, rather than change the code.

pxrl (commented):
IMO a lot of these changes are due to an overly strict config - wdyt about fixing the config instead?

james-a-morris:
> Are these auto-formatted? Are these only lint warnings?
> 
> In general, I think lint warnings are of little to no use, since they don't block merges. I think we should only fix lint errors. If these are only lint warnings, IMO, we should either convert them to lint errors or not fix them.

I agree - these are auto-linted from `yarn lint-fix`

src/pool/poolClient.ts
james-a-morris: Resolved offline

pxrl (approved):

PR #389, opened 2023-09-15
refactor: Hoist HubPool off UMA dependencies

pxrl:
This PR migrates in `Balances` (type and function) and 
`SerializableEvent` from @uma. It also refactors the `reduceEvents`
function to avoid the need for generics. This has a nice side-effect of
removing a lot of duplicated code, so the implementation becomes
simpler.

This hubPool implementation is only used by the poolClient class, and I
have verified the pre- and post-change behaviour via `poolClient.e2e.ts`
- it can still extract the necessary event data after this change.

### DIFF
--- a/src/contracts/utils.ts
+++ b/src/contracts/utils.ts
@@ -1,0 +1,53 @@
+import assert from "assert";
+import { BigNumber, BigNumberish, Event } from "ethers";
+import { isDefined } from "../utils";
+
+/**
+ * Migrated directly from @uma/sdk.
+ * @dev This code is intended to support existing use of contracts/hubPool.ts and should not be used for new code.
+ * @todo Refactor contracts/hubPool.ts to avoid the need for this.
+ */
+
+// useful for maintaining balances from events
+export type Balances = { [key: string]: string };
+
+export function Balances(balances: Balances = {}) {
+  function create(id: string, amount = "0") {
+    assert(!has(id), "balance already exists");
+    return set(id, amount);
+  }
+
+  function has(id: string) {
+    return isDefined(balances[id]);
+  }
+
+  function set(id: string, amount: string) {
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

src/contracts/utils.ts
```diff
@@ -0,0 +1,53 @@
+import assert from "assert";
+import { BigNumber, BigNumberish, Event } from "ethers";
+import { isDefined } from "../utils";
+
+/**
+ * Migrated directly from @uma/sdk.
```
james-a-morris, line 6: Should we document the function and leave this as a `@note`?

src/contracts/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+    return balances[id];
+  }
+
+  function getOrCreate(id: string) {
+    if (has(id)) return get(id);
+    return create(id);
+  }
+
+  return { create, add, sub, get, balances, set, has, getOrCreate };
+}
```
james-a-morris, line 48: This feels as though it should be a class and not a closure. Should we follow up on this in a separate PR?

james-a-morris (commented):
Left a few comments

src/contracts/utils.ts
pxrl: Yeah it does seem like a better fit as a class. I'm not sure whether we want to invest too much time in updating this in isolation, since it's fairly old code that "just works". We can discuss it in the weekly call.

src/contracts/utils.ts
pxrl: Updated here: https://github.com/across-protocol/sdk-v2/pull/389/commits/51f2336d2fe1fb987d4a0de4a548904716af95dd

PR #390, opened 2023-09-18
chore: Export NetworkUtils

pxrl:
Will be used in relayer-v2.


james-a-morris (approved):
Fair call! Should we bump the version as well?

pxrl:
> Fair call! Should we bump the version as well?

Will do in a subsequent change, once we've bundled in a few more improvments :+1:

PR #391, opened 2023-09-18
refactor: Migrate constants from UMA

pxrl:
Import MAX_SAFE_JS_INT, MAX_SAFE_ALLOWANCE and SECONDS_PER_YEAR.

While here, prefer Ethers BN constants over spinning up new BigNumber
instances in UBAFeeCalculator.test.ts.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -6,8 +6,6 @@ export const { AddressZero: ZERO_ADDRESS } = ethersConstants;
 // 2^96 - 1 is a conservative erc20 max allowance.
 export const MAX_SAFE_ALLOWANCE = "79228162514264337593543950335";
 
-export const MAX_SAFE_JS_INT = "4294967297"; // 2^32 + 1
-
 export const SECONDS_PER_YEAR = 31557600; // 365.25 days per year.
 
 /**
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,4 +1,4 @@
-import { constants as ethersConstants } from "ethers";
+import { constants as ethersConstants, BigNumber, utils } from "ethers";
 export { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/contracts-v2/dist/utils/constants";
 
 export const { AddressZero: ZERO_ADDRESS } = ethersConstants;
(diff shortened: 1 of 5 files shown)


src/constants.ts
```diff
@@ -3,6 +3,13 @@ export { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/contracts-v2/dist
 
 export const { AddressZero: ZERO_ADDRESS } = ethersConstants;
 
+// 2^96 - 1 is a conservative erc20 max allowance.
+export const MAX_SAFE_ALLOWANCE = "79228162514264337593543950335";
+
+export const MAX_SAFE_JS_INT = "4294967297"; // 2^32 + 1
```
james-a-morris, line 9: IMO I'm not sure we should keep the name `MAX_SAFE_JS_INT` given that `Number.MAX_SAFE_INTEGER` is defined to be `2^53 - 1`

Would it make sense to reference the fact that this would be a max safe integer for an unsigned 32 bit integer?

james-a-morris (approved):
Left a comment, otherwise LGTM

src/constants.ts
pxrl: I also wondered if we could use this, and I'd be totally happy to ditch `MAX_SAFE_JS_INT` in favour of `MAX_SAFE_INTEGER`. The only reason I didn't do that is because the UBAFeeCalculator test relies on it, so if we're OK to update that implementation then I think we should do it.

nb. one additional consideration: I noticed that `BigNumber.from(Number.MAX_SAFE_INTEGER)` reverts unless `Number.MAX_SAFE_INTEGER` is cast to a string. We should be fine to handle that, but I thought it was worth mentioning.

james-a-morris: Strange. Follow up: would it make sense to directly convert this to a BigNumber instead of a string? AFAIK the string representation isn't actually needed and we almost always directly convert this to a BN

pxrl: Per https://github.com/across-protocol/sdk-v2/pull/394, I'll remove the MAX_SAFE_JS_INT entirely :+1:

PR #392, opened 2023-09-18
refactor: New UMA SDK imports

pxrl:
Import the following from the UMA repo:
 - RateModelDictionary class (+ related definitions).
 - TransactionManager closure.

It was very simple to hoist the ConfigStore interface over to the new local definitions, so that has been applied here. Subsequent changes are needed for the TransactionManager; those will be applied in a follow-up commit so the modifications are explicitly identified.

### DIFF
--- a/src/interfaces/ConfigStore.ts
+++ b/src/interfaces/ConfigStore.ts
@@ -1,12 +1,12 @@
 import { BigNumber } from "ethers";
+import { RateModelDictionary } from "../lpFeeCalculator/rateModel";
 import { SortableEvent } from "./Common";
-import { across } from "@uma/sdk";
 
 export interface ParsedTokenConfig {
   transferThreshold: string;
-  rateModel: across.rateModel.RateModelDictionary;
+  rateModel: RateModelDictionary;
   routeRateModel?: {
-    [path: string]: across.rateModel.RateModelDictionary;
+    [path: string]: RateModelDictionary;
   };
   uba?: UBAOnChainConfigType;
   spokeTargetBalances?: {
--- a/src/lpFeeCalculator/rateModel.ts
+++ b/src/lpFeeCalculator/rateModel.ts
@@ -1,0 +1,183 @@
+import { ethers } from "ethers";
+import { isDefined } from "../utils";
+
+// Each L1 token is mapped to an array of stringified rate models, ordered by the block height at which they were
+// published on-chain. This dictionary is used internally to fetch a rate model for a block height.
+type RateModelEventsDictionary = {
+  [l1TokenAddress: string]: { blockNumber: number; rateModel: string }[];
+};
+
+// Events should be reformatted into this shape to be used as input into methods in this file.
+export type RateModelEvent = {
+  blockNumber: number;
+  transactionIndex: number;
+  logIndex: number;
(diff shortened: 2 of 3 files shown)


pxrl:
> Is this a pure copy/paste? Or was there additional refactors that needed to be done to get things to compile.
> 
> I was hoping to confirm where I should be looking in this PR in terms of code changes. But, if it's a pure refactor, I think this PR is good - but we may need an additional one to refine documentation to bring it in line with the TSDoc formatting that exists on a large chunk of the existing code.

Good that you asked. This is a pure copy/paste with only the absolute minimum of modifications necessary for the code to build. I should have linked the original files, but here they are:
- rateModel.ts: https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/across/rateModel.ts
- TransactionManager.ts: https://github.com/UMAprotocol/protocol/blame/92d726fb1bcdb55002d65f3a43f1a9483c4d5c2a/packages/sdk/src/across/transactionManager.ts

In terms of post-import modifications, here is what a diff between the two shows on my end:

```
--- dev/UMAprotocol/protocol/packages/sdk/src/across/rateModel.ts    2023-06-06 21:01:40.299536043 +0200
+++ dev/across-protocol/sdk-v2/src/lpFeeCalculator/rateModel.ts      2023-09-18 16:54:07.712251309 +0200
@@ -1,6 +1,5 @@
-import { expectedRateModelKeys, RateModel } from "./constants";
-import { exists } from "../utils";
 import { ethers } from "ethers";
+import { isDefined } from "../utils";
 
 // Each L1 token is mapped to an array of stringified rate models, ordered by the block height at which they were
 // published on-chain. This dictionary is used internally to fetch a rate model for a block height.
@@ -17,6 +16,15 @@
   l1Token: string;
 };
 
+interface RateModel {
+  UBar: string; // denote the utilization kink along the rate model where the slope of the interest rate model changes.
+  R0: string; // is the interest rate charged at 0 utilization
+  R1: string; // R_0+R_1 is the interest rate charged at UBar
+  R2: string; // R_0+R_1+R_2 is the interest rate charged at 100% utilization
+}
+
+const expectedRateModelKeys = ["UBar", "R0", "R1", "R2"];
+
 export class RateModelDictionary {
   public rateModelDictionary: RateModelEventsDictionary = {};
 
@@ -92,15 +100,15 @@
           return ethers.utils.getAddress(l1Token);
         else return null;
       })
-      .filter(exists);
+      .filter(isDefined);
   }
 }
+
 /**
  * Helper method that returns parsed rate model from string, or throws.
  * @param rateModelString Stringified rate model to parse.
  * @returns Rate model object. Must conform to `expectedRateModelKeys` format.
  */
-
 export const parseAndReturnRateModelFromString = (rateModelString: string): RateModel => {
   const rateModelFromEvent = JSON.parse(rateModelString);
 
@@ -114,6 +122,7 @@
       );
     }
   }
+
   for (const key of Object.keys(rateModelFromEvent)) {
     if (!expectedRateModelKeys.includes(key)) {
       throw new Error(
```

james-a-morris:
@pxrl Okay perfect - we may want to refactor further, but in terms of this PR, LGTM

PR #395, opened 2023-09-19
fix(constants): Make UBA min version very high so we can slot in preceding upgrades

nicholaspai:


pxrl (approved):
LGTM. It'll be interesting to see whether the relayer-v2 tests just adopt this seamlessly, or whether they have some issues.

PR #396, opened 2023-09-20
refactor: Use BN 0 constants in UBAFeeCalculator test

pxrl:


james-a-morris (approved):
Only question: was this grepped? Just want to confirm we're grabbing everything to convert

pxrl:
> Only question: was this grepped? Just want to confirm we're grabbing everything to convert

This was actually isolated to this file, but I will follow up with a global one.

PR #398, opened 2023-09-20
refactor: Import createFormatFunction + associated from UMA

pxrl:
This unfortunately introduces direct Web3.js dependencies, but we need to import it as-is before we can clean it up.

[update]: No issue anymore - this is a very simple change now.

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -1,5 +1,4 @@
 import { BigNumber, ethers, PopulatedTransaction, providers, VoidSigner } from "ethers";
-import * as uma from "@uma/sdk";
 import Decimal from "decimal.js";
 import { isL2Provider as isOptimismL2Provider } from "@eth-optimism/sdk/dist/l2-provider";
 import { L2Provider } from "@eth-optimism/sdk/dist/interfaces/l2-provider";
@@ -8,12 +7,12 @@ import assert from "assert";
 import { GasPriceEstimate, getGasPriceEstimate } from "../gasPriceOracle";
 import { TypedMessage } from "../interfaces/TypedData";
 import { BN, toBN, BigNumberish } from "./BigNumberUtils";
+import { ConvertDecimals } from "./FormattingUtils";
 
 export type Decimalish = string | number | Decimal;
 export const AddressZero = ethers.constants.AddressZero;
 export const MAX_BIG_INT = BigNumber.from(Number.MAX_SAFE_INTEGER.toString());
 
-const { ConvertDecimals } = uma.utils;
 // These are distances used to traverse when looking for a block with desired lookback.
 // They're meant to be small enough to allow for granularity but large enough to minimize the number of reqests needed
 // to find the desired block.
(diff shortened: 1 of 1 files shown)


src/utils/common.ts
```diff
@@ -88,7 +87,7 @@ export function nativeToToken(
 ): string {
   const priceWei = toBNWei(price);
   const toAmount = toBNWei(fromAmount).div(priceWei);
-  return ConvertDecimals(nativeDecimals, toDecimals)(toAmount);
+  return ConvertDecimals(nativeDecimals, toDecimals)(toAmount.toString()).toString();
```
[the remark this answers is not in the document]
pxrl: Explanation: there are actually two variants of `ConvertDecimals()` - one supporting `Web3.js BN` and the other supporting `ethers BigNumber`. It's not strictly required because raw number and string types are used on both input and output sides, but there's a constraint in that `FormattingUtilsInternal.ts` is not intended to import from other files, and it ideally shouldn't depend directly on `ethers BigNumber` either.

So I propose that we stick with the Web3 variant hidden in `FormattingUtilsInternal.js` and clean this up post-import.

nicholaspai (approved):

src/utils/common.ts
```diff
@@ -71,6 +70,26 @@ export function max(a: BigNumberish, b: BigNumberish): BN {
...
+/**
+ * Promote or demote a number to based on differing number of decimal places.
+ * @dev Not to be confused with a Web3 variant located in FormattingUtilsInternal.
+ */
+export const convertDecimals = (fromDecimals: number, toDecimals: number) => {
+  assert(fromDecimals >= 0, "requires fromDecimals as an integer >= 0");
+  assert(toDecimals >= 0, "requires toDecimals as an integer >= 0");
+  // amount: string, BN, number - integer amount in fromDecimals smallest unit that want to convert toDecimals
+  // returns: string with toDecimals in smallest unit
+  return (amount: BigNumberish): string => {
```
james-a-morris, line 82: Could we return this in terms of `BigNumberish` or directly as a `BigNumber`?

src/utils/FormattingUtilsInternal.ts
james-a-morris, line 1: Do we have a Linear task for this? I like the intermediate step

james-a-morris (approved):
LGTM pending the creation of a linear task for making sure we remove `FormattingUtilsInternal.ts`

src/utils/FormattingUtilsInternal.ts
pxrl: Shouldn't be necessary anymore - I was able to re-work this after your tsdx migration and it's all much simpler now.

src/utils/common.ts
pxrl: I was able to drop this import after merging in your recent changes :+1:

src/utils/common.ts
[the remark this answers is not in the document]
pxrl: Update - it was possible to drop the duplicate imports after merging in the latest changes to master.

PR #400, opened 2023-09-20
fix: Move transfer threshold max supported version down to current version

nicholaspai:
We should support `transferThreshold` before but not including version 2


james-a-morris (commented):
LGTM - this makes sense to me to regress the threshold from 2 to 1 given the explanation. 

~~Should we remove the 0.16.0 release candidate from npm out of an abundance of caution?~~

james-a-morris (approved):

PR #407, opened 2023-09-21
chore: deduplicate yarn lock file

james-a-morris:
Result of running `npx yarn-deduplicate yarn.lock`.

This effectively prunes our dependency list to a list of needed modules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.16.4--beta.1",
+  "version": "0.16.4",
   "license": "AGPL-3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -91,7 +90,8 @@
     "decimal.js": "^10.3.1",
     "ethers": "^5.7.2",
     "lodash.get": "^4.4.2",
-    "superstruct": "^0.15.4"
+    "superstruct": "^0.15.4",
+    "tslib": "^2.6.2"
```
james-a-morris, line 94: This was the functional change to this file. Essentially what is happening is that when this `tslib` is listed as a dev dependency, it doesn't get injected into the `node_modules` file to be used by the `cjs` variant of `sdk-v2`. TSDX handled this, but now we have to do it manually.

nicholaspai (approved):

PR #409, opened 2023-09-27
chore(refactor): move spokepool client tests

james-a-morris:
This PR aims to migrate the `SpokePoolClient` tests to the SDK properly. The impetus of this change is to allow developers to test the spoke pool client in the repo in the codebase it is defined.

Required changes:
* We must now maintain a `.env` for testing on-chain-related unit tests.
* We will need to refactor the relayer-v2 to remove/make use of the code migration
* There is a section of tests commented out due to their use of queryHistoricalDeposit
    * This will require additional refactoring, but I'd recommend we do so in a separate PR.
* We're going to have to maintain these two added contracts for testing


james-a-morris:
@pxrl @nicholaspai the failure in the dependency review is that `"@uma/financial-templates-lib@^2.33.0"` imports `"@google-cloud/logging-winston@^4.1.1"` which imports `"google-gax@^2.24.1"` which imports a vulnerable version of `protobuf` 😔 

I'm not sure if we can get around this except to
* Disable dependency review
* Look to change out our dependency on uma financial lib

nicholaspai:
> @pxrl @nicholaspai the failure in the dependency review is that `"@uma/financial-templates-lib@^2.33.0"` imports `"@google-cloud/logging-winston@^4.1.1"` which imports `"google-gax@^2.24.1"` which imports a vulnerable version of `protobuf` 😔 
> 
> 
> 
> I'm not sure if we can get around this except to
> 
> * Disable dependency review
> 
> * Look to change out our dependency on uma financial lib

Ultimately what modules do we need from financial templates lib? If we find it's small, then we can try to remove it entirely. That'd be a great outcome, fingers crossed.

hardhat.config.ts
```diff
@@ -1,99 +1,21 @@
...
 import "solidity-coverage";
-import "hardhat-deploy";
-import "@openzeppelin/hardhat-upgrades";
-
-dotenv.config();
 
 const solcVersion = "0.8.18";
-const mnemonic = getMnemonic();
 
 const config: HardhatUserConfig = {
```
nicholaspai, line 12: Yeah I don't think we need these extra networks for the SDK since we only want to run unit tests. The relayer's `hardhat.config` is the one where we want all the networks.

package.json
```diff
@@ -44,20 +44,24 @@
...
-    "@typechain/hardhat": "^5.0.0",
+    "@typechain/ethers-v5": "^11.1.1",
+    "@typechain/hardhat": "^6.1.6",
     "@types/chai": "^4.3.6",
     "@types/dotenv": "^8.2.0",
+    "@types/lodash": "^4.14.199",
     "@types/lodash.get": "^4.4.7",
     "@typescript-eslint/eslint-plugin": "^6.6.0",
     "@typescript-eslint/parser": "^6.6.0",
+    "@uma/financial-templates-lib": "^2.33.0",
```
nicholaspai, line 62: Would be really nice to remove this, why exactly do we need it?

nicholaspai (approved):
This is really good. I'd like to push on removing any @uma deps if possible, but if its too intertwined we can do it in a follow up

package.json
james-a-morris: Done

hardhat.config.ts
james-a-morris: Removed

PR #413, opened 2023-09-28
improve: align with #947

james-a-morris:


test/SpokePoolClient.deposits.ts
```diff
@@ -1,83 +1,105 @@
...
-      const destinationChainId = inputDeposit.destinationChainId as number;
-      const outputDeposit = outputDeposits[destinationChainId].find((deposit) => deposit.depositId === depositId);
-      Object.entries(inputDeposit).forEach(([k, v]) => {
-        expect(v).to.not.be.undefined;
-        expect(v).to.deep.equal(outputDeposit?.[k]);
-      });
-    });
+    expect(spokePoolClient.getFills().length).to.equal(6);
+
+    // TODO: Add `getFillsForRepaymentChainId` tests once we update the `fillRelay` method from contracts-v2 to allow
```
nicholaspai, line 97: Is this a TODO you added? 

Also, I think we should try to remove the dependency on any functions from `contracts-v2` so if that means copying the `fiillRelay` code (and all other code in that repo) into this repo, it'd be better. I've found it frustrating in the past to not be able to modify the code because its from another repo. Same reasoning as the @uma functions :)

nicholaspai (commented):
https://github.com/across-protocol/sdk-v2/pull/413/files#r1340820336

test/SpokePoolClient.deposits.ts
james-a-morris: No that was also ported over. I think that may be better suited to do this in another PR.

test/SpokePoolClient.deposits.ts
```diff
@@ -1,83 +1,105 @@
...
+    ));
+    await spokePool.setChainId(destinationChainId); // The spoke pool for a fill should be at the destinationChainId.
+
+    spokePoolClient = new SpokePoolClient(
+      createSpyLogger().spyLogger,
+      spokePool,
+      null,
+      destinationChainId,
+      deploymentBlock
+    );
```
pxrl, line 40: I'm curious about this setup - why is the SpokePool deployed with chainId = `originChainId`, and then later switched to `destinationChainId` ? The `deploySpokePoolWithToken()` helper does some route-based setup, so adjusting the chainIds might have side-effects.

james-a-morris: This was refactored directly from the relayer. Should we follow up in a separate PR?

test/SpokePoolClient.deposits.ts
```diff
@@ -1,83 +1,105 @@
...
+let depositor: SignerWithAddress, relayer1: SignerWithAddress, relayer2: SignerWithAddress;
 let deploymentBlock: number;
-const destinationChainId2 = destinationChainId + 1;
+
+const originChainId2 = originChainId + 1;
 
 let spokePoolClient: SpokePoolClient;
 
-describe("SpokePoolClient: Deposits", async function () {
+describe("SpokePoolClient: Fills", async function () {
```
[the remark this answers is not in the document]
james-a-morris: Resolved

PR #414, opened 2023-09-29
refactor: Import BlockFinder from UMA

pxrl:
BlockFinder required modifications on import because otherwise it wouldn't build. The modifications are fairly straightforward, and hopefully make the code a bit simpler.


pxrl:
BlockFinder (et al.) imported from: https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/blockFinder.ts
https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/utils.ts

nb. BlockFinder was a generic, but I've scoped it to an ethers `Block` type.

src/utils/BlockUtils.ts
```diff
@@ -50,3 +52,117 @@ export async function averageBlockTime(
 
   return { average, blockRange };
 }
+
+async function estimateBlocksElapsed(seconds: number, cushionPercentage = 0.0, provider: Provider): Promise<number> {
```
nicholaspai, line 56: did you change anything in this file or just copy/paste?

PR #416, opened 2023-09-29
refactor: Remove UMA dependencies

pxrl:
Originally proposed to relayer-v2, but re-proposed here due to this ongoing migration:
 - Drop MAX_SAFE_ALLOWANCE and MAX_UINT_VAL (unused in test).
 - Define TokenRolesEnum locally (it's only used here).


james-a-morris (approved):
LGTM

PR #419, opened 2023-09-30
feat: Export updated BAL address

nicholaspai:
Upgrade contracts-v2 version to update new BAL address


pxrl (approved):
LGTM - also nice to flush some cruft from `yarn.lock` !

PR #420, opened 2023-10-02
feat: ABI retrieval utility

pxrl:
Intended to replace UMA's getABI() implementation.


src/utils/abi/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+/**
+ * @notice Retrieve an ABI desription from the set of known contracts.
+ * @param contractName Name of the contract ABI to retrieve.
+ * @returns Contract ABI as an ethers ContractInterface type.
+ */
+export async function getABI(contractName: string): Promise<ContractInterface> {
+  try {
+    return JSON.parse(await fs.readFile(`${__dirname}/contracts/${contractName}.json`, { encoding: "utf8" }));
```
pxrl, line 19: tbd: Will this run OK in serverless? I could imagine scenarios where we might want to point at the output code, rather than src...but maybe this is a non-issue?

pxrl (commented):
Note: This is proposing to merge into one of James' branches, and it's currently in flux, so the diff looks _muuuch_ larger than it really is. The set of changes should consolidate nicely once once the destination branch settles down.

src/utils/abi/index.ts
pxrl: Turns out that it was missing entirely, so I've added it here: https://github.com/across-protocol/sdk-v2/pull/420/commits/2b6437b812485db12eb8c7e38f954edb576a6df1

Interested in your thoughts @james-a-morris. It feels a _little_ bit janky to me, but it works OK. Can it be improved?

package.json
```diff
@@ -13,10 +13,11 @@
   },
   "scripts": {
     "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build'",
-    "build": "yarn run clean && yarn run build:cjs && yarn run build:esm && yarn run build:types",
+    "build": "yarn run clean && yarn run build:cjs && yarn run build:esm && yarn run build:types; yarn copy-abi",
     "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --removeComments --verbatimModuleSyntax false && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
     "build:esm": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm && echo > ./dist/esm/package.json '{\"type\":\"module\",\"sideEffects\":false}'",
     "build:types": "tsc --project tsconfig.build.json --module esnext --declarationDir ./dist/types --emitDeclarationOnly --declaration --declarationMap",
+    "copy-abi": "abi=utils/abi/contracts; for dir in cjs esm; do dstdir=\"./dist/${dir}/${abi}\"; mkdir -p \"${dstdir}\"; cp ./src/${abi}/*.json \"${dstdir}\"; done",
```
pxrl, line 20: nb. I'd rather make copy-abi work for a single output dir (i.e. `./dist/cjs` or `./dist/esm`). Unfortunately `yarn` is a bit limited here - i.e. it doesn't seem feasible to do this via `package.json` without calling externally to a separate script.

pxrl: Update: I was able to implement a workaround for this, and I think it's better than what was initially proposed: https://github.com/across-protocol/sdk-v2/pull/420/commits/4a21390c11e711c3201cba81592e95f607666e6c

test/abi.ts
```diff
@@ -0,0 +1,18 @@
+import { getABI } from "../src/utils/abi";
+import { assertPromiseError, expect } from "./utils";
+
+describe("ABI Utils", () => {
+  describe("getABI", async () => {
+    it("Retrieve Multicall3 ABI", async () => {
+      expect(await getABI("Multicall3"));
```
james-a-morris, line 7: Could we fuzz test for all the available contracts?

src/utils/abi/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  } catch (err) {
+    // @dev fs methods can return errors that are not Error objects (i.e. errno).
+    const msg = isError(err) ? err.message : (err as Record<string, unknown>)?.code ?? "unknown error";
```
james-a-morris, line 22: Nit: the order of operations could be clearer to me. Is the `?? "unknown error"` inside the ternary?

src/utils/abi/index.ts
james-a-morris: The one thing that I would recommend would be to cache these responses once they're read. `fs.readFile` is a heavy call.

Another note on further thought, this will probably fail in the FE. The reason being: `fs` is a node-specific directive and if we call this in via the FE it more than likely be undefined.

james-a-morris (changes requested):
AFAIK this will fail in the FE when imported anywhere that is executed in a web browser. Since the SDK is used heavily in the FE, the usage of `file.readSync` may break the FE.

My recommendation would be to include these ABIs in memory since there's not too many of them.

src/utils/abi/index.ts
pxrl: It applies to the latter part of the ternary statement - i.e. it's a fallback for what comes after the `:` operator.

I made it clearer here: https://github.com/across-protocol/sdk-v2/pull/420/commits/111bb395382641ad87276c44d7fd00a0452b209c

pxrl:
> AFAIK this will fail in the FE when imported anywhere that is executed in a web browser. Since the SDK is used heavily in the FE, the usage of `file.readSync` may break the FE.

I've tested this locally; the FE seems happy enough to import the linked package. If the FE tried to  I haven't seen any side-effects as the FE doesn't actually try to use `getABI()`. Since the FE is not actually an intended target for this, I'm not sure whether that's a showstopper or not. I could alternatively place `getABI()` in relayer-v2, but we do actually have some legitimate use cases for it in the SDK, so I'd really like to have it here.

> My recommendation would be to include these ABIs in memory since there's not too many of them.

I've only added a small handful of ABIs here now, but the intention is that we'd have more. I think the entire `src/common/ContractAddresses.ts` ABI content could be migrated across, for instance, because trying to maintain all those ABIs in a single file is becoming a pain. So the intention of this is that we should be able to add to it without worrying about the RAM overhead in the bots (...or the web browser).

> The one thing that I would recommend would be to cache these responses once they're read. fs.readFile is a heavy call.

We rarely (if ever?) request the same ABI multiple times because we normally instantiate a contract instance once and pass it around instead (with the ABI inside). If this was a hot path that we regularly called to read the same data then I think it'd make sense to cache it, but given our usage pattern it would just imply that we have two copies of the ABI in RAM instead of one.

test/abi.ts
pxrl: My thinking was that it's OK just to test for 1 happy path and 1 sad path, to make sure those two cases work OK. We could iterate over each known contract and verify that it's readable, but then we'd have to maintain that list or load it dynamically via an fs operation. That might detect if someone saves a bad ABI, but they'd anyway catch that as soon as they tried to use it, right?

package.json
```diff
@@ -14,9 +14,12 @@
...
-    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --removeComments --verbatimModuleSyntax false && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm && echo > ./dist/esm/package.json '{\"type\":\"module\",\"sideEffects\":false}'",
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --removeComments --verbatimModuleSyntax false && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}' && yarn copy-abi:cjs",
+    "build:esm": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm && echo > ./dist/esm/package.json '{\"type\":\"module\",\"sideEffects\":false}' && yarn copy-abi:esm",
     "build:types": "tsc --project tsconfig.build.json --module esnext --declarationDir ./dist/types --emitDeclarationOnly --declaration --declarationMap",
+    "copy-abi:cjs": "DIR=cjs yarn copy-abi",
+    "copy-abi:esm": "DIR=esm yarn copy-abi",
+    "copy-abi": "abi=utils/abi/contracts; dstdir=\"./dist/${DIR}/${abi}\"; mkdir -p \"${dstdir}\"; cp ./src/${abi}/*.json \"${dstdir}\"",
```
james-a-morris, line 22: +1 on the usage of ENV variables to prevent code reuse

src/utils/abi/index.ts
james-a-morris: > The one thing that I would recommend would be to cache these responses once they're read. `fs.readFile` is a heavy call.

Could we include a local caching mechanism to prevent having to resolve a specific ABI multiple times? File IO in node is heavy, and since we never change these ABIs, they're a good candidate for caching.
```suggestion
    return 
        CACHE[contractName] ??= JSON.parse(await fs.readFile(`${__dirname}/contracts/${contractName}.json`, { encoding: "utf8" }));
```

src/utils/abi/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ * URL="http://api.etherscan.io/api?module=contract&action=getabi&format=raw&address=${CONTRACT_ADDRESS}"
+ * curl -s "${URL}" | python -m json.tool > "./src/utils/abi/contracts/new-contract.json"
+ */
+
+/**
+ * @notice Retrieve an ABI desription from the set of known contracts.
+ * @param contractName Name of the contract ABI to retrieve.
+ * @returns Contract ABI as an ethers ContractInterface type.
+ */
+export async function getABI(contractName: string): Promise<ContractInterface> {
```
james-a-morris, line 17: Do you think we should be mindful of casing here? Like: `getABI("votingv2")` vs `getABI("VotingV2")`

james-a-morris (commented):
Left a few comments. It looks great so far 🔥

test/abi.ts
pxrl: I modified the test to read all files from the ABI dir and require that it doesn't throw when we read them. That should help to catch any bad files before they get merged :+1: 

https://github.com/across-protocol/sdk-v2/pull/420/commits/713d135b258f07ef6365e0f42a887c0413253d5a

src/utils/abi/index.ts
pxrl: I'd prefer to avoid trying to normalise case here - since this just does an fs lookup we're totally dependent on how the files are named when they're stored. I think it'll work fine  as-is :+1:

james-a-morris (commented):
I think this needs a rebase

james-a-morris (approved):
LGTM - thanks for reviewing my comments & rebasing

PR #422, opened 2023-10-02
refactor: include caching utils

james-a-morris:


src/utils/DepositUtils.ts
```diff
@@ -0,0 +1,73 @@
+import { assert } from "chai";
```
pxrl, line 1:
```suggestion
import assert from "asset";
```

.eslintrc.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
   parserOptions: {
     ecmaVersion: "latest",
     sourceType: "module",
-    project: "./tsconfig.json",
+    project: "./tsconfig.lint.json",
```
nicholaspai, line 17: explain this?

tsconfig.lint.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "include": ["./src/**/*.ts", "./test/**/*.ts"],
+  "compilerOptions": {
+    // Expand the rootDir to include the test directory
```
nicholaspai, line 5: Is this a syntax error to include a comment inside a JSON?

src/utils/DepositUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * not found in the cache, it is queried from the spoke pool client. If the deposit is found, it is cached before
+ * being returned.
+ * @param spokePoolClient The spoke pool client to use to query the deposit.
+ * @param fill The fill to resolve a deposit for.
+ * @param cache An optional cache to use to store the deposit. Optional.
+ * @returns The deposit for the fill, or undefined if the deposit could not be found.
+ * @throws If the fill's origin chain ID does not match the spoke pool client's chain ID.
+ * @throws If the spoke pool client has not been updated.
+ */
+export async function queryHistoricalDepositForFill(
```
nicholaspai, line 26: This is copied from your latest commit in `relayer-v2` right?

src/utils/CachingUtils.ts
```diff
@@ -1,7 +1,42 @@
...
   return latestTime - eventTimestamp >= cachingMaxAge;
 }
+
+/**
+ * Calls the cache's get method and returns the result if it is defined, otherwise returns undefined.
+ * @param key The key to get from the cache.
+ * @param cache The cache to get the key from.
+ * @returns The value associated with the key in the cache, or undefined if the key is not in the cache.
+ */
+export async function getDepositInCache(key: string, cache: CachingMechanismInterface): Promise<Deposit | undefined> {
```
nicholaspai, line 19: These are both copied from relayer-v2?

tsconfig.lint.json
james-a-morris: Nope! TSC reads [JSON5](https://json5.org/) which can allow comments 🙂

.eslintrc.js
james-a-morris: Specific linting tests require running the TSC compiler. We must select a TSC config that builds both the `test` and `src` directories to get linting.

The `tsconfig.lint.json` is a specific linting configuration in TSC that compiles both `src` and `test`. It's a new file because we don't want this configuration during buildtime.

src/utils/CachingUtils.ts
james-a-morris: These are copied from `relayer-v2` with the exception that it was originally `redisClient: RedisClient` and now it's `cache: CachingMechanismInterface`

src/utils/DepositUtils.ts
james-a-morris: This is coped from relayer-v2 on prod with the change that we are now using the caching mechanism interface instead of specifically the `RedisClient`

nicholaspai (approved):

src/utils/CachingUtils.ts
```diff
@@ -1,7 +1,42 @@
...
+  return isDefined(depositRaw) ? JSON.parse(depositRaw, composeRevivers(objectWithBigNumberReviver)) : undefined;
+}
+
+export async function setDepositInCache(
+  deposit: Deposit,
+  currentChainTime: number,
+  cache: CachingMechanismInterface,
+  expirySeconds = DEFAULT_CACHING_TTL
+): Promise<void> {
+  if (shouldCache(deposit.quoteTimestamp, currentChainTime, DEFAULT_CACHING_SAFE_LAG)) {
```
pxrl, line 30: It might be worth a comment to note that `deposit.quoteTimestamp` can be controlled by the user, within certain bounds. If `DEFAULT_CACHING_SAFE_LAG` was set to 1 hour, then a depositor could effectively force something into the cache immediately by setting their `quoteTimestamp` 1 hour in the past.

We do append the block timestamp to the `DepositWithBlock` object type, but that's not available in this case. On that, should we actually consider caching that additional information?

src/utils/ReviverUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @param _ The key to the value being revived. Unused.
+ * @param value The value being revived.
+ * @returns The revived value. If the value is not a stringified BigNumber object, it is returned as-is.
+ */
+export function objectWithBigNumberReviver(_: unknown, value: unknown): unknown {
+  if (value === null || typeof value !== "object") {
+    return value;
+  }
+  const resolvedValue = value as { type: string; hex: string };
+  return resolvedValue.type === "BigNumber" ? toBN(resolvedValue.hex) : resolvedValue;
```
pxrl, line 30: Would `BigNumber.isBigNumber()` be worthwhile here? https://docs.ethers.org/v5/single-page/#/v5/api/utils/bignumber/-%23-BigNumber--BigNumber--methods--inspection

james-a-morris: Unfortunately not. This can be exemplified in this function - the stringify/parsing isn't equivalent:
```
const bn = ethers.BigNumber.from(4);
const stringified = JSON.stringify(bn);
const parsed = JSON.parse(stringified);
console.log(ethers.BigNumber.isBigNumber(parsed)) <- prints false
```

src/utils/CachingUtils.ts
james-a-morris: This is outside the scope of this PR, but worth a check. Wdyt @nicholaspai would this be worth exploring in this cycle?

src/utils/ReviverUtils.ts
pxrl: Thanks for the clear example!

pxrl (approved):
Some minor comments, but no hard blockers :+1:

src/utils/CachingUtils.ts
james-a-morris: Perfect - I also added an additional check just to make sure that the variables themselves are not outside of the range.

https://github.com/across-protocol/sdk-v2/pull/422/commits/e128a05c3179bc913b73cbf47aaa8d0ee73f2d2c

src/utils/DepositUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  let deposit: DepositWithBlock, cachedDeposit: Deposit | undefined;
+  if (cache) {
+    cachedDeposit = await getDepositInCache(getDepositKey(fill), cache);
+  }
+
+  if (isDefined(cachedDeposit)) {
+    deposit = cachedDeposit as DepositWithBlock;
+    // Assert that cache hasn't been corrupted.
+    assert(deposit.depositId === fill.depositId && deposit.originChainId === fill.originChainId);
```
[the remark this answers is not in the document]
james-a-morris: We should circle back to this

PR #423, opened 2023-10-03
chore: make async fn with no await an error

james-a-morris:
This PR makes floating promises an error in the SDK. There was a slight lift to convert non-async functions with the async tag into nonsync counterparts. 

Going forward, the CI will tag these warnings as an error.


test/relayFeeCalculator.test.ts
```diff
@@ -36,12 +36,12 @@ const testCapitalCostsConfig: { [token: string]: any } = {
 // Example of how to write this query class
 class ExampleQueries implements QueryInterface {
   constructor(private defaultGas = "305572") {}
-  async getGasCosts(): Promise<BigNumberish> {
-    return gasCost(this.defaultGas, 1e9); // 1 gwei
+  getGasCosts(): Promise<BigNumberish> {
+    return Promise.resolve(gasCost(this.defaultGas, 1e9)); // 1 gwei
```
nicholaspai, line 40: Why can't you just return `gasCost` if `gasCost` returns a Promise?

nicholaspai (approved):

test/relayFeeCalculator.test.ts
james-a-morris: `gasCost` unfortunately doesn't return a promise:

```
/**
 * Convert a gas amount and gas price to wei.
 *
 * @param {number} gas - gas amount.
 * @param {BigNumberish} gasPrice - gas price in gwei.
 * @returns {BigNumber} - total fees in wei.
 */
export const gasCost = (gas: BigNumberish, gasPrice: BigNumberish): BigNumber => {
  return BigNumber.from(gas).mul(gasPrice);
};
```

package.json
```diff
@@ -19,8 +19,8 @@
     "build:types": "tsc --project tsconfig.build.json --module esnext --declarationDir ./dist/types --emitDeclarationOnly --declaration --declarationMap",
     "test": "hardhat test",
     "test:watch": "hardhat watch test",
-    "lint": "eslint --fix src",
-    "lint-check": "eslint src",
+    "lint": "eslint --fix src test",
```
nicholaspai, line 22: side note: do you think we need to lint `e2e` also?

And do you think we should lint the `hardhat.config.ts`?

nicholaspai (approved):

package.json
james-a-morris: Done

pxrl (approved):
LGTM!

PR #424, opened 2023-10-03
improve: Build in parallel

pxrl:
This runs each of the build pipelines in parallel. On my system, it reduces the build time from ~14 seconds to ~7 seconds.


james-a-morris:
This is 🔥

james-a-morris (approved):

PR #425, opened 2023-10-03
refactor: move uba client tests

james-a-morris:


james-a-morris:
@nicholaspai @pxrl several async functions were converted to non async functions to appease the linter.

nicholaspai:
Are these all testing logic that lives only in the SDK and not in the relayer repo? Or are there functions we need to move over

nicholaspai (approved):

james-a-morris:
> Are these all testing logic that lives only in the SDK and not in the relayer repo? Or are there functions we need to move over

These are purely testing logic within the SDK. There were no functions that needed to be ported over.

PR #426, opened 2023-10-04
feat: Simplify HubPoolClient interface for returning L1 --> L2 token mappings

nicholaspai:
Fixes ACX-1585

# Motivation

The impetus for this PR was to change how the SpokePoolClient assigns `destinationToken` to deposits it queries. 

## UPDATE

After reviewing this PR, we've realized that its still going to be hard/impossible to prevent deposits/fills using all of the same token mappings in the same bundle. This is because bundles get "validated" arbitrarily from the perspective of the relayer, so fills that are 10 blocks apart might see different "latest" token mappings. Moreover, the possibility for old deposits to get filled also makes it difficult for relayers to support if they have to use old `destinationTokens` to fill them.

The simplified change to this PR is to basically keep existing logic in the `Relayer` to use the latest `destinationToken` as the token to fill a deposit with.

This PR still overhauls the interface in the `HubPoolClient` to make it more intuitive. This does mean that the process for upgrading token mappings will be:
- Pause deposits for upgraded token in the FE/API
- Flush out all existing deposits
- Upgrade `PoolRebalanceRoutes`
- Wait some time (i.e. for a bundle to pass)
- Re-enable deposits

### New Interface

The HubPoolClient contains a lot of confusing and overlapping functions that return L2 and L1 token relationships. This PR consolidates them all into two internal functions:
- `protected getL2TokenForL1TokenAtBlock`
- `protected getL1TokenForL2TokenAtBlock`

The new interface that should be called is then:
- `public getL1TokenForDeposit`
- `public getL2TokenForDeposit`

## Deprecated PR comments:

Instead of assigning it to the LATEST destination token for a chain, use the destination token set as of the latest validated mainnet bundle end block. This makes the SpokePoolClient more robust against updated L2 token addresses, unlocking the future ability to:
- Upgrade L2 token addresses without having to pause deposits for those L2 tokens in between bundles

Specifically, the features unlocked from this PR are:

1. The Relayer will fill deposits using the destination token set at the time of the latest validated bundle preceding the deposit. Currently, the relayer uses the latest destination token mapped to a deposited token.
2. The Dataworker will propose root bundles with RelayerRefundRoots where `l2Token` is equal to the mapped token as of the time of the last validated bundle prior to the proposed bundle. Currently the `l2Token` is equal to the latest mapped token, which could cause some accounting issues because the tokens sent from the Hub to the Spoke might be a different L2 token.

### Details

The HubPoolClient contains a lot of confusing and overlapping functions that return L2 and L1 token relationships. This PR consolidates them all into two internal functions:
- `protected getL2TokenForL1TokenAtBlock`
- `protected getL1TokenForL2TokenAtBlock`

The new interface that should be called is then:
- `public getL1TokenForDeposit`
- `public getL2TokenForDeposit`

These public functions make sure that the token mapping is fetched for the nearest bundle end block to the deposit, as opposed to the deposit quote block. This way all deposits  in the same bundle are seamlessly are mapped to the same L1/L2 tokens.

#### Deposits

Like the relayer, the `SpokePoolClient` will set the `destinationToken` value for each `deposit` event it queries equal to the token set for the `destinationChainId` as of the latest validated bundle preceding that deposit.

#### Fills

The dataworker relies on the `SpokePoolClient` function `validateFillForDeposit` which matches the `deposit` event against the passed in fill event. So it also should fall back to the above behavior.

##### PoolRebalanceRoot

PoolRebalanceRoutes use L1 tokens for L2 events for accounting, and the dataworker relies heavily on [this function `getL1TokenCounterpartAtBlock`](https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/PoolRebalanceUtils.ts#L132) to figure out which L1 token to map to a given deposit/fill's token. This will be replaced with `getL1TokenForDeposit` or a call to `getL1TokenForL2TokenAtBlock` to make sure that the L1 token is grabbed as of the time as of the last validated bundle.

##### RelayerRefundRoot

RelayerRefund leaves need to set `l2Token` value, and they rely on [this function `getDestinationTokenForL1Token`](https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/DataworkerUtils.ts#L263) which **doesn't** currently look up the token mapping at a specific block height. This needs to be replaced wth a call to `getL2TokenForDeposit` or `getL2TokenForL1TokenAtBlock` to force the dataworker to use the token mapping as of the latest validated bundle.

This way, the L1 tokens sent from the PoolRebalanceRoot from Hub to Spoke are converted into the same L2 tokens expected to be paid out of the spoke.

##### SlowFillRoot

SlowFill leaves will set the `destinationToken` based on the `FilledRelay` event so this falls back to the above `SpokePoolClient` behavior.

#### How this will affect the Relayer

The relayer will fill the deposit using the `destinationToken` on the `destinationChainId` as set at the time of the latest validated bundle preceding the deposit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.17.11",
+  "version": "0.17.17",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/v/developer-docs/developers/across-sdk",
   "files": [
@@ -12,8 +12,8 @@
     "node": ">=16.18.0"
   },
   "scripts": {
-    "start": "nodemon -e ts,tsx,json,js,jsx --watch . --ignore ./dist --exec 'yarn build'",
-    "build": "yarn run clean && yarn run build:cjs && yarn run build:esm && yarn run build:types",
+    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build'",
+    "build": "yarn run clean && yarn run build:cjs & yarn run build:esm & yarn run build:types; wait",
     "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --removeComments --verbatimModuleSyntax false && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}' && yarn copy-abi:cjs",
     "build:esm": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm && echo > ./dist/esm/package.json '{\"type\":\"module\",\"sideEffects\":false}' && yarn copy-abi:esm",
     "build:types": "tsc --project tsconfig.build.json --module esnext --declarationDir ./dist/types --emitDeclarationOnly --declaration --declarationMap",
@@ -97,8 +97,8 @@
   },
   "dependencies": {
     "@across-protocol/across-token": "^1.0.0",
-    "@across-protocol/constants-v2": "^1.0.4",
-    "@across-protocol/contracts-v2": "^2.4.4",
+    "@across-protocol/constants-v2": "^1.0.7",
+    "@across-protocol/contracts-v2": "^2.4.7",
     "@eth-optimism/sdk": "^2.1.0",
     "@pinata/sdk": "^2.1.0",
     "@types/mocha": "^10.0.1",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "node": ">=16.18.0"
   },
   "scripts": {
-    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build'",
+    "start": "nodemon -e ts,tsx,json,js,jsx --watch . --ignore ./dist --exec 'yarn build'",
     "build": "yarn run clean && yarn run build:cjs && yarn run build:esm && yarn run build:types",
     "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --removeComments --verbatimModuleSyntax false && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}' && yarn copy-abi:cjs",
     "build:esm": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm && echo > ./dist/esm/package.json '{\"type\":\"module\",\"sideEffects\":false}' && yarn copy-abi:esm",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.17.17",
+  "version": "0.18.0",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/v/developer-docs/developers/across-sdk",
   "files": [
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -358,7 +358,8 @@ export class HubPoolClient extends BaseAbstractClient {
       );
 
       // Resolve the HubPool token address, if it isn't already known.
-      const hubPoolToken = hubPoolTokens[originToken] ?? this.getL1TokenForDeposit(deposit);
+      const quoteBlockNumber = quoteBlocks[deposit.quoteTimestamp];
+      const hubPoolToken = hubPoolTokens[originToken] ?? this.getL1TokenForDeposit({ ...deposit, quoteBlockNumber });
       hubPoolTokens[originToken] ??= hubPoolToken;
 
       // Append the quoteTimestamp for this HubPool token, if it isn't already enqueued.
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -1,3 +1,4 @@
+import assert from "assert";
 import { BigNumber, Contract, Event, EventFilter } from "ethers";
 import _ from "lodash";
 import winston from "winston";
@@ -17,13 +18,15 @@ import {
   PendingRootBundle,
   ProposedRootBundle,
   ProposedRootBundleStringified,
+  RealizedLpFee,
   SetPoolRebalanceRoot,
   TokenRunningBalance,
 } from "../interfaces";
 import * as lpFeeCalculator from "../lpFeeCalculator";
 import {
-  BigNumberish,
   BlockFinder,
+  bnZero,
+  dedupArray,
   EventSearchConfig,
   MakeOptional,
   assign,
@@ -31,6 +34,7 @@ import {
   getCachedBlockForTimestamp,
   getCurrentTime,
   isDefined,
+  mapAsync,
   paginatedEventQuery,
   shouldCache,
   sortEventsDescending,
@@ -46,7 +50,6 @@ import { isUBAActivatedAtBlock } from "./UBAClient/UBAClientUtilities";
 type _HubPoolUpdate = {
   success: true;
   currentTime: number;
-  latestBlockNumber: number;
   pendingRootBundleProposal: PendingRootBundle;
   events: Record<string, Event[]>;
   searchEndBlock: number;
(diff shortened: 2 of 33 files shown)


test/HubPoolClient.DepositToDestinationToken.ts
```diff
@@ -22,9 +23,37 @@ import {
 
 let hubPool: Contract, lpTokenFactory: Contract, mockAdapter: Contract;
 let owner: SignerWithAddress;
-let hubPoolClient: HubPoolClient;
+let hubPoolClient: MockHubPoolClient;
 
 describe("HubPoolClient: Deposit to Destination Token", function () {
```
nicholaspai, line 28: All of the new HubPoolClient functions are unit tested rigorously in this file, PTAL

mrice32 (commented):
> use the destination token set as of the latest validated mainnet bundle end block.

Will this accomplish the goal laid out in the proposal? I think there will still be the possibility of bundles with multiple token mappings depending on the deposit, right? Since some deposits came in before the last bundle was validated and some came in after, but all will be included in the next bundle. Presumably, we'd want the mapping to be known at the time of deposit and not change after the fact, right?

pxrl:
> I think there will still be the possibility of bundles with multiple token mappings depending on the deposit, right? Since some deposits came in before the last bundle was validated and some came in after, but all will be included in the next bundle.

I am also curious about this. It seems like a general challenge that we might have latent deposits that get filled in future. If the destinationToken is locked in at the time of their deposit then it's difficult to be sure that we're not impacting some depositors. We have previously seen them come out of the woodwork several months after a Deposit is made, and request that it be filled.

nb. Given that the destinationToken is not strictly identified in the Deposit event, it is tempting to require whatever SpokePool token is configured at the time of the fill. I concede that this might upset some users, though.

> This makes the SpokePoolClient more robust against updated L2 token addresses, unlocking the future ability to:
> 
> * Upgrade L2 token addresses without having to pause deposits for those L2 tokens in between bundles

I'm unsure about the benefit of being able to upgrade without pausing deposits. Is it such a big benefit?

There's still a difficult residual issue that relayers don't know whether the current proposal might be disputed, so it's actually hard for them to have confidence in the start block of any current configuration until liveness has elapsed. Pausing deposits (or disabling individual routes, for a single HubPool token) gives a way of handling this cleanly.

mrice32 (commented):
My only minor push with this updated plan is that using the beginning of the bundle or the "latest" value means that relayers will be unable to predict their refund token, since both of these can change between the time a deposit is made and the repayment is made.

If we use quote timestamp instead, relayers can always predict the refund token by doing a lookup at that timestamp.

Thoughts?

pxrl:
> My only minor push with this updated plan is that using the beginning of the bundle or the "latest" value means that relayers will be unable to predict their refund token, since both of these can change between the time a deposit is made and the repayment is made.
> 
> If we use quote timestamp instead, relayers can always predict the refund token by doing a lookup at that timestamp.
> 
> Thoughts?

Hadn't considered this :thinking:. Do you think it's a showstopper given the current settlement period of ~1.5 - 2 hours? i.e. maximum relayer time exposure is fairly well-bounded, right?
 
I'm still concerned about the problem of weeks or months-old deposits resurfacing, and forcing some potentially unsupported token to be reinjected into an Across bundle. Consider for example the recent BAL issue on Base - what would we do if BAL had been enabled, and someone made a deposit destined for Base that didn't get filled until now? The only way I can see to mitigate that is for the destination token (and refund token) to be determined at the time of the fill.

nicholaspai:
> My only minor push with this updated plan is that using the beginning of the bundle or the "latest" value means that relayers will be unable to predict their refund token, since both of these can change between the time a deposit is made and the repayment is made.
> 
> If we use quote timestamp instead, relayers can always predict the refund token by doing a lookup at that timestamp.
> 
> Thoughts?

The problem with using the `quoteTimestamp` is when a deposit goes unfilled for a long time. Let's say 30 days. At the time of the deposit, the token mapping might be to one token, but 30 days later it might have changed. The relayer might have reallocated its inventory to the new token so it can't fill this deposit anymore (easily).

Also to be clear, not sure if @pxrl caught this, but the current plan is to not pick the token mapping as of the latest validated bundle, but to just use `latestBlockNumber`. This is primarily to reduce complexity, and because even pegging the token mapping to a bundle block doesn't prevent the case where fills in the same bundle can be mapped to different tokens. This is again, because of the old deposit case.

How do you guys think we should proceed? Let's assume we'll be OK to manually pause deposits for updated tokens for a full bundle to allow for the upgrade.

pxrl:
> Also to be clear, not sure if @pxrl caught this, but the current plan is to not pick the token mapping as of the latest validated bundle, but to just use `latestBlockNumber`. This is primarily to reduce complexity, and because even pegging the token mapping to a bundle block doesn't prevent the case where fills in the same bundle can be mapped to different tokens. This is again, because of the old deposit case.

This also implised that deposits _must_ be paused right, since otherwise there's a race condition with relayers who prepare their fills without noticing that the underlying token mapping has changed. It also introduces a link betwen Spoke and Hub timestamps, but I guess we can probably live with that if there's a pause buffer around the config change.

mrice32:
> > Also to be clear, not sure if @pxrl caught this, but the current plan is to not pick the token mapping as of the latest validated bundle, but to just use `latestBlockNumber`. This is primarily to reduce complexity, and because even pegging the token mapping to a bundle block doesn't prevent the case where fills in the same bundle can be mapped to different tokens. This is again, because of the old deposit case.
> 
> This also implies that deposits _must_ be paused right, since otherwise there's a race condition with relayers who prepare their fills without noticing that the underlying token mapping has changed. It also introduces a link betwen Spoke and Hub timestamps, but I guess we can probably live with that if there's a pause buffer around the config change.

If we use latestTimestamp, is it possible to determine correctness in a bundle after there's a token change? How can you verify that the relay used the "right" token mapping for that moment?

nicholaspai:
> > > Also to be clear, not sure if @pxrl caught this, but the current plan is to not pick the token mapping as of the latest validated bundle, but to just use `latestBlockNumber`. This is primarily to reduce complexity, and because even pegging the token mapping to a bundle block doesn't prevent the case where fills in the same bundle can be mapped to different tokens. This is again, because of the old deposit case.
> > 
> > 
> > This also implies that deposits _must_ be paused right, since otherwise there's a race condition with relayers who prepare their fills without noticing that the underlying token mapping has changed. It also introduces a link betwen Spoke and Hub timestamps, but I guess we can probably live with that if there's a pause buffer around the config change.
> 
> If we use latestTimestamp, is it possible to determine correctness in a bundle after there's a token change? How can you verify that the relay used the "right" token mapping for that moment?

Hm that's a good point, so maybe we should really be using the latest mapping as of the latest validated bundle, which is the change I originally made. This also allows really old deposits to be filled with newer token mappings

package.json
```diff
@@ -12,7 +12,7 @@
     "node": ">=16.18.0"
   },
   "scripts": {
-    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build'",
+    "start": "nodemon -e ts,tsx,json,js,jsx --watch . --ignore ./dist --exec 'yarn build'",
```
james-a-morris, line 15: This will watch for ENV file changes  +1

src/clients/HubPoolClient.ts
```diff
@@ -165,74 +165,91 @@ export class HubPoolClient extends BaseAbstractClient {
...
-      return sortEventsDescending(this.l1TokensToDestinationTokensWithBlock[_l1Token][l2ChainId]).find(
-        (mapping: DestinationTokenWithBlock) => mapping.l2Token === l2Token && mapping.blockNumber <= hubPoolBlock
-      );
-    });
-    if (!l1Token) {
+    return l2Token.l2Token;
+  }
+
+  // Returns the latest L1 token to use for an L2 token as of the input hub block.
+  protected getL1TokenForL2TokenAtBlock(
```
pxrl, line 190: @nicholaspai Do you think this needs to be protected? `getXForDeposit()` is really convenient but there are also cases where we don't have a deposit object, and having to coerce the input to look like a deposit is inconvenient. Examples are when we have a Fill (no `originToken` or `quoteBlockNumber`), or in the InventoryClient where we are ignorant of Deposits entirely.

Any objections to removing `protected` here to allow these callers to use it?

src/clients/HubPoolClient.ts
```diff
@@ -165,74 +165,91 @@ export class HubPoolClient extends BaseAbstractClient {
...
+        // If this token doesn't exist on this L2, skip it
+        if (this.l1TokensToDestinationTokensWithBlock[l1Token][destinationChainId] === undefined) {
+          return undefined;
+        }
+        // Return all matching L2 token mappings that are equal to or earlier than the target block.
+        return this.l1TokensToDestinationTokensWithBlock[l1Token][destinationChainId].filter(
+          (mapping) => mapping.l2Token === l2Token && mapping.blockNumber <= latestHubBlock
+        );
+      })
+      .filter(isDefined)
```
pxrl, line 206: Seems like we can use `this.l2TokenEnabledForL1Token()` here?

```suggestion
      .filter((l1Token) => this.l2TokenEnabledForL1Token(l1Token, destinationChainId)
      .map((l1Token) => {
        // Return all matching L2 token mappings that are equal to or earlier than the target block.
        return this.l1TokensToDestinationTokensWithBlock[l1Token][destinationChainId].filter(
          (mapping) => mapping.l2Token === l2Token && mapping.blockNumber <= latestHubBlock
        );
      });
```

src/clients/HubPoolClient.ts
```diff
@@ -163,74 +163,121 @@ export class HubPoolClient extends BaseAbstractClient {
...
+
+  /**
+   * Returns the L1 token that should be used for an L2 Bridge event. This function is
+   * designed to be used by the caller to associate the L2 token with its mapped L1 token
+   * at the HubPool equivalent block number of the L2 event.
+   * @param deposit Deposit event
+   * @param returns string L1 token counterpart for Deposit
+   */
+  getL1TokenForDeposit(deposit: Pick<DepositWithBlock, "blockNumber" | "originToken" | "originChainId">): string {
+    // Destination token should be set equal to the L2 token set as of the latest
```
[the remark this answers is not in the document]
nicholaspai: yep

src/clients/HubPoolClient.ts
nicholaspai: nope lets remove

pxrl: I have made this a public method and updated the mocks to implement the necessary instrumentation for mocking l1 -> l1 token mappings.

pxrl (approved):
@nicholaspai I've made some updates here and have everything testing OK against the relayer. I think we should wait for the current `latestBlockNumber` / `latestBlockSearched` changes to filter in and bump SDK before we merge this, but otherwise I think it's ready to go.

PR #427, opened 2023-10-05
feat: add message handler

james-a-morris:
The aim of this Pull Request is to include gas estimation handling messages within the RelayFeeCalculator class.


src/relayFeeCalculator/chain-queries/arbitrum.ts
```diff
@@ -9,22 +9,12 @@ export class ArbitrumQueries extends QueryBase {
...
-      provider,
-      symbolMapping,
-      spokePoolAddress,
-      usdcAddress,
-      simulatedRelayerAddress,
-      gasMarkup,
-      logger,
-      coingeckoProApiKey
-    );
+    super(provider, symbolMapping, spokePoolAddress, simulatedRelayerAddress, gasMarkup, logger, coingeckoProApiKey);
```
james-a-morris, line 17: A common paradigm in this PR - we can't really specify a specific address anymore since we need this context to be passed into the `handleAcrossMessage` function

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -63,11 +64,13 @@ export default abstract class QueryBase implements QueryInterface {
    * Retrieves the current gas costs of performing a fillRelay contract at the referenced Spoke Pool
    * @returns The gas estimate for this function call (multplied with the optional buffer)
    */
-  async getGasCosts(): Promise<BigNumberish> {
-    const tx = await createUnsignedFillRelayTransaction(this.spokePool, this.usdcAddress, this.simulatedRelayerAddress);
+  async getGasCosts(deposit: Deposit, relayAddress?: string): Promise<BigNumberish> {
+    const relayerToSimulate = relayAddress ?? this.simulatedRelayerAddress;
+    const correspondingFill = buildFillForSimulatingFullDeposit(deposit, relayerToSimulate);
```
james-a-morris, line 69: Note: I am verbose here so that this function can be used to simulate other fills in the future.

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -198,13 +201,38 @@ export class RelayFeeCalculator {
...
-  async gasFeePercent(amountToRelay: BigNumberish, tokenSymbol: string, _tokenPrice?: number): Promise<BigNumber> {
+  async gasFeePercent(
+    amountToRelay: BigNumberish,
+    tokenSymbol: string,
+    originChainId: number,
+    destinationChainId: number,
+    messagePayload?: {
+      message: string;
+      recipientAddress: string;
+    },
```
james-a-morris, line 212: This has been added like this to enforce a mutual requirement. Either we don't include any of these values (and default to the normal gas estimation) or we include `message` and are forced to include the recipient.

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -276,13 +304,26 @@ export class RelayFeeCalculator {
   async relayerFeeDetails(
     amountToRelay: BigNumberish,
     tokenSymbol: string,
-    tokenPrice?: number,
-    _originRoute?: ChainIdAsString,
-    _destinationRoute?: ChainIdAsString
+    originRoute: ChainIdAsString,
+    destinationRoute: ChainIdAsString,
```
james-a-morris, line 306: We don't need these as underlined variables anymore. This is a holdover from when this function was stubbed.

src/utils/AddressUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export async function isContractAddress(address: string, provider: providers.Provider): Promise<boolean> {
+  // A base case for if the address is null
+  if (!address) {
+    return false;
+  }
+  // Retrieve the code at the address
+  const code = await provider.getCode(address);
+  // If the code is not empty, then there is a contract at this address
+  return code !== "0x";
```
james-a-morris, line 17: This heuristic should get us whether or not a contract is deployed at this address

src/utils/AddressUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Checks if a contract is deployed at the given address
+ * @param address The ETH address to check
+ * @param provider A valid Ethers.js provider
+ * @returns A boolean indicating if a contract is deployed at the given address or not (true = contract, false = no contract)
+ */
+export async function isContractAddress(address: string, provider: providers.Provider): Promise<boolean> {
+  // A base case for if the address is null
+  if (!address) {
```
james-a-morris, line 11: This just makes it easier for me to not have to add a conditional when trying to call this function in the event that the calling fn's address is potentially undefined

package.json
```diff
@@ -94,6 +95,7 @@
   },
   "dependencies": {
     "@across-protocol/across-token": "^1.0.0",
+    "@across-protocol/constants-v2": "^1.0.4",
```
nicholaspai, line 98: +1, i forgot about this

src/constants.ts
```diff
@@ -52,3 +52,5 @@ export const DEFAULT_CACHING_SAFE_LAG = 60 * 60 * 24; // 1 Day
 
 export const UBA_BOUNDS_RANGE_MAX = BigNumber.from(String(Number.MAX_SAFE_INTEGER)).mul(utils.parseEther("1.0"));
 export const UBA_BOUNDS_RANGE_MIN = UBA_BOUNDS_RANGE_MAX.mul(-1);
+
+export const RISK_LABS_RELAYER_ADDRESS = "0x428ab2ba90eba0a4be7af34c9ac451ab061ac010";
```
nicholaspai, line 56: nit: should this be checksummed? I also don't like this name
```suggestion
export const DEFAULT_SIMULATED_RELAYER_ADDRESS = "0x428ab2ba90eba0a4be7af34c9ac451ab061ac010";
```

src/utils/AddressUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+    return false;
+  }
+  // Retrieve the code at the address
+  const code = await provider.getCode(address);
```
nicholaspai, line 15: This might be useful rather than rolling our own: https://docs.ethers.org/v5/api/utils/address/#utils-isAddress

src/utils/DepositUtils.ts
```diff
@@ -71,3 +73,38 @@ export async function queryHistoricalDepositForFill(
...
+  const originToken = resolveContractFromSymbol(tokenSymbol, originChainId);
+  const destinationToken = resolveContractFromSymbol(tokenSymbol, destinationChainId);
+  if (!isDefined(originToken) || !isDefined(destinationToken)) {
+    throw new Error(`Could not resolve token contract for ${tokenSymbol} on ${originChainId} or ${destinationChainId}`);
+  }
+
+  return {
+    // We want to set the deposit ID to 0 so that there
+    // is no chance of a collision with a real deposit
+    depositId: 0,
```
nicholaspai, line 97: What's wrong with a collision?

src/utils/FillUtils.ts
```diff
@@ -0,0 +1,33 @@
+import { RISK_LABS_RELAYER_ADDRESS } from "../constants";
+import { Deposit, Fill } from "../interfaces";
+import { bnZero } from "./BigNumberUtils";
+
+export function buildFillForSimulatingFullDeposit(deposit: Deposit, relayerAddress = RISK_LABS_RELAYER_ADDRESS): Fill {
+  return {
+    amount: deposit.amount,
+    fillAmount: deposit.amount, // We're simulating a full fill
```
nicholaspai, line 8: So I'm sure you're considering how to handle this in a future PR but there is a chance that the `deposit.amount > simulated relayer token balance` leading to unexpected reverts FYI

src/utils/common.ts
```diff
@@ -290,33 +292,57 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+  // Only continue sanity checks if the message is not empty
+  if (!isMessageEmpty) {
+    const isRecipientAnAddress = ethers.utils.isAddress(recipientAddress);
+    const isRelayerAnAddress = ethers.utils.isAddress(relayerAddress);
+    if (!isRecipientAnAddress || !isRelayerAnAddress) {
+      throw new Error("Could not simulate message fill. Recipient address or relayer address is not a valid address");
+    }
+    const [isRecipientAContract, relayerBalanceOfToken] = await Promise.all([
+      isContractAddress(recipientAddress, provider),
+      provider.getBalance(fillToSimulate.destinationToken),
```
nicholaspai, line 323: this is incorrect, this is the relayer's `ETH` balance. The token balance is `ERC20.balanceOf`

src/utils/common.ts
```diff
@@ -290,33 +292,57 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+    ]);
+    if (!isRecipientAContract) {
+      throw new Error("Could not simulate message fill. Recipient address is not a contract address");
+    }
+    if (toBN(relayerBalanceOfToken).lt(toBN(amountToRelay))) {
```
nicholaspai, line 328: In this case, you might want to instead reset `amountToRelay = relayerBalanceOfToken` so that the constructed fill still is a "full fill" but it can be simulated.

nicholaspai (commented):
Good start

src/utils/DepositUtils.ts
james-a-morris: This was a result of a conversation I had with @pxrl. He mentioned to avoid potentially causing collisions in the test

src/utils/common.ts
james-a-morris: This is a holdover from a less recent change. It's resolved here: https://github.com/across-protocol/sdk-v2/pull/427/commits/860c2c3c1f3235064e37af1fcb25d5b3d728d077

src/utils/FillUtils.ts
james-a-morris: We handle it in a later case: 
```
  if (!isMessageEmpty) {
    const isRecipientAnAddress = ethers.utils.isAddress(recipientAddress);
    const isRelayerAnAddress = ethers.utils.isAddress(relayerAddress);
    if (!isRecipientAnAddress || !isRelayerAnAddress) {
      throw new Error("Could not simulate message fill. Recipient address or relayer address is not a valid address");
    }
    const [isRecipientAContract, relayerBalanceOfToken] = await Promise.all([
      isContractAddress(recipientAddress, provider),
      getTokenBalance(relayerAddress, fillToSimulate.destinationToken, provider),
    ]);
    if (!isRecipientAContract) {
      throw new Error("Could not simulate message fill. Recipient address is not a contract address");
    }
    if (toBN(relayerBalanceOfToken).lt(toBN(amountToRelay))) {
      throw new Error("Could not simulate message fill. Partial fills are not supported with message relaying");
    }
  }
```

src/utils/AddressUtils.ts
james-a-morris: This function is used to determine if there's a contract deployed at an address. The `isAddress` is just a regex checker and does not determine if the address specified has a contract deployed to it. We'd ideally only want to relay a message to a contract that exposes `handleAcrossMessage` so this is an early signal that we have a malformed reqeust.

src/constants.ts
james-a-morris: We could checksum it, but it shouldn't affect the usage in our logic because ethers internally checksum the addresses it receives. For readability, I've checksum the address & refactored the naming. https://github.com/across-protocol/sdk-v2/pull/427/commits/2b6a569af4bbd07119bde8c56dce11bd58c8a18b

src/utils/common.ts
james-a-morris: Right, but what if the contract that implements the `handleAcrossMessage` is specifically looking at `amount` and `tokenSent` as a conditional path in their logic? Couldn't changing the request intrinsically affect the gas cost estimation?

src/utils/FillUtils.ts
james-a-morris: If `toBN(relayerBalanceOfToken).lt(toBN(amountToRelay))` is true, then we unwind the stack and don't actually simulate the costs.

src/relayFeeCalculator/relayFeeCalculator.ts
nicholaspai: I'm not sure just these should be a mutual requirement. For example, would there really ever be a reason why we'd have `recipientAddress` be undefined? Some of the other parameters seem just as required like the chain.

src/constants.ts
nicholaspai: I just think we should always store hardcoded addresses as checksummmed as it can only help

src/utils/AddressUtils.ts
nicholaspai: hmmm great counterpart, ignore my suggestion then

src/utils/DepositUtils.ts
nicholaspai: I don't understand why we need to avoid collisions can you explain more?

src/utils/common.ts
nicholaspai: ah https://github.com/across-protocol/sdk-v2/commit/860c2c3c1f3235064e37af1fcb25d5b3d728d077#r129265169

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: Hmm, that's fair, but we didn't require them before. Since we're essentially falling back to previous logic when we don't include the message, should we enforce the recipient? We would need to change the calling references in the FE as well.

src/utils/common.ts
james-a-morris: Done and Done https://github.com/across-protocol/sdk-v2/pull/427/commits/72e4c2b53b8c8bf6bfeb752ffa7f2965247798d0 🙂

src/utils/common.ts
james-a-morris: Ah shoot that's a fair point. Thoughts on doing what we did with the `amountTooLow` variable? I.e. we pass another optional query flag to the request like `ignoreMessageFlowFillWarning` and we could relax that restraint conditionally?

src/relayFeeCalculator/relayFeeCalculator.ts
nicholaspai: I do think we might need to introduce a breaking change to this interface and require `recipientAddress` and `message` going forward. What do you think long term is more scalable to better support clients?

src/utils/common.ts
nicholaspai: hmmm that could work. Do you think the parameter/interface is too big and we should keep the calculator simpler? This would mean foisting off the responsibility to the client to determine whether to catch the result and then resubmit with a lower amount? I'm not sure how to lean here, what do you think?

james-a-morris: Right - I'd posit that any developer using this specific endpoint would rather their incorrect assumption of the system return a loud error than for it to return a quote

src/relayFeeCalculator/relayFeeCalculator.ts
mrice32: Agreed. As long as we are okay making a breaking change, I think requiring a recipient address and making the message default to empty would be the best interface going forward.

We could do some defaulting behavior or make them optional, but a required recipient will probably produce more accurate results and require less branching/casework.

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -276,13 +302,26 @@ export class RelayFeeCalculator {
...
+    messagePayload?: {
+      message: string;
+      recipientAddress: string;
+      relayerAddress?: string;
```
mrice32, line 310: This is a little weird since relayer address isn't really part of the message payload. Does it need to be coupled to the other two? Seems like it could just be a separate optional arg (as said above, I think all these should just be decomposed anyway).

src/utils/DepositUtils.ts
mrice32: The issue with an exact collision is that the relay could already be filled, resulting in a revert. A collision is unlikely, though, as all the other params would have to line up. Agreed on avoiding them, though, by making the deposit id unreachable.

Were there deposits with deposit id 0? I forget. If not, 0 is fine.

Note: if people know this number, they could technically try to trigger a collision by sending bad relays. A fix for that would be to hide it via env, but I think this is unlikely and hard to pull off given all the other params would need to match, too.

src/utils/DepositUtils.ts
```diff
@@ -71,3 +73,38 @@ export async function queryHistoricalDepositForFill(
...
+    amount: toBN(amountToRelay),
+    depositor: relayerAddress ?? ZERO_ADDRESS,
```
mrice32, line 99: nit: prefer a known default (like we do today for recipient), random address, or [0xFFFFF....](https://etherscan.io/address/0xffffffffffffffffffffffffffffffffffffffff), since the zero address will probably underestimate the calldata costs (0 bytes are cheaper than nonzero bytes).

src/utils/DepositUtils.ts
```diff
@@ -71,3 +73,38 @@ export async function queryHistoricalDepositForFill(
...
+    destinationChainId: Number(destinationChainId),
+    originChainId: Number(originChainId),
+    message: messagePayload?.message ?? "0x",
+    originToken,
+    destinationToken,
+    recipient: messagePayload?.recipientAddress ?? ZERO_ADDRESS,
```
mrice32, line 105: Same here, zero address will underestimate costs. If we use a default address, we should probably make sure it has no balances, so that token transfers are more costly.

src/utils/DepositUtils.ts
```diff
@@ -71,3 +73,38 @@ export async function queryHistoricalDepositForFill(
 
   return validateFillForDeposit(fill, deposit) ? deposit : undefined;
 }
+
+export function createDepositForSimulatingGas(
```
mrice32, line 77: Do we create a deposit just to pass to the fill creation function? Any reason we can't just create the fill directly and avoid creating the intermediate deposit struct?

src/utils/common.ts
```diff
@@ -290,33 +293,57 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+    if (!isRecipientAnAddress || !isRelayerAnAddress) {
+      throw new Error("Could not simulate message fill. Recipient address or relayer address is not a valid address");
+    }
+    const [isRecipientAContract, relayerBalanceOfToken] = await Promise.all([
+      isContractAddress(recipientAddress, provider),
+      getTokenBalance(relayerAddress, fillToSimulate.destinationToken, provider),
+    ]);
+    if (!isRecipientAContract) {
+      throw new Error("Could not simulate message fill. Recipient address is not a contract address");
+    }
```
mrice32, line 328: Why do we error in this case? Technically across allows messages where the recipient is not a contract.

Is the idea that this error will bubble up in the API preventing users from sending what looks like an erroneous transaction? If so, should that error be at a higher level (in the API code) rather than in the SDK where this code may be used for other things (like in the relayer codebase)?

mrice32 (commented):
Looks great! A few comments

One high level comment -- could you add a brief description to the PR just so the reviewers understand what this PR intends (and doesn't intend) to do:
```
This adds message handling functionality to the relayFeeCalculator module so gas estimations can now be done for deposits that include messages. This is/is not backwards compatible with the existing module interface...
```

Since this is a new feature, we also might want to label it as "feat" instead of "improve" since this is a user facing feature addition.

src/utils/DepositUtils.ts
james-a-morris: I mainly created this extra step so that we can re-use the gas estimation step for an arbitrary fill & so that we could construct an arbitrary deposit. I imagine this may have further usage, but we can definitely close the loop and revisit encapsulating logic in a different PR

src/utils/common.ts
james-a-morris: That's a fair point given how the message is essentially a no-op if the recipient is either not a contract. It was placed here to conform to the tech spec's version of graceful errors. I agree that there's a better place for this specific check and will make some revisions to move it 🙂 

```
relayExecution.updatedRecipient.isContract() && relayExecution.updatedMessage.length > 0
```

src/utils/common.ts
nicholaspai: although, i can see an extra flag being useful so that the caller doesn't have to call into the SDK twice and make 2x RPC calls. The alternative of checking the simulated EOA balance and then calling into the SDK results also in 2x RPC calls, but this is unavoidable, as either the caller checks the EOA balance before calling into the SDK or the SDK will do it inevitably itself.

src/relayFeeCalculator/relayFeeCalculator.ts
nicholaspai: I agree with Matt's instinct that these shouldn't be coupled

src/utils/DepositUtils.ts
nicholaspai: > we can re-use the gas estimation step for an arbitrary fill & so that we could construct an arbitrary deposi

What's the use case for this, out of curiosity? I also thought it was kind of strange to create a deposit to create a fill, rather i'd stick to constructing a fill directly

src/utils/common.ts
nicholaspai: yes so in this case this shouldn't be an error I agree. If the recipient is not a contract, then we should continue with the default behavior and simulating the fill with an empty message. In this case the message is ignored.

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/427/commits/d94450934f46b17d6fdd0a9ce687378490b4bb64

src/utils/DepositUtils.ts
james-a-morris: Sounds good - I believe we should be fine with 0.

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: This has been uncoupled via https://github.com/across-protocol/sdk-v2/pull/427/commits/d94450934f46b17d6fdd0a9ce687378490b4bb64

src/utils/DepositUtils.ts
james-a-morris: Resolved in https://github.com/across-protocol/sdk-v2/pull/427/commits/d94450934f46b17d6fdd0a9ce687378490b4bb64 as the recipient address is now required.

src/utils/DepositUtils.ts
james-a-morris: Fair point - I've made the depositorAddress default to a random address to prevent overestimating with 0xFFFF... unless overestimating would be beneficial.

https://github.com/across-protocol/sdk-v2/pull/427/commits/c08699837a7c78648734ff969f0a74fa275ef890

src/utils/DepositUtils.ts
james-a-morris: This structure is like this because the `QueryInterface` where that abstracts the various chains needs a common data-structure as input. The `ProfitClient` and  Vercel API references deposits so if possible, I will change the `getGasCosts` to reference a `Deposit` instead of a fill.

src/utils/common.ts
james-a-morris: Done: I've kicked the error up to the Vercel API where it matters to check for this case: https://github.com/across-protocol/frontend-v2/pull/862/commits/80cb76f5159902db393a33ce2fc5155571e4ba56

src/utils/DepositUtils.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/427/commits/c08699837a7c78648734ff969f0a74fa275ef890

src/utils/DepositUtils.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/427/commits/c08699837a7c78648734ff969f0a74fa275ef890

src/utils/common.ts
james-a-morris: I'd like to leave this as-is given that this is the best way to effectively estimate gas and ensure that we're not potentially making changes that affect the output of the arbitrary `handleAcrossMessage` contract

src/utils/DepositUtils.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/427/commits/b94563648e64e6a17bd4cb3f89da378cd959a1d6

Ultimately, I was originally trying to make a clean interface and the closest way I thought to do this robustly was to convert `Parameters -> Deposit`, then `Deposit -> Fill`, then simulate that fill. However, we could stop this and only require the values needed so `Parameters -> Fill`

src/utils/DepositUtils.ts
nicholaspai: hmm i guess `MaxUint32` is safer but might cause some errors in the clients. Have you tested that they can handle that large of a deposit id?

src/utils/DepositUtils.ts
nicholaspai: Overestimating can have a bad edge case where relayer refuses to fill a deposit because its too expensive, causing innocent user deposit to be ignored. I'd err on the side of the relayer losing money versus the user

src/utils/DepositUtils.ts
james-a-morris: The `depositId` in the `SpokePool.sol` is defined as a `uint32` so I figure we could max it out. Technically we could probably get away with setting a `depositId = -1` which is going to underflow to somewhere around the max int.

src/utils/common.ts
nicholaspai: So we're going to leave this as is and make caller to catch the error and readjust the `amountToRelay` they pass in? 

Just to confirm

src/utils/common.ts
```diff
@@ -290,33 +292,46 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+  // Only continue sanity checks if the message
+  if (!isMessageEmpty) {
+    const relayerBalanceOfToken = await getTokenBalance(relayerAddress, fillToSimulate.destinationToken, provider);
+    const isRecipientAnAddress = ethers.utils.isAddress(recipientAddress);
+    const isRelayerAnAddress = ethers.utils.isAddress(relayerAddress);
+    if (!isRecipientAnAddress || !isRelayerAnAddress) {
+      throw new Error("Could not simulate message fill. Recipient address or relayer address is not a valid address");
+    }
+    if (toBN(relayerBalanceOfToken).lt(toBN(amountToRelay))) {
+      throw new Error("Could not simulate message fill. Partial fills are not supported with message relaying");
```
nicholaspai, line 318:
```suggestion
      throw new Error(`Could not simulate message fill. Relayer balance (${relayerBalanceOfToken.toString()}) is insufficient to fully fill ${amountToRelay} tokens and there is no support for partial message fills`);
```

test/common.test.ts
```diff
@@ -46,7 +47,17 @@ describe("Utils test", () => {
     const provider = new providers.JsonRpcProvider(rpcUrl, 1);
     const spokePool: SpokePool = SpokePool__factory.connect(spokePoolAddress, provider);
 
-    const unsignedTxn = await createUnsignedFillRelayTransaction(spokePool, usdcAddress, relayerAddress);
+    const fill: Fill = buildFillForSimulatingFullDeposit(
```
nicholaspai, line 50: Should you add a test that this conditional `toBN(relayerBalanceOfToken).lt(toBN(amountToRelay))` throws an error?

src/utils/DepositUtils.ts
nicholaspai: i think that's a tad bit safer than `0`

src/utils/common.ts
```diff
@@ -290,33 +292,46 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
-    "1",
-    "1",
-    "1",
-    [],
+  const { message, recipient: recipientAddress, relayer: relayerAddress, amount: amountToRelay } = fillToSimulate;
+  const provider = spokePool.provider;
+
+  const isMessageEmpty = message.length === 0 || message === "0x";
+
+  // Only continue sanity checks if the message
```
mrice32, line 309: Incomplete comment

src/utils/common.ts
```diff
@@ -290,33 +292,46 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+  if (!isMessageEmpty) {
+    const relayerBalanceOfToken = await getTokenBalance(relayerAddress, fillToSimulate.destinationToken, provider);
+    const isRecipientAnAddress = ethers.utils.isAddress(recipientAddress);
+    const isRelayerAnAddress = ethers.utils.isAddress(relayerAddress);
+    if (!isRecipientAnAddress || !isRelayerAnAddress) {
+      throw new Error("Could not simulate message fill. Recipient address or relayer address is not a valid address");
+    }
```
mrice32, line 316: Sorry, maybe I'm missing something, but why do we only check these if the message isn't empty?

src/utils/common.ts
```diff
@@ -290,33 +292,46 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+    if (toBN(relayerBalanceOfToken).lt(toBN(amountToRelay))) {
```
mrice32, line 317: Same here, wouldn't the simulation also break in the empty message case (if the relayer doesn't have the funds)?

mrice32 (commented):
Looks great! Just a few comments

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -63,11 +59,31 @@ export default abstract class QueryBase implements QueryInterface {
    * Retrieves the current gas costs of performing a fillRelay contract at the referenced Spoke Pool
    * @returns The gas estimate for this function call (multplied with the optional buffer)
    */
-  async getGasCosts(): Promise<BigNumberish> {
-    const tx = await createUnsignedFillRelayTransaction(this.spokePool, this.usdcAddress, this.simulatedRelayerAddress);
+  async getGasCosts(
+    amountToRelay: BN,
```
nicholaspai, line 63: as discussed offline, the interface shoudl require the caller to pass in both `fillAmount` and `amount`. I would rename `amountToRelay` to `fillAmount` to match the smart contracts nomenclature.

src/utils/DepositUtils.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/427/commits/831cf21cb93d8f43981345517bd81815bbf4076a

src/utils/common.ts
james-a-morris: Resolved in newer commit. We now pass in a deposit object and optimistically assume that these are valid values.

src/utils/common.ts
james-a-morris: Yep! But the error would be because the relayer is out of funds - not if there was some out of funds error in the complex message execution to the recipient contract.

In either case, I've reworked the PR in a newer commit 🙂

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -63,11 +60,23 @@ export default abstract class QueryBase implements QueryInterface {
    * Retrieves the current gas costs of performing a fillRelay contract at the referenced Spoke Pool
    * @returns The gas estimate for this function call (multplied with the optional buffer)
    */
-  async getGasCosts(): Promise<BigNumberish> {
-    const tx = await createUnsignedFillRelayTransaction(this.spokePool, this.usdcAddress, this.simulatedRelayerAddress);
+  async getGasCosts(
+    deposit: Deposit,
+    amountToRelay: BigNumberish,
+    relayAddress = DEFAULT_SIMULATED_RELAYER_ADDRESS,
+    _relayerBalanceForToken?: BigNumberish
```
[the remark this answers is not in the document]
james-a-morris: For the `_relayerBalanceForToken`? Yes, this is so that the `relayer` can pass its erc20 balance to the SDK function without having an additional RPC call to resolve the balance.

nicholaspai (commented):
I think this looks good but Paul has better context on how we'd use this in the relayer so will want him to approve

src/utils/common.ts
```diff
@@ -290,33 +293,74 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
-    "1",
-    "1",
-    [],
-    MAX_BIG_INT,
-    { from: simulatedRelayerAddress }
+  // We should perform some basic validation on our deposit.
+
+  // We should ensure that the amount to fill is less than or equal to the
+  // amount that was deposited.
+  if (amountToFill.gt(deposit.amount)) {
```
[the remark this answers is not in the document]
james-a-morris: I agree - but I wanted to test for invalid inputs into the function. The calling code should reasonably not want to overfill a deposit.

In any case I can relax this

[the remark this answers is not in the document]
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/427/commits/2f6f7d11a3395c5f4e4a0d251ee64fd158bd3db9

src/utils/common.ts
```diff
@@ -290,33 +293,74 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+        "Partial fills on deposits with messages are not allowed. If the recipient is a contract, then the amount to fill must be equal to the amount deposited."
+      );
+    }
+  }
+
+  // We should check that the relayer has enough balance to facilitate this
+  // transaction before we populate it.
+  const relayerBalanceForToken =
+    _relayerBalanceForToken ??
+    (await ERC20__factory.connect(deposit.originToken, spokePool.provider).balanceOf(relayerAddress));
```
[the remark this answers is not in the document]
james-a-morris: Fair point. It's been a few days - great call 😓 https://github.com/across-protocol/sdk-v2/pull/427/commits/2f6f7d11a3395c5f4e4a0d251ee64fd158bd3db9

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -11,11 +11,19 @@ import {
...
+import { DEFAULT_SIMULATED_RELAYER_ADDRESS } from "../constants";
+import { Deposit } from "../interfaces";
 
 // This needs to be implemented for every chain and passed into RelayFeeCalculator
 export interface QueryInterface {
-  getGasCosts: () => Promise<BigNumberish>;
+  getGasCosts: (
+    deposit: Deposit,
+    amountToRelay: BigNumberish,
+    relayerAddress: string,
```
pxrl, line 24: qq: this parameter is optional in baseQuery, but mandatory here - is that intentional?

src/utils/common.ts
```diff
@@ -290,33 +294,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
- * @param destinationTokenAddress A valid ERC20 token (system-wide default is UDSC)
- * @param simulatedRelayerAddress The relayer address that relays this transaction
- * @returns A populated (but unsigned) transaction that can be signed/sent or used for estimating gas costs
+ * Create an unsigned transaction to fill a relay. This function is used to simulate the gas cost of filling a relay.
+ * @param spokePool A valid SpokePool contract instance
+ * @param fillToSimulate The fill that this function will use to populate the unsigned transaction
+ * @returns An unsigned transaction that can be used to simulate the gas cost of filling a relay
  */
-export async function createUnsignedFillRelayTransaction(
+export async function createUnsignedFillRelayTransactionFromFill(
```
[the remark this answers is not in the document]
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/427/commits/d5578101c43d8cec7ca84f4748664f44747fe9b0

src/utils/common.ts
```diff
@@ -290,33 +294,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+  const relayerBalanceForToken =
+    _relayerBalanceForToken ?? (await getTokenBalance(relayerAddress, deposit.originToken, spokePool.provider));
+  if (relayerBalanceForToken.lt(amountToFill)) {
+    throw new Error(
+      `Relayer balance for token (${relayerBalanceForToken.toString()}) is less than the amount to fill (${amountToFill.toString()})`
+    );
+  }
+
+  // If we have made it this far, then we can populate the transaction.
+  return spokePool.populateTransaction.fillRelay(
```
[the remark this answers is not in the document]
nicholaspai: correct. Is there a reason we don't handle this switch today?

src/utils/common.ts
```diff
@@ -290,33 +294,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
-    [],
-    MAX_BIG_INT,
-    { from: simulatedRelayerAddress }
+  // We should perform some basic validation on our deposit.
+
+  // We should ensure that if a message is present
+  // in either the original deposit or a sped-up deposit AND the
+  // recipient is a contract, then we should ensure that the amount
+  // being filled is equal to the amount that the recipient expects.
+  // I.E. No partial fills.
```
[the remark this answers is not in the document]
nicholaspai: yeah i thought we decided that the sdk should be unopinonated about partial fills, while the API would never call SDK functions with partial fill parameters, while the relayer might want to (sometimes). Especially when fully filling an already partial filled deposit

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: It is not optional in BaseQuery. There is a defined default so that the value is never `undefined`. We can't really define this in an interface, hence we need to write `relayerAddress: string` rather than `relayerAddress?: string`

src/utils/common.ts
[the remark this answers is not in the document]
james-a-morris: https://github.com/across-protocol/sdk-v2/pull/427/commits/09367ec72ef95f94c78e70dac0a9ae7edb5e28c4

src/utils/common.ts
```diff
@@ -290,33 +294,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+
+  // We can compute all synchronous checks first, and then perform the
+  // asynchronous check last.
+  if (isMessageEmpty(deposit.updatedMessage ?? deposit.message) && !deposit.amount.eq(amountToFill)) {
+    const isRecipientAContract = await isContractDeployedToAddress(
+      deposit.updatedRecipient ?? deposit.recipient,
+      spokePool.provider
+    );
+    if (isRecipientAContract) {
```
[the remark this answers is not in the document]
james-a-morris: This is placed in the FE in later commits. Updated in the SDK here: https://github.com/across-protocol/sdk-v2/pull/427/commits/b5915e7d5d7762af82acc8592048a8a576f8566c

src/utils/common.ts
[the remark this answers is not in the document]
james-a-morris: Removed in favor of placing fully in the FE: https://github.com/across-protocol/sdk-v2/pull/427/commits/b5915e7d5d7762af82acc8592048a8a576f8566c

src/utils/common.ts
```diff
@@ -290,33 +294,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
-    "10",
-    "1",
-    "1",
-    "1",
-    "1",
-    "1",
-    [],
-    MAX_BIG_INT,
-    { from: simulatedRelayerAddress }
+  // We should perform some basic validation on our deposit.
```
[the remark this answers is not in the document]
james-a-morris: This is a spoke pool contract - it doesn't have direct access to its chain Id unless we call asynchronously.

src/utils/common.ts
```diff
@@ -290,34 +293,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
-    "1",
-    "1",
-    [],
-    MAX_BIG_INT,
-    { from: simulatedRelayerAddress }
-  );
+  // We should check that the relayer has enough balance to facilitate this
+  // transaction before we populate it.
+  const relayerBalanceForToken =
+    _relayerBalanceForToken ?? (await getTokenBalance(relayerAddress, deposit.originToken, spokePool.provider));
```
nicholaspai, line 311: I just had a thought: do we really want to perform the check on `depositAmount` versus `_relayerBalanceForToken` in the sdk? Why can't we do this check in the relayer or the API, and then based on the result of that comparison, decide whether to call into the sdk?

src/utils/common.ts
```diff
@@ -290,34 +293,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+  if (relayerBalanceForToken.lt(amountToFill)) {
+    throw new Error(
+      `Relayer balance for token (${relayerBalanceForToken.toString()}) is less than the amount to fill (${amountToFill.toString()})`
+    );
+  }
+
+  // If we have made it this far, then we can populate the transaction.
+  if (isDefined(deposit.speedUpSignature)) {
```
nicholaspai, line 319: if speedUpSignature is defined, then the following MUST be defined. If not there is an error in the clients somewhere: newRelayerFeePct, updatedRecipient, updatedMessage . This is based on how the contract works [here]( https://github.com/across-protocol/contracts-v2/blob/master/contracts/SpokePool.sol#L558) which is the only place where the speedUpSignature is omitted

src/utils/common.ts
```diff
@@ -290,34 +293,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+    return spokePool.populateTransaction.fillRelayWithUpdatedDeposit(
+      deposit.depositor,
+      deposit.recipient,
+      deposit.updatedRecipient ?? "0x",
+      deposit.destinationToken,
+      amountToFill,
```
nicholaspai, line 325: this needs to be swapped with below param. In order, the params are `amount` and then `maxTokensToSend` which are the full deposit amount and then the actual amount to relay

src/utils/common.ts
```diff
@@ -290,34 +293,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+      deposit.amount,
+      deposit.destinationChainId, // Let's assume that the destination chain ID is the same as the repayment chain ID.
+      deposit.originChainId,
+      deposit.realizedLpFeePct ?? bnZero, // Let's assume that the realized LP fee is 0 if it is not present.
+      deposit.relayerFeePct,
+      deposit.realizedLpFeePct ?? bnZero,
```
nicholaspai, line 331: this needs to be `deposit.newRelayerFeePct` which must be defined at this point as explained here https://github.com/across-protocol/sdk-v2/pull/427/files#r1366009439

src/utils/common.ts
```diff
@@ -290,34 +293,65 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+      deposit.updatedMessage ?? "0x",
+      deposit.speedUpSignature,
+      bnUint256Max
+    );
+  } else {
+    return spokePool.populateTransaction.fillRelay(
+      deposit.depositor,
+      deposit.recipient,
+      deposit.destinationToken,
+      amountToFill,
```
nicholaspai, line 343: this param needs to be swapped with one below

nicholaspai: https://github.com/across-protocol/sdk-v2/pull/427/files#r1366010585

src/utils/common.ts
pxrl: From the relayer perspective I am OK if it's not done here, because the relayer already tracks its balance before calling in to this method. If it remains here then I'll just supply `fillAmount` to ensure that the RPC call is never triggered. The RPC balance check is actually not useful for the relayer in any case, since it's balance can change before the fill is executed (i.e. if this is filled in a multicall and other fills come before it).

src/utils/TokenUtils.ts
```diff
@@ -17,3 +19,46 @@ export const getL2TokenAddresses = (l1TokenAddress: string): { [chainId: number]
...
+
+/**
+ * Returns the contract address for a given token symbol and chainId.
+ * @param symbol A case-insensitive token symbol.
+ * @param chainId The chainId to resolve the contract address for.
+ * @returns The contract address for the given token symbol and chainId, or undefined if the token symbol is not supported.
+ */
+export const resolveContractFromSymbol = (symbol: string, chainId: string): string | undefined => {
+  return Object.values(TOKEN_SYMBOLS_MAP).find((details) => {
+    return details.symbol.toLowerCase() === symbol.toLowerCase();
```
pxrl, line 31: Since `TOKEN_SYMBOLS_MAP` always stores in upper-case, how about just promoting the input `symbol` to upper-case instead?

src/relayFeeCalculator/relayFeeCalculator.ts
pxrl: Sorry, clumsy phrasing from my side. By `optional` I meant from the caller's perspective. But I see the reason that it needs to be a required field here :+1:

src/utils/common.ts
james-a-morris: We can drop it. Ultimately the relayer will not incur any additional RPC calls since its passing the balance in - so this is more of a synchronous validation for it.

james-a-morris: https://github.com/across-protocol/sdk-v2/pull/427/commits/f62328a66fcbd8c63b7faf1f760e5aa5fec81619

src/utils/TokenUtils.ts
james-a-morris: My only pushback is that it only stores uppercase so far. I know in the FE we account for USDC.e with the lowercase e. I'm not sure we have a guarantee that `TOKEN_SYMBOLS_MAP.[x].symbol` will always be lowercase

src/utils/common.ts
james-a-morris: https://github.com/across-protocol/sdk-v2/pull/427/commits/b261d69b2d21aa31033bca978392c6eda000db16

src/utils/common.ts
james-a-morris: https://github.com/across-protocol/sdk-v2/pull/427/commits/b261d69b2d21aa31033bca978392c6eda000db16

src/utils/common.ts
james-a-morris: Yep! https://github.com/across-protocol/sdk-v2/pull/427/commits/b261d69b2d21aa31033bca978392c6eda000db16

src/utils/common.ts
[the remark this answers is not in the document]
james-a-morris: @pxrl is this good to resolve the comment? It needs to be resolved before I can merge

src/relayFeeCalculator/chain-queries/baseQuery.ts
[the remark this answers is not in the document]
james-a-morris: This has now been removed. Closing this comment

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: Ah, I see what you mean. 

We had the ability in `BaseQuery` to provide the default relayer. However, since this is an interface that exists outside of BaseQuery, there is a case where we implement this interface but don't have a default in which case the caller would need to supply it every time.

src/utils/TokenUtils.ts
pxrl: Good call on USDC.e and potentially also the Base variant (USDbC or whatever).

pxrl (approved):
LGTM

PR #428, opened 2023-10-12
feat: failover historic query if deposit is malformed

james-a-morris:
We should allow our queries to _self-heal_ if they detect a malformed deposit. This will check if a deposit returned from the cache is malformed. If it is found to be malformed, we will ignore it and re-cache a correct version.

### DIFF
--- a/src/utils/DepositUtils.ts
+++ b/src/utils/DepositUtils.ts
@@ -65,6 +65,9 @@ export async function queryHistoricalDepositForFill(
         fill,
         cachedDeposit,
       });
+      // By setting this value to undefined, we eventually have to pull
+      // the deposit from our spoke pool client. Because this new deposit
+      // is formed correctly, we will cache it below.
       cachedDeposit = undefined;
     }
   }
--- a/src/utils/DepositUtils.ts
+++ b/src/utils/DepositUtils.ts
@@ -1,11 +1,11 @@
-import assert from "assert";
 import { SpokePoolClient } from "../clients";
-import { Fill, DepositWithBlock, Deposit, CachingMechanismInterface } from "../interfaces";
+import { DEFAULT_CACHING_TTL } from "../constants";
+import { CachingMechanismInterface, Deposit, DepositWithBlock, Fill } from "../interfaces";
+import { getDepositInCache, getDepositKey, setDepositInCache } from "./CachingUtils";
 import { validateFillForDeposit } from "./FlowUtils";
 import { getCurrentTime } from "./TimeUtils";
 import { isDefined } from "./TypeGuards";
-import { getDepositInCache, getDepositKey, setDepositInCache } from "./CachingUtils";
-import { DEFAULT_CACHING_TTL } from "../constants";
+import { isDepositFormedCorrectly } from "./ValidatorUtils";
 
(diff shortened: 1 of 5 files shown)


src/utils/DepositUtils.ts
```diff
@@ -55,15 +55,21 @@ export async function queryHistoricalDepositForFill(
   let deposit: DepositWithBlock, cachedDeposit: Deposit | undefined;
   if (cache) {
     cachedDeposit = await getDepositInCache(getDepositKey(fill), cache);
+    if (isDefined(cachedDeposit) && !isDepositFormedCorrectly(cachedDeposit)) {
```
mrice32, line 61: Should we also delete it from redis when this happens?

mrice32 (commented):
Just one comment!

src/utils/DepositUtils.ts
nicholaspai: I was assuming that setting `cachedDeposit = undefined` was James' way to delete it but if not, then we should explicitly delete it

james-a-morris: By setting it to undefined here, we eventually re-set the key on line 77

src/utils/DepositUtils.ts
```diff
@@ -55,15 +55,27 @@ export async function queryHistoricalDepositForFill(
...
+      spokePoolClient.logger.warn({
+        at: "[SDK]:DepositUtils#queryHistoricalDepositForFill",
+        message: "Cached deposit was not formed correctly, removing from cache",
+        fill,
+        cachedDeposit,
+      });
+      // By setting this value to undefined, we eventually have to pull
+      // the deposit from our spoke pool client. Because this new deposit
+      // is formed correctly, we will cache it below.
+      cachedDeposit = undefined;
```
james-a-morris, line 71: @mrice32 @pxrl @nicholaspai added a comment to highlight using the existing logic

src/utils/DepositUtils.ts
mrice32: Ah good point.

mrice32 (approved):

src/utils/ValidatorUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+const AddressValidator = define<string>("AddressValidator", (v) => ethers.utils.isAddress(String(v)));
+const BigNumberValidator = define<BigNumber>("BigNumberValidator", (v) => ethers.BigNumber.isBigNumber(v));
+
+const DepositSchema = object({
+  depositId: Min(Number(), 0),
+  depositor: AddressValidator,
+  recipient: AddressValidator,
+  originToken: AddressValidator,
+  amount: BigNumberValidator,
+  originChainId: Min(Number(), 0),
```
nicholaspai, line 14: Can we make all these `Number` --> `Integer`?

nicholaspai (approved):

src/utils/ValidatorUtils.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/428/commits/fe9d4251b9ca527bcbf384f1525ed895a8d72730

james-a-morris:
Merging as we have new unit tests that cover real world data

PR #437, opened 2023-10-23
improve(constants): Allow HubPoolClient constructor to override default caching safe lag

nicholaspai:
This caching lag is used to store utilization cached values read from the HubPoolClient, and deposit structs

Signed-off-by: nicholaspai <npai.nyc@gmail.com>

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,5 +1,7 @@
 import { constants as ethersConstants, BigNumber, utils } from "ethers";
+import dotenv from "dotenv";
 export { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants-v2";
+dotenv.config();
 
 export const { AddressZero: ZERO_ADDRESS } = ethersConstants;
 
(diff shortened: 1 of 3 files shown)


src/constants.ts
```diff
@@ -48,7 +50,7 @@ export const PUBLIC_NETWORKS: { [chainId: number]: { name: string; etherscan: st
 export const DEFAULT_BLOCKCHAIN_EXPLORER_DOMAIN = "https://etherscan.io";
 
 export const DEFAULT_CACHING_TTL = 60 * 60 * 24 * 7 * 2; // 2 Weeks
-export const DEFAULT_CACHING_SAFE_LAG = 60 * 60 * 24; // 1 Day
+export const DEFAULT_CACHING_SAFE_LAG = Number(process.env.CACHING_SAFE_LAG) ?? 60 * 10; // 10 mins
```
pxrl, line 53: I'd prefer to default conservative, and opt-in via overriding. wdyt about setting it at something like 1 hour, and then we can crank it up as needed?

Also, I'm not super keen on having this read directly from the environment, but I don't have an alternate proposal and I don't think we should let that hold this up, but it would be good to handle it in another way in future.

nicholaspai: ok i'm fine with that, defaulted back to 1 hour. In the relayer-v2 repo this would be set via Config. I can make it an input var

james-a-morris, line 53:
```suggestion
export const DEFAULT_CACHING_SAFE_LAG = Number(process.env.CACHING_SAFE_LAG) ?? 60 * 10; // 10 mins
if(!Number.isNumber(DEFAULT_CACHING_SAFE_LAG) || DEFAULT_CACHING_SAFE_LAG < 0) {
    throw new Error("Invalid default caching safe lag")
}
```

james-a-morris: Thoughts on performing some sanitization so that we fail on invalid parameters

james-a-morris (commented):
One comment: https://github.com/across-protocol/sdk-v2/pull/437/files#r1369233700

pxrl (approved):
This looks OK to me, pending any comments from others.

mrice32 (approved):

james-a-morris (approved):

PR #438, opened 2023-10-23
feat: include cache in SDK block finder

james-a-morris:
This change introduces the function `getCachedBlockForTimestamp` into the SDK. All calls to blockfinder's `getBlockForTimestamp` have been diverted to either an available `hubPool.getBlock(timestamp: number)` function (which leverages `getCachedBlockForTimestamp` or with `getCachedBlockForTimestamp` directly.

The original code block is:

```
/**
 * @notice Get the block number for a given timestamp fresh from on-chain data if not found in redis cache.
 * If redis cache is not available, then requests block from blockFinder.
 * @param chainId Chain to load block finder for.
 * @param timestamp Approximate timestamp of the to requested block number.
 * @param blockFinder Caller can optionally pass in a block finder object to use instead of creating a new one
 * or loading from cache. This is useful for testing primarily.
 * @returns Block number for the requested timestamp.
 */
export async function getBlockForTimestamp(
  chainId: number,
  timestamp: number,
  blockFinder?: BlockFinder
): Promise<number> {
  blockFinder ??= await getBlockFinder(chainId);
  const redisClient = await getRedis();

  // If no redis client, then request block from blockFinder. Otherwise try to load from redis cache.
  if (redisClient === undefined) {
    return (await blockFinder.getBlockForTimestamp(timestamp)).number;
  }

  const key = `${chainId}_block_number_${timestamp}`;
  const result = await redisClient.get(key);
  if (result === null) {
    const provider = await getProvider(chainId);
    const [currentBlock, { number: blockNumber }] = await Promise.all([
      provider.getBlock("latest"),
      blockFinder.getBlockForTimestamp(timestamp),
    ]);

    // Expire key after 90 days.
    if (shouldCache(timestamp, currentBlock.timestamp)) {
      await setRedisKey(key, blockNumber.toString(), redisClient, 60 * 60 * 24 * 90);
    }
    return blockNumber;
  } else {
    return parseInt(result);
  }
}

```

### DIFF
--- a/src/utils/BlockUtils.ts
+++ b/src/utils/BlockUtils.ts
@@ -193,29 +193,29 @@ export async function getCachedBlockForTimestamp(
   if (!isDefined(cache)) {
     return resolver();
   }
+
   // Cache exists. We should first check if it's possible to retrieve the block number from cache.
-  else {
-    // Resolve the key for the block number.
-    const key = `${chainId}_block_number_${timestamp}`;
-    // See if it's even possible to retrieve the block number from cache.
-    if (shouldCache(timestamp, getCurrentTime(), DEFAULT_CACHING_SAFE_LAG)) {
-      // Attempt to retrieve the block number from cache.
-      const result = await cache.get<string>(key);
-      // If the block number is in cache, then return it.
-      if (result !== null) {
-        return parseInt(result);
-      }
-      // Otherwise, we need to resolve the block number and cache it.
-      else {
-        const blockNumber = await resolver();
-        // Expire key after 90 days.
-        await cache.set(key, blockNumber.toString(), 60 * 60 * 24 * 90); // 90 days
-        return blockNumber;
-      }
+
+  // Resolve the key for the block number.
+  const key = `${chainId}_block_number_${timestamp}`;
+  // See if it's even possible to retrieve the block number from cache.
+  if (shouldCache(timestamp, getCurrentTime(), DEFAULT_CACHING_SAFE_LAG)) {
+    // Attempt to retrieve the block number from cache.
+    const result = await cache.get<string>(key);
+    // If the block number is in cache, then return it.
+    if (result !== null) {
+      return parseInt(result);
     }
-    // It's too early to cache this key. Resolve the block number and return it.
+    // Otherwise, we need to resolve the block number and cache it.
     else {
-      return resolver();
+      const blockNumber = await resolver();
+      // Expire key after 90 days.
+      await cache.set(key, blockNumber.toString(), 60 * 60 * 24 * 90); // 90 days
+      return blockNumber;
     }
+  }
+  // It's too early to cache this key. Resolve the block number and return it.
+  else {
+    return resolver();
   }
 }
--- a/src/utils/BlockUtils.ts
+++ b/src/utils/BlockUtils.ts
@@ -186,23 +186,39 @@ export async function getCachedBlockForTimestamp(
   provider: Provider,
   cache?: CachingMechanismInterface
 ): Promise<number> {
-  // If no redis client, then request block from blockFinder. Otherwise try to load from redis cache.
+  // Resolve a convenience function to directly compute what we're
+  // looking for.
+  const resolver = async () => (await blockFinder.getBlockForTimestamp(timestamp)).number;
+
+  // If no redis client, then request block from blockFinder.
   if (!isDefined(cache)) {
-    return (await blockFinder.getBlockForTimestamp(timestamp)).number;
+    return resolver();
   }
-  const key = `${chainId}_block_number_${timestamp}`;
-  const result = await cache.get<string>(key);
-  if (result === null) {
-    const [currentBlock, { number: blockNumber }] = await Promise.all([
-      provider.getBlock("latest"),
-      blockFinder.getBlockForTimestamp(timestamp),
-    ]);
-    // Expire key after 90 days.
-    if (shouldCache(timestamp, currentBlock.timestamp, DEFAULT_CACHING_SAFE_LAG)) {
-      await cache.set(key, blockNumber.toString(), 60 * 60 * 24 * 90); // 90 days
+  // Cache exists. We should first check if it's possible to retrieve the block number from cache.
+  else {
+    // Grab the latest block number
+    const latestBlockNumber = await provider.getBlock("latest");
+    // Resolve the key for the block number.
+    const key = `${chainId}_block_number_${timestamp}`;
+    // See if it's even possible to retrieve the block number from cache.
+    if (shouldCache(timestamp, latestBlockNumber.timestamp, DEFAULT_CACHING_SAFE_LAG)) {
+      // Attempt to retrieve the block number from cache.
+      const result = await cache.get<string>(key);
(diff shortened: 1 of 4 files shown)


src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,40 @@ export class BlockFinder {
...
+ * @param chainId Chain to load block finder for.
+ * @param timestamp Approximate timestamp of the to requested block number.
+ * @param blockFinder Caller can optionally pass in a block finder object to use instead of creating a new one
+ * or loading from cache. This is useful for testing primarily.
+ * @returns Block number for the requested timestamp.
+ */
+export async function getCachedBlockForTimestamp(
+  chainId: number,
+  timestamp: number,
+  blockFinder: BlockFinder,
```
james-a-morris, line 185: Made this a required field

src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,40 @@ export class BlockFinder {
...
+  provider: Provider,
+  cache?: CachingMechanismInterface
```
james-a-morris, line 187: Added this as an optional item.

src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,40 @@ export class BlockFinder {
...
+): Promise<number> {
+  // If no redis client, then request block from blockFinder. Otherwise try to load from redis cache.
+  if (!isDefined(cache)) {
```
james-a-morris, line 190: Changed from `redisClient === undefined`

src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,40 @@ export class BlockFinder {
...
+    return (await blockFinder.getBlockForTimestamp(timestamp)).number;
+  }
+  const key = `${chainId}_block_number_${timestamp}`;
+  const result = await cache.get<string>(key);
+  if (result === null) {
+    const [currentBlock, { number: blockNumber }] = await Promise.all([
+      provider.getBlock("latest"),
```
james-a-morris, line 197: We needed need to call `getProvider()` since we pass it as input now.

src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,40 @@ export class BlockFinder {
...
+ * @param timestamp Approximate timestamp of the to requested block number.
+ * @param blockFinder Caller can optionally pass in a block finder object to use instead of creating a new one
+ * or loading from cache. This is useful for testing primarily.
+ * @returns Block number for the requested timestamp.
+ */
+export async function getCachedBlockForTimestamp(
+  chainId: number,
+  timestamp: number,
+  blockFinder: BlockFinder,
+  provider: Provider,
```
james-a-morris, line 186: Passing a provider now.

nicholaspai (commented):
This will replace the `getBlockForTimestamp` arrow function in `relayer-v2` right?

src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,40 @@ export class BlockFinder {
...
+  cache?: CachingMechanismInterface
+): Promise<number> {
+  // If no redis client, then request block from blockFinder. Otherwise try to load from redis cache.
+  if (!isDefined(cache)) {
+    return (await blockFinder.getBlockForTimestamp(timestamp)).number;
+  }
+  const key = `${chainId}_block_number_${timestamp}`;
```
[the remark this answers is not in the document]
nicholaspai: yeah so we're going to have to break up the Promise.all into two calls:
- first get current block.
- Check `shouldCache`.
- If shouldCache, try to load from cache, etc.
- otherwise, load fresh

james-a-morris:
> getBlockForTimestamp

The goal is to replace this one: https://github.com/across-protocol/relayer-v2/blob/cb7e4e62ab060187a648e69f7a5cbc4752a0775c/src/utils/BlockUtils.ts#L28

src/utils/BlockUtils.ts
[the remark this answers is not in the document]
mrice32: Why do we need current block? Can we just compute the local (js) timestamp? Wouldn't that be good enough?

src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,40 @@ export class BlockFinder {
...
+  const result = await cache.get<string>(key);
+  if (result === null) {
+    const [currentBlock, { number: blockNumber }] = await Promise.all([
+      provider.getBlock("latest"),
+      blockFinder.getBlockForTimestamp(timestamp),
+    ]);
+    // Expire key after 90 days.
+    if (shouldCache(timestamp, currentBlock.timestamp, DEFAULT_CACHING_SAFE_LAG)) {
```
mrice32, line 201: Once again, I think it would be ideal to avoid extra get latest block calls if possible -- any reason we can't just use the current local machine's time?

src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,56 @@ export class BlockFinder {
...
+  const resolver = async () => (await blockFinder.getBlockForTimestamp(timestamp)).number;
+
+  // If no redis client, then request block from blockFinder.
+  if (!isDefined(cache)) {
+    return resolver();
+  }
+  // Cache exists. We should first check if it's possible to retrieve the block number from cache.
+  else {
+    // Grab the latest block number
+    const latestBlockNumber = await provider.getBlock("latest");
```
nicholaspai, line 200: can  we avoid this? Can we just use getTime() ?

james-a-morris: https://github.com/across-protocol/sdk-v2/pull/438/commits/2015da59e59943f7babf309bc60ceab5601fe315

src/utils/BlockUtils.ts
james-a-morris: Resolved https://github.com/across-protocol/sdk-v2/pull/438/commits/2015da59e59943f7babf309bc60ceab5601fe315

src/utils/BlockUtils.ts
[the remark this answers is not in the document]
james-a-morris: https://github.com/across-protocol/sdk-v2/pull/438/commits/2015da59e59943f7babf309bc60ceab5601fe315

nicholaspai (approved):

src/utils/BlockUtils.ts
```diff
@@ -166,3 +169,53 @@ export class BlockFinder {
...
+  // Resolve a convenience function to directly compute what we're
+  // looking for.
+  const resolver = async () => (await blockFinder.getBlockForTimestamp(timestamp)).number;
+
+  // If no redis client, then request block from blockFinder.
+  if (!isDefined(cache)) {
+    return resolver();
+  }
+  // Cache exists. We should first check if it's possible to retrieve the block number from cache.
+  else {
```
pxrl, line 197: This else clause seems redundant - OK to drop it to reduce the nesting here?

james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/438/commits/147a409aef4f77eb6d5a5695550ccd93206d39fe

nicholaspai (approved):

PR #439, opened 2023-10-23
improve(BlockUtils): Better robustness on average block time resolution

pxrl:
The underlying Provider instance is actually backed by multiple RPC provider backends, and they can run be out of sync at various times. This change will catch and report those instances, and will also help to avoid them by subtracting an offset to the latest block number to give a greater chance of getting coherent responses from all RPC providers.

### DIFF
--- a/src/utils/BlockUtils.ts
+++ b/src/utils/BlockUtils.ts
@@ -49,6 +49,9 @@ export async function averageBlockTime(
     return { average: cache.average, blockRange: cache.blockRange };
   }
 
+  // If the caller was not specific about latestBlockNumber, resolve it via the
+  // RPC provider. Subtract an offset to account for various RPC provider sync
+  // issues that might occur when querying the latest block.
   if (!isDefined(latestBlockNumber)) {
     latestBlockNumber = await provider.getBlockNumber();
     latestBlockNumber -= latestBlockOffset ?? defaultLatestBlockOffset;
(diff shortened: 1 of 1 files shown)


src/utils/BlockUtils.ts
```diff
@@ -40,13 +49,25 @@ export async function averageBlockTime(
     return { average: cache.average, blockRange: cache.blockRange };
   }
 
-  latestBlockNumber ??= await provider.getBlockNumber();
+  if (!isDefined(latestBlockNumber)) {
```
nicholaspai, line 52: what's the reasoning behind this change inside this if {} branch?

src/utils/BlockUtils.ts
```diff
@@ -40,13 +49,25 @@ export async function averageBlockTime(
...
+    latestBlockNumber = await provider.getBlockNumber();
```
nicholaspai, line 53: Can you add a comment here that you are fetching the latest block from the primary provider, and then trying to reach quorum on it later on? This is why we add an offset to be safe in case the primary is lagging

nicholaspai (approved):
This LGTM but I think we want comments to help debug in future

src/utils/BlockUtils.ts
pxrl: The block offset is only subtracted from `latestBlockNumber` if `latestBlockNumber` was resolved via getBlockNumber(). If the caller supplies `latestBlockNumber` then I thought it was OK to avoid applying the offset, since they're apparently explicit about wanting to know the timing between two specific blocks.

I described it briefly here, but I will update with additional context.
```
// Default offset to the latest block number. This is subtracted from the block number
// of the latest block when it is queried from the network, rather than having been
// specified by the caller.
```

nicholaspai (approved):

src/utils/BlockUtils.ts
pxrl: Done :+1:

PR #442, opened 2023-10-24
improve(spokePoolClient): If UBA is not active, don't query RPC for block.timestamps

nicholaspai:
This reduces eth_getBlockByNumber calls pre-UBA. 

Signed-off-by: nicholaspai <npai.nyc@gmail.com>

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -581,13 +581,12 @@ export class SpokePoolClient extends BaseAbstractClient {
     events.forEach((events: Event[]) => sortEventsAscendingInPlace(events));
 
     // Load block timestamps if the UBA is active so that the UBAClient can order events using blockTimestamp.
-    // Otherwise skip this RPC call.
+    // Otherwise skip these extra RPC calls.
     let blocks: { [blockNumber: number]: Block } | undefined = undefined;
-    let blockNumbers: number[] = [];
     const isUBAActivated = isDefined(this.hubPoolClient?.configStoreClient.getUBAActivationBlock());
     if (isUBAActivated) {
       // Collate the relevant set of block numbers and filter for uniqueness, then query each corresponding block.
-      blockNumbers = Array.from(
+      const blockNumbers = Array.from(
         new Set(
           ["FundsDeposited", "FilledRelay", "RefundRequested"]
             .filter((eventName) => eventsToQuery.includes(eventName))
@@ -599,14 +598,12 @@ export class SpokePoolClient extends BaseAbstractClient {
             .flat()
         )
       );
-      blocks = isUBAActivated
-        ? Object.fromEntries(
-            await mapAsync(blockNumbers, async (blockNumber) => {
-              const block = await this.spokePool.provider.getBlock(blockNumber);
-              return [blockNumber, block];
-            })
-          )
-        : undefined;
+      blocks = Object.fromEntries(
+        await mapAsync(blockNumbers, async (blockNumber) => {
+          const block = await this.spokePool.provider.getBlock(blockNumber);
+          return [blockNumber, block];
+        })
+      );
     }
 
     return {
@@ -701,6 +698,8 @@ export class SpokePoolClient extends BaseAbstractClient {
           quoteBlockNumber: dataForQuoteTime[index].quoteBlock,
           blockTimestamp: 0,
         };
+        // Override the default blockTimestamp of 0 only if the UBA is active and we have pre-queried block times
+        // for each event.
         if (isDefined(blocks)) {
           deposit.blockTimestamp = blocks[event.blockNumber]?.timestamp ?? (await event.getBlock()).timestamp;
         }
@@ -748,6 +747,8 @@ export class SpokePoolClient extends BaseAbstractClient {
           ...rawFill,
           blockTimestamp: 0,
         };
+        // Override the default blockTimestamp of 0 only if the UBA is active and we have pre-queried block times
+        // for each event.
         if (isDefined(blocks)) {
           fill.blockTimestamp = blocks[event.blockNumber].timestamp;
         }
@@ -774,6 +775,8 @@ export class SpokePoolClient extends BaseAbstractClient {
           repaymentChainId: this.chainId, // repaymentChainId is not part of the on-chain event, so add it here.
           blockTimestamp: 0,
         };
+        // Override the default blockTimestamp of 0 only if the UBA is active and we have pre-queried block times
+        // for each event.
         if (isDefined(blocks)) {
           refundRequest.blockTimestamp = blocks[event.blockNumber].timestamp;
         }
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -583,31 +583,34 @@ export class SpokePoolClient extends BaseAbstractClient {
     // Sort all events to ensure they are stored in a consistent order.
     events.forEach((events: Event[]) => sortEventsAscendingInPlace(events));
 
-    // Collate the relevant set of block numbers and filter for uniqueness, then query each corresponding block.
-    const blockNumbers = Array.from(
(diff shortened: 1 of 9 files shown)


mrice32 (approved):
LGTM!

mrice32:
We should be really careful about ever using this variable outside of the UBA context. It could make sense in a follow-up to set this to some other value than 0 just in case.

pxrl (approved):
This should be OK (subject to some tweaks for test + lint failures). I checked in sdk-v2 and relayer-v2 and can't see anywhere that we actually use this field in the pre-UBA world.

nicholaspai:
> We should be really careful about ever using this variable outside of the UBA context. It could make sense in a follow-up to set this to some other value than 0 just in case.

that's a good point, do you think we could make `blockTimestamp` an optional param in `DepositWithBlock` and set it to undefined?

nicholaspai:
@mrice32 @james-a-morris @pxrl I'm not sure if this typing is "correct" but I tried enforcing that `blockTimestamp` can be undefined for non-UBA events and required for UBA events. Is there a more typescript native way to do this?

In particular, look at the type changes in `src/interfaces/SpokePool.ts` and how we leave `blockTimestamp = undefined` in `src/clients/SpokePoolClient.ts`

src/clients/SpokePoolClient.ts
```diff
@@ -688,8 +699,11 @@ export class SpokePoolClient extends BaseAbstractClient {
           realizedLpFeePct: dataForQuoteTime[index].realizedLpFeePct,
           destinationToken: this.getDestinationTokenForDeposit(rawDeposit),
           quoteBlockNumber: dataForQuoteTime[index].quoteBlock,
-          blockTimestamp: blocks[event.blockNumber]?.timestamp ?? (await event.getBlock()).timestamp,
         };
+        if (blocks !== undefined) {
```
james-a-morris, line 703:
```suggestion
        if (isDefined(blocks)) {
```

src/clients/SpokePoolClient.ts
```diff
@@ -688,8 +699,11 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+          (deposit as UBADepositWithBlock).blockTimestamp =
+            blocks[event.blockNumber]?.timestamp ?? (await event.getBlock()).timestamp;
```
james-a-morris, line 705: QQ: we've had issues in the past of `getBlock()` failing due to an undefined/null block. Should we be doing any form of type checking this far downstream or should we fully rely on the upstream error checks?

src/clients/SpokePoolClient.ts
```diff
@@ -732,8 +746,10 @@ export class SpokePoolClient extends BaseAbstractClient {
         const rawFill = spreadEventWithBlockNumber(event) as FillWithBlock;
         const fill: FillWithBlock = {
           ...rawFill,
-          blockTimestamp: blocks[event.blockNumber].timestamp,
         };
+        if (blocks !== undefined) {
```
james-a-morris, line 750:
```suggestion
        if (isDefined(blocks)) {
```
This also checks for blocks === null which isn't necessarily required but it's a good way to round off the falsy values.

james-a-morris (commented):
Left some nits and a comment.

src/clients/SpokePoolClient.ts
nicholaspai: I didn't want to change the logic at this point, what do you suggest?

pxrl: `null` blocks should now be prevented by this change at the provider layer: https://github.com/across-protocol/relayer-v2/commit/97fe05b883b062c3edd816f6b38b00ee06fe5ed2

src/clients/SpokePoolClient.ts
```diff
@@ -591,12 +596,18 @@ export class SpokePoolClient extends BaseAbstractClient {
           .flat()
```
pxrl, line 596: We only need the above `blockNumbers` array in the UBA world, so it might be nice to relocate into the `isUBAActivated` if statement. Then we'll avoid parsing over the set of events and filtering their block numbers for uniqueness as well.

src/interfaces/SpokePool.ts
```diff
@@ -25,8 +25,12 @@ export interface Deposit {
...
 
 export interface DepositWithBlock extends Deposit, SortableEvent {
-  blockTimestamp: number;
   quoteBlockNumber: number;
+  blockTimestamp?: number;
+}
+
+export interface UBADepositWithBlock extends DepositWithBlock {
+  blockTimestamp: number;
 }
```
pxrl, line 34: Rather than adding another type, is it worth just setting `blockTimestamp` in `DepositWithBlock` to 0 for pre-UBA scenarios? It's a bit clunky, but it might be nice to avoid all the conditional typecasting.

src/clients/SpokePoolClient.ts
nicholaspai: good point done

src/interfaces/SpokePool.ts
nicholaspai: yeah i originally wanted to avoid creating problems where a blockTimestamp is set to 0 but is understood to be something else by another client, but i think the overall changes we have to make to support that aren't worth it

src/clients/SpokePoolClient.ts
```diff
@@ -578,25 +580,34 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-      })
-    );
+    // Load block timestamps if the UBA is active so that the UBAClient can order events using blockTimestamp.
+    // Otherwise skip this RPC call.
+    let blocks: { [blockNumber: number]: Block } | undefined = undefined;
+    let blockNumbers: number[] = [];
+    const isUBAActivated = isDefined(this.hubPoolClient?.configStoreClient.getUBAActivationBlock());
+    if (isUBAActivated) {
+      // Collate the relevant set of block numbers and filter for uniqueness, then query each corresponding block.
+      blockNumbers = Array.from(
```
pxrl, line 590:
```suggestion
    // Load block timestamps if the UBA is active so that the UBAClient can order events using blockTimestamp.
    // Otherwise skip this RPC call.
    let blocks: { [blockNumber: number]: Block } | undefined = undefined;
    const isUBAActivated = isDefined(this.hubPoolClient?.configStoreClient.getUBAActivationBlock());
    if (isUBAActivated) {
      // Collate the relevant set of block numbers and filter for uniqueness, then query each corresponding block.
      const blockNumbers = Array.from(
```

pxrl (approved):
LGTM - this is pretty simple.

src/clients/SpokePoolClient.ts
```diff
@@ -578,25 +580,34 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+            .filter((eventName) => eventsToQuery.includes(eventName))
+            .map((eventName) => {
+              const idx = eventsToQuery.indexOf(eventName);
+              // tsc needs type hints on this map...
+              return (events[idx] as Event[]).map(({ blockNumber }) => blockNumber);
+            })
+            .flat()
+        )
+      );
+      blocks = isUBAActivated
```
james-a-morris, line 602: Is this needed? We are already inside the conditional that `isUBAActivated` is true.

src/clients/SpokePoolClient.ts
nicholaspai: ah yes

src/clients/SpokePoolClient.ts
nicholaspai: oh shoot sorry i need to push my latest commit

src/clients/SpokePoolClient.ts
nicholaspai: Sorry I forgot to push a commit I had ready can you re review latest? its a simplifying refactor

james-a-morris (approved):

PR #445, opened 2023-10-27
improve: allow quotes for non-messages if relayer balance is too low

james-a-morris:
We received a ticket that a large token holder couldn't receive a quote due to the relayer not having enough balance. This is the case when we need to partially fill the deposit & the simulated gas cost does not take this into account today. 

To circumvent this issue, we can pass a balance amount to be used for simulating the relay. This can be set optionally to be a full fill (in the event that the caller knows there is a message). By default, this is set to 1wei to simulate a zeroFill.


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.17.8",
+  "version": "0.17.9",
```
pxrl, line 4:
```suggestion
  "version": "0.17.10",
```

pxrl (approved):

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -197,6 +198,7 @@ export class RelayFeeCalculator {
   async gasFeePercent(
     deposit: Deposit,
     amountToRelay: BigNumberish,
+    amountToSimulate: BigNumberish,
```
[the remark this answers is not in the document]
james-a-morris: Updated to make this a boolean that defaults to true. I agree that we may want to add an override type to these functions, but it may be better suited in a separate PR.

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -284,6 +286,7 @@ export class RelayFeeCalculator {
   async relayerFeeDetails(
     deposit: Deposit,
     amountToRelay?: BigNumberish,
+    amountToSimulate = bnOne as BigNumberish,
```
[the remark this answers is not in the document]
james-a-morris: Removed in lieu of a simple approcah

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -207,10 +228,12 @@ export class RelayFeeCalculator {
...
     if (toBN(amountToRelay).eq(0)) return MAX_BIG_INT;
 
-    const getGasCosts = this.queries.getGasCosts(deposit, amountToRelay, relayerAddress).catch((error) => {
-      this.logger.error({ at: "sdk-v2/gasFeePercent", message: "Error while fetching gas costs", error });
-      throw error;
-    });
+    const getGasCosts = this.queries
+      .getGasCosts(deposit, simulateZeroFill ? bnOne : amountToRelay, relayerAddress)
+      .catch((error) => {
+        this.logger.error({ at: "sdk-v2/gasFeePercent", message: "Error while fetching gas costs", error });
```
pxrl, line 234: While here, should we include in this log message?
 - deposit information
 - simulateZeroFill flag

pxrl (approved):
This looks pretty clean :+1:

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: Done: https://github.com/across-protocol/sdk-v2/pull/445/commits/901d590328278ef5e1fb2e7c4aa03e4614ea0823

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -207,12 +229,26 @@ export class RelayFeeCalculator {
...
+        throw error;
+      });
     const getTokenPrice = this.queries.getTokenPrice(tokenInformation.symbol).catch((error) => {
-      this.logger.error({ at: "sdk-v2/gasFeePercent", message: "Error while fetching token price", error });
+      this.logger.error({
+        at: "sdk-v2/gasFeePercent",
+        message: "Error while fetching token price",
+        error,
+        simulateZeroFill,
+        deposit: JSON.stringify(deposit, jsonReplacerWithBigNumbers),
```
pxrl, line 250: `simulateZeroFill` and the complete `deposit` object aren't really necessary here - how about just logging the `destinationToken` ?

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -194,9 +195,29 @@ export class RelayFeeCalculator {
...
+   *       on a zero fill. However, the percentage will be returned as a percentage of the
+   *       amount to relay. This is useful for determining the maximum gas fee % that a
+   *       relayer may need to make on a regular fill. You will get differing results if
+   *       a message & recipient contract is provided as this function may not simulate with
+   *       the correct parameters to see a full fill.
+   */
   async gasFeePercent(
     deposit: Deposit,
     amountToRelay: BigNumberish,
+    simulateZeroFill = true,
```
nicholaspai, line 220: As mentioned offline, we should rename this to indicate its a 1 wei fill and also simulate a full fill instead of a partial

pxrl: fwiw, just for alignment, "zero fill" is the term used in the relayer to indicate a 1 wei fill (counter-intuitive). As I understand it, zero-fill also means 1 wei in this context.

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: Done

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: @nicholaspai this is using the `relayer-v2` terminology of a `zeroFill` being `1wei`. Setting `simulateZeroFill = false` will attempt to `fillRelay( )` with `amountToRelay` as the `MaxTokensToSend`.

james-a-morris: Should we add an additional flag to override this so that `amountToRelay` is ignored for the simulation purpose?

nicholaspai: oki'm on board with naming it `zeroFill` and no I don't think another flag is the correct way to do this. 

Why not do:

- if `simulateZeroFill` is `true`:
- pass in `{ ...deposit, amount: 1 wei }` when simulating the gas costs to simulate the full fill and then you can still derive `gasFeePercent` from `gasCost / amountToRelay`?

james-a-morris: I think I'd be okay with this. 

As an alternative, per @mrice32 we can usually assume that if a message is non-empty, then we should attempt to make a full fill, but if the message is not empty, we can default back to the original usage and attempt to fill with one unit. 

In the above case, we could also fix deposit.amount = 1. 

It's a shame that we moved the check against partial fills outside of the SDK because that would be a good decider of whether or not we should reject if the relayer balance is not enough to full fill a message.

TLDR: THoughts on removing `simulateZeroFill` in lieu of checking if `isMessageEmpty(deposit)` as the conditional for whether or not we attempt to simulate with a zero fill?

mrice32: I like having an explicit flag, personally. I don't love the idea of inferring based on other variables. We should be consistent on deciding this outside of this function or inside, and we've decided that outside is the way to go.

Secondly, 1 wei fills will probably not give an accurate estimation. They can round to 0, causing the math to change and no tokens to be transferred. I would suggest 100 units (small no matter which token) or similar. There _may not be_ rounding if the fees are all 0 with 1 wei, but I'm not 100% sure (we could test this). Thoughts on either testing this or making it 100 units instead of 1?

mrice32 (commented):
Looks good, just one comment!

james-a-morris:
@nicholaspai @mrice32 I've updated this PR so that our zero fills are making full fills rather than partial. This differs from the terminology used in `relayer-v2` in that `zeroFills` can initiate a slow fill by relaying `1wei`.

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -207,12 +227,33 @@ export class RelayFeeCalculator {
 
     if (toBN(amountToRelay).eq(0)) return MAX_BIG_INT;
 
-    const getGasCosts = this.queries.getGasCosts(deposit, amountToRelay, relayerAddress).catch((error) => {
-      this.logger.error({ at: "sdk-v2/gasFeePercent", message: "Error while fetching gas costs", error });
-      throw error;
-    });
+    const getGasCosts = this.queries
+      .getGasCosts(
```
mrice32, line 231: Sorry, one last question: did you change this simulation to using nonzero fees (1 wei should be fine), just to match the previous behavior?

mrice32 (commented):
Just one more comment!

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: It matches the deposit, but the API does pass in a non-zero fee. It's visible in the PR on the FE.

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -194,9 +194,29 @@ export class RelayFeeCalculator {
...
+   *       on a zero fill. However, the percentage will be returned as a percentage of the
+   *       amount to relay. This is useful for determining the maximum gas fee % that a
+   *       relayer may need to make on a regular fill. You will get differing results if
+   *       a message & recipient contract is provided as this function may not simulate with
+   *       the correct parameters to see a full fill.
+   */
   async gasFeePercent(
     deposit: Deposit,
     amountToRelay: BigNumberish,
+    simulateZeroFill = true,
```
nicholaspai, line 219: why make this default `true`? I would think this should be default `false`

nicholaspai (commented):
One question

src/relayFeeCalculator/relayFeeCalculator.ts
mrice32: Gotcha, okay, cool.

src/relayFeeCalculator/relayFeeCalculator.ts
mrice32: Good point -- agreed.

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: Resolved offline with @mrice32

src/relayFeeCalculator/relayFeeCalculator.ts
james-a-morris: My thought is that our default case so far is that we're dealing with non-message fields. However, I am of two minds.

I've updated the PR to make this default to false https://github.com/across-protocol/sdk-v2/pull/445/commits/801cfaac30169e380f6aa5b65191a23cbe423bb4

mrice32 (approved):
LGTM!

src/relayFeeCalculator/relayFeeCalculator.ts
mrice32: That may be the common case for the API, but other users, like the relayer, would probably want to do this estimate on the un-modified deposit, right? I think the setting that is what an uninformed caller might expect is probably the better default.

PR #448, opened 2023-10-29
refactor: Use isDepositSpedUp in unsigned fill generation

pxrl:


pxrl:
Dropping this one - there are some annoying type implications of this change.

PR #452, opened 2023-11-01
chore: Don't force lint, build & test on commit

pxrl:
Enforcing this on commit is a productivity killer and I have to keep typing --no-verify when committing.

If we want additional checks before code is submitted to GitHub, I'd probably be keen to explore a pre-push hook instead. But I am generally happier if CI is instrumented to catch everything so that we don't waste local developer time waiting for standard checks to occur.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,6 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-yarn lint-staged
-yarn build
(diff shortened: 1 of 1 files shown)


james-a-morris (approved):
I agree with this change. Honestly, if we could get a similar circle CI configuration like in the relayer this could be done entirely within GH

nicholaspai (approved):

mrice32 (approved):

PR #453, opened 2023-11-01
improve(relayFeeCalculator): Return cost estimates in gas units

pxrl:
relayFeeCalculator.getGasCosts() only returns the gas cost in the native gas token. For the purpose of setting its transaction gasLimit, the relayer would also like to know the underlying cost in units of gas.


src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -63,22 +63,21 @@ export default abstract class QueryBase implements QueryInterface {
     deposit: Deposit,
     amountToRelay: BigNumberish,
     relayAddress = DEFAULT_SIMULATED_RELAYER_ADDRESS
-  ): Promise<BigNumberish> {
+  ): Promise<{ gasCost: BigNumberish; gasTokenCost: BigNumberish }> {
```
pxrl, line 66: fwiw I'm not in love with the naming `gasCost` and `gasTokenCost`. I'm wondering whether `gasCostNative` and `gasCostToken` might be a better division? We do already have a concept of the "native" gas cost in the relayer, but that can be adjusted if it makes the e2e naming more cohesive.

james-a-morris: I think I'd prefer it to be something like `nativeGasCost` and `tokenGasCost`. There's a few ways to work this, but I think it would be a cleaner naming scheme not to leave one value as `gasCost`

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -63,22 +63,21 @@ export default abstract class QueryBase implements QueryInterface {
...
     const relayerToSimulate = relayAddress ?? this.simulatedRelayerAddress;
     const tx = await createUnsignedFillRelayTransactionFromDeposit(
       this.spokePool,
       deposit,
       toBN(amountToRelay),
       relayerToSimulate
     );
-    const estimatedGas = await estimateTotalGasRequiredByUnsignedTransaction(
+    return await estimateTotalGasRequiredByUnsignedTransaction(
```
james-a-morris, line 74: Any reason for this change other than style? I agree with it, and we could go one step further and remove the `await`
```suggestion
    return estimateTotalGasRequiredByUnsignedTransaction(
```

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -18,7 +18,11 @@ import { Deposit } from "../interfaces";
 
 // This needs to be implemented for every chain and passed into RelayFeeCalculator
 export interface QueryInterface {
-  getGasCosts: (deposit: Deposit, amountToRelay: BigNumberish, relayerAddress: string) => Promise<BigNumberish>;
+  getGasCosts: (
+    deposit: Deposit,
+    amountToRelay: BigNumberish,
+    relayerAddress: string
+  ) => Promise<{ gasCost: BigNumberish; gasTokenCost: BigNumberish }>;
```
james-a-morris, line 25: Since we use this in two+ places, we should place this in a type

james-a-morris (commented):
Left a few comments

src/utils/common.ts
```diff
@@ -268,28 +268,32 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+  // Scale the results by the computed multiplier.
+  gasCost = gasCost.mul(gasTotalMultiplier).div(fixedPointAdjustment);
+  gasTokenCost = gasTokenCost.mul(gasTotalMultiplier).div(fixedPointAdjustment);
 
-  // Find the total gas cost by taking the product of the gas price & the
-  // estimated number of gas units needed.
-  return BigNumber.from(gasPrice).mul(gasTotalMultiplier).mul(estimatedGasUnits).div(toBNWei(1)).toString();
+  return {
+    gasCost, // Estimate in units of gas
+    gasTokenCost, // Estimate in units of native gas token
```
nicholaspai, line 295: So what's the difference between `gasCost` and `gasTokenCost`? `gasTokenCost` = `gasCost` * `gasUnitPriceUsd`?

nicholaspai (commented):
Some question

src/relayFeeCalculator/relayFeeCalculator.ts
pxrl: Applied: https://github.com/across-protocol/sdk-v2/pull/453/commits/4d6c8ad2cea64444f9d4642003fa7bf395550723

src/relayFeeCalculator/chain-queries/baseQuery.ts
pxrl: I settled on nativeGasCost and tokenGasCost in the end :+1:

src/utils/common.ts
```diff
@@ -244,22 +244,27 @@ export function retry<T>(call: () => Promise<T>, times: number, delayS: number):
   return promiseChain;
 }
 
+export type TransactionCostEstimate = {
+  nativeGasCost: BigNumber; // gas
+  tokenGasCost: BigNumber; // wei/gas
```
nicholaspai, line 249: How can this be in Wei or gas? Or am I misunderstanding the slash? Can you make this comment a bit more clear

src/utils/common.ts
```diff
@@ -268,28 +273,32 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+  // Scale the results by the computed multiplier.
+  nativeGasCost = nativeGasCost.mul(gasTotalMultiplier).div(fixedPointAdjustment);
+  tokenGasCost = tokenGasCost.mul(gasTotalMultiplier).div(fixedPointAdjustment);
 
-  // Find the total gas cost by taking the product of the gas price & the
-  // estimated number of gas units needed.
-  return BigNumber.from(gasPrice).mul(gasTotalMultiplier).mul(estimatedGasUnits).div(toBNWei(1)).toString();
+  return {
+    nativeGasCost, // Units: gas
+    tokenGasCost, // Units: wei (nativeGasCost * wei/gas)
```
nicholaspai, line 300: This comment is more clear can you copy to struct above?

nicholaspai (approved):
https://github.com/across-protocol/sdk-v2/pull/453#discussion_r1380864432

Much prefer this naming just one comment on a Comment

src/utils/common.ts
pxrl: Bad comment - good catch. Fixed: https://github.com/across-protocol/sdk-v2/pull/453/commits/e7398df1162e0243b065dfe98d2e1a297fee88ca

PR #454, opened 2023-11-02
improve(HubPoolClient): Process LP fees in batch

pxrl:
This change removes a source of significant inefficiency in the HubPoolClient. The pre-existing implementation processes each Deposit's realizedLpFeePct independently, but in parallel. In cases where there are n deposits with the same quoteTimestamp, this results in n concurrent searches for the underlying block number. Additionally, when there are n deposits with matching originToken and quoteTimestamp (as can easily happen with the popular tokens like (W)ETH and USDC) this results in similarly duplicate queries for the "opening" HubPool utilization at the resolved block number.

This commit updates the implementation as follows:
 - Pass the set of Deposits to be computed by the HubPoolClient as an array, rather than individually.
 - Filter quoteTimestamps for uniqueness before resolving the corresponding HubPool block number.
 - Map HubPool tokens to quoteTimestamps and eventually block numbers in batch.
 - Query "opening" HubPool utilization for each HubPool token at the relevant block once, and reference the result later for each unique deposit.

Debug logging added indicates a significant drop in the number of HubPool block numbers that need to be resolved; this is especially the case on popular deposit chains like zkSync and Arbitrum.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -25,6 +25,8 @@ import * as lpFeeCalculator from "../lpFeeCalculator";
 import {
   BigNumberish,
   BlockFinder,
+  bnZero,
+  dedupArray,
   EventSearchConfig,
   MakeOptional,
   assign,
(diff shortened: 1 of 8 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -345,6 +347,188 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    // Map SpokePool tokens to HubPool tokens, and each HubPool token to a quoteTimestamp.
+    // This will later be reconciled to a HubPool blockNumber.
+    const hubPoolTokens: { [originToken: string]: string } = {};
+    const utilizationTimestamps: { [hubPoolToken: string]: number[] } = {};
+    const { originChainId: firstOriginChainId } = deposits[0];
+    deposits.forEach((deposit) => {
+      const { originChainId, originToken, quoteTimestamp } = deposit;
+      assert(
+        originChainId === firstOriginChainId,
+        "Cannot compute bulk realizedLpFeePct for disparate origin chains" +
```
nicholaspai, line 424:
```suggestion
        "Cannot compute bulk realizedLpFeePct for different origin chains" +
```

Is "different" more precise of a word than "disparate"?

src/clients/SpokePoolClient.ts
```diff
@@ -675,9 +676,13 @@ export class SpokePoolClient extends BaseAbstractClient {
       });
       this.earlyDeposits = earlyDeposits;
 
-      const dataForQuoteTime: { realizedLpFeePct: BigNumber | undefined; quoteBlock: number }[] = await Promise.all(
-        depositEvents.map((event) => this.computeRealizedLpFeePct(event))
```
nicholaspai, line 679: Can we remove `computeRealizedLpFeePct` function now?

nicholaspai (commented):
How often does this happen?
> cases where there are n deposits with the same quoteTimestamp

Assuming this is a non trivial amount of deposits, I can definitely see the benefits of this PR.

src/clients/SpokePoolClient.ts
pxrl: Yeah, that would be the intention (or convert it to a wrapper for `batchCompute...()` for a computing a single realizedLpFeePct, as we do when searching for historical deposits).

I've left it in there for the moment because it allowed me to quickly A/B test in my local branch.

nicholaspai (commented):
This algorithm makes sense but the `batchComputeRealizedLpFeePct` function is a bit hard to read. Can you break it down into smaller sub functions perhaps?

pxrl:
> This algorithm makes sense but the `batchComputeRealizedLpFeePct` function is a bit hard to read. Can you break it down into smaller sub functions perhaps?

Yeah I'm also not happy with the length of it, so splitting some parts out would make it easier to read :+1:. I'll implement that.

src/clients/HubPoolClient.ts
```diff
@@ -345,6 +347,188 @@ export class HubPoolClient extends BaseAbstractClient {
...
+   * @param blocknumber Block number to query utilization at.
+   * @param amount Amount to query. If set to 0, the closing utilization at blockNumber is returned.
+   * @param amount timestamp Associated quoteTimestamp for query, used for caching evaluation.
+   * @param timeToCache Age at which the response is able to be cached.
+   * @returns HubPool utilization at `blockNumber` after optional `amount` increase in utilization.
+   */
+  protected async _getUtilization(
+    hubPoolToken: string,
+    blockNumber: number,
+    amount: BigNumber,
```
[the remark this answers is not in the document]
pxrl: Implemented here: https://github.com/across-protocol/sdk-v2/pull/454/commits/6dcb5565e2b460c6d2807bd0f15ecb296b5d6fda

pxrl:
@nicholaspai I verified 100 bundles with this script and all checked out OK - see the attached `verification.txt` for the corresponding `REQUEST_TIME` timestamps.

Based on this, I think this is OK to go in now.

[verification.txt](https://github.com/across-protocol/sdk-v2/files/13495872/verification.txt)

PR #455, opened 2023-11-03
improve(SpokePoolClient): Optimize algorithm for appending speed up data to deposits

nicholaspai:
This PR optimizes compute time for the case where there is a speed up deposit in a SpokePoolClient update.

Currently the algorithm for adding speed up data to deposits is:
- load all speed up events into memory
- loop through all deposits already stored in memory, detect if there is a speed up event newly saved into memory for the deposit, and update the deposit

We needlessly loop through all deposits since there are expected to very few speed ups in a given lookback window relative to the number of deposits loaded. Instead, the proposed algorithm is:
- load all speed up events into memory
- for each new speed up data newly saved, detect if there is a deposit data stored in memory matching the speed up, and update the deposit
Merge state

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.17.12",
+  "version": "0.17.11",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/v/developer-docs/developers/across-sdk",
   "files": [
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.17.11",
+  "version": "0.17.12",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/v/developer-docs/developers/across-sdk",
   "files": [
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -412,7 +412,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @note This hash is used to match deposits and fills together.
    * @note This hash takes the form of: `${depositId}-${originChainId}`.
    */
-  public getDepositHash(event: Deposit | Fill | DepositWithBlock | FillWithBlock): string {
+  public getDepositHash(event: { depositId: number; originChainId: number }): string {
     return `${event.depositId}-${event.originChainId}`;
   }
 
@@ -725,11 +725,17 @@ export class SpokePoolClient extends BaseAbstractClient {
       for (const event of speedUpEvents) {
         const speedUp: SpeedUp = { ...spreadEvent(event.args), originChainId: this.chainId };
         assign(this.speedUps, [speedUp.depositor, speedUp.depositId], [speedUp]);
-      }
 
-      // Traverse all deposit events and update them with associated speedups, If they exist.
-      for (const deposits of Object.entries(this.depositHashes)) {
-        this.depositHashes[deposits[0]] = this.appendMaxSpeedUpSignatureToDeposit(deposits[1]);
+        // Find deposit hash matching this speed up event and update the deposit data associated with the hash,
+        // if the hash+data exists.
+        const depositHash = this.getDepositHash(speedUp);
+
+        // We can assume all deposits in this lookback window are loaded in-memory already so if the depositHash
+        // is not mapped to a deposit, then we can throw away the speedup as it can't be applied to anything.
+        const depositDataAssociatedWithSpeedUp = this.depositHashes[depositHash];
+        if (isDefined(depositDataAssociatedWithSpeedUp)) {
+          this.depositHashes[depositHash] = this.appendMaxSpeedUpSignatureToDeposit(depositDataAssociatedWithSpeedUp);
+        }
       }
     }
 
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -729,7 +729,7 @@ export class SpokePoolClient extends BaseAbstractClient {
         newSpeedUps.push(speedUp);
       }
 
-      // `appendMaxSpeedUpSignatureToDeposit` assu,es that all this.speedUps has all speedups stored into memory.
+      // `appendMaxSpeedUpSignatureToDeposit` assumes that all this.speedUps has all speedups stored into memory.
       for (const speedUp of newSpeedUps) {
         // Find deposit hash matching this speed up event and update the deposit data associated with the hash,
         // if the hash+data exists.
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -729,7 +729,7 @@ export class SpokePoolClient extends BaseAbstractClient {
         newSpeedUps.push(speedUp);
       }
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.17.11",
+  "version": "0.17.12",
```
pxrl, line 4: I haven't produced a new build yet, so don't necessarily need to bump this :+1:

src/clients/SpokePoolClient.ts
```diff
@@ -725,11 +725,17 @@ export class SpokePoolClient extends BaseAbstractClient {
       for (const event of speedUpEvents) {
         const speedUp: SpeedUp = { ...spreadEvent(event.args), originChainId: this.chainId };
         assign(this.speedUps, [speedUp.depositor, speedUp.depositId], [speedUp]);
-      }
 
-      // Traverse all deposit events and update them with associated speedups, If they exist.
-      for (const deposits of Object.entries(this.depositHashes)) {
```
pxrl, line 731: I wonder if it would make sense to do this in two loops - i.e. first ensure all speedUp events get assigned to `this.speedUps`, and then have a 2nd loop over `speedUpEvents` that evaluates for each relevant deposit.

The reason I'm curious about this is that `this.appendMaxSpeedUpSignatureToDeposit()` expects that `this.speedUps` is complete. I'm not sure whether that creates an issue or not, but there is one test failing as a result of this change :thinking:

pxrl: nb. the above would still save a lot of loops, because we expect that nDeposits >> nSpeedUps.

test/SpokePoolClient.SpeedUp.ts
```diff
@@ -130,6 +130,9 @@ describe("SpokePoolClient: SpeedUp", function () {
     // Deposit is not returned until we reach deposit.quoteTimestamp.
     expect(spokePoolClient.getDepositsForDestinationChain(destinationChainId).length).to.equal(0);
     await spokePool.setCurrentTime(deposit.quoteTimestamp);
+
+    // Clear out spoke pool client data and re-update so that it should now see the formerly "early" deposit.
+    spokePoolClient = new SpokePoolClient(createSpyLogger().spyLogger, spokePool, null, originChainId, deploymentBlock);
```
nicholaspai, line 135: @pxrl this fixes the test because currently it does the following which I think is unrealistic in prod:
- updates with an early deposit. Chooses not to store the early deposit in memory
- updates with a speed up. There is no early deposit in memory so doesn't apply the speedup
- updates so that the deposit is no longer early. However, since its the still the same spoke pool client, it doesn't re query older blocks to look for deposit

Is it not more realistic to use a new spoke pool client before step 3?

src/clients/SpokePoolClient.ts
nicholaspai: right, good point

nicholaspai: 17dbca4dc128b1785fa4f97979029903319a51ed

src/clients/SpokePoolClient.ts
```diff
@@ -722,14 +722,25 @@ export class SpokePoolClient extends BaseAbstractClient {
...
       for (const event of speedUpEvents) {
         const speedUp: SpeedUp = { ...spreadEvent(event.args), originChainId: this.chainId };
         assign(this.speedUps, [speedUp.depositor, speedUp.depositId], [speedUp]);
+        newSpeedUps.push(speedUp);
       }
 
-      // Traverse all deposit events and update them with associated speedups, If they exist.
-      for (const deposits of Object.entries(this.depositHashes)) {
-        this.depositHashes[deposits[0]] = this.appendMaxSpeedUpSignatureToDeposit(deposits[1]);
+      // `appendMaxSpeedUpSignatureToDeposit` assu,es that all this.speedUps has all speedups stored into memory.
```
pxrl, line 732:
```suggestion
      // `appendMaxSpeedUpSignatureToDeposit` assumes that all this.speedUps has all speedups stored into memory.
```

test/SpokePoolClient.SpeedUp.ts
pxrl: What if we just iterate over `this.speedUps` instead of `newSpeedUps` ? In the worst case we might end up hanging onto some irrelevant deposits, but this isn't an issue in serverless mode. It does however seem a bit more compatible with polling mode (which would otherwise be affected by this).

nicholaspai: https://github.com/across-protocol/sdk-v2/pull/455/commits/37c406b25e120fa4901867ae965f553b780befb6

So I tried this originally the only issue is that it creates a very ugly but not performance-affecting triple loop, because `this.speedUps` is a three-level deep object

src/clients/SpokePoolClient.ts
```diff
@@ -727,9 +727,24 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+            // We can assume all deposits in this lookback window are loaded in-memory already so if the depositHash
+            // is not mapped to a deposit, then we can throw away the speedup as it can't be applied to anything.
+            const depositDataAssociatedWithSpeedUp = this.depositHashes[depositHash];
+            if (isDefined(depositDataAssociatedWithSpeedUp)) {
+              this.depositHashes[depositHash] = this.appendMaxSpeedUpSignatureToDeposit(
+                depositDataAssociatedWithSpeedUp
+              );
+            }
+          }
+        }
```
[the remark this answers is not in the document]
nicholaspai: https://github.com/across-protocol/sdk-v2/pull/455#discussion_r1381943652

src/clients/SpokePoolClient.ts
nicholaspai: After offline discussions I've decided that its better to not support looping mode, which isn't used by anyone in prod IIUC, and maintain cleaner code

test/SpokePoolClient.SpeedUp.ts
nicholaspai: https://github.com/across-protocol/sdk-v2/pull/455/files#r1383630447

src/clients/SpokePoolClient.ts
```diff
@@ -725,11 +725,17 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+        // Find deposit hash matching this speed up event and update the deposit data associated with the hash,
+        // if the hash+data exists.
+        const depositHash = this.getDepositHash(speedUp);
+
+        // We can assume all deposits in this lookback window are loaded in-memory already so if the depositHash
+        // is not mapped to a deposit, then we can throw away the speedup as it can't be applied to anything.
+        const depositDataAssociatedWithSpeedUp = this.depositHashes[depositHash];
+        if (isDefined(depositDataAssociatedWithSpeedUp)) {
+          this.depositHashes[depositHash] = this.appendMaxSpeedUpSignatureToDeposit(depositDataAssociatedWithSpeedUp);
+        }
```
james-a-morris, line 738: Given that we are no longer being mindful of a looping mechanism, this looks good. And, since we build forward in time, we shouldn't run into the case where a deposit is loaded at a later call to `update()` and after its `speedUp` events are retrieved.

james-a-morris (approved):

pxrl (approved):

PR #457, opened 2023-11-09
chore: Define Optimism Goerli query class

pxrl:


src/relayFeeCalculator/chain-queries/optimism.ts
```diff
@@ -26,3 +31,25 @@ export class OptimismQueries extends QueryBase {
     );
   }
 }
+
+export class OptimismGoerliQueries extends QueryBase {
```
james-a-morris, line 35: Thoughts on making this a child class of `OptimismQueries`?

james-a-morris (commented):
Left a comment - otherwise lgtm

src/relayFeeCalculator/chain-queries/optimism.ts
pxrl: I thought about it but we've only done that for the Polygon because of the special way we query gas for that chain. Otherwise I'm not sure what the point would be, since there's no real inherited behaviour from one to the other.

james-a-morris: Agreed, but it would be more cohesive with the design (following Polygon) and follow a semantic OOP structure (it is a type of OptimismQueries but on testnet). 

We wouldn't need to worry about that `asL2Provider(provider)` in the constructor.

james-a-morris: Thoughts for another PR: `Base` is on the OP stack - we should probably be extending it from `OptimismQueries`

james-a-morris (approved):

src/relayFeeCalculator/chain-queries/optimism.ts
pxrl: Good call on Base/Optimism - I've described that in a ticket for some follow-up work :+1:

pxrl:
Applied `do not merge` because this PR is also dependent on a bump in both contracts-v2 and probably also constants-v2. I will wait until those builds are produced and will then bump here.

src/relayFeeCalculator/chain-queries/optimism.ts
pxrl: This is a good point - we have been impacted in other areas where an OP-stack change was made that didn't catch Base. I'm keen to get this in as-is because it's one link in a dependency chain for some ongoing work, but I've created a ticket to circle back and implement your suggestion :+1:

PR #461, opened 2023-11-14
improve(NetworkUtils): Expose production and test chain IDs

pxrl:


nicholaspai (approved):

src/utils/NetworkUtils.ts
```diff
@@ -1,5 +1,8 @@
 import { PublicNetworks } from "@uma/common/dist/PublicNetworks";
 
+export const productionChainIds = [1, 10, 137, 324, 8453, 42161];
+export const testnetChainIds = [5, 280, 420, 80001, 84531, 421613];
```
james-a-morris, line 4: Thoughts on adding this to `constants.ts` since it's very general and could be useful elsewhere?

james-a-morris (approved):
LGTM - left one comment about the placement of the production/testnet chain list. Also, we may want to bump the version number.

src/utils/NetworkUtils.ts
pxrl: Makes sense - good call :+1: https://github.com/across-protocol/sdk-v2/pull/461/commits/768139211d72a27ac4827cf10181106356088357

PR #462, opened 2023-11-14
feat: Add dedupArray()

pxrl:


nicholaspai (approved):

src/utils/ArrayUtils.ts
```diff
@@ -1,3 +1,12 @@
+/**
+ * De-duplicates an array by filtering it via a Set.
+ * @param array The array to de-duplicate.
+ * @returns A new array, filtered for uniqueness.
```
james-a-morris, line 4: If we're adding this function, should we include an optional comparator?

```suggestion
 * @returns A new array, filtered for uniqueness.
 * @note This filtering is using primitive comparisons. DO NOT USE for complex array types.
```

james-a-morris (commented):
One comment about potentially complex data structures being passed into this function.

PR #463, opened 2023-11-21
improve(ConfigStoreClient): Improve logs and optimistically don't load `latestBlockNumber` if passed in as search config toBlock

nicholaspai:
Precursor to support https://github.com/across-protocol/relayer-v2/pull/1085


mrice32 (approved):
LGTM!

PR #465, opened 2023-11-27
improve(BlockFinder): Allow the caller to supply hints

pxrl:
If the caller supplies a sensible hint then this will skip the initial lower-bound block estimation, which can be wildly inaccurate in cases where the average block time has changed (especially reduced) over time. This is also nice because one of the main beneficiaries of this is in the relayer, the hinted lowBlock can be set statically, so it has a much greater chance of being served out of the RPC provider cache.

### DIFF
--- a/src/utils/BlockUtils.ts
+++ b/src/utils/BlockUtils.ts
@@ -91,6 +91,11 @@ async function estimateBlocksElapsed(seconds: number, cushionPercentage = 0.0, p
   return Math.floor((seconds * cushionMultiplier) / average);
 }
 
+export type BlockFinderHint = {
+  lowBlock?: number;
+  highBlock?: number;
+};
+
 export class BlockFinder {
   constructor(
     private readonly provider: Provider,
@@ -101,7 +106,7 @@ export class BlockFinder {
    * @notice Gets the latest block whose timestamp is <= the provided timestamp.
    * @param {number} timestamp timestamp to search.
    */
-  public async getBlockForTimestamp(timestamp: number | string): Promise<Block> {
+  public async getBlockForTimestamp(timestamp: number | string, hints: BlockFinderHint): Promise<Block> {
     timestamp = Number(timestamp);
     assert(timestamp !== undefined && timestamp !== null, "timestamp must be provided");
     // If the last block we have stored is too early, grab the latest block.
@@ -110,7 +115,11 @@ export class BlockFinder {
       if (timestamp >= block.timestamp) return block;
     }
 
-    // Check the first block. If it's grater than our timestamp, we need to find an earlier block.
+    // Prime the BlockFinder cache with any supplied hints.
+    // If the hint is accurante then this will bypass the subsequent estimation.
+    await Promise.all(Object.values(hints).map((blockNumber) => this.getBlock(blockNumber)));
+
+    // Check the first block. If it's greater than our timestamp, we need to find an earlier block.
     if (this.blocks[0].timestamp > timestamp) {
       const initialBlock = this.blocks[0];
       // We use a 2x cushion to reduce the number of iterations in the following loop and increase the chance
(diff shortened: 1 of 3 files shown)


src/utils/BlockUtils.ts
```diff
@@ -101,7 +106,7 @@ export class BlockFinder {
    * @notice Gets the latest block whose timestamp is <= the provided timestamp.
    * @param {number} timestamp timestamp to search.
    */
-  public async getBlockForTimestamp(timestamp: number | string): Promise<Block> {
+  public async getBlockForTimestamp(timestamp: number | string, hints: BlockFinderHint): Promise<Block> {
```
james-a-morris, line 109: Do we want to make this optional as well? Right now we'd have to pass `getBlockForTimestamp(..., {})` to be type compliant. It may be convenient to provide a default.
```suggestion
  public async getBlockForTimestamp(timestamp: number | string, hints: BlockFinderHint = {}): Promise<Block> {
```

james-a-morris, line 109: We should update the doc comment

src/utils/BlockUtils.ts
```diff
@@ -110,7 +115,11 @@ export class BlockFinder {
       if (timestamp >= block.timestamp) return block;
     }
 
-    // Check the first block. If it's grater than our timestamp, we need to find an earlier block.
+    // Prime the BlockFinder cache with any supplied hints.
+    // If the hint is accurante then this will bypass the subsequent estimation.
```
james-a-morris, line 119:
```suggestion
    // If the hint is accurate, then this will bypass the subsequent estimation.
```

james-a-morris (commented):
Left a few comments. I suppose I'm a bit confused as to why we'd want to prime the hints when they'd eventually get primed anyway. But I assume that logic is outside the scope of this PR and that there's an intentional reason for this PR.

pxrl:
> Left a few comments. I suppose I'm a bit confused as to why we'd want to prime the hints when they'd eventually get primed anyway. But I assume that logic is outside the scope of this PR and that there's an intentional reason for this PR.

The BlockFinder's estimation is sometimes inaccurate because it relies on consistent average block times in order to produce a sensible lower bound for the subsequent search. This is fine for mainnet, but on the various L2s and alt-L1s the block times often haven't been as consistent, so the estimate can be wildly wrong (I've seen it estimate 0 as a sensible start block on some testnets).

At the same time, this is the best initial estimate it can make because it doesn't have any of the relevant context about the timestamp, so it makes sense for the caller to supply some extra context about where the target block might lie. This relegates the initial estimate to being a fallback behaviour in the event that no hint is supplied, and in that scenario is still works, albeit with some undesirable worst-case performance.

Note that this is also a nice feature to have if we were to sub out the interpolation search for a binary search, which would round off the worst-case performance but isn't quite as effective on mainnet.

src/utils/BlockUtils.ts
mrice32, line 109: tiny nit: I think hint is singular.

```suggestion
  public async getBlockForTimestamp(timestamp: number | string, hint: BlockFinderHint): Promise<Block> {
```

mrice32 (approved):
LGTM!

src/utils/BlockUtils.ts
```diff
@@ -110,7 +115,11 @@ export class BlockFinder {
...
+    await Promise.all(Object.values(hints).map((blockNumber) => this.getBlock(blockNumber)));
```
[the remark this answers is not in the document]
pxrl: Hardcoding it doesn't work super well because both `lowBlock` and `highBlock` can be undefined, and this would cause an issue in `getBlock()`. `mapAsync()` seems like an OK candidate, but I've updated the code to include a filter stage to prevent supplying an undefined block number, so I don't really see much opportunity to simplify this now anyway :shrug:

src/utils/BlockUtils.ts
pxrl: Good call: https://github.com/across-protocol/sdk-v2/pull/465/commits/d2bfc9bb16dd977dab2d30fa940234a82f3debb4

pxrl: I tried this but reverted to using plural everywhere, so at least it's consistent.

PR #468, opened 2023-12-03
fix(HubPoolClient): Avoid utilization cache collisions

pxrl:
The last change to HubPool utilization computation ended up reusing the same cache key for post-relay utilization, but storing the value with a different format (1 number instead of 2 comma-separated).

This PR replaces the existing getUtilization() function because it's no longer used, and adds a temporarily change to the post-relay utilization key to avoid collisions until the original keys are ejected.

This was caught during final local testing of the PR to bump sdk-v2 within relayer-v2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.17.16",
+  "version": "0.17.17",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/v/developer-docs/developers/across-sdk",
   "files": [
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

james-a-morris (approved):
LGTM - pending that one change we talked about by adding a //FIXME

pxrl:
> LGTM - pending that one change we talked about by adding a //FIXME

https://github.com/across-protocol/sdk-v2/pull/468/commits/954edb91b09dd6df99a81724237a052bb0e175ff

pxrl:
I verified this again against recent bundles and it produced results consistent with prod - including the 2 recent disputes.

PR #469, opened 2023-12-09
refator(SpokePoolClient): Make caller optional in events

pxrl:
The caller field is typically defaulting to the Multicall3 address. This has been eliminated for the USS and is not really useful pre-USS anyway.

### DIFF
--- a/src/interfaces/SpokePool.ts
+++ b/src/interfaces/SpokePool.ts
@@ -165,7 +165,7 @@ export interface RelayerRefundExecution {
   leafId: number;
   l2TokenAddress: string;
   refundAddresses: string[];
-  caller: string;
+  caller?: string;
 }
 
 export interface RelayerRefundExecutionWithBlock extends RelayerRefundExecution, SortableEvent {}
@@ -228,7 +228,7 @@ export interface TokensBridged extends SortableEvent {
   chainId: number;
   leafId: number;
   l2TokenAddress: string;
-  caller: string;
+  caller?: string;
 }
 
 export type TokensBridgedStringified = Omit<TokensBridged, "amountToReturn"> & {


nicholaspai (approved):

nicholaspai:
Do you think we should just eliminate this field entirely if it's not used in any other place? This way we don't even set it in the object and it makes it slightly easier to update?

pxrl:
> Do you think we should just eliminate this field entirely if it's not used in any other place? This way we don't even set it in the object and it makes it slightly easier to update?

We definitely don't need it and I don't think any code even evaluates this field, but the way that events are unpacked into our objects means that we will always end up with a `caller` field for pre-USS executions, so I was hesitant to remove it outright.

I am onboard with removing it though, so I'll test that and see how that goes.

nicholaspai:
> > Do you think we should just eliminate this field entirely if it's not used in any other place? This way we don't even set it in the object and it makes it slightly easier to update?
> 
> We definitely don't need it and I don't think any code even evaluates this field, but the way that events are unpacked into our objects means that we will always end up with a `caller` field for pre-USS executions, so I was hesitant to remove it outright.
> 
> I am onboard with removing it though, so I'll test that and see how that goes.

https://github.com/across-protocol/sdk-v2/pull/467

pxrl:
> > > Do you think we should just eliminate this field entirely if it's not used in any other place? This way we don't even set it in the object and it makes it slightly easier to update?
> > 
> > 
> > We definitely don't need it and I don't think any code even evaluates this field, but the way that events are unpacked into our objects means that we will always end up with a `caller` field for pre-USS executions, so I was hesitant to remove it outright.
> > I am onboard with removing it though, so I'll test that and see how that goes.
> 
> #467

Ah sorry - I hadn't realised you already had proposed that. I'll close this one.

PR #470, opened 2023-12-12
chore: Flush unused functions

pxrl:


james-a-morris (approved):
Nice cleanup. We may want to visit the SDK in the future to clean up additional functionality.

PR #472, opened 2023-12-12
feat: add ComposableBridging simulation unit tests

james-a-morris:
The purpose of these unit tests is to prevent regressions in our relayFeeCalculator as we make further changes.


contracts/MockAcrossMessageContract.sol
```diff
@@ -0,0 +1,40 @@
...
+contract MockAcrossMessageContract is AcrossMessageHandler {
+  function handleAcrossMessage(
+    address tokenSent,
+    uint256 amount,
+    bool fillCompleted,
+    address relayer,
+    bytes memory message
+  ) external view override {
+    // "Use" the variables to avoid a compilation warning.
+    tokenSent;
```
nicholaspai, line 26: The more correct fix to this is to remove the variable name in the function declaration. For example instead of declaring:

```sol
function x(uint256 unusedVar, ...)
```

change it to
```sol
function x(uint256, ...
```

contracts/MockAcrossMessageContract.sol
```diff
@@ -0,0 +1,40 @@
...
+    amount;
+    fillCompleted;
+    relayer;
+    message;
+    // For the case that we want to test a revert.
+    // We consider a message with a single byte 0x02 to be a revert.
+    require(!(message.length == 1 && message[0] == 0x02), "MockAcrossMessageContract: revert");
```
nicholaspai, line 33: Why not make something more intuitive than 0x02 a message, like `bytes("REVERT")`

contracts/MockAcrossMessageContract.sol
```diff
@@ -0,0 +1,40 @@
...
+    // Iterate from 0 to 1000 to simulate a long-running operation.
```
nicholaspai, line 34: is this long running? How much gas does this expect to spend? Since we seem to want to test that a message adds a non-zero amount of gas, why not just do somethingn even simpler like incrementing the loop from 0 --> 100 and doing nothing. The increment of the memory variable `i` is sufficient so I think removing the extra `address(this).balance` is kind of overkill

nicholaspai: Even calling into this function incurs a fixed cost so we don't even need the loop arguably

test/relayFeeCalculator.test.ts
```diff
@@ -272,3 +290,73 @@ describe("RelayFeeCalculator", () => {
...
+  });
+  it("should not revert if no message is passed", async () => {
+    await assertPromisePasses(testGasFeePct());
+  });
+  it("should revert if the contract message fails", async () => {
+    // Per our test contract, a single byte message of 0x02 will revert.
+    const message = "0x02";
+    await assertPromiseError(testGasFeePct(message), "MockAcrossMessageContract: revert");
+  });
+  it("should be more gas to call a contract with a message", async () => {
```
nicholaspai, line 357: Can we be more precise? Can we get tighter bounds around the expected gas expenditure to test that the relayFeeCalculator correctly adds the gas cost to run the message to the total gas fee?

nicholaspai (commented):
I left several questions on the implementation but directionally +1 to unit testing this to prevent regressions!

contracts/MockAcrossMessageContract.sol
james-a-morris: That's fair. I've added a more intuitive message that will cause a revert. https://github.com/across-protocol/sdk-v2/pull/472/commits/37b60fd432221fb0c1d1134f78b98e70221cb1f0

contracts/MockAcrossMessageContract.sol
james-a-morris: The loop exists so that there is an easily viewed reason that we are expending gas outside the intrinsic gas costs of calling any function.

I wanted this function to touch the chain in some way so that it could be marked as `view` and not `virtual` to simulate what a realistic function could look like.

I've removed the `address(this).balance`, but I'd recommend we keep the look so that this function showcases that it's meant to expend more gas.

test/relayFeeCalculator.test.ts
james-a-morris: Good call - added further bounds

nicholaspai (approved):

PR #473, opened 2023-12-13
improve: rename `latestBlockNumber` to `latestBlockSearched`

nicholaspai:
The current name `latestBlockNumber` usually implies that it should return the HEAD block for the provider but the way its used, it 
should actually return the latest block searched by the client.

### DIFF
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -75,8 +75,6 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
   public cumulativeDisabledChainUpdates: DisabledChainsUpdate[] = [];
 
   protected rateModelDictionary: across.rateModel.RateModelDictionary;
-  public firstBlockToSearch: number;
-  public latestBlockSearched = 0;
 
   public hasLatestConfigStoreVersion = false;
   public chainId: number | undefined;
@@ -89,6 +87,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
   ) {
     super();
     this.firstBlockToSearch = eventSearchConfig.fromBlock;
+    this.latestBlockSearched = 0;
     this.rateModelDictionary = new across.rateModel.RateModelDictionary();
   }
 
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -76,7 +76,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
 
   protected rateModelDictionary: across.rateModel.RateModelDictionary;
   public firstBlockToSearch: number;
-  public latestBlockNumber = 0;
+  public latestBlockSearched = 0;
 
   public hasLatestConfigStoreVersion = false;
   public chainId: number | undefined;
@@ -519,7 +519,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     this.rateModelDictionary.updateWithEvents(this.cumulativeRateModelUpdates);
 
     this.hasLatestConfigStoreVersion = this.hasValidConfigStoreVersionForTimestamp();
-    this.latestBlockNumber = result.searchEndBlock;
+    this.latestBlockSearched = result.searchEndBlock;
     this.firstBlockToSearch = result.searchEndBlock + 1; // Next iteration should start off from where this one ended.
     this.chainId = this.chainId ?? chainId; // Update on the first run only.
     this.isUpdated = true;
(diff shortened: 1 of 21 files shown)


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -76,7 +76,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
 
   protected rateModelDictionary: across.rateModel.RateModelDictionary;
   public firstBlockToSearch: number;
-  public latestBlockNumber = 0;
+  public latestBlockSearched = 0;
```
pxrl, line 79: Should we put this in `BaseAbstractClient.ts` ? This generic update functionality feels like it could belong there and would give greater consistency across all of the inheriting classes.

src/clients/HubPoolClient.ts
```diff
@@ -107,7 +107,7 @@ export class HubPoolClient extends BaseAbstractClient {
     cachingMechanism?: CachingMechanismInterface
   ) {
     super(cachingMechanism);
-    this.latestBlockNumber = deploymentBlock === 0 ? deploymentBlock : deploymentBlock - 1;
+    this.latestBlockSearched = deploymentBlock === 0 ? deploymentBlock : deploymentBlock - 1;
```
pxrl, line 110: Opportunistic tweak:

```suggestion
    this.latestBlockSearched = Math.min(deploymentBlock - 1, 0);
```

src/utils/BlockUtils.ts
pxrl, line 1: All of the changes in this file are unrelated to the {Hub,Spoke}PoolClient updates so they should be reverted.

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
nicholaspai: you mean move these generic class properties like this? 7a07154d3f828583ef3c7a0dd577fa3cc474a48f

src/utils/BlockUtils.ts
nicholaspai: I think these are correct though because this PR is essentially deprecating `latestBlockNumber` in favor of a variable named `latestBlockSearched`, thoughts?

pxrl: I think something like `highBlock` or `endBlock` is a better fit, since block number in this context is only choosing the end block that's used for determining an average block time over a block range - it's not related to searching.

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
pxrl: Yeah - the batched update feature is basically common between all of the classes, but I have noticed that we have some slight differences between them (and duplicated code, to some extent). We can improve on that by pulling it out to the common class that they all inherit from :+1:.

src/clients/BaseAbstractClient.ts
```diff
@@ -6,6 +6,8 @@ import { isDefined } from "../utils";
  */
 export abstract class BaseAbstractClient {
   protected _isUpdated: boolean;
+  public firstBlockToSearch = 0;
+  public latestBlockSearched: number | undefined;
```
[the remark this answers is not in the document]
nicholaspai: Ok but we'll have to be careful to change any checks in relayer-v2 for this being undefined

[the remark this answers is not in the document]
nicholaspai: I'm not sure where I see that type assertion in `BundleUtils` but this commit makes `latestBlockSearched` type number d983bcb482bf2f0f62cdf7ae7473ab18a8623c2c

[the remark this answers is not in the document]
james-a-morris: This may be a good push to add the tsconfig option to check for undefined/null types in the relayer. We currently don't check for it

src/transfers-history/services/SpokePoolEventsQueryService.ts
```diff
@@ -23,8 +23,8 @@ export class SpokePoolEventsQueryService {
     let speedUpDepositEvents: RequestedSpeedUpDepositEvent[] = [];
     let blockTimestampMap: { [blockNumber: number]: number } = {};
 
-    if (this.latestBlockNumber) {
-      from = this.latestBlockNumber + 1;
+    if (this.latestBlockSearched) {
```
james-a-morris, line 26: Do we want to check if this is undefined, or is it okay to jump to the else statement if `latestBlockSearched === 0`?

My assumption is that we consider latestBlockSearched=0 an erroneous state, but it's worth double checking

james-a-morris (approved):
Nice refactor - I only have one high level question about the relayer accounting for null/undefined states.

src/utils/BlockUtils.ts
nicholaspai: a5eb37cec01216a07db86193188d4095274dab55

thoughts

pxrl (approved):
LGTM!

PR #476, opened 2023-12-17
refactor: Remove SpokePoolClient dependency from depositId search

pxrl:
The core binary search logic doesn't depend on the SpokePoolClient, so make a clean break.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -437,7 +437,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    *        // where the deposit with deposit ID = targetDepositId was created.
    */
   public _getBlockRangeForDepositId(
-    targetDepositId: number,
+    depositId: number,
     initLow: number,
     initHigh: number,
     maxSearches: number
@@ -445,7 +445,7 @@ export class SpokePoolClient extends BaseAbstractClient {
     low: number;
     high: number;
   }> {
-    return getBlockRangeForDepositId(targetDepositId, initLow, initHigh, maxSearches, this);
+    return getBlockRangeForDepositId(depositId, initLow, initHigh, maxSearches, this.spokePool, this.deploymentBlock);
   }
 
   /**
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -1,5 +1,5 @@
 import { Contract } from "ethers";
-import { SpokePoolClient } from "../clients";
+import { getDeployedBlockNumber } from "@across-protocol/contracts-v2";
 
 /**
  * Find the block range that contains the deposit ID. This is a binary search that searches for the block range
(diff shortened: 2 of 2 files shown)


src/utils/SpokeUtils.ts
```diff
@@ -21,35 +21,26 @@ export async function getBlockRangeForDepositId(
...
       [initLow <= initHigh, "Binary search failed because low > high"],
-      // Sanity check to ensure that init Low is greater than or equal to zero.
       [initLow >= deploymentBlock, "Binary search failed because low must be >= deploymentBlock"],
-      // Sanity check to ensure that maxSearches is greater than zero.
       [maxSearches > 0, "maxSearches must be > 0"],
-      // Sanity check to ensure that deploymentBlock is greater than or equal to zero.
       [deploymentBlock >= 0, "deploymentBlock must be >= 0"],
     ] as [boolean, string][]
   ).forEach(([condition, errorMessage]) => {
-    // If the condition is false, then we will throw an error.
```
pxrl, line 52: These comments were just 100% mirroring what the code does (and what the error messages said the code was doing), so I opportunistically deleted them but wanted to flag it just in case.

src/utils/SpokeUtils.ts
```diff
@@ -62,15 +53,14 @@ export async function getBlockRangeForDepositId(
   // queriedIds cache to see if the deposit ID at the block number has already been queried. If not, it will
   // make an eth_call request to get the deposit ID at the block number. It will then cache the deposit ID
   // in the queriedIds cache.
-  const _getDepositIdAtBlock = async (blockNumber: number): Promise<number> => {
-    queriedIds[blockNumber] ??= await spokePool._getDepositIdAtBlock(blockNumber);
-    return queriedIds[blockNumber];
+  const getDepositIdAtBlock = async (blockTag: number): Promise<number> => {
+    return queriedIds[blockTag] ??= await spokePool.numberOfDeposits({ blockTag });
```
pxrl, line 57: Intuitively this implementation is wrong to me, since numberOfDeposits != "highest depositId". I'd expect that we should be doing something like:

```
const nDeposits = await spokePoolClient.numberOfDeposits({ blockTag });
return Math.max(nDeposits - 1, 0);
```

nb. the above ignores the in-memory cache, but it's implied that that is used.

pxrl (commented):
A few explanatory comments.

src/clients/SpokePoolClient.ts
```diff
@@ -437,15 +437,15 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    depositId: number,
     initLow: number,
     initHigh: number,
     maxSearches: number
   ): Promise<{
     low: number;
     high: number;
   }> {
-    return getBlockRangeForDepositId(targetDepositId, initLow, initHigh, maxSearches, this);
+    return getBlockRangeForDepositId(depositId, initLow, initHigh, maxSearches, this.spokePool, this.deploymentBlock);
```
james-a-morris, line 448: This is a good call - I thought we did this previously. +1

james-a-morris (approved):
LGTM - +1

PR #478, opened 2023-12-18
fix: SpokePoolClient oops

pxrl:


src/clients/SpokePoolClient.ts
```diff
@@ -138,7 +138,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @returns A list of deposits.
    * @note This method returns all deposits, regardless of destination chain ID in sorted order.
    */
-  public getDeposits(filter?: { fromBlock: number, toBlock: number }): DepositWithBlock[] {
+  public getDeposits(filter?: { fromBlock: number; toBlock: number }): DepositWithBlock[] {
```
pxrl, line 141: fwiw it's curious that this passed CI previously. Could we have a blind spot ?

nicholaspai: yep

nicholaspai (approved):

src/clients/SpokePoolClient.ts
james-a-morris: Eesh if this passed then it's more of a linter issue - it's nonstandard to use `,` to separate objects but isn't necessarily a compiler issue. It may be worth having _at least_ a `yarn lint` pre-commit hook so this can be caught prior to submitting to git

james-a-morris (approved):
Left feedback

PR #480, opened 2023-12-19
feat(HubPoolClient): Support testing SpokePool token equivalency

pxrl:
This enables the bots to determine equivalency between input token and output token in Across v3.

Closes ACX-1666

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -358,7 +358,8 @@ export class HubPoolClient extends BaseAbstractClient {
       );
 
       // Resolve the HubPool token address, if it isn't already known.
-      const hubPoolToken = hubPoolTokens[originToken] ?? this.getL1TokenForDeposit(deposit);
+      const quoteBlockNumber = quoteBlocks[deposit.quoteTimestamp];
+      const hubPoolToken = hubPoolTokens[originToken] ?? this.getL1TokenForDeposit({ ...deposit, quoteBlockNumber });
       hubPoolTokens[originToken] ??= hubPoolToken;
 
       // Append the quoteTimestamp for this HubPool token, if it isn't already enqueued.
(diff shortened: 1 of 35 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -483,6 +483,30 @@ export class HubPoolClient extends BaseAbstractClient {
...
+      const l1TokenA = this.getL1TokenForL2TokenAtBlock(tokenA, chainIdA, hubPoolBlock);
+      const l1TokenB = this.getL1TokenForL2TokenAtBlock(tokenB, chainIdB, hubPoolBlock);
+      if (l1TokenA !== l1TokenB) {
+        return false;
+      }
+
+      // Resolve both HubPool tokens back to a current SpokePool token and verify that they match.
+      const _tokenA = this.getL2TokenForL1TokenAtBlock(l1TokenA, chainIdA, hubPoolBlock);
+      const _tokenB = this.getL2TokenForL1TokenAtBlock(l1TokenB, chainIdB, hubPoolBlock);
+      return tokenA === _tokenA && tokenB === _tokenB;
```
pxrl, line 504: nb. this might seem like a counter-intuitive check, but the current HubPoolClient implementation will retain old mappings of SpokePool token -> HubPool token even after the HubPool -> SpokePool mapping has been updated. There might be a more intelligent way to achieve the same check; this just seemed simple.

nicholaspai (approved):
I think this handles the weird edge case where L2 tokens have the same address but map to different L1 tokens. For example BOBA weth and OP eth

pxrl:
> I think this handles the weird edge case where L2 tokens have the same address but map to different L1 tokens. For example BOBA weth and OP eth

DAI & BAL were also considerations here, since they share the same token address on some chains (but not all).

PR #481, opened 2023-12-20
improve(EventManager): Auto-enqueue generated events

pxrl:
This frees the caller from having to track generated events.

### DIFF
--- a/src/clients/mocks/MockSpokePoolClient.ts
+++ b/src/clients/mocks/MockSpokePoolClient.ts
@@ -144,12 +144,12 @@ export class MockSpokePoolClient extends SpokePoolClient {
 
     const { blockNumber, transactionIndex } = deposit;
     let { depositId, depositor, destinationChainId } = deposit;
-    depositId ??= this.numberOfDeposits;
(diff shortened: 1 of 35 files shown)


src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -67,10 +66,6 @@ export class MockSpokePoolClient extends SpokePoolClient {
     this.latestBlockSearched = blockNumber;
   }
 
-  addEvent(event: Event): void {
```
nicholaspai, line 70: Is this manual "push" fallback function useful at all/worth keeping? Or is it the programmer's "bug" if they're having to call this manual function

nicholaspai (approved):
This is a really nice QoL, just one question: https://github.com/across-protocol/sdk-v2/pull/481/files#r1440722799

src/clients/mocks/MockSpokePoolClient.ts
pxrl: It's been relocated to the `eventManager´ object, and that's public in this class so it is still available to use if needed - albeit a little bit less direct to get to.

PR #486, opened 2024-01-02
feat: Propagate invalid fill reasons to caller

pxrl:
This change provides additional information to the caller of queryHistoricalDepositForFill(). In the case that a fill is not found, a cause code is associated:
- 0: The deposit ID was invalid for the SpokePool (i.e. cannot exist).
- 1: The deposit ID was valid, but it wasn't found within scraped data (i.e. maybe the RPCs aren't serving it in eth_getLogs() requests).
- 2: The deposit ID was valid, but one or more fields in the corresponding deposit were mismatched (i.e. the fill was invalid).

This is most interesting for case (1), where there's a reasonable indication that the deposit should be held within the SpokePoolClient's deposit map, but isn't. This indirectly acts like a qualitative measure for RPC provider responses and will allow the upper layers to dynamically revert to more conservative behaviour.

### DIFF
--- a/src/utils/DepositUtils.ts
+++ b/src/utils/DepositUtils.ts
@@ -69,7 +69,7 @@ export async function queryHistoricalDepositForFill(
 
     return {
       found: false,
-      code: InvalidFill.DepositIdNotFound,
+      code: isDefined(deposit) ? InvalidFill.FillMismatch : InvalidFill.DepositIdNotFound,
       reason: `Deposit ID ${depositId} not found in SpokePoolClient event buffer.`
     };
   }
@@ -104,10 +104,7 @@ export async function queryHistoricalDepositForFill(
   }
 
   if (validateFillForDeposit(fill, deposit)) {
-    return {
-      found: true,
-      deposit,
-    };
+    return { found: true, deposit };
   }
 
   return {
--- a/src/utils/DepositUtils.ts
+++ b/src/utils/DepositUtils.ts
@@ -17,11 +17,11 @@ export enum InvalidFill {
   DepositIdInvalid = 0, // Deposit ID seems invalid for origin SpokePool
   DepositIdNotFound, // Deposit ID not found (bad RPC data?)
   FillMismatch, // Fill does not match deposit parameters for deposit ID.
-};
+}
 
 export type DepositSearchResult =
-  | { found: true; deposit: DepositWithBlock; }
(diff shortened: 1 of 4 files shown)


test/SpokePoolClient.ValidateFill.ts
```diff
@@ -469,9 +475,13 @@ describe("SpokePoolClient: Fill Validation", function () {
...
     // Override the first spoke pool deposit ID that the client thinks is available in the contract.
     await spokePoolClient1.update();
-    spokePoolClient1.firstDepositIdForSpokePool = 1;
+    spokePoolClient1.firstDepositIdForSpokePool = deposit.depositId + 1;
     expect(fill.depositId < spokePoolClient1.firstDepositIdForSpokePool).is.true;
-    await queryHistoricalDepositForFill(spokePoolClient1, fill);
+    const search = await queryHistoricalDepositForFill(spokePoolClient1, fill);
+    expect(search.found).is.false;
+    if (search.found === false) {
+      expect(search.code).to.equal(InvalidFill.DepositIdInvalid);
```
pxrl, line 483: Reminder to also test for `DepositIdNotFound` and `FillMismatch`.

nicholaspai (approved):
This is a great solution. Do you want to add unit tests for the other InvalidDeposit codes in this PR or a follow up? I think it's ok to do a follow up since this updates all existing logic

src/utils/DepositUtils.ts
```diff
@@ -82,7 +103,18 @@ export async function queryHistoricalDepositForFill(
     }
   }
 
-  return validateFillForDeposit(fill, deposit) ? deposit : undefined;
+  if (validateFillForDeposit(fill, deposit)) {
```
mrice32, line 106: Passing back the reasons is really great! Not necessary for this PR, but we may want to also deal with errors in the findDeposit call similarly, so they can be attributed with reasons instead of just throwing an error in the middle of this function.

mrice32 (approved):
LGTM!

test/SpokePoolClient.ValidateFill.ts
pxrl: Tests were updated to verify `FIllMismatch`. `DepositIdNotFound` is currently difficult to test but I am working on that downstream in `relayer-v2`, and those tests might be backported to the SDK.

PR #488, opened 2024-01-04
feat: Helper to resolve the block number for a fill

pxrl:
Given the valid relayData that can be constructed from a Deposit, resolve whether the deposit was ever filled on the destination chain without relying on events. This can be used as a sanity check on RPC responses that sometimes drop logs.

The "left-most" variant of a binary search is implemented. The logic is very simple and has been repeatedly tested to resolve in ~20 iterations with a fill made randomly within a range of 1M blocks. This can be verified locally by adjusting the _nBlocks_ variable within the accompanying test case (remember also to bump the test timeout via `this.timeout(ms)` in the relevant test case).

The criteria for determining when the fill was made is the block at which is was considered complete. This effectively ignores partial fills entirely, but that's fortunately forwards-compatible with the direction of Across.

### DIFF
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -2,6 +2,7 @@ import { getRelayHash } from "@across-protocol/contracts-v2/dist/test-utils";
 import { BigNumber, Contract } from "ethers";
 import { RelayData } from "../interfaces";
 import { SpokePoolClient } from "../clients";
+import { getNetworkName } from "./NetworkUtils";
 
 /**
  * Find the block range that contains the deposit ID. This is a binary search that searches for the block range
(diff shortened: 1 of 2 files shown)


src/utils/SpokeUtils.ts
```diff
@@ -177,3 +178,45 @@ export function relayFilledAmount(
...
 }
+
+export async function findFillBlock(
+  spokePool: Contract,
+  relayData: RelayData,
+  lowBlockNumber: number,
+  highBlockNumber?: number
+): Promise<number | undefined> {
+  const { provider } = spokePool;
+  highBlockNumber ??= await provider.getBlockNumber();
```
pxrl, line 189: todo: Should probably assert `highBlockNumber > lowBlockNumber`.

src/utils/SpokeUtils.ts
```diff
@@ -177,3 +178,45 @@ export function relayFilledAmount(
 
   return spokePool.relayFills(hash, { blockTag });
 }
+
+export async function findFillBlock(
```
pxrl, line 182: todo: missing documentation for `fiindFillBlock()`.

nicholaspai (approved):

src/utils/SpokeUtils.ts
pxrl: https://github.com/across-protocol/sdk-v2/pull/488/commits/1e6d86d95b2e1377959bb7100bbdb851c8923594

src/utils/SpokeUtils.ts
pxrl: https://github.com/across-protocol/sdk-v2/pull/488/commits/0b4a2dd1b6dbe33160cc02a343297939a4463215

PR #489, opened 2024-01-09
chore: Decouple transfers-history from apiClient

pxrl:
The ChainId definition is only a number, so define it directly in the apiClient. Done in anticipation of the transfers-client implementation being removed from the sdk, pending confirmation as to whether anyone actually uses it.


james-a-morris (approved):
Good call

PR #490, opened 2024-01-09
chore: Remove transfers-history implementation

pxrl:
This is unsued code that will impose unnecessary work in the v2 -> v3 migration.


pxrl:
To be merged subject to confirmation from @amateima

amateima (approved):
🥲

PR #491, opened 2024-01-09
improve(relayFeeCalculator): Skip redundant decimals lookup

pxrl:
getFeePercent() already obtains the destination token decimals via a call to getTokenInformationFromAddress(), so there's no need for a separate lookup via getTokenDecimals().


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -221,8 +221,8 @@ export class RelayFeeCalculator {
     relayerAddress = DEFAULT_SIMULATED_RELAYER_ADDRESS,
     _tokenPrice?: number
   ): Promise<BigNumber> {
-    const tokenInformation = getTokenInformationFromAddress(deposit.originToken);
```
pxrl, line 224: Renamed to avoid reflowing line 265.

nicholaspai (approved):

PR #493, opened 2024-01-12
improve: Identify reason for deposit <-> fill mismatches

pxrl:
The combination of abiCoder.encode() and keccak256() seems to be
relatively expensive, and this can be a real performance suck when
called repeatedly in a tight loop. Additionally, performing a per-field
comparison exposes the underlying error when two RelayData objects don't
match; this makes it easy to identify the root cause for invalid fills.


src/utils/DepositUtils.ts
```diff
@@ -103,14 +113,15 @@ export async function queryHistoricalDepositForFill(
     }
   }
 
-  if (validateFillForDeposit(fill, deposit)) {
+  const match = validateFillForDeposit(fill, deposit)
```
pxrl, line 116:
```suggestion
  const match = validateFillForDeposit(fill, deposit);
```

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -134,13 +134,12 @@ describe("SpokePoolClient: Fill Validation", function () {
 
     // Some fields are expected to be dynamically populated by the client, but aren't in this environment.
     // Fill them in manually from the fill struct to get a valid comparison.
```
pxrl, line 136:
```suggestion
    // Populate them manually from the fill struct to get a valid comparison.
```

nicholaspai (approved):

src/utils/FlowUtils.ts
```diff
@@ -69,18 +69,25 @@ export function filledSameDeposit(fillA: Fill, fillB: Fill): boolean {
...
   }
 
   // Note: this short circuits when a key is found where the comparison doesn't match.
   // TODO: if we turn on "strict" in the tsconfig, the elements of FILL_DEPOSIT_COMPARISON_KEYS will be automatically
   // validated against the fields in Fill and Deposit, generating an error if there is a discrepency.
-  return FILL_DEPOSIT_COMPARISON_KEYS.every((key) => {
-    if (fillFieldsToIgnore.includes(key)) {
-      return true;
+  for (const key of FILL_DEPOSIT_COMPARISON_KEYS) {
+    if (fill[key]?.toString() === deposit[key]?.toString() || fillFieldsToIgnore.includes(key)) {
```
nicholaspai, line 85: +1

src/clients/SpokePoolClient.ts
```diff
@@ -335,10 +335,21 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+
+    const originChain = getNetworkName(fill.originChainId);
+    this.logger.debug({
+      at: "SpokePoolClient::getDepositForFill",
+      message: `Rejected fill for ${originChain} deposit ${fill.depositId}.`,
+      reason: match.reason,
+      deposit,
+      fill
+    });
+    return undefined;
```
james-a-morris, line 352: Nit: do we need to explicitly return undefined?

src/utils/FlowUtils.ts
james-a-morris, line 85: Minor nit: I wonder if the resulting code block would be more intuitive using a `find()` for the error case because that's what we're doing here.

james-a-morris (approved):
I don't have anything actionable to mention about this PR - LGTM. Left a few nits about using the built-in array operators/return statements.

src/utils/FlowUtils.ts
pxrl: I migrated to using `find` here: https://github.com/across-protocol/sdk-v2/pull/493/commits/0820373148e5e70c036bcc348d9a9afe631c107b

I'd initially avoided it because it required inverting the negating all the comparison logic, but I do agree that it's an improvement :+1:

src/clients/SpokePoolClient.ts
pxrl: *shrug* We're probably better to be explicit on return types, no?

nicholaspai (approved):

pxrl:
@bmzig This is a _very_ old PR that I've just revived. It subs out RelayData comparison by hash to RelayData comparison by evaluating each individual field.

I see a pretty decent performance improvement with this, in the order of ~2 seconds per loop. I think in hindsight that the `abiCoder.encode()` call that's needed as part of computing a RelayData hash is probably very expensive, especially when coupled with the subsequent `keccak256()` call.

Before updating this I did alternatively consider whether we could pre-load the ABI format into an ethers interface, so it doesn't need to re-parse the format on each call. It seems like this specific use case isn't really served by ethers Interfaces, but I might have missed something.

bmzig (approved):

PR #494, opened 2024-01-15
chore: Temporarily import select v3 typechain defs

pxrl:
For preparing v3 SpokePoolClient functionality in advance of bumping contracts-v2.

These definitions should be removed as part of the eventual bump in contracts-v2.

fwiw, this is copied from the contracts-v2 typechain exports.


pxrl:
Test is currently failing due to an unrelated issue with Infura.

james-a-morris:
Could we create a `beta` version of the `contracts-v2` and import things in the same pattern as the rest of the typechain imports?

I think it would be good to be as close to the end build as possible

james-a-morris (commented):
Could we create a `beta` version of the `contracts-v2` and import things in the same pattern as the rest of the typechain imports?

I think it would be good to be as close to the end build as possible

pxrl:
> Could we create a `beta` version of the `contracts-v2` and import things in the same pattern as the rest of the typechain imports?
> 
> I think it would be good to be as close to the end build as possible

I was hoping to avoid having to deal with 2 concurrent versions of the contracts-v2 package, but maybe it's better to take that approach. I think we can achieve it with selective dependency resolutions in yarn: https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions

PR #495, opened 2024-01-16
feat: Introduce v3 types

pxrl:
This PR shuffles some existing types around to make room for the new v3
v3 types. It also introduces some convenience functions that will be
useful for discriminating between v2 and v3 types, as well as extracting
data from them. The v3 types are made available by this change, but are 
not imposed on any of the underlying implementation. No functional
change intended.

Fixes ACX-1671


src/utils/MigrationUtils.ts
```diff
@@ -0,0 +1,63 @@
+import { FillType, v2Deposit, v3Deposit, v2Fill, v3Fill, v2SpeedUp, v3SpeedUp } from "../interfaces";
```
nicholaspai, line 1: Is this file named `MigrationUtils` because it will be renamed or removed post-upgrade?

nicholaspai (approved):
Very nice +1. I agree with the naming of current functions as "v2" to make it more clear they are distinct from v3

src/utils/MigrationUtils.ts
james-a-morris, line 1: A lot of this will eventually be in the SDK, right? This logic will be directly relevant to the FE

james-a-morris (approved):
Just one comment on the migration utils. Otherwise, lgtm

src/utils/MigrationUtils.ts
pxrl: This is proposed for the SDK, but I had intended to remove it later.

If you need to retain it in the FE then it might make sense to name it something other than `MigrationUtils`, though it could also be relocated to the FE when we don't need it in the bots anymore.

src/utils/MigrationUtils.ts
pxrl: Yeah, it was my intention to remove this once we remove the v2/2.5 functions from the contracts (onchain). Per James' comment below it seems like it might be useful to retain some or all of the functionality, so I guess we can decide whether to find a better name for it, or drop it from the SDK and relocate to the FE at that point.

src/interfaces/SpokePool.ts
```diff
@@ -87,20 +167,65 @@ export type FillWithBlockStringified = Omit<
   updatableRelayData: RelayerRefundExecutionInfoStringified;
 };
 
-export interface SpeedUp {
+export type v3FillWithBlockStringified = Omit<
+  v3FillWithBlock,
+  "inoutAmount" | "outputAmount" | "updatableRelayData"
```
pxrl, line 172:
```suggestion
  "inputAmount" | "outputAmount" | "updatableRelayData"
```

src/utils/MigrationUtils.ts
nicholaspai: I think it might be better to rename it V3Utils or something

pxrl: Sure, will do.

pxrl: Done: https://github.com/across-protocol/sdk-v2/pull/495/commits/7901cb5c8b6082a70828192fd8c3c91cad6040f0

PR #496, opened 2024-01-16
chore(SpokePoolClient): Refactor before v3 introduction

pxrl:
This change makes some non-functional refactoring to the SpokePoolClient to make it easier to introduce the support for v3. Doing it now just reduces the size and complexity of the subsequent diff.


src/clients/SpokePoolClient.ts
```diff
@@ -664,25 +679,29 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+      // Filter out any early v2 deposits (quoteTimestamp > HubPoolClient.currentTime). Early deposits are no longer a
+      // critical risk in v3, so don't worry about filtering those. This will reduce complexity in several places.
+      const { earlyDeposits = [], v2DepositEvents = [] } = groupBy(
+        [
+          ...this.earlyDeposits,
+          ...((queryResults[eventsToQuery.indexOf("FundsDeposited")] ?? []) as FundsDepositedEvent[]),
+        ],
+        (depositEvent) => (this._isEarlyDeposit(depositEvent, currentTime) ? "earlyDeposits" : "v2DepositEvents")
+      );
+      if (earlyDeposits.length > 0) {
```
pxrl, line 691: One minor change here - we now log once for the sum of any early deposits, rather than once per deposit. The same information is available in the new consolidated log. Ultimately this will be removed once the v2 deposit functions are flushed; I anticipcate that we probably won't be concerned about early deposits in a v3-only world.

src/clients/SpokePoolClient.ts
```diff
@@ -404,17 +420,16 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-        : fillB.totalFilledAmount.lt(fillA.totalFilledAmount)
-        ? -1
-        : 0
-    );
+    const fillsOrderedByTotalFilledAmount = validFills.sort((fillA, fillB) => {
+      const totalFilledA = getTotalFilledAmount(fillA);
+      const totalFilledB = getTotalFilledAmount(fillB);
+
+      return totalFilledB.gt(totalFilledA) ? 1 : totalFilledB.lt(totalFilledA) ? -1 : 0;
+    });
```
pxrl, line 428: This change is added as a simple way to silence tsc warnings. In practice we should never see v3 fills here because they guarantee to fill to completion, but tsc doesn't know that.

src/clients/SpokePoolClient.ts
```diff
@@ -298,24 +305,33 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+        ...deposit,
+        speedUpSignature: maxSpeedUp.depositorSignature,
+        newRelayerFeePct: maxSpeedUp.newRelayerFeePct,
+        updatedRecipient: maxSpeedUp.updatedRecipient,
+        updatedMessage: maxSpeedUp.updatedMessage,
+      };
+
+      return updatedDeposit;
+    }
+    assert(false); // v3 is coming.
```
pxrl, line 334: A follow-up commit will swap this out for v3 speedups.

src/clients/SpokePoolClient.ts
```diff
@@ -664,25 +679,29 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-            currentTime,
-            deposit: { args, transactionHash },
-          });
-          return "earlyDeposits";
-        } else {
-          return "depositEvents";
-        }
-      });
+      // Filter out any early v2 deposits (quoteTimestamp > HubPoolClient.currentTime). Early deposits are no longer a
+      // critical risk in v3, so don't worry about filtering those. This will reduce complexity in several places.
```
nicholaspai, line 683: This is a really nice observation about early deposits in v3, just confirming that this is because the relayer no longer needs to set `realizedLpFeePct` correctly, only approximate it?

src/clients/SpokePoolClient.ts
nicholaspai: This is a great QoL improvement, the logs were really busy

nicholaspai (approved):
Love seeing these changes, also like these changes being small enough to review and building upon previous PR's. +1

src/clients/SpokePoolClient.ts
```diff
@@ -298,24 +305,33 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-    );
+    const { depositId, depositor } = deposit;
 
-    // We assume that the depositor authorises SpeedUps in isolation of each other, which keeps the relayer
-    // logic simple: find the SpeedUp with the highest relayerFeePct, and use all of its fields
-    if (!maxSpeedUp || maxSpeedUp.newRelayerFeePct.lte(deposit.relayerFeePct)) {
-      return deposit;
-    }
+    if (isV2Deposit(deposit)) {
+      const v2SpeedUps = (this.speedUps[depositor]?.[depositId]?.filter(() => isV2SpeedUp) ?? []) as v2SpeedUp[];
```
james-a-morris, line 311:
```suggestion
      const v2SpeedUps = (this.speedUps[depositor]?.[depositId]?.filter(isV2SpeedUp) ?? []) as v2SpeedUp[];
```
Do you think this will work as it stands? We're not passing anything from the filter & we're not invoking the function.

james-a-morris (commented):
One comment related to filtering values.

src/clients/SpokePoolClient.ts
pxrl: Yeah - the relayer only needs to know `realizedLpFeePct` for its internal profitability logic. Handling of "early" deposits clutters the code in a few places, so I'll advocate to remove it once `deposit()` has been removed.

PR #497, opened 2024-01-17
improve(findFillBlock): Source destinationChainId from SpokePool

pxrl:
This is transparently compatible with v3 RelayData, which does not directly include the destinationChainId.

### DIFF
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -205,7 +205,7 @@ export async function findFillBlock(
   const { provider } = spokePool;
   highBlockNumber ??= await provider.getBlockNumber();
   assert(highBlockNumber > lowBlockNumber, `Block numbers out of range (${lowBlockNumber} > ${highBlockNumber})`);
-  const { chainId: destinationChainId } = await spokePool.provider.getNetwork();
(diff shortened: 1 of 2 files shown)


src/utils/SpokeUtils.ts
```diff
@@ -205,6 +205,7 @@ export async function findFillBlock(
   const { provider } = spokePool;
   highBlockNumber ??= await provider.getBlockNumber();
   assert(highBlockNumber > lowBlockNumber, `Block numbers out of range (${lowBlockNumber} > ${highBlockNumber})`);
+  const { chainId: destinationChainId } = await spokePool.provider.getNetwork();
```
nicholaspai, line 208: You could also call spokePool.chainId() I believe, which might be less provider-opinionated

pxrl: I think it's preferable to use `provider.getNetwork()` because for a StaticJsonRpcProvider instance, the chainId is served directly out of the provider and never results in an external RPC call (nb. the `await` statement is misleading in this sense). Calling `SpokePool.chainId()` would instead result in an external call.

pxrl, line 208:
```suggestion
  const { chainId: destinationChainId } = await provider.getNetwork();
```

PR #498, opened 2024-01-17
feat: Support querying v3 RelayData hash

pxrl:
Upgrade the various RelayData-dependent functions to work with v3 types. This will importantly be used for slow fill requests in the pending v3 bump.


nicholaspai (approved):
Looks correct

src/utils/SpokeUtils.ts
```diff
@@ -1,8 +1,11 @@
 import assert from "assert";
-import { getRelayHash } from "@across-protocol/contracts-v2/dist/test-utils";
-import { BigNumber, Contract } from "ethers";
-import { RelayData } from "../interfaces";
+import { getRelayHash as _getV2RelayHash } from "@across-protocol/contracts-v2/dist/test-utils";
```
james-a-morris, line 2: We may want to avoid these `dist` imports. Is there a way we can avoid this type of import?

pxrl: This is a good prompt and I'm 100% onboard for that. We should refocus on using more from contracts-v2 rather than defining duplicates in sdk-v2, as we do now (i.e. a lot of the types). I think the reason for the status quo that we have at the moment is that the contracts exports are contaminated with hardhat exports that are used internally for test, so we'd need to separate those into test-specific functionalities and export-focused functionalities.

One additional issue - we can't get any benefit from this right now because contracts-v2 has already moved on past where we could update to. We'll have to wait for the v2 -> v3 migration to occur. I've created ACX-1826 to track this, though.

pxrl: I think its correct home is in contracts-v2 - especially since we use it over there for test purposes. It's just not practical to do that now though, so for the time being we need a local copy. This is also necessary because I've found that importing from contracts-v2/dist subtly breaks some functionality in the relayer. PR to fix it here: https://github.com/across-protocol/sdk-v2/pull/505.

src/utils/SpokeUtils.ts
```diff
@@ -190,12 +206,68 @@ export function relayFilledAmount(
...
+      [
+        "tuple(" +
+          "address depositor," +
+          "address recipient," +
+          "address exclusiveRelayer," +
+          "address inputToken," +
+          "address outputToken," +
+          "uint256 inputAmount," +
+          "uint256 outputAmount," +
+          "uint256 originChainDI," +
```
pxrl, line 230:
```suggestion
          "uint256 originChainId," +
```

PR #501, opened 2024-01-19
feat: Unify v2 & v3 types

pxrl:
This PR rolls out v3 types everywhere by unifying them under the Deposit, Fill, ... types that are handled throughout the SDK and relayer codebases. With this change, all consumers of these datatypes must have implemented specific handling to discriminate between v2 and v3 events. In many cases these objects are passed transparently though functions because they are not evaluating any of the fields which are different between the two. tsc is very helpful to identify which codepaths need to be upgraded. Much of that work has occurred in advance of this change.


src/clients/HubPoolClient.ts
```diff
@@ -224,12 +228,12 @@ export class HubPoolClient extends BaseAbstractClient {
    * @param deposit Deposit event
    * @param returns string L1 token counterpart for Deposit
    */
-  getL1TokenForDeposit(deposit: Pick<DepositWithBlock, "quoteBlockNumber" | "originToken" | "originChainId">): string {
+  getL1TokenForDeposit(deposit: Pick<v3DepositWithBlock, "originChainId" | "inputToken" | "quoteBlockNumber">): string {
```
nicholaspai, line 231: How come we don't need a `getL1TokenForV2Deposit` AND a `getL1TokenForV3Deposit`?

src/clients/UBAClient/UBAClientUtilities.ts
```diff
@@ -668,18 +680,20 @@ export async function refundRequestIsValid(
   // refund was sent.
   const fill = destSpoke.getFillsForRelayer(relayer).find((fill) => {
     // prettier-ignore
```
pxrl, line 682: note to self - delete this.

src/clients/mocks/MockSpokePoolClient.ts
pxrl, line 1: nb. The changes in this file probably belong in branch pxrl/types-prep.

src/clients/SpokePoolClient.ts
```diff
@@ -331,7 +344,29 @@ export class SpokePoolClient extends BaseAbstractClient {
 
       return updatedDeposit;
     }
-    assert(false); // v3 is coming.
+
+    const v3SpeedUps = (this.speedUps[depositor]?.[depositId]?.filter(() => isV3SpeedUp) ?? []) as v3SpeedUp[];
```
nicholaspai, line 348: Do you think its worth storing v3 and v2 speedups separately in memory? Just a thought, im not opinionated

pxrl: The v2 | v3 union setup seems to work OK with speedups so far, so unless that proves to be flawed or difficult then I think we're OK to leave them consolidated.

nicholaspai: ok, i don' t think its a big decision

src/clients/UBAClient/UBAClientUtilities.ts
pxrl: Done: https://github.com/across-protocol/sdk-v2/pull/501/commits/f7cc09614b72bcf31430469d6d25cf71fbe766b1.

src/clients/SpokePoolClient.ts
```diff
@@ -710,17 +765,23 @@ export class SpokePoolClient extends BaseAbstractClient {
 
       const dataForQuoteTime = await this.batchComputeRealizedLpFeePct(depositEvents);
       for (const [index, event] of Array.from(depositEvents.entries())) {
-        const rawDeposit = spreadEventWithBlockNumber(event) as DepositWithBlock;
-
-        // Derive and append the destination token, LP fee, quote block number and block timestamp from the event.
-        // @dev Deposit events may _also_ include early deposits, in which case we did not retrieve a block.
-        const deposit: DepositWithBlock = {
-          ...rawDeposit,
-          realizedLpFeePct: dataForQuoteTime[index].realizedLpFeePct,
```
nicholaspai, line 719: Unfortunately I think we're still going to want to compute the realizedLpFeePct but we're going to want to associate them with the `Fill`. Or, we could leave this part in. This is to help aid the Dataworker, which needs to add an lpFee for each Fill that it is refunding, and probably also the relayer, which needs to compute the lpFee to determine profitability.

So I think we might want to keep this in here. This is something I'm realizing while scoping out the dataworker work.

We either need to compute realizedLpFeePct here, or in the dataworker step when constructing relayer refund roots

src/utils/FlowUtils.ts
```diff
@@ -74,13 +138,23 @@ export function validateFillForDeposit(fill: Fill, deposit?: Deposit, fillFields
...
+      if (fillFieldsToIgnore.includes(key)) {
+        return true;
+      }
+      return fill[key] !== undefined && fill[key].toString() === deposit[key]?.toString();
+    });
+  }
+
+  if (isV3Deposit(deposit) && isV3Fill(fill)) {
+    return V3_DEPOSIT_COMPARISON_KEYS.every((key) => {
+      if (fillFieldsToIgnore.includes(key)) {
```
nicholaspai, line 152: We will need to add logic (in this PR or a follow up) to account for when `outputToken = 0x0` on the `Deposit`, then the `outputToken` must be the canonical destination token on the `Fill` and the `outputToken` will technically not match but will still be valid.

nicholaspai: However, we also should be able to use this function to validate SlowFillRequests with Deposits in which case the slow fill request CAN specify `outputToken = 0x0` so the `outputTokens` will match in this case

pxrl: I've updated the SpokePool so that it overwrites outputToken 0x0 immediately on ingestion. This is basically doing that at the lowest possible level and should save everyone else from having to worry about it :+1:.

src/clients/SpokePoolClient.ts
pxrl: Agree on keeping realizedLpFeePct - I might be a bit behind with this remote branch, but in the local state I have I am setting realizedLpFeePct unconditionally. There's actually no reason not to do it - we have the information available anyway.

src/utils/FlowUtils.ts
nicholaspai: - A slow fill is valid iff `deposit.outputToken = requestSlowFill.outputToken` && `outputToken = <canonical destination token for deposit.inputToken> OR `deposit.inputToken = 0x0` and `fill.outputToken = <canonical destination token for deposit.inputToken>`.
- A fast fill is valid iff `deposit.outputToken == fill.outputToken` OR `deposit.inputToken = 0x0` and `fill.outputToken = <canonical destination token for deposit.inputToken>`

src/clients/HubPoolClient.ts
pxrl: I thought I could get away with just normalising everything to v3 types (as was done in the SpokePoolClient), but this blew up in relayer-v2. I've subsequently updated this to accept both types. It's pretty ugly because it's a union of `Pick<...>` statements, and ti has to be exported to relayer-v2 since we override this method there as well. However, it's only going to be short-lived so I'm OK with it.

src/clients/mocks/MockSpokePoolClient.ts
pxrl: Implemented: https://github.com/across-protocol/sdk-v2/pull/495/commits/06fc1ff8f484f1c854f8faefe5de0fd842366988.

src/utils/FlowUtils.ts
```diff
@@ -74,13 +138,23 @@ export function validateFillForDeposit(fill: Fill, deposit?: Deposit, fillFields
...
-  // Note: this short circuits when a key is found where the comparison doesn't match.
-  // TODO: if we turn on "strict" in the tsconfig, the elements of FILL_DEPOSIT_COMPARISON_KEYS will be automatically
-  // validated against the fields in Fill and Deposit, generating an error if there is a discrepency.
-  return FILL_DEPOSIT_COMPARISON_KEYS.every((key) => {
-    if (fillFieldsToIgnore.includes(key)) {
-      return true;
-    }
-    return fill[key] !== undefined && fill[key].toString() === deposit[key]?.toString();
-  });
+  if (isV2Deposit(deposit) && isV2Fill(fill)) {
```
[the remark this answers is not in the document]
pxrl: Yep! This is one of the advantages of passing all types around as a union of v2 and v3 - we inherit a lot of the existing functionality for free without having to duplicate it, and tsc is pretty helpful in identifying where things obviously don't work. There's a delta that we need to verify via test, but it should be overall less work and less risk than having to duplicate a bunch of functionality.

src/clients/SpokePoolClient.ts
pxrl: It's updated now. There's still a missing `relayerFeePct`. We have the option of computing it here, but that only works for like tokens, so I feel like it should be determined outside of the SpokePoolClient. It's only relevant to the relayer anyway, so it might slot into the ProfitClient instead (where we already have most of the tools to normalise to common prices anyway).

diff

src/utils/FlowUtils.ts
pxrl: Per https://github.com/across-protocol/sdk-v2/pull/501#discussion_r1464828125, the SpokePoolClient subs in the right outputToken in the case that the deposit set it to 0x0 :+1:.

src/clients/SpokePoolClient.ts
nicholaspai: I agree its only relevant for the relayer. I think ProfitCLient makes sense!

src/clients/HubPoolClient.ts
pxrl: Resolved here: https://github.com/across-protocol/sdk-v2/pull/501/commits/8884af524de2eac197de32348b7696e55e5e39a4

I was able to simplify quite a bit by taking advantage of some of the new utility functions in V3Utils.

src/interfaces/SpokePool.ts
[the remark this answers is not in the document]
pxrl: Stringified types have been removed, so this is no longer an issue.

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -122,6 +122,16 @@ describe("SpokePoolClient: Fill Validation", function () {
     expect(filled.eq(deposit.amount)).is.true;
   });
 
+  it.skip("Tracks v3 fill status", async function () {
```
pxrl, line 125: Flagging a todo - this should be resolved in this PR.

test/SpokePoolClient.v3Events.ts
pxrl, line 1: Flagging a todo - add a test for Filledv3SlowRelay events.

src/interfaces/SpokePool.ts
pxrl, line 1: Flagging a todo - missing types for v3 RelayerRefundLeaves.

src/clients/SpokePoolClient.ts
```diff
@@ -822,14 +903,22 @@ export class SpokePoolClient extends BaseAbstractClient {
...
     }
 
-    const deposits = depositEvents.map(({ args, blockNumber }) => {
+    const deposits = depositEvents.map((event) => {
+      let inputToken: string, inputAmount: BigNumber;
+
+      if (this.isV3DepositEvent(event)) {
+        ({ inputToken, inputAmount } = event.args);
+      } else {
+        // Coerce v2 deposit objects into v3 format.
```
pxrl, line 912: This needs a dedicated test to ensure that v2 deposits and v3 deposits are handled the same by batchComputeRealizezdLpFeePct(), and also that the translation into a v3 object is handled correctly.

pxrl:
I've shaved ~200 lines off this PR by proposing some of the changes separately. I've also identified a few todos that need to be resolved before this can be merged. I'm going to throw it open for general comments now.

src/interfaces/SpokePool.ts
```diff
@@ -47,8 +47,8 @@ export interface v3DepositWithBlock extends v3Deposit, SortableEvent {
   quoteBlockNumber: number;
 }
 
-export type Deposit = v2Deposit; // @todo: Extend with v2Deposit | v3Deposit.
-export type DepositWithBlock = v2DepositWithBlock; // @todo Extend with v2DepositWithBlock | v3DepositWithBlock.
+export type Deposit = v2Deposit | v3Deposit;
+export type DepositWithBlock = v2DepositWithBlock | v3DepositWithBlock;
```
nicholaspai, line 51: Brought up offline but wdyt about moving `realizedLpFeePct` into the v2Deposit type and out of DepositCommon since we won't be able to compute it upfront until we match a fill with a deposit once we implement https://linear.app/uma/issue/ACX-1838/update-lp-fee-computation-to-replace-destinationchainid-with

src/interfaces/SpokePool.ts
```diff
@@ -230,8 +230,7 @@ export interface v3RelayData extends RelayDataCommon {
   exclusivityDeadline: number;
 }
 
-// @todo: Extend with v2RelayData | v3RelayData.
-export type RelayData = v2RelayData;
+export type RelayData = v2RelayData | v3RelayData;
```
nicholaspai, line 233: Should we merge the v3RelayData and Fill objects somehow? At the contract level, the following functions all take in a `V3RelayData` object:
- `fillV3Relay`
- `fillV3RelayWithUpdatedDeposit`
- `requestV3SlowFill`
- `executeV3SlowRelayLeaf`
- `_getV3RelayHash`

src/clients/SpokePoolClient.ts
```diff
@@ -706,19 +770,34 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-        const rawFill = spreadEventWithBlockNumber(event) as FillWithBlock;
-        const fill: FillWithBlock = {
-          ...rawFill,
-        };
+        const fill = this.isV3FillEvent(event)
+          ? { ...(spreadEventWithBlockNumber(event) as v3FillWithBlock) }
+          : { ...(spreadEventWithBlockNumber(event) as v2FillWithBlock) };
 
         assign(this.fills, [fill.originChainId], [fill]);
         assign(this.depositHashesToFills, [this.getDepositHash(fill)], [fill]);
```
nicholaspai, line 803: For v3, should we be storing fills according to the relay hash? At the contract level, v3 deposits, fills, slow fills all use the same V3RelayData to identifiy bridge txns

src/utils/FlowUtils.ts
```diff
@@ -1,28 +1,59 @@
...
+      fillA.amount.eq(fillB.amount) &&
+      fillA.destinationChainId === fillB.destinationChainId &&
+      fillA.relayerFeePct.eq(fillB.relayerFeePct) &&
+      fillA.recipient === fillB.recipient &&
+      fillA.depositor === fillB.depositor &&
+      fillA.message === fillB.message
+    );
+  } else if (isV3Fill(fillA) && isV3Fill(fillB)) {
+    return (
+      fillA.depositId === fillB.depositId &&
```
nicholaspai, line 43: We also need to compare `exclusiveRelayer`, `fillDeadline`, `exclusivityDeadline`. Basically everything in `V3RelayData`. So should we use `getV3RelayHash` to compare?

nicholaspai (commented):
This looks really useful, have left some comments about trying to condense the keys that we can use to compare deposits, fills, slow fills, and speed up requests for V3 events.

test/utils/utils.ts
```diff
@@ -303,27 +309,55 @@ export async function buildDepositStruct(
     blockNumber: await getLastBlockNumber(),
   };
 }
+
+// export function buildV3Deposit(
```
nicholaspai, line 313: building a v3 deposit struct will be very useful when integrating into relayer-v2, did you just want to add in a followup?

src/utils/common.ts
```diff
@@ -302,6 +303,8 @@ export function createUnsignedFillRelayTransactionFromDeposit(
   amountToFill: BN,
   relayerAddress: string
 ): Promise<PopulatedTransaction> {
+  assert(isV2Deposit(deposit));
```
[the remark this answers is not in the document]
pxrl: Safety + hinting to tsc that it can narrow the types. It will be removed after the migration and v2 deprecation.

test/utils/utils.ts
pxrl: Yeah - we'll need the new ABI for that so it's not included here, but it will follow shortly.

src/utils/FlowUtils.ts
pxrl: That's a good proposal and would make life easier when comparing deposits or fills. I'll implement that.

src/utils/DepositUtils.ts
```diff
@@ -2,12 +2,13 @@ import assert from "assert";
 import { SpokePoolClient } from "../clients";
 import { DEFAULT_CACHING_TTL, EMPTY_MESSAGE } from "../constants";
 import { CachingMechanismInterface, Deposit, DepositWithBlock, Fill } from "../interfaces";
-import { getNetworkName } from "../utils";
```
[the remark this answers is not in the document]
pxrl: Yeah - we've had issues in the past when importing from the directory where a file is. It wasn't causing any issues in this case, but all of the other imports from this dir in this file are sourced from the specific files.

src/utils/DepositUtils.ts
```diff
@@ -129,7 +130,14 @@ export function isMessageEmpty(message = EMPTY_MESSAGE): boolean {
...
 export function isDepositSpedUp(deposit: Deposit): boolean {
-  return isDefined(deposit.speedUpSignature) && isDefined(deposit.newRelayerFeePct);
+  if (!isDefined(deposit.speedUpSignature)) {
+    return false;
+  }
+
+  return (
+    (isV2Deposit(deposit) && isDefined(deposit.newRelayerFeePct)) ||
+    (isV3Deposit(deposit) && isDefined(deposit.updatedOutputAmount))
+  );
```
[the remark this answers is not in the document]
pxrl: Implemented here: https://github.com/across-protocol/sdk-v2/pull/501/commits/981fae04f60f3ae3d3f306efdc23fe9cb5b14240

I had to overrule prettier because it kept wanting to make the code look dumb :(

src/clients/SpokePoolClient.ts
```diff
@@ -628,7 +682,7 @@ export class SpokePoolClient extends BaseAbstractClient {
     // new deposits that were found in the searchConfig (new from the previous run). This is important as this operation
     // is heavy as there is a fair bit of block number lookups that need to happen. Note this call REQUIRES that the
     // hubPoolClient is updated on the first before this call as this needed the the L1 token mapping to each L2 token.
-    if (eventsToQuery.includes("FundsDeposited")) {
+    if (eventsToQuery.includes("FundsDeposited") || eventsToQuery.includes("V3FundsDeposited")) {
```
nicholaspai, line 685: Do we have to add `V3FundsDeposited` to `queryableEventNames`?

test/SpokePoolClient.ValidateFill.ts
pxrl: Update - this can't be resolved within this PR because this test queries the SpokePool contract directly via `buildFill()` and `relayFilledAmount()`. We need to bump contracts-v2 before we're able to do that.

pxrl:
> This is great work! I think we should stick with the prevailing stylization and use Upper Camel Case for all of our interfaces.
> 
> DeFacto Stylization guide (by Google): https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type

Good call - I've proposed to update the naming in this PR: https://github.com/across-protocol/sdk-v2/pull/528.

src/clients/SpokePoolClient.ts
pxrl: I investigated this and it's not really achievable or useful until we bump contracts-v2. We can survive without it when using the MockSpokePoolClient because it overrides the SpokePoolClient._update() method, which is where we use these filters.

test/utils/utils.ts
pxrl: PR for adding the new ABI here: https://github.com/across-protocol/sdk-v2/pull/526

But since we also need to deploy and run the contracts in a hardhat environment, we'll also need a mocked SpokePool instance (or a separate copy of the updated contracts-v2 repo).

src/clients/SpokePoolClient.ts
```diff
@@ -660,16 +713,23 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-          ...rawDeposit,
-          realizedLpFeePct: dataForQuoteTime[index].realizedLpFeePct,
-          destinationToken: this.getDestinationTokenForDeposit(rawDeposit),
-          quoteBlockNumber: dataForQuoteTime[index].quoteBlock,
-        };
+        const rawDeposit = spreadEventWithBlockNumber(event);
+        let deposit: DepositWithBlock;
+
+        if (this.isV3DepositEvent(event)) {
+          deposit = { ...(rawDeposit as v3DepositWithBlock) };
```
mrice32, line 720: Out of scope of the v3 change, but I would really like to have a system where the spread event functions preserve type information. But this is a larger change and not specific to v3.

mrice32 (commented):
Looks great! A few comments on how we do types. Given how tricky some of these v2 and v3 types are, I think a good goal should be to "lie" less to typescript.

src/clients/SpokePoolClient.ts
pxrl: Yeah I definitely agree on this. The current method forces us to add a lot of type assertions.

I would FWIW be in favour of just supporting each event type individually by unpacking via an Ethers interface, rather than the coercion that we currently do.

src/clients/SpokePoolClient.ts
```diff
@@ -305,7 +317,29 @@ export class SpokePoolClient extends BaseAbstractClient {
 
       return updatedDeposit;
     }
-    assert(false); // v3 is coming.
+
+    const v3SpeedUps = (this.speedUps[depositor]?.[depositId]?.filter(() => isV3SpeedUp) ?? []) as v3SpeedUp[];
+    const maxSpeedUp = v3SpeedUps.reduce(
+      (prev, current) => (prev.updatedOutputAmount.lt(current.updatedOutputAmount) ? prev : current),
+      { updatedOutputAmount: deposit.outputAmount } as v3SpeedUp
```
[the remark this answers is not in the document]
pxrl: Yep, good catch! There was some outdated code in the V3 part that had previously been resolved in the V2 SpeedUp handling, but it was also an error to supply an initial value for the reduce loop. Having removed that, tsc is now happy and there are no type assertions in this code.

Change: https://github.com/across-protocol/sdk-v2/pull/501/commits/3d64f09fd9fcb557da88d2e5ebad0d536facddfb

src/clients/HubPoolClient.ts
```diff
@@ -224,12 +241,16 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    deposit:
+      | Pick<v2DepositWithBlock, "originChainId" | "originToken" | "quoteBlockNumber">
+      | Pick<v3DepositWithBlock, "originChainId" | "inputToken" | "quoteBlockNumber">
+  ): string {
     // L1-->L2 token mappings are set via PoolRebalanceRoutes which occur on mainnet,
     // so we use the latest token mapping. This way if a very old deposit is filled, the relayer can use the
     // latest L2 token mapping to find the L1 token counterpart.
-
-    return this.getL1TokenForL2TokenAtBlock(deposit.originToken, deposit.originChainId, deposit.quoteBlockNumber);
+    const inputToken = getDepositInputToken(deposit as DepositWithBlock);
```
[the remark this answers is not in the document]
pxrl: This is a good call and it is definitely a weakness in the current approach. The issue arises partly due to the fact that `getL1TokenForDeposit()` requires only a partial type, and `getDepositInputToken()` expects a complete type. If the implementation of `getDepositInputToken()` were to change (i.e. if it resolved true or false based on a different member of the `Deposit` type then it could incorrectly make the wrong conclusion. I've implicitly accepted those risks in this PR, but it is a definite risk and if you have a suggestion on how to remove these sharp edges then I'm all ears.

src/interfaces/SpokePool.ts
pxrl: Yep, I like this. Have done some initial work towards it here: https://github.com/across-protocol/sdk-v2/commit/e7d67a50175884f70fe3d98cd7caf7d6457cd6c6

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
mrice32: I'm putting together some generic type guards to see if we can make typescript give us these guarantees. Will link once I have an example!

src/clients/SpokePoolClient.ts
pxrl: Yeah, we need to bump contracts-v2 here before we can use it here.

src/utils/common.ts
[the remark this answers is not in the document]
pxrl: This is no longer a part of this PR btw, so I'll mark it resolved.

src/clients/SpokePoolClient.ts
nicholaspai: matt thanks for the push on this, i think unpacking events individually would be useful here as well as a utility function that takes in an object, an expected type, and then unpacks it prop by prop

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
pxrl: Since this is now tracked in a separate PR I'll mark this thread as resolved.

PR: https://github.com/across-protocol/sdk-v2/pull/537

src/interfaces/SpokePool.ts
```diff
@@ -51,16 +50,20 @@ export interface V2DepositWithBlock extends V2Deposit, SortableEvent {
 export interface V3Deposit extends V3RelayData {
   destinationChainId: number;
   quoteTimestamp: number;
-  updatedOutputAmount?: BigNumber;
+  realizedLpFeePct?: BigNumber;
```
pxrl, line 53: For this commit to go in, we'll leave realizedLpFeePct defined here. A follow-up change will remove it, and the consumers of the realizedLpFeePct on v3 events will have to compute it as needed.

src/clients/SpokePoolClient.ts
```diff
@@ -283,10 +301,11 @@ export class SpokePoolClient extends BaseAbstractClient {
 
     if (isV2Deposit(deposit)) {
       const v2SpeedUps = this.speedUps[depositor]?.[depositId]?.filter(isV2SpeedUp);
-      const maxSpeedUp = v2SpeedUps?.reduce(
-        (prev, current) => (prev.newRelayerFeePct.gt(current.newRelayerFeePct) ? prev : current),
-        { newRelayerFeePct: deposit.relayerFeePct } as V2SpeedUp
-      );
+      const maxSpeedUp = v2SpeedUps?.reduce((prev, current) => {
+        assert(isV2SpeedUp(prev) && isV2SpeedUp(current)); // tsc hinting.
```
mrice32, line 305: Would this not work to convince typescript that these are the input and output types? And if not, what is its complaint?

```ts
const maxSpeedUp = v2SpeedUps?.reduce((prev: V2SpeedUp, current: V2SpeedUp):  V2SpeedUp => {
...
```

src/clients/SpokePoolClient.ts
```diff
@@ -305,7 +324,30 @@ export class SpokePoolClient extends BaseAbstractClient {
...
       return updatedDeposit;
     }
-    assert(false); // v3 is coming.
+
+    const v3SpeedUps = this.speedUps[depositor]?.[depositId]?.filter(isV3SpeedUp);
+    const maxSpeedUp = v3SpeedUps?.reduce((prev, current) => {
+      assert(isV3SpeedUp(prev) && isV3SpeedUp(current)); // tsc hinting.
+      return prev.updatedOutputAmount.lt(current.updatedOutputAmount) ? prev : current;
+    });
+    assert(!isDefined(maxSpeedUp) || isV3SpeedUp(maxSpeedUp)); // tsc hinting.
```
mrice32, line 333: Similar to the above comment, I think we should be able to get the right types if we help typescript know the input and output types of the function passed to reduce, right? It should infer

test/SpokePoolClient.v3Events.ts
pxrl, line 1: Just to put the size of this PR into context, this test file accounts for about 50% of the added code.

src/clients/SpokePoolClient.ts
```diff
@@ -617,6 +670,20 @@ export class SpokePoolClient extends BaseAbstractClient {
     return depositEvent.args.quoteTimestamp > currentTime;
   }
 
+  // Temporary type discriminator for v2 -> v3 transition.
+  protected isV3DepositEvent(event: FundsDepositedEvent | V3FundsDepositedEvent): event is V3FundsDepositedEvent {
+    const { event: eventName } = event;
+    assert(isDefined(eventName) && ["FundsDeposited", "V3FundsDeposited"].includes(eventName));
```
mrice32, line 676: Why assert? Don't the FundsDepositedEvent and V3FundsDepositedEvent types already have the eventName field?

src/clients/SpokePoolClient.ts
```diff
@@ -617,6 +670,20 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    return event.event === "V3FundsDeposited";
+  }
+
+  // Temporary type discriminator for v2 -> v3 transition.
+  protected isV3FillEvent(event: FilledRelayEvent | FilledV3RelayEvent): event is FilledV3RelayEvent {
+    const { event: eventName } = event;
+    assert(isDefined(eventName) && ["FilledRelay", "FilledV3Relay"].includes(eventName));
```
mrice32, line 683: Same question here

mrice32 (commented):
Added some comments on the use of asserts for ts hinting.

I think that practice is _okay_ (better than casting), but if we can fix it through explicit function types, I think that's preferable. If we force the types on the reduce function, for instance, typescript might show an error that reveals that there is a case we aren't covering in our type assumptions.

src/clients/SpokePoolClient.ts
pxrl: Resolved here: https://github.com/across-protocol/sdk-v2/pull/501/commits/c3b6cb7cd50ad315a53ff5a642aa0c9dbd910ff4

Thanks for the insights on this!!

src/clients/SpokePoolClient.ts
pxrl: As above: https://github.com/across-protocol/sdk-v2/pull/501#discussion_r1481529634

src/clients/SpokePoolClient.ts
pxrl: As above :+1:

src/utils/FlowUtils.ts
```diff
@@ -1,28 +1,62 @@
...
+  "relayerFeePct",
+  "realizedLpFeePct",
+] as const;
+
+export const V3_DEPOSIT_COMPARISON_KEYS = [
+  ...FILL_DEPOSIT_COMPARISON_KEYS,
+  "inputToken",
+  "inputAmount",
+  "outputToken",
+  "outputAmount",
```
nicholaspai, line 26: Missing `exclusiveRelayer`, `exclusivityDeadline`, `fillDeadline`

test/SpokePoolClient.ValidateFill.ts
nicholaspai, line 125: did you mean to leave this as skip-ed?

nicholaspai (approved):
Latest round of commits looks good, ptal at my comments

test/SpokePoolClient.ValidateFill.ts
pxrl: Yeah - we need to bump contracts before we can test this because it queries the SpokePool contract directly (bypassing the SpokePoolClient).

src/utils/FlowUtils.ts
pxrl: Good catch - https://github.com/across-protocol/sdk-v2/pull/501/commits/0d0018c2287975292d35b601eb341b6e70a2bcb8

src/clients/SpokePoolClient.ts
pxrl: Definition here: https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/contracts/src.ts/index.ts#L78

mrice32: As mentioned offline, I sort of think it might be cleaner and easier to just use the unique fields of these types to distinguish rather than the event name, which relies on an assertion.

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
pxrl: Resolved: https://github.com/across-protocol/sdk-v2/pull/501/commits/6953b372e9d237b4a64b67b46b08a828711f4c54

src/clients/SpokePoolClient.ts
pxrl: Resolved: https://github.com/across-protocol/sdk-v2/pull/501/commits/6953b372e9d237b4a64b67b46b08a828711f4c54

src/clients/SpokePoolClient.ts
pxrl: Resolved: https://github.com/across-protocol/sdk-v2/pull/501/commits/3e260c134a0f9e49e0cd986bf3e708e232260c88

src/clients/SpokePoolClient.ts
pxrl: Further update: https://github.com/across-protocol/sdk-v2/pull/501/commits/3e260c134a0f9e49e0cd986bf3e708e232260c88

src/interfaces/SpokePool.ts
pxrl: We'll tackle this in a follow up commit :+1:

PR #503, opened 2024-01-23
refactor: Upstream HubPoolClient deposit quoteTimestamp check

pxrl:
Relocating this from the relayer-v2 HubPoolClient class to ease v3 integrations.


pxrl:
fwiw the current relayer-v2 implementation is here: https://github.com/across-protocol/relayer-v2/blob/9806bf5b83f72e82416fe0a654edde2ef5fc7d4a/src/clients/HubPoolClient.ts#L41.

pxrl:
This one is also difficult to upstream for annoying reasons in test. Will abandon it for now - it'll be easier once we're over the v3 hump and can remove early quoteTimestamp detection.

PR #504, opened 2024-01-23
refactor: Upstream SpokePoolClient early quoteTimestamp detection

pxrl:
This is relocated from the relayer-v2 SpokePoolClient implementation. It's preferable to centralise the logic here in advance of its eventual removal of early deposit detection.

The the proposer will still need to be conscious of future quoteTimestamps to avoid computing fees on early deposits. Alternatively, an UMIP change could be made to permit the proposer to clamp the quoteTimestamp at the toBlock in the relevant proposal.


pxrl:
This turns out to be hard to pull off for some really annoying reasons. Abandoning it for now, since it's easier to just live with the status quo.

PR #506, opened 2024-01-23
feat: Add oldest time searched to SpokePoolClient

nicholaspai:
These will be useful for the dataworker functions (beginning to be implemented in https://github.com/across-protocol/relayer-v2/pull/1155 ) that have to deal with expired deposits. We need to be able to know when a deposit is expired (comparing its fill deadline TIMESTAMP with a bundle BLOCK NUMBER), and also whether our spoke pool clients have enough of a lookback to fetch expired deposits for a bundle

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -583,7 +583,7 @@ export class SpokePoolClient extends BaseAbstractClient {
     const [numberOfDeposits, currentTime, oldestTime, ...events] = await Promise.all([
       this.spokePool.numberOfDeposits({ blockTag: searchConfig.toBlock }),
       this.spokePool.getCurrentTime({ blockTag: searchConfig.toBlock }),
-      this.spokePool.getCurrentTime({ blockTag: searchConfig.fromBlock }),
+      this.spokePool.getCurrentTime({ blockTag: Math.max(searchConfig.fromBlock, this.deploymentBlock) }),
       ...eventSearchConfigs.map((config) => paginatedEventQuery(this.spokePool, config.filter, config.searchConfig)),
     ]);
     this.log("debug", `Time to query new events from RPC for ${this.chainId}: ${Date.now() - timerStart} ms`);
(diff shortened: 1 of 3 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -573,9 +577,13 @@ export class SpokePoolClient extends BaseAbstractClient {
     });
 
     const timerStart = Date.now();
-    const [numberOfDeposits, currentTime, ...events] = await Promise.all([
+    // TODO: Once the SpokePools are updated and have the `fillDeadlineBuffer()` method, load it dynamically here.
+    const fillDeadlineBuffer = 2 * 60 * 60; // 2 hours. We're assuming this is hardcoded to the same value that the
```
mrice32, line 581: What happens if the fill deadline buffer changes?

There may be nothing we can do to guard against that other than make sure we add a cushion here and don't change it dramatically at any one time.

mrice32 (approved):
Looks great -- just one question!

pxrl (approved):

james-a-morris (commented):
LGTM - but we should add these variables in the `toJSON` and `fromJSON` serialization functions.

src/clients/SpokePoolClient.ts
nicholaspai: I don't know how to handle it since the `fillDeadlineBuffer` is a constant/immutable in the contract so we don't have an easy to track its updates. We could try to track Upgraded events but its not super easy. I also think the value here doesn't need to align with the contracts exactly, as long as its no smaller.

nicholaspai: I'll add a comment here to that^

nicholaspai: thoughts on this treatment? 2a3963097a3cbd56ee1d0e9349e33a5456fde5fc

nicholaspai: The latest implementation now removes `fillDeadlineBuffer` from being loaded by the `SpokePoolClient`. Instead, we load it at the `dataworker` time by querying `spokePool.fillDeadlineBuffer({ blockTag: start|end })` and taking the maximum of the value at the start and end of a block range. 

(We assume this value doesn't change in the middle of a bundle, which is a fair assumption 99% of the time and even if its not correct, probably still protects us if the admin updating `fillDeadlineBuffer` is reasonable.

https://github.com/across-protocol/relayer-v2/blob/017f51a833d4f5504522b6e546903a0ad912804d/src/dataworker/DataworkerUtils.ts#L100

mrice32: Yeah, I think that's the best we can do.

mrice32 (approved):
LGTM!

src/clients/SpokePoolClient.ts
mrice32: Thanks for adding that comment -- I agree. I also forgot that it was immutable, so probably even less likely to be changed.

pxrl:
> LGTM - but we should add these variables in the `toJSON` and `fromJSON` serialization functions.

These were subsequently removed in this commit, so it should be OK now :+1:

https://github.com/across-protocol/sdk-v2/commit/fd22f6a2886190d8207b6e3ffe94a6953e33df0f

PR #509, opened 2024-01-24
feat: Export V3 version

nicholaspai:
`isV3` can be used by the dataworker and relayer to switch how it creates RelayerRefundLeaves/SlowFillLeaves and also to know when v3 interfaces are available to be called on-chain


james-a-morris (approved):
LGTM. This aligns well with how `isUBA` is built.

pxrl (approved):
fwiw I think the relayer can escape without this because it just responds to the onchain events emitted (so it knows implicitly when v3 is live) but it does make sense as a controlled config for the dataworker.

PR #510, opened 2024-01-24
refactor: Nuke SpokePool RefundRequests

pxrl:
RefundRequests were never rolled out in v2, and they're currently not a part of v3. Retaining the logic complicates the v3 updates, so remove them for now.

Note: There are a couple of straggling RefundRequest references - i.e. in src/interfaces/SpokePool.ts. I've consciously left those in for now, since they're harmless but will be useful in the event that we do re-add RefundRequests.


nicholaspai (commented):
I think we should remove the types from interfaces. If we need to add them again in the future, then we should redefine them or grab them from git history easily. no need to pollute what's on `master`

nicholaspai (approved):
LGTM assuming you remove all RefundRequest traces in this PR or a follow up, including in the type definitions

pxrl:
> LGTM assuming you remove all RefundRequest traces in this PR or a follow up, including in the type definitions

Updated here - have eliminated all traces of RefundRequests: https://github.com/across-protocol/sdk-v2/pull/510/commits/5f7f2fceefb5f11e9200c139f81f567b2913015c.

nicholaspai (approved):

PR #512, opened 2024-01-25
improve: Remove toJSON/updateFromJSON functions

nicholaspai:


james-a-morris (commented):
Looks good, but since we're removing the `updateFromJSON` in the `AcrosConfigStoreClient` we may also want to remove it from the `SpokePoolClient`. 

https://github.com/across-protocol/sdk-v2/blob/a34c18c818f22ec3e9d0abcf6b689d217637e51d/src/clients/SpokePoolClient.ts#L976

pxrl (approved):

PR #514, opened 2024-01-26
improve(scan-link): add new block explorer links

james-a-morris:


nicholaspai (approved):

src/constants.ts
```diff
@@ -47,6 +47,10 @@ export const PUBLIC_NETWORKS: { [chainId: number]: { name: string; etherscan: st
   43114: { name: "avalanche", etherscan: "https://snowtrace.io" },
   84531: { name: "base-goerli", etherscan: "https://basescan.org" },
   421613: { name: "arbitrum-goerli", etherscan: "https://goerli.arbiscan.io" },
+  11155111: { name: "sepolia", etherscan: "https://sepolia.etherscan.io" },
+  84532: { name: "base-sepolia", etherscan: "https://sepolia.basescan.org" },
+  534352: { name: "scroll", etherscan: "https://scrollscan.com" },
+  534351: { name: "scroll-sepolia", etherscan: "https://sepolia.scrollscan.com" },
```
pxrl, line 53: It looks like the keys in this mapping are sorted numerically - OK to maintain that with this change?

pxrl (approved):

src/constants.ts
james-a-morris: Good call. I've made this numerically sorted and added more outstanding block explorer links.

dohaki (approved):

amateima (approved):

PR #515, opened 2024-01-29
refactor: Remove UBA functionality

pxrl:
This is complicating the v3 upgrade path. Remove it now; it can be restored post-migration when v2 support has been removed.

I've cast the net wide on this one and have attempted to flush as much UBA-related functionality as I can find. It could be the case that we don't want to remove this much - we can selectively restore anything that falls into that category.

nb. This is a huge diff, and in practice all of this functionality is unused in the current system. It feels somewhat violent to be removing this now, but it does also represent a risk to have so much deactivated code in the codebase.


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -334,29 +326,8 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
 
       try {
         const parsedValue = parseJSONWithNumericString(args.value) as ParsedTokenConfig;
-
         const l1Token = args.key;
 
-        // For now use the presence of `uba` or `rateModel` to decide which configs to parse.
-        if (parsedValue?.uba !== undefined) {
```
pxrl, line 341: Note - assuming we also want to remove ConfigStore support for the UBA, we should be sure to also flush the config in the onchain update(s).

nicholaspai: agreed

nicholaspai (approved):

james-a-morris (approved):
I like this. There's a lot of tested-but-unproven code paths that are now removed - this feels like a net good.

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
pxrl: Have recorded this in ACX-1846 :+1:

PR #516, opened 2024-01-29
improve: Add lpFeeCalculator type annotations

pxrl:


nicholaspai (approved):

amateima (approved):
Just out of curiosity, why do we need explicit type annotations?

james-a-morris (approved):
Fair call

pxrl:
> Just out of curiosity, why do we need explicit type annotations?

Fair question. It's a pattern that's followed fairly consistently throughout the repository, so I was surprised they were missing here. As to whether they're explicitly needed - I guess there could be an argument that they aren't? Is it preferable to just outsource it to tsc?

PR #517, opened 2024-01-29
improve: Make deposit realizedLpFeePct mandatory

pxrl:
Now that the UBA-related functionality has been removed, the realizedLpFeePct member of the Deposit-family objects can be made mandatory. This eases the process of introducing v3.

### DIFF
--- a/src/interfaces/SpokePool.ts
+++ b/src/interfaces/SpokePool.ts
@@ -16,14 +16,13 @@ export interface DepositCommon {
   speedUpSignature?: string; // appended after initialization, if deposit was speedup (not part of Deposit event).
   updatedRecipient?: string;
   updatedMessage?: string;
-  realizedLpFeePct?: BigNumber; // appended after initialization (not part of Deposit event).
+  realizedLpFeePct: BigNumber; // appended after initialization (not part of Deposit event).
 }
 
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):
+1

src/interfaces/SpokePool.ts
```diff
@@ -16,14 +16,13 @@ export interface DepositCommon {
   speedUpSignature?: string; // appended after initialization, if deposit was speedup (not part of Deposit event).
   updatedRecipient?: string;
   updatedMessage?: string;
-  realizedLpFeePct?: BigNumber; // appended after initialization (not part of Deposit event).
+  realizedLpFeePct: BigNumber; // appended after initialization (not part of Deposit event).
```
james-a-morris, line 19: OOC does the comment line up with the code? If we're appending this post-initialization then it should be `?:`?

james-a-morris (commented):
Left a comment

src/interfaces/SpokePool.ts
pxrl: It's still correct; the comment refers to the fact that `realizedLpFeePct` is not emitted as part of the `FundsDeposited` event, so the `SpokePoolClient` computes and appends it to the event when it's being ingested.

PR #524, opened 2024-01-31
feat: update FillsToRefund type to account for V3

nicholaspai:
I'd  like to re-use this dictionary in the BundleDataClient for V3

### DIFF
--- a/src/interfaces/SpokePool.ts
+++ b/src/interfaces/SpokePool.ts
@@ -250,7 +250,7 @@ export interface Refund {
 export type FillsToRefund = {
   [repaymentChainId: number]: {
     [l2TokenAddress: string]: {
-      fills: Fill[];
+      fills: Fill[] | v3Fill[];
       refunds?: Refund;
       totalRefundAmount: BigNumber;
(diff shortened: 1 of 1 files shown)


src/interfaces/SpokePool.ts
```diff
@@ -250,7 +250,7 @@ export interface Refund {
 export type FillsToRefund = {
   [repaymentChainId: number]: {
     [l2TokenAddress: string]: {
-      fills: Fill[];
+      fills: Fill[] | v3Fill[];
```
pxrl, line 253: It's preferable to use `v2Fill` here because `Fill` will soon be defined as `v2Fill | v3Fill`.

Also just to confirm, is it important that the `Fills` array is _either_ v2 or v3 fills, but never a mix? If so, this is definitely the right approach. I do however wonder if we could have them interleaved - that seems to be working OK for the other data types so far.

```suggestion
      fills: v2Fill[] | v3Fill[];
```

nicholaspai: AH shoot good point it should be `(v2Fill | v3Fill)[]`

PR #528, opened 2024-02-02
refactor: Re-organise types around RelayData

pxrl:
This PR makes two changes:
- Reorganise common type definitions around v*RelayData types.
- Rename type definitions v* -> V* to fit type style norms (proposed by
  James).

The former is done out of recognition that RelayData definitions
actually drive a lot of the activity in the bots, and this is even
moreso the case with v3. Anchoring types around RelayData definitions
shortens the set of definitions needed in the SpokePool interfaces file,
and also gives assurance that the bots can correctly derive a RelayData
object out of a Deposit, Fill, or even a SlowFillRequest.

### DIFF
--- a/src/clients/mocks/MockSpokePoolClient.ts
+++ b/src/clients/mocks/MockSpokePoolClient.ts
@@ -9,12 +9,12 @@ import {
   FundsDepositedEvent,
   RealizedLpFee,
   SlowFillRequestWithBlock,
-  v2DepositWithBlock,
-  v2FillWithBlock,
-  v2SpeedUp,
-  v3DepositWithBlock,
-  v3FillWithBlock,
-  v3SpeedUp,
+  V2DepositWithBlock,
+  V2FillWithBlock,
+  V2SpeedUp,
+  V3DepositWithBlock,
+  V3FillWithBlock,
+  V3SpeedUp,
 } from "../../interfaces";
 import {
   bnZero,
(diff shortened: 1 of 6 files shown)


src/interfaces/SpokePool.ts
pxrl, line 1: Unfortunately this is ++cluttered and the diff is pretty tricky to review. I can split it up into separate parts if that helps to review it.

The change was FWIW motivated by this comment to another PR: https://github.com/across-protocol/sdk-v2/pull/501#discussion_r1473066444

pxrl: Also, note that this PR shaves off 50 lines - mostly from src/interfaces/SpokePool.ts.

src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -221,7 +221,7 @@ export class MockSpokePoolClient extends SpokePoolClient {
       outputAmount,
       quoteTimestamp,
       fillDeadline: deposit.fillDeadline ?? quoteTimestamp + 3600,
-      relayer: deposit.relayer ?? ZERO_ADDRESS,
+      relayer: deposit.exclusiveRelayer ?? ZERO_ADDRESS,
```
nicholaspai, line 224: why not just name this `exclusiveRelayer` instead of `relayer`?

nicholaspai (approved):
Looks great LGTM

PR #535, opened 2024-02-05
feat: Add helper methods for usage in Dataworker

nicholaspai:
For usage in BundleDataClient
- getSlowFillRequestsForOriginChain: return list of slow fill requests
- getV3RelayHashFromEvent: get the RelayData hash from a Deposit/Fill/SlowFillRequest


src/utils/SpokeUtils.ts
```diff
@@ -240,6 +240,24 @@ function getV3RelayHash(relayData: V3RelayData, destinationChainId: number): str
...
+    outputToken: e.outputToken,
+    inputAmount: e.inputAmount,
+    outputAmount: e.outputAmount,
+    originChainId: e.originChainId,
+    depositId: e.depositId,
+    fillDeadline: e.fillDeadline,
+    exclusivityDeadline: e.exclusivityDeadline,
+    message: e.message,
+  };
+  return getV3RelayHash(relayData, e.destinationChainId);
```
pxrl, line 258: `SlowFillRequest` is currently just a 1:1 match for `V3RelayData`, so it lacks `destinationChainId. I'm not sure why tsc isn't complaining about it. I've proposed to add `destinationChainId` here: https://github.com/across-protocol/sdk-v2/pull/536

common

relayData

slowFillrequest

pxrl: Have merged in the above commit, so we should be OK now.

src/utils/SpokeUtils.ts
```diff
@@ -240,6 +240,24 @@ function getV3RelayHash(relayData: V3RelayData, destinationChainId: number): str
...
+    exclusiveRelayer: e.exclusiveRelayer,
+    inputToken: e.inputToken,
+    outputToken: e.outputToken,
+    inputAmount: e.inputAmount,
+    outputAmount: e.outputAmount,
+    originChainId: e.originChainId,
+    depositId: e.depositId,
+    fillDeadline: e.fillDeadline,
+    exclusivityDeadline: e.exclusivityDeadline,
+    message: e.message,
```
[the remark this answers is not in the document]
nicholaspai: What do you propose as a change?

src/utils/SpokeUtils.ts
nicholaspai: Should we instead just pass in `destinationChainId` into this function?

nicholaspai: https://github.com/across-protocol/sdk-v2/pull/535/commits/78a64ab0670428446d0506933a8346fa6179c4b9

nicholaspai: I realized that even though `V3FilledRelay` has a `destinationChainId`, it doesn't get populated by the `SpokePoolClient` because this param doesn't exist in the onchain event. I think the issue is the type casting we do here overrides the typeguard 
Screenshot 2024-02-06 at 13 43 57

nicholaspai: I think you need this fix https://github.com/across-protocol/sdk-v2/commit/d429f1f93f5bbfd79416cf9c644c82c726d6390d

src/clients/SpokePoolClient.ts
```diff
@@ -816,7 +827,7 @@ export class SpokePoolClient extends BaseAbstractClient {
 
       for (const event of fillEvents) {
         const fill = this.isV3FillEvent(event)
-          ? { ...(spreadEventWithBlockNumber(event) as V3FillWithBlock) }
+          ? { ...(spreadEventWithBlockNumber(event) as V3FillWithBlock), destinationChainId: this.chainId }
```
nicholaspai, line 830: @pxrl i think this was a bug before that was covered up by the `as` cast. Otherwise the new V3Fill doesn't get assigned a `destinationChain`

src/interfaces/SpokePool.ts
```diff
@@ -102,7 +102,6 @@ export interface V2Fill extends V2RelayData {
 
 export interface V3Fill extends V3RelayData {
   destinationChainId: number;
-  realizedLpFeePct: BigNumber;
```
nicholaspai, line 105: This doesn't get set by the spoke pool client. In fact, I don't think its possible to know this until the downstream client matches the fill with a deposit

src/clients/SpokePoolClient.ts
```diff
@@ -1064,4 +1075,79 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    // for a spoke pool and we're looking for a deposit <5 days older than HEAD.
+    const searchBounds = await this._getBlockRangeForDepositId(
+      depositId,
+      this.deploymentBlock,
+      this.latestBlockSearched,
+      7
+    );
+
+    const tStart = Date.now();
+    const query = await paginatedEventQuery(
```
nicholaspai, line 1096: I'm having trouble testing this in `relayer-v2` because this query doesn't seem to work with the Mock SpokePool client and we don't have the latest ABI's in relayer-v2 yet

test/HubPoolClient.DepositToDestinationToken.ts
```diff
@@ -109,7 +109,7 @@ describe("HubPoolClient: Deposit to Destination Token", function () {
   it("Gets L1 token for deposit", async function () {
     const depositData = {
       originChainId,
-      originToken: randomOriginToken,
+      inputToken: randomOriginToken,
```
nicholaspai, line 112: This might be a. merge conflict? or I need to rebase something

src/clients/SpokePoolClient.ts
nicholaspai: trying to test here but the `queryFilter` is returning nothing https://github.com/across-protocol/relayer-v2/pull/1155/files#diff-e73c5f246f260ad0f49e8990b9602cfb17f18b3e5f3c1627ec20e48db048c284R908

src/clients/SpokePoolClient.ts
pxrl: Yeah this is a major pitfall of these type assertions :( Good catch - I will fix this in my PR.

src/utils/SpokeUtils.ts
pxrl: Good catch - this is of course one of the pitfalls of the prevailing method of unpacking events in the SpokePoolClient. I've applied the corresponding fix to v3types and added tests to verify it.

https://github.com/across-protocol/sdk-v2/pull/501/commits/960e6f01fc3aaaad6b8b9ff8aa76873845b24b14

src/clients/SpokePoolClient.ts
pxrl: https://github.com/across-protocol/sdk-v2/pull/501/commits/960e6f01fc3aaaad6b8b9ff8aa76873845b24b14

src/clients/SpokePoolClient.ts
pxrl: This resolves down to Contract.filters.<event-name>, so the contract needs to be instantiated with the correct ABI. We're still instantiating with the old v2 ABIs, which is why it currently doesn't work if you're querying outside of the MockSpokePoolClient context (the mocked variant ignores these filters because it overrides the `SpokePoolClient._update()` method). We do have the option of instantiating a standalone contract with the new SpokePoolV3 ABI, but we still can't interact with it in hre so it wouldn't gain us much ground.

Thinking about this a bit more, it's arguably a layering violation that we sometimes query via the SpokePoolClient and other times query directly to the SpokePool. This prevents us from having being able to entirely encapsulate/abstract the SpokePool contract, which is what the MockSpokePoolClient tries to do for events.

For now I think the best way forward is to bump contracts-v2. After the recent updates it _should_ be ABI backwards compatible with current v2 deployments. Once we get that new package then we can deploy the contract in hre and interact with it as we normally would.

src/clients/HubPoolClient.ts
```diff
@@ -271,6 +272,23 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    tokenB: string,
+    chainIdB: number,
+    latestHubBlock = Number.MAX_SAFE_INTEGER
+  ): boolean {
+    try {
+      return (
+        this.getL1TokenForL2TokenAtBlock(tokenA, chainIdA, latestHubBlock) ===
+        this.getL1TokenForL2TokenAtBlock(tokenB, chainIdB, latestHubBlock)
+      );
+    } catch (err) {
```
nicholaspai, line 287: Convenience method used in BundleDataClient to validate a slow fill request, but I don't love how it hides the error.

src/utils/V3Utils.ts
```diff
@@ -88,6 +89,10 @@ export function isSlowFill(fill: Fill): boolean {
   return isV2Fill(fill) ? fill.updatableRelayData.isSlowRelay : fill.updatableRelayData.fillType === FillType.SlowFill;
 }
 
+export function isSlowFillRequest(fill: Fill | SlowFillRequest): fill is SlowFillRequest {
+  return !isV2Fill(fill as Fill) && !isV3Fill(fill as Fill);
+}
+
```
mrice32, line 95: I don't think this works. I think it will always return false. isV2Fill and isV3Fill are only meant to differentiate V2 and V3 fills from one another. Since SlowFillRequest and V3Fill share the field used to test in isV3Fill (inputToken).

Here's my way to test the type narrowing to see if ts agrees with me (I hover over the fill variable in vscode). If ts agrees with the conditions laid out above, I should see the type of fill as SlowFillRequest when used within the inner if. Instead, ts says never (meaning it thinks I can never reach this, screenshot below).
```ts
  if (!isV3Fill(fill)) {
    if (!isV2Fill(fill)) {
      fill
    }
  }
```

Screenshot 2024-02-07 at 11 58 47 PM

```suggestion
// If needed, this can be made generic in the future.
export function isSlowFillRequest(fill: Fill | SlowFillRequest): fill is SlowFillRequest {
  // Relayer field exists in V2Fill, V3Fill, but not SlowFillRequest.
  return (fill as Fill).relayer === undefined;
}
```

mrice32 (commented):
Looks good, some comments. Only bug I noticed was that I don't think the type guard works as intended.

src/clients/HubPoolClient.ts
pxrl: This is partially duplicating the upstream function `HubPoolClient.tokensAreEquivalent()` (but not completely). Is there a reason it can't be used?

https://github.com/across-protocol/sdk-v2/blob/9a2f0f464936dbc28ce0552ad36922fc3d9ede59/src/clients/HubPoolClient.ts#L512

nicholaspai: oops didn't see this. will replace

src/clients/SpokePoolClient.ts
```diff
@@ -1095,4 +1112,79 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+      throw new Error(
+        `Could not find deposit ${depositId} for ${dstChain} fill` +
+          ` between ${srcChain} blocks [${searchBounds.low}, ${searchBounds.high}]`
+      );
+    }
+    const partialDeposit = spreadEventWithBlockNumber(event) as V3DepositWithBlock;
+    const { realizedLpFeePct, quoteBlock: quoteBlockNumber } = (await this.batchComputeRealizedLpFeePct([event]))[0]; // Append the realizedLpFeePct.
+
+    // Append destination token and realized lp fee to deposit.
+    const deposit: V3DepositWithBlock = {
```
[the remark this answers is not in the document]
nicholaspai: Yeah we really want some sort of helper typeguard here:

```
const partialDeposit = spreadEventWithBlockNumber(event) as V3DepositWithBlock;
```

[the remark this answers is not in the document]
nicholaspai: do you mean something like this but better? https://github.com/across-protocol/sdk-v2/commit/06ffbf54b9c53cdba6ed7cf2e355c4a767a61e68

src/utils/ValidatorUtils.ts
```diff
@@ -28,9 +29,38 @@ const DepositSchema = object({
...
+  logIndex: Min(integer(), 0),
+  quoteBlockNumber: Min(integer(), 0),
+  transactionHash: HexValidator,
 });
 
 export function isDepositFormedCorrectly(deposit: unknown): deposit is DepositWithBlock {
-  return DepositSchema.is(deposit);
+  return (
+    (V2DepositSchema.is(deposit) && isV2Deposit(deposit as DepositWithBlock)) ||
+    (V3DepositSchema.is(deposit) && isV3Deposit(deposit as DepositWithBlock))
```
mrice32, line 64: If we're validating the entire V2 or V3 schema, why do we need to call the isV2/V3Deposit type guards? Aren't those just a subset of the superstruct schema?

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
mrice32: We spoke about this offline, but I would prefer doing that only if we can't get the right types out of spread. But we should probably do that in a separate PR, as it's an unrelated change.

mrice32 (approved):
LGTM!

src/interfaces/SpokePool.ts
```diff
@@ -105,7 +105,7 @@ export interface V3Fill extends V3RelayData {
   destinationChainId: number;
   relayer: string;
   repaymentChainId: number;
-  updatableRelayData: V3RelayExecutionEventInfo;
+  relayExecutionInfo: V3RelayExecutionEventInfo;
```
nicholaspai, line 108: @pxrl This was a bug I found once I sent a real fillV3 to the SpokePool (not via the MockSpkePoolClient) since the onchain event has the `relayExecutionInfo` prop not `updatableRelayData`.

src/utils/FlowUtils.ts
```diff
@@ -64,27 +64,41 @@ export function filledSameDeposit(fillA: Fill, fillB: Fill): boolean {
 
 // Ensure that each deposit element is included with the same value in the fill. This includes all elements defined
 // by the depositor as well as the realizedLpFeePct and the destinationToken, which are pulled from other clients.
-export function validateFillForDeposit(fill: Fill, deposit?: Deposit, fillFieldsToIgnore: string[] = []): boolean {
```
pxrl, line 67: I think we can make some simplifications for v3 types here - I've proposed that in a separate PR: https://github.com/across-protocol/sdk-v2/pull/549/files.

src/utils/ValidatorUtils.ts
```diff
@@ -28,9 +29,38 @@ const DepositSchema = object({
...
+  destinationChainId: Min(integer(), 0),
+  quoteTimestamp: Min(integer(), 0),
+  fillDeadline: Min(integer(), 0),
+  exclusivityDeadline: Min(integer(), 0),
+  exclusiveRelayer: AddressValidator,
+  realizedLpFeePct: optional(BigNumberValidator),
+  outputToken: AddressValidator,
+  outputAmount: BigNumberValidator,
+  message: string(),
+  speedUpSignature: optional(string()),
```
pxrl, line 50: Could speedUpSignature also be validated via HexValidator?

src/utils/ValidatorUtils.ts
pxrl: The type assertion on input to `isV2Deposit()` should be unnecessary now, in any case.

src/interfaces/SpokePool.ts
pxrl: @nicholaspai I'd pulled that out from this PR and have proposed separately: https://github.com/across-protocol/sdk-v2/pull/550

src/clients/SpokePoolClient.ts
```diff
@@ -1098,4 +1125,80 @@ export class SpokePoolClient extends BaseAbstractClient {
 
     return deposit;
   }
+
+  async findDepositV3(depositId: number, destinationChainId: number, depositor: string): Promise<V3DepositWithBlock> {
+    // Binary search for block. This way we can get the blocks before and after the deposit with
```
pxrl, line 1130: This comment looks like it should have been removed when the original functionality was copied - there's no binary search here. ooc, do you intend for this to revert to a binary search in future?

nicholaspai: Replaced the language with "binary search for event query search bounds" which I think is more accurate

src/clients/SpokePoolClient.ts
```diff
@@ -863,9 +889,6 @@ export class SpokePoolClient extends BaseAbstractClient {
       }
     }
 
-    // Exact sequencing of relayer refund executions doesn't seem to be important. There are very few consumers of
-    // these objects, and they are typically used to search for a specific rootBundleId & leafId pair. Therefore,
-    // relayerRefundExecutions don't need exact sequencing and parsing of v2/v3 events can occur without sorting.
```
pxrl, line 868: This has been removed upstream fwiw.

src/clients/SpokePoolClient.ts
```diff
@@ -1086,4 +1109,80 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    if (event === undefined) {
+      const srcChain = getNetworkName(this.chainId);
+      const dstChain = getNetworkName(destinationChainId);
+      throw new Error(
+        `Could not find deposit ${depositId} for ${dstChain} fill` +
+          ` between ${srcChain} blocks [${searchBounds.low}, ${searchBounds.high}]`
+      );
+    }
+    const partialDeposit = spreadEventWithBlockNumber(event) as V3DepositWithBlock;
+    const { realizedLpFeePct, quoteBlock: quoteBlockNumber } = (await this.batchComputeRealizedLpFeePct([event]))[0]; // Append the realizedLpFeePct.
```
pxrl, line 1165: I was hoping to submit a PR shortly that removes the realizedLpFeePct entirely from the V3Deposit object. Do you think that's achievable or do you currently depend on it?

src/utils/SpokeUtils.ts
```diff
@@ -240,6 +240,23 @@ export function getV3RelayHash(relayData: V3RelayData, destinationChainId: numbe
   );
 }
 
+export function getV3RelayHashFromEvent(e: V3Deposit | V3Fill | SlowFillRequest): string {
```
pxrl, line 243: I'm pretty sure it's not necessary to define `relayData` here. Each of the types `V3Deposit | V3Fill | SlowFillRequest` are a superset of a `RelayData` type, and the hash function is abi-aware so it's smart enough to pick out the right fields and place them in the correct order - even if additional fields are supplied. So I think the entire function can probably just be removed, and the caller can call `getV3RelayHash()` directly.

src/utils/V3Utils.ts
```diff
@@ -83,6 +85,11 @@ export function isSlowFill(fill: Fill): boolean {
   return isV2Fill(fill) ? fill.updatableRelayData.isSlowRelay : fill.relayExecutionInfo.fillType === FillType.SlowFill;
 }
 
+// If needed, this can be made generic in the future.
+export function isSlowFillRequest(fill: Fill | SlowFillRequest): fill is SlowFillRequest {
+  // Relayer field exists in V2Fill, V3Fill, but not SlowFillRequest.
+  return (fill as Fill).relayer === undefined;
```
pxrl, line 91: This might be a cleaner way of handling this:

```suggestion
export function isSlowFillRequest(fill: Fill | SlowFillRequest): fill is SlowFillRequest {
  return !isV2Fill(fill) && !isV3Fill(fill);
```

src/utils/V3Utils.ts
```diff
@@ -83,6 +85,11 @@ export function isSlowFill(fill: Fill): boolean {
   return isV2Fill(fill) ? fill.updatableRelayData.isSlowRelay : fill.relayExecutionInfo.fillType === FillType.SlowFill;
 }
 
+// If needed, this can be made generic in the future.
+export function isSlowFillRequest(fill: Fill | SlowFillRequest): fill is SlowFillRequest {
```
pxrl, line 89: ooc, why are we unsure of whether an object is a Fill or a SlowFillRequest? They're very separate events and they aren't mixed within the SpokePoolClient. My initial impression is that it seems unusual to mix them like this.

src/clients/SpokePoolClient.ts
nicholaspai: i do depend on it a lot so this will be a bbig change i might have to merge separately

src/utils/SpokeUtils.ts
nicholaspai: i think this is useful so you don't have to pass in the destination chain

src/clients/SpokePoolClient.ts
pxrl: OK by me. The relayer doesn't depend on it, it was just ignore whatever value is there and will recompute based on its own logic. That makes it forwards-compatible with the LP fee update :+1:

src/utils/CachingUtils.ts
```diff
@@ -44,10 +44,10 @@ export async function setDepositInCache(
...
- * Resolves the key for caching either a deposit or a fill.
- * @param depositOrFill Either a deposit or a fill. In either case, the depositId and originChainId are used to generate the key.
- * @returns The key for caching the deposit or fill.
+ * Resolves the key for caching a bridge event.
+ * @param bridgeEvent The depositId and originChainId are used to generate the key.
+ * @returns The key for caching the event.
  */
-export function getDepositKey(depositOrFill: Deposit | Fill): string {
-  return `deposit_${depositOrFill.originChainId}_${depositOrFill.depositId}`;
+export function getDepositKey(bridgeEvent: Deposit | Fill | SlowFillRequest): string {
```
pxrl, line 51: Is it OK to use this for SlowFillRequest objects? In the SpokePoolClient I concluded that we needed a full RelayData hash, because SlowFillRequest permits the user to control _all_ of the fields that would be used in a lookup (i.e. the `depositId` and `originChainId`). So it's very easy for someone to cause confusion in the bots. By using a RelayData hash then we should also get the depositor included, so it's not possible for them to spoof or confuse the bots about a separate event.

pxrl (approved):
OK subject to a conclusion on this one: https://github.com/across-protocol/sdk-v2/pull/535/files#r1491058411

src/clients/SpokePoolClient.ts
nicholaspai: the dataworker uses it so i'll have to replace those times it fetches `lpFee` from the deposit and compute it there

src/utils/CachingUtils.ts
nicholaspai: I thnk a better fix for this is to switch on the version of the event and either use the existing key or the new getV3Relayhash to create a unique cache key: https://github.com/across-protocol/sdk-v2/pull/535/commits/7ed442207538be35ec264567085467efe6684e50

PR #537, opened 2024-02-06
feat: replace type guards with generics to avoid casting

mrice32:
This is the first PR in the push to use generics to avoid casting: https://github.com/across-protocol/sdk-v2/pull/501#discussion_r1476621443.

This allows you to pass in, for instance, a DepositWithBlock, a Deposit, or any union of structs with "originToken" | "inputToken" to isV2Deposit(), and it will correctly pick the right member of the union.

### DIFF
--- a/src/utils/V3Utils.ts
+++ b/src/utils/V3Utils.ts
@@ -119,15 +119,17 @@ export function isV3RelayerRefundLeaf<T extends MinV3RelayerRefundLeaf, U extend
 
 type MinV2RelayerRefundExecution = Pick<V2RelayerRefundExecution, "amountToReturn">;
 type MinV3RelayerRefundExecution = Pick<V3RelayerRefundExecution, "fillsRefundedRoot" | "fillsRefundedHash">;
-export function isV2RelayerRefundExecution<T extends MinV2RelayerRefundExecution, U extends MinV3RelayerRefundExecution>(
-  leaf: T | U
-): leaf is T {
+export function isV2RelayerRefundExecution<
+  T extends MinV2RelayerRefundExecution,
+  U extends MinV3RelayerRefundExecution,
+>(leaf: T | U): leaf is T {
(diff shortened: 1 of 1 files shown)


src/utils/V3Utils.ts
```diff
@@ -12,13 +12,23 @@ import {
   V3SpeedUp,
 } from "../interfaces";
 import { BN } from "./BigNumberUtils";
-import { isDefined } from "./TypeGuards";
 
-type Deposit = V2Deposit | V3Deposit;
 type Fill = V2Fill | V3Fill;
-type SpeedUp = V2SpeedUp | V3SpeedUp;
-type RelayData = V2RelayData | V3RelayData;
-type SlowFillLeaf = V2SlowFillLeaf | V3SlowFillLeaf;
```
pxrl, line 21: ooc, is `Fill` intentionally the only locally-defined union that wasn't removed?

src/utils/V3Utils.ts
```diff
@@ -103,6 +103,34 @@ export function isV3SlowFillLeaf<T extends MinV3SlowFillLeaf, U extends MinV2Slo
...
+): leaf is T {
+  return unsafeIsType<T, U>(leaf, "amountToReturn") && !isV3RelayerRefundExecution(leaf);
+}
+
+export function isV3RelayerRefundExecution<T extends MinV3RelayerRefundExecution, U extends MinV2RelayerRefundExecution>(
+  leaf: T | U
+): leaf is T {
+  return unsafeIsType<T, U>(leaf, "fillsRefundedRoot");
+}
+
```
pxrl: Just flagging that I added these after merging this PR (https://github.com/across-protocol/sdk-v2/commit/0278a7d0a3addf720eee9d6cf977bd0989a0cab4), so these lines probably warrant extra scrutiny.

pxrl (approved):
I'd like to get this in so that it can be used in the types PR; I expect we'll make adjustments to the underlying implementation as required.

PR #540, opened 2024-02-07
ci(DX): :construction_worker: confirm that SDK builds before merging

james-a-morris:
We should be testing that the SDK builds properly via `yarn build`. This done when a new version is created, but we should have the check at the PR level.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,5 @@
 name: CI
 on: [push]
-env:
-  NODE_OPTIONS: "--max_old_space_size=4096"
 jobs:
   build:
     name: Builds
@@ -13,6 +11,4 @@ jobs:
           node-version: 16
       - run: yarn install
       - run: yarn build
-        env:
-          NODE_URL_1: ${{ secrets.NODE_URL_1 }}
 


.github/workflows/build.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: yarn install
+      - run: yarn build
+        env:
+          NODE_URL_1: ${{ secrets.NODE_URL_1 }}
```
pxrl, line 17: ooc do we use this to build?

pxrl: I was curious about NODE_URL_1. Does this env persist into our tests or is it only for building? If only for building then I suspect it's not required, but we obviously need it for tests.

.github/workflows/build.yml
```diff
@@ -0,0 +1,18 @@
+name: CI
+on: [push]
+env:
+  NODE_OPTIONS: "--max_old_space_size=4096"
```
pxrl, line 4: I also wonder if we could get away without this, since here we only test the SDK.

pxrl (approved):
LGTM - a couple of outstanding questions with the aim to shrink the config a bit, but no blockers.

PR #541, opened 2024-02-09
feat: Import new contracts-v2 beta

pxrl:
This change imports the new contracts-v2 package, aliased under contracts-v2-beta. It can be used side-by-side with the existing contracts-v2.


package.json
```diff
@@ -98,13 +98,14 @@
   "dependencies": {
     "@across-protocol/across-token": "^1.0.0",
     "@across-protocol/constants-v2": "^1.0.9",
-    "@across-protocol/contracts-v2": "^2.4.7",
+    "@across-protocol/contracts-v2": "^2.4.8",
```
pxrl, line 101: For reference, the delta from v2.4.7 to v2.4.8 is this commit: https://github.com/across-protocol/contracts-v2/commit/65a007c5e6dbb51b1a54b770f246d2002dafb1b0

nicholaspai (approved):

pxrl:
The follow-up PR here shows how this can be used: https://github.com/across-protocol/sdk-v2/pull/545

nicholaspai (changes requested):
I think there's an issue with this approach. It doesn't test the clients against the realistic scenario where a single spoke pool emits both V2 and V3 events. This PR instead tests the clients against V2 and V3 events in isolation but there might be regressions when combined. I think a better approach with low code diff is to modify the contracts-v2 code exported out of dir/test-utils to handle this

nicholaspai:
I recommend closing this in favor of https://github.com/across-protocol/sdk-v2/pull/546 which uses an updated contracts-v2 import with a depositV2 function https://github.com/across-protocol/contracts-v2/pull/429

pxrl:
Closed in favour of https://github.com/across-protocol/sdk-v2/commit/04be89449dd177d6c58385dd5f6af62cf089ef90

PR #543, opened 2024-02-09
feat: Update to >=contracts-v2@0.2.5

nicholaspai:
## Reviewer's guide
- The current unit tests heavily rely on importing functions from `@across-protocol/contracts-v2/dist/` which means that the unit tests are reliant on outdated code. This PR removes almost all imports from this repo except for some that have not changed and will not change for a long time.
- Most of the diff is moving test helper functions around and/or copying test helpers from `contracts-v2`. There are no logical changes in the test/ directory
- contracts-v2 is updated. This is a breaking change.
- Updates MockSpokePool.sol to emit the old FundsDeposted event so we can still unit test V2 logic


test/fixtures/UmaEcosystemFixture.ts
```diff
@@ -1,7 +1,7 @@
-import * as utils from "@across-protocol/contracts-v2/dist/test-utils";
 import { Contract } from "ethers";
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
-import { utf8ToHex, identifier, refundProposalLiveness } from "@across-protocol/contracts-v2/dist/test-utils";
+import { getContractFactory, utf8ToHex } from "@across-protocol/contracts-v2";
+import { constants } from "../utils/index";
```
pxrl, line 4:
```suggestion
import { constants } from "../utils";
```

test/utils/HubPoolUtils.ts
```diff
@@ -1,8 +1,24 @@
...
+  AcrossConfigStore,
+  getContractFactory,
+  utf8ToHex,
+  constants,
+  seedWallet,
+  MAX_L1_TOKENS_PER_POOL_REBALANCE_LEAF,
+  MAX_REFUNDS_PER_RELAYER_REFUND_LEAF,
+  getSampleRateModel,
+  TokenRolesEnum,
+} from "./index";
```
pxrl, line 19: Does this risk creating a circular dependency on imports?

test/utils/SpokePoolUtils.ts
```diff
@@ -68,3 +103,669 @@ export function v3FillFromDeposit(deposit: V3DepositWithBlock, relayer: string):
...
+}
+
+/**
+ * Find the amount filled for a deposit at a particular block.
+ * @param spokePool SpokePool contract instance.
+ * @param relayData Deposit information that is used to complete a fill.
+ * @param blockTag Block tag (numeric or "latest") to query at.
+ * @returns The amount filled for the specified deposit at the requested block (or latest).
+ */
+export async function relayFilledAmount(
```
pxrl, line 671: I'm OK to remove `relayFilledAmount()` and `findFillBlock()` for now - I will ressurrect them in v3 versions.

pxrl: Or we can just nuke the v2-specific parts of `relayFilledAmount()` and make it v3-specific, since it _should_ already work.

package.json
```diff
@@ -98,7 +98,7 @@
   "dependencies": {
     "@across-protocol/across-token": "^1.0.0",
     "@across-protocol/constants-v2": "^1.0.9",
-    "@across-protocol/contracts-v2": "^2.4.7",
+    "@across-protocol/contracts-v2": "^2.5.0-beta.2",
```
pxrl, line 101: One consideration: Doing this on master could leave us in a situation where we can't bump sdk-v2 despite needing to get a change into production. I wonder if it might be safer to have this on a development branch.

nicholaspai: Hmm

test/utils/HubPoolUtils.ts
nicholaspai: Is there a specific circular dep I created or are there many?

I was thinking that index should export everything out of utils while the individual utils files should load from index to get things from other utils files

nicholaspai:
Deprecating for https://github.com/across-protocol/sdk-v2/pull/546 which maintains the existing contracts-v2/dir/ import but has a lot smaller code diff

PR #546, opened 2024-02-09
feat: Update to contracts-v2@2.5.0

nicholaspai:
This  PR takes advantage of an updated MockSpokePool and updated test-utils exports from the contracts-v2 repo that make existing unit tests in this repo pass out of the box


pxrl (approved):
@nicholaspai I pushed some changes to this PR, specifically to:
 - Update the fill utility functions to work _only_ with v3 types.
 - Complete the v3 fill status test (was previously @todo) and deprecated the v2 fill status test (will never be used).
 - Add depositV3() and fillV3Relay() utility functions.

It's probably still my preference to import v3 contracts separately as an aliased package, but I don't have any reason to think that this approach won't work, and I am really satisfied to be able to test against v3 contracts in this PR.

PR #547, opened 2024-02-09
feat(arweave): add arweave client for persisting/retrieving data

james-a-morris:


.github/workflows/test.yml
```diff
@@ -10,8 +10,10 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 18
```
nicholaspai, line 13: will this break the relayer? didn't we have an issue bumping this before?

src/caching/Arweave/ArweaveClient.ts
```diff
@@ -0,0 +1,110 @@
...
+    const transaction = await this.client.createTransaction(
+      { data: JSON.stringify(value, jsonReplacerWithBigNumbers) },
+      this.arweaveJWT
+    );
+    // Add tags to the transaction
+    transaction.addTag("Content-Type", "application/json");
+    // Sign the transaction
+    await this.client.transactions.sign(transaction, this.arweaveJWT);
+    // Send the transaction
+    const result = await this.client.transactions.post(transaction);
```
nicholaspai, line 48: how do fees work?

test/arweaveClient.ts
```diff
@@ -0,0 +1,121 @@
...
+    const expectedValue = { test: "value", bigNumber: "1000000000000000000" };
+    expect(retrievedValue).to.deep.equal(expectedValue);
+  });
+
+  it("should fail to get a non-existent record", async () => {
+    const retrievedValue = await client.get("non-existent");
+    expect(retrievedValue).to.be.null;
+  });
+
+  it("should validate the record with a struct validator", async () => {
```
nicholaspai, line 92: Can you add a test where the validator is of type  `[ objA[], objB[], key ]`? Replicating what we're gonna store in production: a list of Fills, SlowFills, and a bundle block range maybe

nicholaspai (approved):
This looks really good

src/caching/Arweave/ArweaveClient.ts
```diff
@@ -0,0 +1,110 @@
...
+    // Ensure that the result is successful. If it is not, the retrieved value is not our expected type
+    // but rather a {status: string, statusText: string} object. We can detect that and return null.
+    if (data.status === 400) {
+      return null;
+    }
+    if (validator && !is(data, validator)) {
+      this.logger.warn("Retrieved value from Arweave does not match the expected type");
+      return null;
+    }
+    return data as T;
```
dohaki, line 90: Is this `as` type cast required? Shouldn't the `is` function infer the type of `data` as `T` automatically?

src/caching/Arweave/ArweaveClient.ts
```diff
@@ -0,0 +1,110 @@
+// import { Struct } from "superstruct";
+// import { CachingMechanismInterface } from "../../interfaces";
```
dohaki, line 2: Can be removed?

dohaki (commented):
Looks good to me 👍 Left two small nits

dohaki (approved):

src/caching/Arweave/ArweaveClient.ts
nicholaspai: Yes we should send an error to the caller if there aren't funds to pay for the txn

nicholaspai (approved):
Doesn't have to be in this PR, but what are your thoughts on:
- catching out of fees errors
- handling file organization so we can aggregate files posted by the dataworker easily

nicholaspai:
> > Doesn't have to be in this PR, but what are your thoughts on:
> > 
> > * catching out of fees errors
> > * handling file organization so we can aggregate files posted by the dataworker easily
> 
> We have custom tags now via ([578f74c](https://github.com/across-protocol/sdk-v2/commit/578f74cc8b0b205c99694c0c1b8b477743109b28)). Because Arweave is tangential from our normal day-to-day blockchain interactions, I'd think we should put a balance check in the monitor and dump that to slack if it gets close to zero.

it is tangential but it has the ability to stop operations if we run out of arweave gas and either page us until we fill up gas or propose a bundle silently without posting to arweave

PR #549, opened 2024-02-12
improve: Upgrade fill/deposit comparison utils

pxrl:
This pivots fill comparison for v3 towards using relay data hashes. This is simpler code and should also be safer, since this code inherits existing assurances about the RelayData being complete.

### DIFF
--- a/src/utils/FlowUtils.ts
+++ b/src/utils/FlowUtils.ts
@@ -36,7 +36,7 @@ export function filledSameDeposit(fillA: Fill, fillB: Fill): boolean {
   } else if (isV3Fill(fillA) && isV3Fill(fillB)) {
     const { destinationChainId: chainA } = fillA;
     const { destinationChainId: chainB } = fillB;
-    return chainA === chainB && getV3RelayHash(fillA, chainA) === getV3RelayHash(fillB, chainB);
+    return getV3RelayHash(fillA, chainA) === getV3RelayHash(fillB, chainB);
   }
 
   return false;
--- a/src/utils/FlowUtils.ts
+++ b/src/utils/FlowUtils.ts
@@ -1,5 +1,6 @@
-import { Deposit, Fill } from "../interfaces";
-import { isV2Deposit, isV3Deposit, isV2Fill, isV3Fill } from "./V3Utils";
+import { Deposit, Fill, RelayData, SlowFillRequest, V2RelayData, V3Fill } from "../interfaces";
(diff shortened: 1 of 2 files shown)


src/utils/FlowUtils.ts
```diff
@@ -31,61 +32,47 @@ export const V3_DEPOSIT_COMPARISON_KEYS = [
...
+    : validateV3FillForDeposit(relayData, deposit);
+}
+
+function validateV2FillForDeposit(fill: V2RelayData, deposit: Deposit, fillFieldsToIgnore: string[] = []): boolean {
+  if (!isV2Deposit(deposit)) {
+    return false;
   }
 
-  if (isV3Deposit(deposit) && isV3Fill(fill)) {
-    return V3_DEPOSIT_COMPARISON_KEYS.every((key) => {
```
pxrl, line 82: Note - technically we could remove V3_DEPOSIT_COMPARISON_KEYS in this PR. I've left that definition in because I foresee reusing it in this PR (https://github.com/across-protocol/sdk-v2/pull/493) in order to produce the underlying reason in the event of a mismatch. For example, if the hashes mismatch, then we proceed to iterate over the keys to identify which field(s) disagree.

src/utils/FlowUtils.ts
```diff
@@ -31,61 +32,47 @@ export const V3_DEPOSIT_COMPARISON_KEYS = [
...
-      fillA.message === fillB.message &&
-      fillA.inputAmount.eq(fillB.inputAmount) &&
-      fillA.outputAmount.eq(fillB.outputAmount) &&
-      fillA.fillDeadline === fillB.fillDeadline &&
-      fillA.exclusivityDeadline === fillB.exclusivityDeadline &&
-      fillA.exclusiveRelayer === fillB.exclusiveRelayer
-    );
+    const { destinationChainId: chainA } = fillA;
+    const { destinationChainId: chainB } = fillB;
+    return chainA === chainB && getV3RelayHash(fillA, chainA) === getV3RelayHash(fillB, chainB);
```
nicholaspai, line 39: Is `chainA === chainB` neccessary since its part of the hash?

pxrl: Removed: https://github.com/across-protocol/sdk-v2/pull/549/commits/b16bd440f5cf31e9bed28488ac8630ddcb2c3bd7

We can also remove this function entirely in the not-too-distant future :raised_hands:

PR #551, opened 2024-02-12
refactor: Cleanup test import/exports

pxrl:
Primarily, avoid re-exporting the entirety of contracts-v2 from multiple places in the test directory. This gives better control over what is imported, and where from. Additionally, re-organise so that test/constants only exports actual constants and not utility functions, then relocate the functions to test/utils/utils.ts.


nicholaspai (commented):
Why are we increasing the dependency on `contracts-v2`?

nicholaspai (approved):

PR #552, opened 2024-02-12
[WIP] chore: Update UnfilledDeposit types for V3

pxrl:
The Dataworker needs a V3 variant to track the realizedLpFeePct of the deposit with payout on the destination chain.

### DIFF
--- a/src/interfaces/SpokePool.ts
+++ b/src/interfaces/SpokePool.ts
@@ -173,12 +173,18 @@ export interface V3RelayerRefundExecutionWithBlock extends V3RelayerRefundExecut
 export type RelayerRefundExecution = V2RelayerRefundExecution | V3RelayerRefundExecution;
 export type RelayerRefundExecutionWithBlock = V2RelayerRefundExecutionWithBlock | V3RelayerRefundExecutionWithBlock;
 
-export interface UnfilledDeposit {
-  deposit: Deposit;
+export interface V2UnfilledDeposit {
+  deposit: V2Deposit;
   unfilledAmount: BigNumber;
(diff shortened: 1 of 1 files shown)


src/interfaces/SpokePool.ts
```diff
@@ -173,13 +173,19 @@ export interface V3RelayerRefundExecutionWithBlock extends V3RelayerRefundExecut
...
-  deposit: Deposit;
+export interface V2UnfilledDeposit {
+  deposit: V2Deposit;
   unfilledAmount: BigNumber;
   hasFirstPartialFill?: boolean;
   relayerBalancingFee?: BigNumber;
 }
 
+export interface V3UnfilledDeposit extends V3Deposit {
+  realizedLpFeePct: BigNumber; // Assumes repayment on the destination chain.
```
nicholaspai, line 184: When we start computing this based on repayment chain, will this stay in this type and be set by the Relayer?

nicholaspai (approved):
https://github.com/across-protocol/sdk-v2/pull/552/files#r1486083383

src/interfaces/SpokePool.ts
pxrl: In terms of what the relayer will do, I think the realizedLpFeePct will only be needed for a very short while, so I don't actually think it needs to be bundled into any types - it can probably just be passed around as an additional parameter to the ProfitClient (etc).

nicholaspai: This type is not being used by my dataworker implementation, I have a new type defined so if you're not using this you can remove

PR #554, opened 2024-02-12
feat(arweave): store fill refund leaves

james-a-morris:
Adds functionality to get/set fill refund leaves with validation.


test/arweaveClient.ts
```diff
@@ -118,4 +142,25 @@ describe("ArweaveClient", () => {
     const retrievedValue = await client.get(txID!, validatorStruct);
     expect(retrievedValue).to.eq(null);
   });
+
+  it.only("should be able to set and retrieve a complex record", async () => {
```
nicholaspai, line 146: leftover .only

PR #558, opened 2024-02-15
improve(arweave): handle out of gas error

james-a-morris:
We should explicitly handle the path where we are out of gas. This should unwind the stack to the calling function.


nicholaspai (approved):
Can we set up an automatic way to replenish the arweave account? Similar to how we do so for the dataworker in the Monitor code

james-a-morris:
> Can we set up an automatic way to replenish the arweave account? Similar to how we do so for the dataworker in the Monitor code

Tracked in ACX-1903

PR #564, opened 2024-02-20
feat(HubPoolClient): Support arbitrary payment LP fee payment chain

pxrl:
This permits the caller to specify a payment chain, rather than just assuming payment occurs on the destination chain.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -72,13 +72,13 @@ type L1TokensToDestinationTokens = {
 // Temporary type for v2 -> v3 transition. @todo: Remove.
 type V2PartialDepositWithBlock = Pick<
   V2DepositWithBlock,
-  "originChainId" | "originToken" | "amount" | "quoteTimestamp"
+  "originChainId" | "originToken" | "amount" | "quoteTimestamp" | "blockNumber"
 >;
 
 // Temporary type for v2 -> v3 transition. @todo: Remove.
 type V3PartialDepositWithBlock = Pick<
   V3DepositWithBlock,
-  "originChainId" | "inputToken" | "inputAmount" | "quoteTimestamp"
+  "originChainId" | "inputToken" | "inputAmount" | "quoteTimestamp" | "blockNumber"
 >;
 
 export type LpFeeRequest = (V2PartialDepositWithBlock | V3PartialDepositWithBlock) & { paymentChainId?: number };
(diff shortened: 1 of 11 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -70,17 +70,19 @@ type L1TokensToDestinationTokens = {
 };
 
 // Temporary type for v2 -> v3 transition. @todo: Remove.
-export type V2PartialDepositWithBlock = Pick<
+type V2PartialDepositWithBlock = Pick<
   V2DepositWithBlock,
-  "originChainId" | "destinationChainId" | "originToken" | "amount" | "quoteTimestamp" | "blockNumber"
```
pxrl, line 75: fwiw `blockNumber` was not actually required in this type and was probably there by error.

src/clients/HubPoolClient.ts
```diff
@@ -70,17 +70,19 @@ type L1TokensToDestinationTokens = {
...
 
 // Temporary type for v2 -> v3 transition. @todo: Remove.
-export type V3PartialDepositWithBlock = Pick<
+type V3PartialDepositWithBlock = Pick<
   V3DepositWithBlock,
-  "originChainId" | "destinationChainId" | "inputToken" | "inputAmount" | "quoteTimestamp" | "blockNumber"
+  "originChainId" | "inputToken" | "inputAmount" | "quoteTimestamp"
 >;
 
+export type LpFeeRequest = (V2PartialDepositWithBlock | V3PartialDepositWithBlock) & { paymentChainId: number };
```
pxrl, line 84: `paymentChainId` selected because this also covers slow fills, in which case there's no `repayment` per se.

nicholaspai (approved):

src/clients/HubPoolClient.ts
```diff
@@ -428,14 +426,18 @@ export class HubPoolClient extends BaseAbstractClient {
 
     // Helper compute the realizedLpFeePct of an individual deposit based on pre-retrieved batch data.
     const computeRealizedLpFeePct = async (deposit: (typeof deposits)[0]) => {
-      const { originChainId, destinationChainId, inputToken, inputAmount, quoteTimestamp } = deposit;
+      const { originChainId, paymentChainId, inputToken, inputAmount, quoteTimestamp } = deposit;
       const quoteBlock = quoteBlocks[quoteTimestamp];
 
+      if (paymentChainId === undefined) {
```
nicholaspai, line 432: @pxrl I added this here so that the SpokePoolClient doesn't needlessly compute a realizedLpFeePct for V3 deposit events

src/clients/HubPoolClient.ts
nicholaspai: I actually think this is important to get the MockHubPoolClient to be able to mock an LP fee: 262ba0762b87d8801e9d2efa0eef5f62033deb3d

This is required to make the tests easier in the BundleDataClient because it calls `hubPoolClient.computeLpFee` instead of loading the realized lp fee stored in Deposits at SpokePoolClient.update time

PR #568, opened 2024-02-21
feat(relayFeeCalculator): Support estimating v3 fills

pxrl:
This is a simple change to permit estimating v3 fills. This is used by the relayer, but also the FE. It'll be necessary for simulating messages with fills to recipients who only support handleV3AcrossMessage(), and the relayer also needs it prior to v3 going live.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -93,6 +93,9 @@ export const DEFAULT_LOGGER: Logger = {
   error: (...args) => console.error(args),
 };
 
+// Small amount to simulate filling with. Should be low enough to guarantee a successful fill.
+const safeOutputAmount = toBN(100);
+
 export class RelayFeeCalculator {
   private queries: QueryInterface;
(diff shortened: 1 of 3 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -234,26 +237,24 @@ export class RelayFeeCalculator {
...
-      .catch((error) => {
-        this.logger.error({
-          at: "sdk-v2/gasFeePercent",
-          message: "Error while fetching gas costs",
-          error,
-          simulateZeroFill,
-          deposit,
-        });
-        throw error;
+    if (simulateZeroFill) {
```
pxrl, line 240: `simulateZeroFill` is a bit of a misnomer in v3, since we're simulating a reduced fill rather than a `requestV3SlowFill()` call. I've left it as-is because the logic is mostly relevant to the FE, and I think there's plenty of opportunity to revisit that logic once the v2 logic is removed.

mrice32 (approved):
LGTM!

pxrl:
Have been testing this OK on testnet.

dohaki (approved):

PR #573, opened 2024-02-22
fix: Add `originChainId` to partial deposit queried from eth_getLogs

nicholaspai:
Missing `originChainId` needs to be appended to this partial event queried from eth_getLogs

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.22.4",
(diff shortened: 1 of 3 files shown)


james-a-morris (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk-v2",
   "author": "UMA Team",
-  "version": "0.22.4",
+  "version": "0.22.5",
```
james-a-morris, line 4: We should bump this to 0.22.6

PR #578, opened 2024-03-06
feat: Support querying relay fill status as an array

pxrl:
This is an efficient way of determining bulk fill status in a single RPC call. Both the relayer and dataworker currently make these requests per deposit, so this will reduce RPC consumption and latency.

### DIFF
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -1,11 +1,14 @@
 import assert from "assert";
-import { Contract, PopulatedTransaction, utils as ethersUtils } from "ethers";
+import { BigNumber, BytesLike, Contract, PopulatedTransaction, providers, utils as ethersUtils } from "ethers";
 import { CHAIN_IDs, ZERO_ADDRESS } from "../constants";
 import { FillStatus, RelayData, SlowFillRequest, V2RelayData, V3Deposit, V3Fill, V3RelayData } from "../interfaces";
 import { SpokePoolClient } from "../clients";
+import { toBN } from "./BigNumberUtils";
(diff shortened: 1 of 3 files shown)


src/utils/SpokeUtils.ts
```diff
@@ -307,12 +310,38 @@ export async function relayFillStatus(
...
+    (result: BytesLike) => spokePool.interface.decodeFunctionResult(fillStatuses, result)[0]
+  );
+
+  const bnUnfilled = toBN(FillStatus.Unfilled);
+  const bnFilled = toBN(FillStatus.Filled);
+
+  return status.map((status: unknown) => {
+    return BigNumber.isBigNumber(status) && status.gte(bnUnfilled) && status.lte(bnFilled)
+      ? status.toNumber()
+      : undefined;
```
nicholaspai, line 341: Can these ever be undefined? Wouldn't that indicate a contract error?

nicholaspai (approved):
One comment about the return type allowing `undefined` array values

src/utils/SpokeUtils.ts
pxrl: The contract itself should never permit this to occur, but I don't think we can make that assumption of the RPC provider, so that's why the check is there. I don't expect that we'd ever see this arise in production, but the options would be to throw or return undefined and force the caller to check it and decide what they want to do.

pxrl: Also, the check is not exactly looking for whether the status returned is `undefined` - it's moreso checking that it fits within the range of values that we consider valid for this mapping.

mrice32 (approved):
LGTM!

PR #595, opened 2024-03-19
refactor(test): Upgrade additional tests to v3

pxrl:
In advance of removing v2 fills from the SpokePoolClient.


test/SpokePoolClient.deposits.ts
pxrl, line 1: Removed because this test seems redundant - it's entirely duplicated by `test/SpokePoolClient.fills.ts`.

test/SpokePoolClient.fills.ts
```diff
@@ -53,91 +45,54 @@ describe("SpokePoolClient: Fills", function () {
     await setupTokensForWallet(spokePool, relayer1, [erc20, destErc20], weth, 10);
     await setupTokensForWallet(spokePool, relayer2, [erc20, destErc20], weth, 10);
 
-    quoteTimestamp = Number(await spokePool.getCurrentTime());
-    fillDeadline = quoteTimestamp + 600;
-  });
-
-  it("Correctly fetches fill data single fill, single chain", async function () {
-    const deposit: Deposit = {
+    const spokePoolTime = Number(await spokePool.getCurrentTime());
```
james-a-morris, line 48: OOC why did we change this from `quoteTimestamp`?

james-a-morris (commented):
one question - relatively minor nit

test/SpokePoolClient.fills.ts
pxrl: The value is the same but the name changed to `spokePoolTime` because we are setting multiple inputs in the Deposit() based on it (quoteTimestamp, fillDeadline).

pxrl:
This PR was made redundant by its changes being merged by another PR, so it now has 0 diff to master.

PR #596, opened 2024-03-19
refactor(SpokePoolClient): Drop v2 fill update support

pxrl:


src/clients/SpokePoolClient.ts
```diff
@@ -799,10 +796,10 @@ export class SpokePoolClient extends BaseAbstractClient {
       // @note The type assertions here suppress errors that might arise due to incomplete types. For now, verify via
       // test that the types are complete. A broader change in strategy for safely unpacking events will be introduced.
       for (const event of fillEvents) {
-        const fill = this.isV3FillEvent(event)
-          ? { ...(spreadEventWithBlockNumber(event) as V3FillWithBlock), destinationChainId: this.chainId }
-          : { ...(spreadEventWithBlockNumber(event) as V2FillWithBlock) };
-
+        const fill = {
+          ...(spreadEventWithBlockNumber(event) as V3FillWithBlock),
```
james-a-morris, line 800: Out of scope question: I wonder if we can add an additional generic parameter on `spreadEventWithBlockNumber` so that we don't need to typecast as much

src/clients/SpokePoolClient.ts
```diff
@@ -784,11 +784,8 @@ export class SpokePoolClient extends BaseAbstractClient {
       }
     }
 
-    if (eventsToQuery.includes("FilledRelay") || eventsToQuery.includes("FilledV3Relay")) {
-      const fillEvents = [
-        ...((queryResults[eventsToQuery.indexOf("FilledRelay")] ?? []) as FilledRelayEvent[]),
-        ...((queryResults[eventsToQuery.indexOf("FilledV3Relay")] ?? []) as FilledV3RelayEvent[]),
-      ];
+    if (eventsToQuery.includes("FilledV3Relay")) {
+      const fillEvents = [...((queryResults[eventsToQuery.indexOf("FilledV3Relay")] ?? []) as FilledV3RelayEvent[])];
```
james-a-morris, line 788: Would something like this give us any tsc errors? I'm not sure the spread operator is necessary here

```suggestion
      const fillEvents: FilledV3RelayEvent[] = queryResults[eventsToQuery.indexOf("FilledV3Relay")] ?? [];
```

pxrl: tsc isn't happy about this because `queryResults` is an ethers Event, whereas `FilledV3RelayEvent` is a more tightly-scoped type.

src/clients/SpokePoolClient.ts
pxrl: Yeah; I think there's general consensus on the value of doing this :+1:

pxrl:
This PR was made redundant by its changes being merged by another PR, so it now has 0 diff to master.

PR #603, opened 2024-03-26
improve: Verify that outputToken is correctly set

pxrl:
Resolves ACX-1898


src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -33,8 +34,14 @@ export class MockSpokePoolClient extends SpokePoolClient {
...
   public blocks: Record<number, Block> = {};
 
-  constructor(logger: winston.Logger, spokePool: Contract, chainId: number, deploymentBlock: number) {
-    super(logger, spokePool, null, chainId, deploymentBlock);
+  constructor(
+    logger: winston.Logger,
+    spokePool: Contract,
+    chainId: number,
+    deploymentBlock: number,
+    opts: { hubPoolClient: HubPoolClient | null } = { hubPoolClient: null }
```
james-a-morris, line 42: Thoughts on making the HubPoolClient optional? It should have the same intended effect unless you explicitly need `null` instead of `undefined`.
```suggestion
    opts: { hubPoolClient?: HubPoolClient } = {}
```

test/SpokePoolClient.v3Events.ts
```diff
@@ -129,6 +130,35 @@ describe("SpokePoolClient: Event Filtering", function () {
...
+    const outputToken = randomAddress();
+    hubPoolClient.setTokenMapping(hubPoolToken, originChainId, inputToken);
+    hubPoolClient.setTokenMapping(hubPoolToken, destinationChainId, outputToken);
+    hubPoolClient.setDefaultRealizedLpFeePct(toBNWei("0.0001"));
+
+    const _deposit = spokePoolClient.depositV3({
+      originChainId,
+      destinationChainId,
+      inputToken,
+      outputToken: ZERO_ADDRESS,
```
james-a-morris, line 148: Thoughts on adding a small comment here to highlight that this line is the main testing point?

```suggestion
      outputToken: ZERO_ADDRESS, // Set empty address 0x0 to test outputToken substitution
```

james-a-morris (approved):
Two nits, otherwise LGTM

src/clients/mocks/MockSpokePoolClient.ts
pxrl: It specifically needs to be null per SpokePoolClient definition: https://github.com/across-protocol/sdk-v2/blob/b26b1feacf0c00e7ed13ed4303faf618790207ec/src/clients/SpokePoolClient.ts#L91.

We could change that if we want...

test/SpokePoolClient.v3Events.ts
pxrl: https://github.com/across-protocol/sdk-v2/pull/603/commits/0232d53ec8bb879872c2f1505f7679273607c15d

src/clients/mocks/MockSpokePoolClient.ts
pxrl: Will merge this as-is and follow up out of band.

PR #605, opened 2024-03-27
fix(relay-execution-info): patch parsing of nested type

james-a-morris:
The `relayExecutionInfo` is not being parsed properly when a fill message is non-empty. The discrepancy was detected in [this arweave bundle](https://m5gtkkyrzrvyyd6ec554mbwiplhapbwt2jrmaqs33dycjd3puj2q.arweave.net/Z001KxHMa4wPxBd7xgbIes4HhtPSYsBCW9jwJI9vonU).

This patch performs a secondary check on `relayExecutionInfo` to manually update the event into a more stable version. A more comprehensive fix will follow.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -18,6 +18,7 @@ import {
   isV3Deposit,
   isV3SpeedUp,
   toBN,
+  isV3Fill,
 } from "../utils";
 import {
   paginatedEventQuery,
@@ -835,6 +836,15 @@ export class SpokePoolClient extends BaseAbstractClient {
           ? { ...(spreadEventWithBlockNumber(event) as V3FillWithBlock), destinationChainId: this.chainId }
           : { ...(spreadEventWithBlockNumber(event) as V2FillWithBlock) };
 
+        if (isV3Fill(fill) && Array.isArray(fill.relayExecutionInfo)) {
+          const correctedRelayExecutionInfo = {
+            updatedRecipient: fill.relayExecutionInfo[0],
+            updatedMessage: fill.relayExecutionInfo[1],
+            updatedOutputAmount: fill.relayExecutionInfo[2],
+            fillType: fill.relayExecutionInfo[3],
+          };
+          fill.relayExecutionInfo = correctedRelayExecutionInfo;
+        }
         assign(this.fills, [fill.originChainId], [fill]);
         assign(this.depositHashesToFills, [this.getDepositHash(fill)], [fill]);
       }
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -835,6 +836,15 @@ export class SpokePoolClient extends BaseAbstractClient {
           ? { ...(spreadEventWithBlockNumber(event) as V3FillWithBlock), destinationChainId: this.chainId }
           : { ...(spreadEventWithBlockNumber(event) as V2FillWithBlock) };
 
+        if (isV3Fill(fill) && Array.isArray(fill.relayExecutionInfo)) {
```
nicholaspai, line 839: @pxrl I think if you merge in https://github.com/across-protocol/sdk-v2/pull/596 we won't need this isV3Fill check

src/clients/SpokePoolClient.ts
```diff
@@ -835,6 +836,15 @@ export class SpokePoolClient extends BaseAbstractClient {
           ? { ...(spreadEventWithBlockNumber(event) as V3FillWithBlock), destinationChainId: this.chainId }
           : { ...(spreadEventWithBlockNumber(event) as V2FillWithBlock) };
```
nicholaspai, line 837: Why doesn't spreadEvent work? Shouldn't we put this logic in there if its broken?

nicholaspai (commented):
https://github.com/across-protocol/sdk-v2/pull/605#discussion_r1541795551

src/clients/SpokePoolClient.ts
james-a-morris: I'm working on a larger fix to see why `spreadEvent` isn't working. This is a patch - the next PR will be a fix

nicholaspai (approved):
Can we add a test to prevent regressions? Would also protect us in the larger spreadEvent fix.

Some simple fix that `relayExecutionInfo` is always in the expected shape post .update()

src/clients/SpokePoolClient.ts
pxrl: Yeah agree. I'm fine for this to go in as-is though - I can update my PR to subsequently remove it.

pxrl, line 839: ooc, did you manage to confirm that this type is never an array on other chains?

Also, worth an `assert(fill.relayExecutionInfo.length === 4)` ?

pxrl (approved):

src/clients/SpokePoolClient.ts
james-a-morris: I found it also on Arbitrum now

james-a-morris:
Closing in lieu of #606

PR #606, opened 2024-03-27
hotfix(spreadEvent): fix recurrence step

james-a-morris:
There is a bug in the spreadEvent function right before the recurrence. Previously, we were getting the keys of 
```
args[key].filter((key: string) => isNaN(+key)
```
which would first filter the array of non-numeric keys. Whereas we should have gotten the keys of `args[key]` and filtered on the keys.

### DIFF
--- a/src/utils/EventUtils.ts
+++ b/src/utils/EventUtils.ts
@@ -22,7 +22,7 @@ export function spreadEvent(args: Result = {} as Result): any {
       case "object":
         if (Array.isArray(args[key])) {
           returnedObject[key] =
-            Object.keys(args[key].filter((key: string) => isNaN(+key))).length > 0
+            Object.keys(args[key]).filter((key: string) => isNaN(+key)).length > 0
               ? spreadEvent(args[key]) // Record/array hybrid...
               : args[key]; // Just an array
         } else {
--- a/test/relayFeeCalculator.test.ts
+++ b/test/relayFeeCalculator.test.ts
@@ -1,7 +1,17 @@
 import dotenv from "dotenv";
 import hre from "hardhat";
 import { RelayFeeCalculator, QueryInterface } from "../src/relayFeeCalculator/relayFeeCalculator";
-import { toBNWei, toBN, toGWei, TransactionCostEstimate, bnOne, bnZero, getCurrentTime } from "../src/utils";
+import {
+  toBNWei,
+  toBN,
+  toGWei,
+  TransactionCostEstimate,
+  bnOne,
+  bnZero,
+  getCurrentTime,
+  spreadEvent,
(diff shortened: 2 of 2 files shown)


james-a-morris:
This change supersedes #605

nicholaspai (approved):

pxrl (approved):

james-a-morris:
# Post Mortem: 

### Issue:
There is a bug in the `EventUtils#spreadEvent` function where certain fields do not trigger the recurrence step to parse nested objects.

### Problem Code Segment:
```
Object.keys(args[key].filter((key: string) => isNaN(+key))).length > 0
```

### Reasoning:
This line is used in a ternary to either return the `args[key]` object directly (to assume it's already been parsed) or to execute the recurrence step with `spreadEvent(args[key])`. The intended purpose of this ternary is to check if **any keys in `args[key]` are non-numeric**. This constraint is not being tested, however. Let's decompose:

```
// Errant Line:
Object.keys(args[key].filter((key: string) => isNaN(+key))).length > 0

// Break down the statement into variables
const arrayToTest = args[key].filter((key: string) => isNaN(+key))
const keys = Object.keys(arrayToTest);
keys.length > 0
```
What we are actually testing is whether **args[key] has any non-numeric string in the array.** This is wrong.

### Why this hasn't been resolved yet:

This is a subtle bug because 
 1.  We don't have many nested types in our contract events.
 2. Of the types we do have, we only refer to their nested fields in 3 locations.
 3. For this bug, only fills with non-empty messages can trigger this (reasons explained below)
 4. All logic related to referencing the `relayExecutionInfo` is related to slow filling & we don't slow fill messages.

For fills with empty messages, this is a non-issue. The reason is as followed:

```
    // Empty Message
    const relayExecutionInfo = [
        "0xabcdef....", // Address 
        "0x", // Empty Message
        "0xabcdef...", // Output Amount 
        0 // Filltype
    ]

    // Run down of `Object.keys(args[key].filter((key: string) => isNaN(+key))).length > 0`
   args[key].filter((key: string) => isNaN(+key)) -> ["0x"] // 0x is not a number.
   Object.keys( ^ ).length > 0 // true
```
Whereas
```
    // Populated Message
    const relayExecutionInfo = [
        "0xabcdef....", // Address 
        "0xabcdef...", // Filled Message
        "0xabcdef...", // Output Amount 
        0 // Filltype
    ]

    // Run down of `Object.keys(args[key].filter((key: string) => isNaN(+key))).length > 0`
   args[key].filter((key: string) => isNaN(+key)) -> [] // isNaN considers hex strings valid
   Object.keys( ^ ).length > 0 // false & recurrence step not entered
```

nicholaspai (approved):

PR #608, opened 2024-03-27
refactor: Remove all identified v2-specific code

pxrl:
Summary:
- Remove all V2* type definitions.
- All V3 prefixes are removed (i.e. V3Deposit -> Deposit).
- All helpers for v2/v3 type discrimination are removed.


src/clients/SpokePoolClient.ts
```diff
@@ -364,11 +345,9 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @param fill The fill to find a corresponding deposit for.
    * @returns The corresponding deposit if found, undefined otherwise.
    */
-  public getDepositForFill(fill: Fill, fillFieldsToIgnore: string[] = []): DepositWithBlock | undefined {
+  public getDepositForFill(fill: Fill): DepositWithBlock | undefined {
     const depositWithMatchingDepositId = this.depositHashes[this.getDepositHash(fill)];
-    return validateFillForDeposit(fill, depositWithMatchingDepositId, fillFieldsToIgnore)
-      ? depositWithMatchingDepositId
-      : undefined;
+    return validateFillForDeposit(fill, depositWithMatchingDepositId) ? depositWithMatchingDepositId : undefined;
```
nicholaspai, line 350: Should we have this function return V3 type?

src/clients/SpokePoolClient.ts
```diff
@@ -954,7 +797,7 @@ export class SpokePoolClient extends BaseAbstractClient {
     return this.oldestTime;
   }
 
-  async findDepositV3(depositId: number, destinationChainId: number, depositor: string): Promise<V3DepositWithBlock> {
+  async findDeposit(depositId: number, destinationChainId: number, depositor: string): Promise<DepositWithBlock> {
```
nicholaspai, line 800: Should this be a V3 deposit

nicholaspai (approved):
I think this might break a lot of things in relayer V2 but curious if you've tried that locally

src/clients/SpokePoolClient.ts
pxrl: This change removes all V2 types, so everything is implicitly a V3 type anyway :+1:

pxrl:
> I think this might break a lot of things in relayer V2 but curious if you've tried that locally

Here's the minimum diff that works for me on relayer-v2: https://github.com/across-protocol/relayer-v2/pull/1360.

src/clients/SpokePoolClient.ts
pxrl: Same as https://github.com/across-protocol/sdk-v2/pull/608#discussion_r1542164480

PR #611, opened 2024-03-28
improve(SpokePoolClient): Support update timeout

pxrl:
Allow the caller to specify a timeout on SpokePoolClient.update(). The caller can then use `SpokePoolClient.isUpdated` to determine whether the update was successful.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -1,6 +1,7 @@
 import assert from "assert";
 import { BigNumber, Contract, Event, EventFilter, ethers } from "ethers";
 import winston from "winston";
+import { isError, isEthersError } from "../typeguards";
 import {
   AnyObject,
   bnZero,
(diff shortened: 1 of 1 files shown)


pxrl:
> OOC is there a way to kill the other promise if we do experience a timeout? Or is that the purpose of `Promise.race()`

I wasn't able to find any simple way of doing this unfortunately. Promise.race() just leaves it running in the background, but it seems de-anchored from where it was called, so it just resolves and then gets thrown away. I've made sure that `SpokePoolClient._update()` doesn't modify global SpokePoolClient state, so there shouldn't be any side-effects from this. It would certainly be nice if we could proactively axe it in the event of a timeout though; from what I've understood we'd need a separate utility to handle this.

src/clients/SpokePoolClient.ts
```diff
@@ -677,16 +686,30 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    if (timeout !== Number.POSITIVE_INFINITY) {
+      assert(timeout > 0);
+      const updateTimeout = async (timeout: number): Promise<SpokePoolUpdate> => {
+        await delay(timeout);
+        return { success: false, reason: `Timed out after ${timeout} seconds` };
+      };
+      updates.push(updateTimeout(timeout));
+    }
+
+    const update: SpokePoolUpdate = await Promise.race(updates);
```
pxrl, line 707: todo: Need to wrap this in a `try...catch` to also handle RPC errors.

pxrl: Fixed: https://github.com/across-protocol/sdk-v2/pull/611/commits/9282a96b8ab0ee8fb34ebcae21490fec7327ffe9.

nicholaspai (approved):

PR #614, opened 2024-04-02
refactor: Remove realizedLpFeePct from Deposit type

pxrl:
This has some breaking changes for the relayer repository so it's marked WIP for now.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -600,8 +600,7 @@ export class SpokePoolClient extends BaseAbstractClient {
 
         // Derive and append the common properties that are not part of the onchain event.
         const { quoteBlock: quoteBlockNumber } = dataForQuoteTime[index];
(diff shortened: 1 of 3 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -599,10 +584,6 @@ export class SpokePoolClient extends BaseAbstractClient {
       }
     }
 
-    // For each depositEvent, compute the realizedLpFeePct. Note this means that we are only finding this value on the
-    // new deposits that were found in the searchConfig (new from the previous run). This is important as this operation
-    // is heavy as there is a fair bit of block number lookups that need to happen. Note this call REQUIRES that the
-    // hubPoolClient is updated on the first before this call as this needed the the L1 token mapping to each L2 token.
```
pxrl, line 605: We throw on entry to `_update()` if the HubPoolClient is not updated, so I think we can remove this comment.

PR #615, opened 2024-04-02
refactor(utils): Defer to constants-v2 for chain definitions

pxrl:
This centalises our chain definitions and eliminates an annoying import workaround that was previously needed for PUBLIC_NETWORKS to be used in the FE.

### DIFF
--- a/src/utils/NetworkUtils.ts
+++ b/src/utils/NetworkUtils.ts
@@ -1,5 +1,10 @@
 import { CHAIN_IDs, PRODUCTION_CHAIN_IDS, PUBLIC_NETWORKS, TESTNET_CHAIN_IDS } from "../constants";
 
+const hreNetworks: Record<number, string> = {
+  666: "Hardhat1",
+  1337: "Hardhat2",
+};
+
 /**
  * Resolves a network name from a network id.
  * @param networkId The network id to resolve the name for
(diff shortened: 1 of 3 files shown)


dohaki (approved):
Awesome!

src/utils/NetworkUtils.ts
```diff
@@ -1,39 +1,22 @@
...
  * Resolves a network name from a network id.
  * @param networkId The network id to resolve the name for
  * @returns The network name for the network id. If the network id is not found, returns "unknown"
  */
 export function getNetworkName(networkId: number | string): string {
+  networkId = Number(networkId);
   try {
-    const networkName = PublicNetworks[Number(networkId)].name;
+    const networkName = PUBLIC_NETWORKS[networkId].name;
     return networkName.charAt(0).toUpperCase() + networkName.slice(1);
```
pxrl, line 12: @dohaki @james-a-morris Are we dependent on this behaviour anywhere in the FE? I'd rather zap it and just inherit the naming specified in constants-v2 but just want to make sure that it doesn't break anything.

dohaki: No, AFAICT we don't use that function in the FE

nicholaspai (approved):

PR #626, opened 2024-04-11
feat(network): include chainIs* functions

james-a-morris:
This will help consolidate repeated code in the finalizer.


pxrl (approved):
It might be dejavu, but I thought I'd already approved this exact PR yesterday :thinking:

PR #628, opened 2024-04-12
chore(clients): Export SpokePoolUpdate type

pxrl:
This is needed in relayer-v2 where the SpokePoolClient will be overridden.


james-a-morris (approved):
Should we also bump the version?

PR #632, opened 2024-04-16
chore: Remove v2 deposit validation

pxrl:


test/validatorUtils.test.ts
```diff
@@ -8,29 +8,9 @@ import { Deposit } from "../src/interfaces";
 
 describe("validatorUtils", () => {
   describe("isDeposit", () => {
-    let deposit: interfaces.V2DepositWithBlock, depositV3: interfaces.V3DepositWithBlock;
```
pxrl, line 11: `V3DepositWithBlock` hasn't been defined for quite some time. Why did this not error before :face_exhaling:

james-a-morris: This is strange. Should we bump the strictness for TSC in the tests?

PR #636, opened 2024-04-23
improve(clients): Centralise EventSearchConfig update

pxrl:
Don't log any warning if update() is called multiple times within the same block. This permits the relayer to be updated/looped repeatedly without emitting noisy warnings.

### DIFF
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -273,9 +273,6 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     const searchConfig = await this.updateSearchConfig(this.configStore.provider);
     if (isUpdateFailureReason(searchConfig)) {
       const reason = searchConfig;
-      if (reason === UpdateFailureReason.BadRequest) {
-        this.logger.warn({ at: "AcrossConfigStore", message: "Invalid update() searchConfig." });
-      }
       return { success: false, reason };
     }
 
@@ -308,6 +305,11 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
   async update(): Promise<void> {
     const result = await this._update();
     if (!result.success) {
+      if (result.reason !== UpdateFailureReason.AlreadyUpdated) {
+        throw new Error(`Unable to update ConfigStoreClient: ${result.reason}`);
+      }
+
+      // No need to touch `this.isUpdated` because it should already be set from a previous update.
       return;
     }
     const { chainId } = result;
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -489,6 +489,7 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     this.hasLatestConfigStoreVersion = this.hasValidConfigStoreVersionForTimestamp();
     this.latestBlockSearched = result.searchEndBlock;
     this.firstBlockToSearch = result.searchEndBlock + 1; // Next iteration should start off from where this one ended.
+    this.eventSearchConfig.toBlock = undefined; // Caller can re-set on subsequent updates if necessary
     this.chainId = this.chainId ?? chainId; // Update on the first run only.
     this.isUpdated = true;
 
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -33,9 +33,9 @@ import {
   UBAParsedConfigType,
 } from "../../interfaces";
 import { parseJSONWithNumericString } from "../../utils/JSONUtils";
-import { BaseAbstractClient } from "../BaseAbstractClient";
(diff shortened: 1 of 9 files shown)


src/clients/BaseAbstractClient.ts
```diff
@@ -37,6 +51,34 @@ export abstract class BaseAbstractClient {
...
+  public async updateSearchConfig(provider: providers.Provider): Promise<EventSearchConfig | UpdateFailureReason> {
+    const fromBlock = this.firstBlockToSearch;
+    let { toBlock } = this.eventSearchConfig;
+    if (isDefined(toBlock)) {
+      if (fromBlock > toBlock) {
+        return UpdateFailureReason.BadRequest;
+      }
+    } else {
+      try {
+        toBlock = await provider.getBlockNumber();
```
nicholaspai, line 69: Would this result in failure cases where we send 2x the # of rpc requests per failure? in other words, do you see this  extra rpc request making failures even noisier?

src/clients/HubPoolClient.ts
```diff
@@ -730,20 +735,19 @@ export class HubPoolClient extends BaseAbstractClient {
...
 
-    const searchConfig = {
-      fromBlock: this.firstBlockToSearch,
-      toBlock: this.eventSearchConfig.toBlock || (await this.hubPool.provider.getBlockNumber()),
-      maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
-    };
-    if (searchConfig.fromBlock > searchConfig.toBlock) {
-      this.logger.warn({ at: "HubPoolClient#_update", message: "Invalid update() searchConfig.", searchConfig });
-      return { success: false };
+    if (isUpdateFailureReason(searchConfig)) {
```
nicholaspai, line 740: This code block is repeated, should we refactor?

nicholaspai (approved):
I like this PR just have some suggestions on refactoring

src/clients/HubPoolClient.ts
```diff
@@ -807,9 +816,7 @@ export class HubPoolClient extends BaseAbstractClient {
     }
     const update = await this._update(eventsToQuery);
     if (!update.success) {
-      // This failure only occurs if the RPC searchConfig is miscomputed, and has only been seen in the hardhat test
-      // environment. Normal failures will throw instead. This is therefore an unfortunate workaround until we can
-      // understand why we see this in test. @todo: Resolve.
+      this.isUpdated = update.reason === UpdateFailureReason.AlreadyUpdated;
```
mrice32, line 819: Does this mean that we will go from being updated to not updated in some cases? I think previously, we were crashing on RPC errors instead of returning success=false.

IIUC, that means that RPC errors will show up as the hub pool not being updated, causing the program to crash when that is checked. This will also mean the error will say that the hub pool was not updated.

Is that desired?

The concern is two-fold:
1. Less transparency about _what_ the RPC error is when it crashes.
2. Error message about hub pool not being updated is a bit of a red herring.

Thoughts? Do you agree with this analysis or am I misunderstanding the change?

mrice32 (commented):
Looks good! One question around error handling

src/clients/BaseAbstractClient.ts
```diff
@@ -37,6 +51,34 @@ export abstract class BaseAbstractClient {
...
+      } catch (err) {
+        return UpdateFailureReason.RPCError;
+      }
+      if (toBlock < fromBlock) {
```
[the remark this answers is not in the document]
pxrl: It happens when we do multiple update() calls in the same block. At the conclusion of the previous update, we set `this.firstBlockToSearch = latestBlockSearched + 1`. So then when update() is subsequently called again, it resolves that `latestBlock === latestBlockSearched`, which gives `fromBlock = toBlock + 1`.

src/clients/BaseAbstractClient.ts
pxrl: I don't think it sends any more requests than the current implementation, but I might be missing a scenario. At present the various clients typically issue the `getBlockNumber()` call on each update because in most cases `eventSearchConfig.toBlock` is not set.

src/clients/HubPoolClient.ts
pxrl: Fair call RE suppressing RPC errors. I think we'd be logging these at the Provider level, but it's not a good idea to rely on that, and if an update fails on ConfigStore or HubPool then we probably want to bail anyway. I've updated the PR to back that part out, such that RPC provider exceptions won't be caught anymore.

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -309,6 +305,11 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
   async update(): Promise<void> {
     const result = await this._update();
     if (!result.success) {
+      if (result.reason !== UpdateFailureReason.AlreadyUpdated) {
+        throw new Error(`Unable to update ConfigStoreClient: ${result.reason}`);
+      }
+
+      // No need to touch `this.isUpdated` because it should already be set from a previous update.
```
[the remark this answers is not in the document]
pxrl: If it "fails" due to the mainnet `toBlock` being unchanged from previous, then it remains as-is. But if it fails for other reasons, then it should just throw.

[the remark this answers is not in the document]
james-a-morris: I like it - +1

james-a-morris (approved):

mrice32 (approved):
LGTM!

PR #637, opened 2024-04-23
improve(HubPoolClient): multicall currentTime + pendingRootBundleProposal

pxrl:


src/clients/HubPoolClient.ts
```diff
@@ -768,11 +768,21 @@ export class HubPoolClient extends BaseAbstractClient {
...
-      this.hubPool.rootBundleProposal({ blockTag: searchConfig.toBlock }),
+
+    const { hubPool } = this;
+    const multicallFunctions = ["getCurrentTime", "rootBundleProposal"];
+    const [multicallOutput, ...events] = await Promise.all([
+      hubPool.callStatic.multicall(
+        multicallFunctions.map((f) => hubPool.interface.encodeFunctionData(f)),
+        { blockTag: searchConfig.toBlock }
+      ),
       ...eventSearchConfigs.map((config) => paginatedEventQuery(this.hubPool, config.filter, config.searchConfig)),
```
pxrl, line 779:
```suggestion
      ...eventSearchConfigs.map((config) => paginatedEventQuery(hubPool, config.filter, config.searchConfig)),
```

nicholaspai (approved):

src/clients/HubPoolClient.ts
```diff
@@ -768,11 +768,21 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    const [multicallOutput, ...events] = await Promise.all([
+      hubPool.callStatic.multicall(
+        multicallFunctions.map((f) => hubPool.interface.encodeFunctionData(f)),
+        { blockTag: searchConfig.toBlock }
+      ),
+      ...eventSearchConfigs.map((config) => paginatedEventQuery(hubPool, config.filter, config.searchConfig)),
     ]);
+    const [currentTime, pendingRootBundleProposal] = multicallFunctions.map((fn, idx) => {
+      const output = hubPool.interface.decodeFunctionResult(fn, multicallOutput[idx]);
+      return BigNumber.isBigNumber(output) ? output.toNumber() : spreadEvent(output);
```
mrice32, line 783: Why is this conversion needed? Doesn't seem like that was happening in the previous version.

mrice32 (commented):
Looks good! Just one question

src/clients/HubPoolClient.ts
```diff
@@ -768,11 +768,21 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    });
```
james-a-morris, line 784: Because we have a small, fixed number of functions that we want to multicall here, could we manually write out the operations? 

```suggestion
    const { hubPool } = this;
    const multicallFunctions = ["getCurrentTime", "rootBundleProposal"];
    const [[currentTimeMulticall, rootBundleMulticall], ...events] = await Promise.all([
      hubPool.callStatic.multicall(
        [
            hubPool.interface.encodeFunctionData("getCurrentTime"),
            hubPool.interface.encodeFunctionData("rootBundleProposal")
        ],
        { blockTag: searchConfig.toBlock }
      ),
      ...eventSearchConfigs.map((config) => paginatedEventQuery(hubPool, config.filter, config.searchConfig)),
    ]);
    const [currentTime, pendingRootBundleProposal] = [
        hubPool.interface.decodeFunctionResult("getCurrentTime", currentTimeMulticall).toNumber(),
        spreadEvent(hubPool.interface.decodeFunctionResult("rootBundleMulticall", rootBundleMulticall))
    ]
```

james-a-morris (commented):
One comment spurred by the number of functions we're multicalling

src/clients/HubPoolClient.ts
bmzig: Isn't this just because `multicall` returns `bytes[]` and not the deserialized types?

src/clients/HubPoolClient.ts
bmzig: This does make the code more readable, too.

bmzig (approved):

src/clients/HubPoolClient.ts
mrice32: Sorry, I'm talking about the conversion of line 783. The one on 782 makes sense (decoding the multicall).

src/clients/HubPoolClient.ts
pxrl: I avoided that construction because it requires the ordering of function names to be correct in 3 separate places, two of which are as string literals. The advantage of specifying it up front in the `multicallFunctions` array is that the subsequent decoding is guaranteed to occur in the correct order, so there's only 2 places that need to be aligned (the ordering of `multicallFunctions` and the resulting dereferencing at line 781), rather than 3.

PR #639, opened 2024-04-24
improve(SpokePoolClient): Protected setter for deposit route updates

pxrl:
Makes it easier to update SpokePoolClient.depositRoutes from other places.


nicholaspai (approved):

pxrl:
Not required anymore, since we're dropping these events entirely.

PR #640, opened 2024-04-26
improve(SpokePoolClient): Delete some depositRoute functions

nicholaspai:
We don't rely on these functions [anymore](https://github.com/across-protocol/relayer-v3/commit/d1a6c65478fed946df57958a74d2a4f6a4f6192a) in the Relayer so delete them to avoid future dependencies on them.

We favor instead mapping l1 and l2 tokens using PoolRebalanceRoutes which requires querying only the Hub chain as opposed to all Spoke chains, which is overall much faster.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -169,24 +169,6 @@ export class SpokePoolClient extends BaseAbstractClient {
   }
 
   /**
-   * Determines whether a deposit route is enabled for the given origin token and destination chain ID.
-   * @param originToken The origin token address.
-   * @param destinationChainId The destination chain ID.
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -168,24 +168,6 @@ export class SpokePoolClient extends BaseAbstractClient {
     return this.depositRoutes;
   }
```
pxrl, line 169: Should we also drop this, and `this.depositRoutes` entirely?

nicholaspai: We use this in the `scripts/disableDepositRoutes` script that's why I kept it

pxrl (approved):
LGTM

PR #643, opened 2024-05-10
[WIP] improve(HubPoolClient): Don't throw on token resolution failure

pxrl:
WIP - there's lots of users of this function so there's a bit of fallout from the change. I'll update it progressively as each case is addressed.

This is too violent for a lot of use cases and requires callers to wrap their calls in try/catch statements, which has its own set of problems. Instead, just return undefined if the requested token can't be resolved and allow the caller to decide on how to respond to that.

### DIFF
--- a/src/clients/BundleDataClient/utils/DataworkerUtils.ts
+++ b/src/clients/BundleDataClient/utils/DataworkerUtils.ts
@@ -18,6 +18,7 @@ import {
   fixedPointAdjustment,
   count2DDictionaryValues,
   count3DDictionaryValues,
+  isDefined,
 } from "../../../utils";
 import {
   addLastRunningBalance,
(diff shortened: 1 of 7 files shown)


src/clients/BundleDataClient/utils/DataworkerUtils.ts
```diff
@@ -160,8 +161,10 @@ export function _buildPoolRebalanceRoot(
           mainnetBundleEndBlock
         );
 
-        updateRunningBalance(runningBalances, repaymentChainId, l1TokenCounterpart, totalRefundAmount);
-        updateRunningBalance(realizedLpFees, repaymentChainId, l1TokenCounterpart, totalRealizedLpFee);
+        if (isDefined(l1TokenCounterpart)) {
```
bmzig, line 164: Leaving a comment here since this is just a temporary fix to let the code compile. I'm pretty sure this is a bad idea to do since if `l1TokenCounterpart` is undefined, then we probably do want to throw an error. (Maybe we should just assert that here).

PR #647, opened 2024-05-16
chore(refactor): relayFeeCalculator queries

james-a-morris:
These classes are dynamically resolved with minimal hardcoding. We can condense into a single factory that will make adding new chains more dynamic.

Instead of `new EthereumQueries(...)` we'd call `QueryBase__factory.create(1, ...)`

Closes ACX-2147.


nicholaspai (approved):

dohaki (approved):
LGTM 👍

PR #657, opened 2024-05-31
[WIP] improve(utils): Avoid recomputing relayData hashes

pxrl:
This will save redundant recomputation cycles of a relayData hash when it is needed more than once.

This PR will be updated with tests.

### DIFF
--- a/src/interfaces/SpokePool.ts
+++ b/src/interfaces/SpokePool.ts
@@ -19,6 +19,7 @@ export interface RelayData {
   fillDeadline: number;
   exclusiveRelayer: string;
   exclusivityDeadline: number;
+  _hash?: Record<number, string>;
 }
 
 export interface Deposit extends RelayData {
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -216,7 +216,9 @@ export async function getDepositIdAtBlock(contract: Contract, blockTag: number):
  * @returns The corresponding RelayData hash.
  */
 export function getRelayDataHash(relayData: RelayData, destinationChainId: number): string {
-  return ethersUtils.keccak256(
+  relayData._hash ??= {};
+
+  return relayData._hash[destinationChainId] ??= ethersUtils.keccak256(
     ethersUtils.defaultAbiCoder.encode(
       [
         "tuple(" +


src/interfaces/SpokePool.ts
```diff
@@ -19,6 +19,7 @@ export interface RelayData {
   fillDeadline: number;
   exclusiveRelayer: string;
   exclusivityDeadline: number;
+  _hash?: Record<number, string>;
```
pxrl, line 22: It'd be a fair bit more efficient for this to be a simple string, but the `RelayData` type doesn't include the `destinationChainId`, so the same RelayData object can map to multiple hashes depending on the value of `destinationChainId`. In the interests of safety it seems best to use a record, but I'd love to find a way to use a string instead.

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -652,22 +652,24 @@ describe("SpokePoolClient: Fill Validation", function () {
...
-    expect(validateFillForDeposit({ ...validFill, depositId: 1337 }, deposit_2)).to.be.false;
+    expect(validateFillForDeposit({ ...validFill, depositId: 1337, _hash }, deposit_2)).to.be.false;
 
     // Changed the depositor.
-    expect(validateFillForDeposit({ ...validFill, depositor: relayer.address }, deposit_2)).to.be.false;
+    expect(validateFillForDeposit({ ...validFill, depositor: relayer.address, _hash }, deposit_2)).to.be.false;
 
     // Changed the recipient.
-    expect(validateFillForDeposit({ ...validFill, recipient: relayer.address }, deposit_2)).to.be.false;
+    expect(validateFillForDeposit({ ...validFill, recipient: relayer.address, _hash }, deposit_2)).to.be.false;
```
pxrl, line 673: These tests exposed a sharp edge - producing a new relayData event based on an existing one, with some modified fields, risks inadvertently copying the old _hash. It needs to be manually unset. It's a pretty big foot gun...

src/interfaces/SpokePool.ts
mrice32: Does that mean the hash really should be one level higher where the destination chain is present (i.e. the hash should be in fill and deposit)?

mrice32 (commented):
This is awesome! +1

src/interfaces/SpokePool.ts
pxrl: Yeah, that's a good option. It introduces a little bit of duplication but it's nothing major. I'll think about another way of passing in the hash if it's not defined in `RelayData`.

PR #660, opened 2024-06-11
refactor: improve `getGasCosts` rpc calls

dohaki:
While monitoring the speed of the `GET /suggested-fees` handler, I noticed some areas of improvement. For example, in the trace of the attached screenshot, you can see all RPC requests triggered by the `relayFeeCalculator`. This PR aims to reduce the number of those and parallelize a few by:

- allow passing in `gasPrice` to `getGasCosts` to allow caching of that value to prevent RPC calls
- allow passing `chainId` to `estimateTotalGasRequiredByUnsignedTransaction` to prevent additional RPC if chain already known
- for OP stack:
  - set `gasLimit` with already retrieved value to prevent redundant RPC call
  - concurrently call parts of `estimateTotalGasCost` for speed up

image

### DIFF
--- a/src/relayFeeCalculator/chain-queries/baseQuery.ts
+++ b/src/relayFeeCalculator/chain-queries/baseQuery.ts
@@ -58,16 +58,22 @@ export class QueryBase implements QueryInterface {
    * Retrieves the current gas costs of performing a fillRelay contract at the referenced SpokePool.
    * @param deposit V3 deposit instance.
    * @param relayerAddress Relayer address to simulate with.
+   * @param gasPrice Optional gas price to use for the simulation.
    * @returns The gas estimate for this function call (multplied with the optional buffer).
    */
-  async getGasCosts(deposit: Deposit, relayer = DEFAULT_SIMULATED_RELAYER_ADDRESS): Promise<TransactionCostEstimate> {
+  async getGasCosts(
+    deposit: Deposit,
+    relayer = DEFAULT_SIMULATED_RELAYER_ADDRESS,
+    gasPrice = this.fixedGasPrice
+  ): Promise<TransactionCostEstimate> {
     const tx = await populateV3Relay(this.spokePool, deposit, relayer);
     return estimateTotalGasRequiredByUnsignedTransaction(
       tx,
       relayer,
       this.provider,
       this.gasMarkup,
-      this.fixedGasPrice
+      gasPrice,
+      deposit.destinationChainId
     );
   }
 
--- a/src/relayFeeCalculator/chain-queries/baseQuery.ts
+++ b/src/relayFeeCalculator/chain-queries/baseQuery.ts
@@ -67,14 +67,7 @@ export class QueryBase implements QueryInterface {
     gasPrice = this.fixedGasPrice
   ): Promise<TransactionCostEstimate> {
     const tx = await populateV3Relay(this.spokePool, deposit, relayer);
-    return estimateTotalGasRequiredByUnsignedTransaction(
-      tx,
-      relayer,
-      this.provider,
-      this.gasMarkup,
-      gasPrice,
-      deposit.destinationChainId
-    );
+    return estimateTotalGasRequiredByUnsignedTransaction(tx, relayer, this.provider, this.gasMarkup, gasPrice);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/utils/common.ts
```diff
@@ -267,11 +269,20 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
-    tokenGasCost = await provider.estimateTotalGasCost(populatedTransaction);
+    const populatedTransaction = await voidSigner.populateTransaction({
+      ...unsignedTx,
+      gasLimit: nativeGasCost, // prevents additional gas estimation call
+    });
+    // Concurrently estimate the gas cost on L1 and L2 instead of calling
+    // `provider.estimateTotalGasCost` to improve performance.
+    const [l1GasCost, l2GasCost] = await Promise.all([
+      provider.estimateL1GasCost(populatedTransaction),
+      provider.estimateL2GasCost(populatedTransaction),
```
melisaguevara, line 280: Where are these functions defined?

dohaki: These are from the Optimism SDK https://sdk.optimism.io/modules#estimateL2GasCost. But can't find the source code on GitHub because the SDK is superseded now by https://viem.sh/op-stack

pxrl: These are OP-specific implementations (note `if (chainIsOPStack(chainId))` a few lines above)- see https://sdk.optimism.io/modules#estimateL1GasCost

melisaguevara: Thank u both!

melisaguevara (approved):
LGTM!

src/utils/common.ts
```diff
@@ -242,21 +242,23 @@ export type TransactionCostEstimate = {
...
+  gasPrice?: BigNumberish,
+  chainId?: number
 ): Promise<TransactionCostEstimate> {
   assert(
     gasMarkup > -1 && gasMarkup <= 4,
     `Require -1.0 < Gas Markup (${gasMarkup}) <= 4.0 for a total gas multiplier within (0, +5.0]`
   );
   const gasTotalMultiplier = toBNWei(1.0 + gasMarkup);
-  const { chainId } = await provider.getNetwork();
+  chainId ??= (await provider.getNetwork()).chainId;
```
pxrl, line 261: Did you confirm whether this actually produced an RPC call? I'd understood that we are using StaticJsonRpcProvider instances under the hood, and they _should_ suppress chainId lookups because they cache it internally. But it's also harder to know when the provider is typed broadly as `Provider` - and this means we don't get a concrete guarantee on it.

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -58,16 +58,22 @@ export class QueryBase implements QueryInterface {
...
    * @param deposit V3 deposit instance.
    * @param relayerAddress Relayer address to simulate with.
+   * @param gasPrice Optional gas price to use for the simulation.
    * @returns The gas estimate for this function call (multplied with the optional buffer).
    */
-  async getGasCosts(deposit: Deposit, relayer = DEFAULT_SIMULATED_RELAYER_ADDRESS): Promise<TransactionCostEstimate> {
+  async getGasCosts(
+    deposit: Deposit,
+    relayer = DEFAULT_SIMULATED_RELAYER_ADDRESS,
+    gasPrice = this.fixedGasPrice
```
pxrl, line 67: Is `fixedGasPrice` supplied by the FE during instantiation?

src/utils/common.ts
pxrl: @dohaki Running these in paralell is a nice observation based on the implementation of `estimateTotalGasCost()` here: https://github.com/ethereum-optimism/optimism/blob/ee058e413e0c7ff92d000593af648f3d79ca1a31/packages/sdk/src/l2-provider.ts#L133

pxrl (approved):
LGTM

src/utils/common.ts
dohaki: No, you are right actually. I checked and there is no `eth_chainId` call. I will revert this change then

image

src/relayFeeCalculator/chain-queries/baseQuery.ts
dohaki: No, not yet. But was planning to cache that somehow and pass that in

PR #662, opened 2024-06-17
feat(lite-chains): add lite chain support

james-a-morris:
This PR adds lite chain support. Additional PRs to follow:
1. Add ability to inject lite chain list for testing
2. Improve testing across this client to include chain id list + lite chain list.

### DIFF
--- a/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
+++ b/src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
@@ -58,6 +58,7 @@ export enum GLOBAL_CONFIG_STORE_KEYS {
   VERSION = "VERSION",
   DISABLED_CHAINS = "DISABLED_CHAINS",
   CHAIN_ID_INDICES = "CHAIN_ID_INDICES",
+  LITE_CHAIN_ID_INDICES = "LITE_CHAIN_ID_INDICES",
 }
 
 export class AcrossConfigStoreClient extends BaseAbstractClient {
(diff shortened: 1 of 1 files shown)


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -407,6 +420,29 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+        if (new Set(chainIndices).size !== chainIndices.length) {
+          this.logger.warn({
+            at: "ConfigStore",
+            message: `The array ${chainIndices} contains duplicates making it invalid.`,
+          });
+          // If not a valid array, skip.
+          continue;
+        }
+        // If all else passes, we can add this update.
+        this.liteChainIndicesUpdates.push({ ...args, value: chainIndices });
```
nicholaspai, line 445: Should this list be a subset of the enabled chain id list?

nicholaspai (commented):
Can we accompany this pr with a umip change draft so we can better verify it implements the umip change?

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: We could refine it to be this - this kind of check would need to happen after the original loop but this isn't out of scope

james-a-morris: Will see how we can fit this into the UMIP

nicholaspai (approved):
Lgtm but Alex and Melisa should review

melisaguevara (approved):
lgtm!

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -153,6 +155,17 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
   }
 
+  /**
+   * Resolves the lite chain ids that were available to the protocol at a given block range.
+   * @param blockNumber Block number to search for. Defaults to latest block.
+   * @returns List of lite chain IDs that were available to the protocol at the given block number.
+   * @note This dynamic functionality has been added after the launch of Across.
+   */
+  getLiteChainIdIndicesForBlock(blockNumber: number = Number.MAX_SAFE_INTEGER): number[] {
+    const liteChainIdList = sortEventsDescending(this.liteChainIndicesUpdates);
```
pxrl, line 165: Should we consider sorting this on update() so that we don't need to do it each time it's queried? I could imagine we might query this a lot in the relayer.

james-a-morris: Agreed - we also do this in other cases, like for the `disabledChainId` and `chainIdIndices.` It may be worth squeezing more performance by amortizing this in the `update`.

amateima: ➕

james-a-morris:
will follow this PR with optimizations from @pxrl

PR #663, opened 2024-06-18
feat(lite-chain): ensure all inbound fills have lite repayments

james-a-morris:
We can do this in the SDK upstream by changing the way the spokepool extracts fill information from on-chain. We can determine whether the fill originated from a lite-chain enabled deposit route and, if so, enforce the repayment on the origin.


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -166,6 +167,37 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+    return liteChainIdList.find((update) => update.timestamp <= timestamp)?.value ?? [];
+  }
+
+  /**
+   * Checks if a chain ID was a lite chain at a given block.
+   * @param chainId The chain ID to check.
+   * @param blockNumber The block number to check. Defaults to latest block.
+   * @returns True if the chain ID was a lite chain at the given block. False otherwise.
+   */
+  isChainLiteChainAtBlock(chainId: number, blockNumber: number = Number.MAX_SAFE_INTEGER): boolean {
```
nicholaspai, line 187: Where is this function used?

nicholaspai (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -866,4 +869,19 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+   * @param deposit The deposit to evaluate.
+   * @returns True if the deposit originates from a lite chain, false otherwise. If the hub pool client is not defined,
+   *          this method will return false.
+   */
+  protected doesDepositOriginateFromLiteChain(deposit: DepositWithBlock): boolean {
+    return (
+      this.hubPoolClient?.configStoreClient.isChainLiteChainAtTimestamp(
+        deposit.originChainId,
+        deposit.quoteTimestamp
+      ) ?? false
```
pxrl, line 884: I'm not sure about this logic - it forces the deposit to be a lite deposit for eternity. Isn't the "lite chain" check relevant for the fill instead? We could in theory have a deposit with long expiry made on a lite chain, but the fill could occur after it is removed from the set of lite chains.

james-a-morris: This is something I discussed with Nick. I initially thought I would put the check on the current block time because this is when the fill is made.

I think there's tradeoffs to determining the lite chain inclusion at fill time or at deposit time - will discuss offline

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: No longer used in the new update: removed here https://github.com/across-protocol/sdk/pull/663/commits/d5c74a91cd9cd7082ca5b4dce0931ecc6d39ad97

src/clients/SpokePoolClient.ts
james-a-morris: Resolved offline. Will move forward with this current direction

src/clients/SpokePoolClient.ts
```diff
@@ -634,6 +634,12 @@ export class SpokePoolClient extends BaseAbstractClient {
           ...(spreadEventWithBlockNumber(event) as FillWithBlock),
           destinationChainId: this.chainId,
         };
+
+        // If the fill originates on a lite chain, then the repayment chain ID must be set to the origin chain ID.
+        if (this.hubPoolClient?.configStoreClient?.isChainLiteChainAtBlock(fill.originChainId, fill.blockNumber)) {
```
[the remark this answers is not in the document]
james-a-morris: Correct - this is outdated code now, however

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -166,6 +167,27 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
 
+  /**
+   * Resolves the lite chain ids that were available to the protocol at a given timestamp.
+   * @param timestamp Timestamp to search for. Defaults to latest time - in seconds.
+   * @returns List of lite chain IDs that were available to the protocol at the given timestamp.
+   * @note This dynamic functionality has been added after the launch of Across.
+   */
+  getliteChainIdIndicesForTimestamp(timestamp: number = Number.MAX_SAFE_INTEGER): number[] {
+    const liteChainIdList = sortEventsDescending(this.liteChainIndicesUpdates);
+    return liteChainIdList.find((update) => update.timestamp <= timestamp)?.value ?? [];
```
nicholaspai, line 178: what does .value resolve to? are you sure that's the right syntax?

src/clients/SpokePoolClient.ts
```diff
@@ -866,4 +869,19 @@ export class SpokePoolClient extends BaseAbstractClient {
...
     return deposit;
   }
+
+  /**
+   * Determines whether a deposit originates from a lite chain.
+   * @param deposit The deposit to evaluate.
+   * @returns True if the deposit originates from a lite chain, false otherwise. If the hub pool client is not defined,
+   *          this method will return false.
+   */
+  protected doesDepositOriginateFromLiteChain(deposit: DepositWithBlock): boolean {
```
nicholaspai, line 879: can you add a regression test for this function? Also we need one in the sdk if it doesn't exist yet

nicholaspai (commented):
- one syntax comment
- one request for unit tests

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
james-a-morris: The variable `liteChainIndicesUpdates is defined as 
```
public liteChainIndicesUpdates: LiteChainsIdListUpdate[] = [];
```

The call to 
```
liteChainIdList.find((update) => update.timestamp <= timestamp)
```
returns
```
LiteChainsIdListUpdate<...> | undefined
```
which is defined as
```
export interface LiteChainsIdListUpdate<ValueStore = number[]> extends GlobalConfigUpdate<ValueStore> {
  timestamp: number;
}
```
and has its value stored in 
```
export interface GlobalConfigUpdate<ValueStore = number> extends SortableEvent {
  value: ValueStore;
}
```
which in this case `ValueStore = number[]`

src/clients/SpokePoolClient.ts
james-a-morris: We can't easily add a regression test to this specific function, but the underlying config store function has regression tests

nicholaspai: why can't you test this function?

test/ConfigStoreClient.ts
```diff
@@ -367,6 +367,43 @@ describe("AcrossConfigStoreClient", function () {
...
+      // Test the getliteChainIdIndicesForTimestamp function
+      expect(configStoreClient.getliteChainIdIndicesForTimestamp(timestampBeforeNewLiteChains)).to.deep.equal([]);
+      expect(configStoreClient.getliteChainIdIndicesForTimestamp(timestampAfterNewLiteChains)).to.deep.equal([1]);
+      // Test the getliteChainIdIndicesForBlock function
+      expect(configStoreClient.getLiteChainIdIndicesForBlock(blockBeforeNewLiteChains)).to.deep.equal([]);
+      expect(configStoreClient.getLiteChainIdIndicesForBlock(blockAfterNewLiteChains)).to.deep.equal([1]);
+
+      const update2 = await configStore.updateGlobalConfig(
+        utf8ToHex(GLOBAL_CONFIG_STORE_KEYS.LITE_CHAIN_ID_INDICES),
+        JSON.stringify([1, 15])
```
nicholaspai, line 395: What if there is an invalid update? i.e. what if there is a chain ID that isn't valid? or not a number?

james-a-morris: Resolved below

src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -166,6 +167,27 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
     return liteChainIdList.find((update) => update.blockNumber <= blockNumber)?.value ?? [];
   }
 
+  /**
+   * Resolves the lite chain ids that were available to the protocol at a given timestamp.
+   * @param timestamp Timestamp to search for. Defaults to latest time - in seconds.
+   * @returns List of lite chain IDs that were available to the protocol at the given timestamp.
+   * @note This dynamic functionality has been added after the launch of Across.
+   */
+  getliteChainIdIndicesForTimestamp(timestamp: number = Number.MAX_SAFE_INTEGER): number[] {
```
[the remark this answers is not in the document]
james-a-morris: https://github.com/across-protocol/sdk/pull/663/commits/4022cfda62d193039e592b071985ac99626162ba

src/clients/SpokePoolClient.ts
james-a-morris: Tested in follow up commits

amateima (approved):

PR #675, opened 2024-07-02
chore: Bump contracts

pxrl:


src/typechain.ts
pxrl, line 1: These changes were necessary after this commit in contracts: https://github.com/across-protocol/contracts/pull/543

dohaki (approved):

PR #678, opened 2024-07-03
chore: Add Blast constants

pxrl:


src/utils/NetworkUtils.ts
```diff
@@ -60,10 +60,12 @@ export function chainIsOPStack(chainId: number): boolean {
   return [
```
pxrl, line 60: nb. this will treat Blast as a generic OP stack implementation. I _think_ that's OK, but there will definitely be places where we need to override OP stack defaults (i.e. finalization challenge window settings).

bmzig (approved):

james-a-morris (commented):
Is Blast added to the list of PRODUCTION_CHAINS in constants?

bmzig:
> Is Blast added to the list of PRODUCTION_CHAINS in constants?

It is. The SDK is at the latest version of constants.

PR #679, opened 2024-07-04
fix(utils): update deposit schema with to/fromLiteChain

bmzig:
We are emitting a bunch of warning logs since we are saving deposits with `toLiteChain` and `fromLiteChain` to cache, which do not match with our current `V3DepositSchema`. A simple fix is to add these fields to the schema.


pxrl (approved):

james-a-morris (commented):
Good call - did we want to make this optional though?

What is the stack trace of the error and is it resolved in the integration tests?

pxrl (approved):

PR #680, opened 2024-07-05
fix: set gas limit and price for fill estimation

dohaki:
Some integrators flagged an issue for fee estimation when setting custom messages. After some digging, a workaround is to set a large enough hardcoded gas limit and a gas price when calling `eth_estimateGas`. E.g. the following `eth_estimateGas` call fails where `gas` and `gasPrice` are not set:
```bash
curl https://arb1.arbitrum.io/rpc \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"method":"eth_estimateGas","params":[{"from":"0x428ab2ba90eba0a4be7af34c9ac451ab061ac010","to":"0xe35e9842fceaca96570b734083f4a58e8f7c5f2a","data":"0x2e3781150000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000a4b1000000000000000000000000ccb0be73bc7386db701006762dd7f8638d176d48000000000000000000000000ccb0be73bc7386db701006762dd7f8638d176d4800000000000000000000000000000000000000000000000000000000000000000000000000000000000000003c499c542cef5e3811e1192ce70d8cc03d5c3359000000000000000000000000af88d065e77c8cc2239327c5edb3a432268e58310000000000000000000000000000000000000000000000000000000005f5e1000000000000000000000000000000000000000000000000000000000005f5e100000000000000000000000000000000000000000000000000000000000000008900000000000000000000000000000000000000000000000000000000ffffffff00000000000000000000000000000000000000000000000000000000ffffffff0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000000400000000000000000000000005f851f67d24419982ecd7b7765defd64fbb50a97000000000000000000000000eefef4a7992d93a3d6f1135cc8ea02cd5b863c2a"}],"id":1,"jsonrpc":"2.0"}'
```
This one on the other hand succeeds with limit and price set
```bash
curl https://arb1.arbitrum.io/rpc \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"method":"eth_estimateGas","params":[{"from":"0x428ab2ba90eba0a4be7af34c9ac451ab061ac010","to":"0xe35e9842fceaca96570b734083f4a58e8f7c5f2a","gas":"0x989680","gasPrice":"0x989680","data":"0x2e3781150000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000a4b1000000000000000000000000ccb0be73bc7386db701006762dd7f8638d176d48000000000000000000000000ccb0be73bc7386db701006762dd7f8638d176d4800000000000000000000000000000000000000000000000000000000000000000000000000000000000000003c499c542cef5e3811e1192ce70d8cc03d5c3359000000000000000000000000af88d065e77c8cc2239327c5edb3a432268e58310000000000000000000000000000000000000000000000000000000005f5e1000000000000000000000000000000000000000000000000000000000005f5e100000000000000000000000000000000000000000000000000000000000000008900000000000000000000000000000000000000000000000000000000ffffffff00000000000000000000000000000000000000000000000000000000ffffffff0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000000400000000000000000000000005f851f67d24419982ecd7b7765defd64fbb50a97000000000000000000000000eefef4a7992d93a3d6f1135cc8ea02cd5b863c2a"}],"id":1,"jsonrpc":"2.0"}'
```

This PR therefore proposes a change to allow us to pass a custom gas limit and price into the fee calculator.

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -242,6 +242,7 @@ export type TransactionCostEstimate = {
  * @param provider A valid ethers provider - will be used to reason the gas price.
  * @param gasMarkup Markup on the estimated gas cost. For example, 0.2 will increase this resulting value 1.2x.
  * @param gasPrice A manually provided gas price - if set, this function will not resolve the current gas price.
+ * @param gasLimit A manually provided gas limit - if set, this function will not estimate the gas limit.
  * @returns Estimated cost in units of gas and the underlying gas token (gasPrice * estimatedGasUnits).
  */
 export async function estimateTotalGasRequiredByUnsignedTransaction(
@@ -249,7 +250,8 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
   senderAddress: string,
   provider: providers.Provider | L2Provider<providers.Provider>,
   gasMarkup: number,
-  gasPrice?: BigNumberish
+  gasPrice?: BigNumberish,
+  gasLimit?: BigNumberish
 ): Promise<TransactionCostEstimate> {
   assert(
     gasMarkup > -1 && gasMarkup <= 4,
@@ -262,8 +264,8 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
   // Estimate the Gas units required to submit this transaction.
   let nativeGasCost = await voidSigner.estimateGas({
     ...unsignedTx,
-    gasPrice: gasPrice ?? undefined,
-    gasLimit: 8_000_000, // This prevents out of gas errors.
+    gasPrice,
+    gasLimit,
   });
   let tokenGasCost: BigNumber;
 
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -260,13 +260,16 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
   const voidSigner = new VoidSigner(senderAddress, provider);
 
   // Estimate the Gas units required to submit this transaction.
-  let nativeGasCost = await voidSigner.estimateGas(unsignedTx);
+  let nativeGasCost = await voidSigner.estimateGas({
+    ...unsignedTx,
+    gasPrice: gasPrice ?? undefined,
+    gasLimit: 8_000_000, // This prevents out of gas errors.
+  });
   let tokenGasCost: BigNumber;
 
   // OP stack is a special case; gas cost is computed by the SDK, without having to query price.
   if (chainIsOPStack(chainId)) {
     assert(isOptimismL2Provider(provider), `Unexpected provider for chain ID ${chainId}.`);
-    assert(gasPrice === undefined, `Gas price (${gasPrice}) supplied for Optimism gas estimation (unused).`);
     const populatedTransaction = await voidSigner.populateTransaction({
       ...unsignedTx,
       gasLimit: nativeGasCost, // prevents additional gas estimation call
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -278,10 +278,11 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
     });
     // Concurrently estimate the gas cost on L1 and L2 instead of calling
     // `provider.estimateTotalGasCost` to improve performance.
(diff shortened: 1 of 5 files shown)


src/utils/common.ts
```diff
@@ -260,13 +262,16 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
+    ...unsignedTx,
+    gasPrice,
+    gasLimit,
+  });
   let tokenGasCost: BigNumber;
 
   // OP stack is a special case; gas cost is computed by the SDK, without having to query price.
   if (chainIsOPStack(chainId)) {
     assert(isOptimismL2Provider(provider), `Unexpected provider for chain ID ${chainId}.`);
-    assert(gasPrice === undefined, `Gas price (${gasPrice}) supplied for Optimism gas estimation (unused).`);
```
pxrl, line 269: Not sure how to handle this, since the gas price is never used for the OP stack deployments... If the caller supplies a gas price and we return a result, I think it's reasonable for them to expect that their gas price was actually used.

dohaki: That's a good point. I added a commit here https://github.com/across-protocol/sdk/pull/680/commits/87575e72ba58f400e52b6ecae06a0d8b958b57d9 that passes on the gas price if provided. This even saves some RPC calls which is also nice.

src/utils/common.ts
```diff
@@ -278,10 +278,11 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
     });
     // Concurrently estimate the gas cost on L1 and L2 instead of calling
     // `provider.estimateTotalGasCost` to improve performance.
-    const [l1GasCost, l2GasCost] = await Promise.all([
+    const [l1GasCost, l2GasPrice] = await Promise.all([
       provider.estimateL1GasCost(populatedTransaction),
-      provider.estimateL2GasCost(populatedTransaction),
```
dohaki, line 278: Here is the implementation in the optimism SDK.
```ts
/**
 * Estimates the L2 gas cost for a given L2 transaction in wei.
 *
 * @param l2Provider L2 provider to query the gas usage from.
 * @param tx Transaction to estimate L2 gas cost for.
 * @returns Estimated L2 gas cost.
 */
export const estimateL2GasCost = async (
  l2Provider: ProviderLike,
  tx: TransactionRequest
): Promise<BigNumber> => {
  const parsed = toProvider(l2Provider)
  const l2GasPrice = await parsed.getGasPrice()
  const l2GasCost = await parsed.estimateGas(tx)
  return l2GasPrice.mul(l2GasCost)
}

```

james-a-morris (commented):
Is this PR ready to be merged or is this a WiP? @pxrl

james-a-morris (commented):
Would this be giving the user of the sdk too much flexibility?

dohaki:
> Would this be giving the user of the sdk too much flexibility?

Hmm, can you elaborate what you mean with too much flexibility? 

But I think we could break this PR up into two parts:
1. Allowing the gasPrice override - this is actually also a possible performance boost. This allows us to cache and use prefetched gas prices which will save some RPC calls.
2. Allowing the gasLimit override

I think 1. should definitely get in. With 2. I am indifferent but I don't see any risk tbh. Especially if it is allowed at RPC call level.

pxrl (approved):
> But I think we could break this PR up into two parts:
> 
> 1. Allowing the gasPrice override - this is actually also a possible performance boost. This allows us to cache and use prefetched gas prices which will save some RPC calls.
> 2. Allowing the gasLimit override
> 
> I think 1. should definitely get in. With 2. I am indifferent but I don't see any risk tbh. Especially if it is allowed at RPC call level.

Agree; I think #1 is a good addition. The use case for #2 has evaporated a bit since the underlying issue w/ the integrator was resolved by a small tweak in their contract, so I guess we can live without it, but we may find a use case for it in future.

PR #682, opened 2024-07-11
improve(NetworkUtils): add chainIsL1Function

bmzig:
As we begin to introduce `tryMulticall` into the relayer, the `MulticallerClient` will need to be aware of whether it is calling a spoke pool or hub pool contract (to determine whether it can even call `tryMulticall`). One approach is to first check if the contract's chain ID we are calling is one which has a hub pool implementation. It would be ideal to introduce this code into the sdk, where there already exists a set of functions which performs similar checks with our other chain IDs.

### DIFF
--- a/src/utils/NetworkUtils.ts
+++ b/src/utils/NetworkUtils.ts
@@ -90,6 +90,15 @@ export function chainIsLinea(chainId: number): boolean {
 }
 
 /**
+ * Determines whether a chain ID has a corresponding hub pool contract.
+ * @param chainId Chain ID to evaluate.
+ * @returns True if chain corresponding to chainId has a hub pool implementation.
+ */
+export function chainIsL1(chainId: number): boolean {
+  return [CHAIN_IDs.MAINNET, CHAIN_IDs.SEPOLIA].includes(chainId);
+}
+
+/**
  * Determines whether a chain ID has the capacity for having its USDC bridged via CCTP.
  * @param chainId Chain ID to evaluate.
  * @returns True if chainId is a CCTP-bridging enabled chain, otherwise false.
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "3.1.11",
+  "version": "3.1.12",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "files": [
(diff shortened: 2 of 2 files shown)


dohaki (approved):

pxrl:
> One approach is to first check if the contract's chain ID we are calling is one which has a hub pool implementation.

wdyt about just checking for the existence of tryMulticall in the contract interface? I think something like that is still needed (or alternatively a check to verify that the target address is a SpokePool), and in the case of mainnet we still need to distinguish between Hub and Spoke, so I'm not sure how knowing whether the chain is mainnet is helpful.

bmzig:
> > One approach is to first check if the contract's chain ID we are calling is one which has a hub pool implementation.
> 
> wdyt about just checking for the existence of tryMulticall in the contract interface? I think something like that is still needed (or alternatively a check to verify that the target address is a SpokePool), and in the case of mainnet we still need to distinguish between Hub and Spoke, so I'm not sure how knowing whether the chain is mainnet is helpful.

The issue with this is that we get the contract interface from `dist/typechain/factories/contracts` in the contracts repo, so this will falsely claim that every spoke pool has tryMulticall (since the ABIs cached are of the most up-to-date contract). The other option is to fetch the deployed bytecode and compare that with the tryMulticall function signature, but since this is asynchronous (and the context in which we want to check whether a contract has tryMulticall is synchronous), it would require some extra refactoring on the relayer + an RPC call (this also assumes that the bytes in tryMulticall's function selector do not appear anywhere else in the contract). 

I am adding a function to determine if a contract is a hub pool or spoke pool, but this will be in https://github.com/across-protocol/relayer/pull/1466.

src/utils/NetworkUtils.ts
```diff
@@ -89,6 +89,15 @@ export function chainIsLinea(chainId: number): boolean {
...
 
+/**
+ * Determines whether a chain ID has a corresponding hub pool contract.
+ * @param chainId Chain ID to evaluate.
+ * @returns True if chain corresponding to chainId has a hub pool implementation.
+ */
+export function chainIsL1(chainId: number): boolean {
+  return [CHAIN_IDs.MAINNET, CHAIN_IDs.SEPOLIA].includes(chainId);
+}
+
```
james-a-morris, line 100: Good call +1

james-a-morris (approved):
We should include the version bump before merging.

bmzig:
> We should include the version bump before merging.

I don't need this in now, so I'm actually going to just wait until this gets merged in, so that version bump will include this as well. https://github.com/across-protocol/sdk/pull/683

PR #683, opened 2024-07-12
improve(utils): Inherit OP stack chain properties

pxrl:
This should lead to fewer commits required to add new OP stack chains, and can later be extended to other chain stacks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "3.1.11",
+  "version": "3.1.12",
   "license": "AGPL-3.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -99,7 +99,7 @@
   },
   "dependencies": {
     "@across-protocol/across-token": "^1.0.0",
-    "@across-protocol/constants": "^3.1.5",
+    "@across-protocol/constants": "^3.1.7",
```
pxrl, line 102: This version isn't published yet, so CI currently fails.

pxrl: Fixed: https://github.com/across-protocol/sdk/pull/683/commits/6758639b9c8eaa05c300cc548f386393e7ae1ed4

bmzig (approved):

PR #685, opened 2024-07-16
feat(cg): add historical price lookup

james-a-morris:


src/coingecko/Coingecko.ts
```diff
@@ -89,6 +89,37 @@ export class Coingecko {
...
+   * @throws An error if the date is invalid (malformed or in the future).
+   */
+  getContractHistoricDayPrice(contractAddress: string, date: string, platform_id = "ethereum", currency = "usd") {
+    // Resolve the date into starting/ending timestamps for the day and validate the date.
+    const [day, month, year] = date.split("-").map(Number);
+    const startDate = new Date(year, month - 1, day);
+    const endDate = new Date(year, month - 1, day, 23, 59, 59, 999); // Use 999 ms for end of the day
+    const startTimestamp = Math.floor(startDate.getTime() / 1000);
+    const endTimestamp = Math.floor(endDate.getTime() / 1000);
+    assert(isNaN(startTimestamp) || isNaN(endTimestamp) || startDate > new Date(), "Invalid date provided");
```
pxrl, line 109: I don't understand this `assert` - the logic looks inverted to me :thinking:. Why is it OK for `startTimestamp` or `endTimestamp` to be NaN? I'd also expect `startDate` to pre-date `new Date()`.

I feel like I'm missing something obvious :\

pxrl (approved):
Nice!

src/coingecko/Coingecko.ts
amateima: > the logic looks inverted to me

➕

james-a-morris: Updated - the response of the market range didn't prove to have parity with the scraper so we should use the same API surface and move to eventually deprecate the non-contract coingecko api calls

james-a-morris: In the end - we did not need this check since the datetimes should error themselves out given that the endpoint will fail if we pass invalid dates

amateima (approved):

src/coingecko/Coingecko.ts
```diff
@@ -89,6 +89,33 @@ export class Coingecko {
...
+    const coingeckoTokenIdentifier = getCoingeckoTokenIdByAddress(contractAddress);
+    assert(date, "Requires date string");
+    // Build the path for the Coingecko API request
+    const url = `coins/${coingeckoTokenIdentifier}/history`;
+    // Build the query parameters for the Coingecko API request
+    const queryParams = {
+      date,
+      localization: "false",
+    };
+    console.log(`${url}?${new URLSearchParams(queryParams).toString()}`);
```
nicholaspai, line 111: leftover log?

nicholaspai (approved):

james-a-morris:
Reverted package version since we don't have a 3.1.12 version cut

PR #687, opened 2024-07-18
feat: add scroll to gas price oracle

dohaki:
According to the last paragraph in  https://docs.scroll.io/en/developers/transaction-fees-on-scroll/#l1-originated-transactions, we can use the legacy RPC call `eth_gasPrice` to determine the gas price on Scroll:
> If your system supports off-chain mechanisms, you can also call eth_estimateGas and eth_gasPrice on any Scroll RPC node to get an estimate of the gas required for a given transaction.

Closes ACX-2356

### DIFF
--- a/e2e/oracle.e2e.ts
+++ b/e2e/oracle.e2e.ts
@@ -51,6 +51,7 @@ const networks: { [chainId: number]: string } = {
   324: "https://mainnet.era.zksync.io",
   8453: "https://mainnet.base.org",
   42161: "https://rpc.ankr.com/arbitrum",
+  534352: "https://rpc.scroll.io",
 };
 
 const stdGasPrice = ethersUtils.parseUnits("10", 9);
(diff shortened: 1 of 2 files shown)


james-a-morris (approved):

dohaki:
> How does this perform in test btw? Does it produce a gas price that kinda-resembles the onchain price advertised by Scrollscan?
> That's in the `L1 Originated Transactions` section, which I think is messaging from mainnet -> Scroll, so I'm, not sure whether it applies to our use case 🤔
> In any case, it's probably sensible to just see how legacy pricing performs; if it's reasonably accurate (or over-estimates, but with an insignificant difference) then I think it's probably fine to proceed with that. If we need to go into more detail we can implement a Scroll-specific adapter that queries their gas oracle contracts.

Ah damn, yea you are right... It is less than the gas price displayed on Scrollscan. So yea I will re-work this PR and implement the adapter

PR #689, opened 2024-07-31
fix: cg lookup edge case

james-a-morris:
The `gasFeePercent()` function always attempts to call into CoinGecko. It's possible that its API call returns faster than our logic and triggers a 500 internal server error.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -256,19 +256,20 @@ export class RelayFeeCalculator {
       });
       throw error;
     });
-    const getTokenPrice = this.queries.getTokenPrice(token.symbol).catch((error) => {
-      this.logger.error({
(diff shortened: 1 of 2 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -256,19 +256,20 @@ export class RelayFeeCalculator {
...
-        error,
-        destinationChainId: deposit.destinationChainId,
-        inputToken,
-      });
-      throw error;
-    });
     const [{ tokenGasCost }, tokenPrice] = await Promise.all([
       getGasCosts,
-      _tokenPrice !== undefined ? _tokenPrice : getTokenPrice,
+      _tokenPrice !== undefined
```
nicholaspai, line 261: Nice fix and +1 for parallelizing this with the getGasCosts call

nicholaspai (approved):

PR #694, opened 2024-08-09
chore: Bump contracts

pxrl:
For a redeployed Redstone SpokePool.

### DIFF
--- a/src/relayFeeCalculator/chain-queries/factory.ts
+++ b/src/relayFeeCalculator/chain-queries/factory.ts
@@ -1,9 +1,10 @@
+import assert from "assert";
 import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants";
 import { getDeployedAddress } from "@across-protocol/contracts";
 import { asL2Provider } from "@eth-optimism/sdk";
 import { providers } from "ethers";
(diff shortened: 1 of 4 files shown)


src/relayFeeCalculator/chain-queries/factory.ts
```diff
@@ -28,6 +29,8 @@ export class QueryBase__factory {
     gasMarkup = 0,
     coingeckoBaseCurrency = "eth"
   ): QueryBase {
+    assert(isDefined(spokePoolAddress));
```
bmzig, line 32: What is this check for? Is there now a chance we don't have it defined with the contract bump?

bmzig (approved):

src/relayFeeCalculator/chain-queries/factory.ts
pxrl: The function that returns it can now return `undefined` _if_ the no SpokePool is recorded for the given chain, _and_ if the caller permits to not throw in that case (by setting the third argument false, which is the default). So in practice, we would assert before ever hitting this and this assert is mostly just tsc appeasement.

(This was changed such that the deployment scripts can check whether there is a SpokePool deployed to a chain without needing try/catch rails).

nicholaspai (approved):

PR #698, opened 2024-08-14
fix(ArweaveClient): Fix precision issue

nicholaspai:
We are multiplying arweave balances by 1e12 too much

### DIFF
--- a/src/caching/Arweave/ArweaveClient.ts
+++ b/src/caching/Arweave/ArweaveClient.ts
@@ -1,11 +1,11 @@
 import Arweave from "arweave";
 import { JWKInterface } from "arweave/node/lib/wallet";
 import axios from "axios";
-import { ethers } from "ethers";
+import { BigNumber, ethers } from "ethers";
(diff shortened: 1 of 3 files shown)


src/caching/Arweave/ArweaveClient.ts
```diff
@@ -230,9 +230,9 @@ export class ArweaveClient {
     if (balanceInFloat.includes("e")) {
       const [balance, exponent] = balanceInFloat.split("e");
       const resultingBN = ethers.BigNumber.from(balance).mul(toBN(10).pow(exponent.replace("+", "")));
-      return parseWinston(resultingBN.toString());
+      return BigNumber.from(resultingBN.toString());
```
james-a-morris, line 233: Fair call - does this work on integration tests?

nicholaspai: yeah its the same type signature it just removes the extra 1e12 multiplication

james-a-morris (approved):
lgtm - thanks for clarifying my question

PR #699, opened 2024-08-15
fix: Correctly resolve historical deposits

pxrl:
When the fill is totally malformed and filters on values that never existed (i.e. the real deposit for this deposit ID specified an alternative destinationChainId), it will skip returning the _actual_ deposit for this deposit ID.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -814,15 +814,8 @@ export class SpokePoolClient extends BaseAbstractClient {
         null,
         null,
         null,
-        destinationChainId,
+        null,
         depositId,
-        null,
-        null,
-        null,
-        depositor,
-        null,
-        null,
-        null
       ),
       {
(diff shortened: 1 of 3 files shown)


bmzig (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -809,21 +809,7 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-        depositId,
-        null,
-        null,
-        null,
-        depositor,
-        null,
-        null,
-        null
-      ),
+      this.spokePool.filters.V3FundsDeposited(null, null, null, null, null, depositId),
```
nicholaspai, line 812: I think this is fine:
- Same # of RPC requests are sent via paginatedEventQuery
- Since `depositId` is never duplicated in the SpokePool, the filter won't return too many events

nicholaspai (approved):

PR #700, opened 2024-08-15
improve: switch simulated relayer to fast relayer address

bmzig:
As more funds get migrated to the fast relayer, it will be better to start simulating with the fast relayer address so less deposits are skipped in simulation mode due to shortfalls.


src/constants.ts
```diff
@@ -47,7 +47,7 @@ export const DEFAULT_CACHING_SAFE_LAG = 60 * 60; // 1 hour
 export const UBA_BOUNDS_RANGE_MAX = BigNumber.from(String(Number.MAX_SAFE_INTEGER)).mul(utils.parseEther("1.0"));
 export const UBA_BOUNDS_RANGE_MIN = UBA_BOUNDS_RANGE_MAX.mul(-1);
 
-export const DEFAULT_SIMULATED_RELAYER_ADDRESS = "0x428AB2BA90Eba0a4Be7aF34C9Ac451ab061AC010";
+export const DEFAULT_SIMULATED_RELAYER_ADDRESS = "0x07aE8551Be970cB1cCa11Dd7a11F47Ae82e70E67";
```
pxrl, line 50: I realised that we need to coordinate this with the FE, since they use this address to simulate all tokens and the fast relayer can't yet handle that. We could fix that in the short term by just dusting the fast relayer with > 100 units of each token on each supported chain. This would be BAL, POOL, SNX, ...

bmzig: Checked on the front end and all the tokens can be quoted.

nicholaspai: yeah this should work now

nicholaspai: since we've transferred BAL/POOL/UMA/ACX to this EOA. SNX uses a custom override, hardcoded in the frontend here https://github.com/across-protocol/frontend/blob/1c65e3ce4bf34c6e785ff2e0ad942eb41f35bf61/api/_constants.ts#L118

nicholaspai (approved):

PR #702, opened 2024-08-15
improve: remove uma logger dep

james-a-morris:
We can assume that the user needs to pass a logger to get logging, and since the uma logger extends Winston we can just use that.

Change:
1. Remove uma/logger (vuln package)
2. Use `Logger` type in lieu of uma/logger type
3. RateLimitedProvider will not have a fallback logger

### DIFF
--- a/src/providers/rateLimitedProvider.ts
+++ b/src/providers/rateLimitedProvider.ts
@@ -19,7 +19,9 @@ export class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider 
   constructor(
     maxConcurrency: number,
     readonly pctRpcCallsLogged: number,
-    readonly logger: Logger = winston.createLogger(),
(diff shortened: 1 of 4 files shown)


src/providers/rateLimitedProvider.ts
```diff
@@ -20,7 +19,7 @@ export class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider
   constructor(
     maxConcurrency: number,
     readonly pctRpcCallsLogged: number,
-    readonly logger: Logger = umaLogger,
+    readonly logger?: Logger,
```
nicholaspai, line 22: Can't the fallback just be the default Logger construction?

james-a-morris:
> I believe this is incorrectly labeled a hotfix as its more of a long-term improvement and removal of a dependency

changed the name

nicholaspai (approved):

PR #703, opened 2024-08-16
chore: remove rails around inverting capital cost config

james-a-morris:
We're attempting to invert the capital cost model in the FE - these rails should be removed.


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -196,7 +196,6 @@ export class RelayFeeCalculator {
    */
   static validateCapitalCostsConfig(capitalCosts: CapitalCostConfig): void {
     assert(toBN(capitalCosts.upperBound).lt(toBNWei("0.01")), "upper bound must be < 1%");
```
nicholaspai, line 198: Is cap on the upper bound needed anymore?

nicholaspai (approved):

PR #704, opened 2024-08-16
feat: Cheat codes for resolving provider URLs

pxrl:
Alchemy has recently added support for reusing the same API key across multiple networks. Observing that both Alchemy and Infura use reasonably-predictable symbolic names for their endpoints (with a small number of exceptions), it's fairly advantageous to programatically resolve the endpoint URLs from a chain ID and an API key. This permits a single API key to be defined per provider, rather than the existing requirement of having to define each individual RPC configuration in the environment.

This change deliberately avoids trying to encode an env-var format for reading keys from the environment; that should permit maximum reuse amongst different applications.

### DIFF
--- a/src/providers/alchemy.ts
+++ b/src/providers/alchemy.ts
@@ -1,4 +1,5 @@
 import { CHAIN_IDs, PUBLIC_NETWORKS } from "../constants";
+import { RPCTransport } from "./types";
 
 // Chain-specific overrides for when the Alchemy endpoint does not match the canonical chain name.
 const endpoints: { [chainId: string]: string } = {
@@ -7,11 +8,11 @@ const endpoints: { [chainId: string]: string } = {
   [CHAIN_IDs.OPTIMISM]: "opt-mainnet",
 };
 
-export function getURL(chainId: number, apiKey: string): string {
+export function getURL(chainId: number, apiKey: string, transport: RPCTransport): string {
   const host = endpoints[chainId] ?? PUBLIC_NETWORKS[chainId]?.name;
   if (!host) {
     throw new Error(`No known Alchemy provider for chainId ${chainId}`);
   }
 
-  return `https://${host.toLowerCase().replace(" ", "-")}.g.alchemy.com/v2/${apiKey}`;
+  return `${transport.toLowerCase()}://${host.toLowerCase().replace(" ", "-")}.g.alchemy.com/v2/${apiKey}`;
 }
--- a/src/providers/alchemy.ts
+++ b/src/providers/alchemy.ts
@@ -6,8 +6,11 @@ const MAINNET_CHAIN_IDs = Object.values(_MAINNET_CHAIN_IDs);
 // Chain-specific overrides for when the Alchemy endpoint does not match the canonical chain name.
 const endpoints: { [chainId: string]: string } = {
   [CHAIN_IDs.ARBITRUM]: "arb",
+  [CHAIN_IDs.ARBITRUM_SEPOLIA]: "arb",
   [CHAIN_IDs.MAINNET]: "eth",
+  [CHAIN_IDs.SEPOLIA]: "eth-sepolia",
   [CHAIN_IDs.OPTIMISM]: "opt",
+  [CHAIN_IDs.OPTIMISM_SEPOLIA]: "opt-sepolia",
 };
 
 export function getURL(chainId: number, apiKey: string, transport: RPCTransport): string {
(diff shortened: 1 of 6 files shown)


pxrl:
Example implementation in the relayer repository: https://github.com/across-protocol/relayer/pull/1758

src/providers/utils.ts
```diff
@@ -1,8 +1,29 @@
...
+const PROVIDERS = {
+  ALCHEMY: alchemy.getURL,
+  INFURA: infura.getURL,
+};
+
+export function isSupportedProvider(provider: string): provider is RPCProvider {
+  return ["ALCHEMY", "INFURA"].includes(provider);
+}
+
+export function getURL(provider: RPCProvider, chainId: number, apiKey?: string): string {
```
pxrl, line 20: Probably also want the option to specify the protocol here - i.e. HTTPS or WSS.

pxrl, line 20: Supplying `apiKey` should also be mandatory, and will avoid the need for an assert.

pxrl: Done: https://github.com/across-protocol/sdk/pull/704/commits/cfb1ccaaca4d9551831fabd6a92f98c66c58e77f

pxrl: Done: https://github.com/across-protocol/sdk/pull/704/commits/cfb1ccaaca4d9551831fabd6a92f98c66c58e77f

src/providers/alchemy.ts
```diff
@@ -0,0 +1,28 @@
+import { CHAIN_IDs, MAINNET_CHAIN_IDs as _MAINNET_CHAIN_IDs, PUBLIC_NETWORKS } from "../constants";
+import { RPCTransport } from "./types";
+
+const MAINNET_CHAIN_IDs = Object.values(_MAINNET_CHAIN_IDs);
+
+// Chain-specific overrides for when the Alchemy endpoint does not match the canonical chain name.
+const endpoints: { [chainId: string]: string } = {
+  [CHAIN_IDs.ARBITRUM]: "arb",
+  [CHAIN_IDs.ARBITRUM_SEPOLIA]: "arb",
```
pxrl, line 9:
```suggestion
  [CHAIN_IDs.ARBITRUM_SEPOLIA]: "arb-sepolia",
```

bmzig (approved):
I believe this could also work for DRPC, although, we admittedly don't use them as much. I think you also need to bump the version for https://github.com/across-protocol/relayer/pull/1758?

pxrl:
> I see you are still adding more networks? I believe this could also work for DRPC, although, we admittedly don't use them as much. I think you also need to bump the version for [across-protocol/relayer#1758](https://github.com/across-protocol/relayer/pull/1758)?

Yeah; DRPC should be compatible with this. We can definitely consider adding them. @bmzig Do you want to just push that to the SDK pr?

pxrl:
todo: Verify compatibility w/ Sepolia endpoints.

src/providers/alchemy.ts
```diff
@@ -0,0 +1,28 @@
...
+export function getURL(chainId: number, apiKey: string, transport: RPCTransport): string {
+  let host = endpoints[chainId] ?? PUBLIC_NETWORKS[chainId]?.name;
+  if (!host) {
+    throw new Error(`No known Alchemy provider for chainId ${chainId}`);
+  }
+
+  if (MAINNET_CHAIN_IDs.includes(chainId)) {
+    host = `${host}-mainnet`;
+  }
+  host = host.toLowerCase().replace(" ", "-");
```
james-a-morris, line 25: OOC, what is the purpose of this line?

src/providers/alchemy.ts
```diff
@@ -0,0 +1,28 @@
...
+};
+
+export function getURL(chainId: number, apiKey: string, transport: RPCTransport): string {
+  let host = endpoints[chainId] ?? PUBLIC_NETWORKS[chainId]?.name;
+  if (!host) {
+    throw new Error(`No known Alchemy provider for chainId ${chainId}`);
+  }
+
+  if (MAINNET_CHAIN_IDs.includes(chainId)) {
+    host = `${host}-mainnet`;
```
james-a-morris, line 23: Could we remove the `-` from this line if we're replacing it below
```suggestion
    host = `${host}mainnet`;
```

james-a-morris (commented):
left a few comments - OOC can we do this in a more consolidated way? I.e. these providers seem to have a similar flow (at least the ones we are working with at the moment) of `(https/wss)://(host).(provider-base-url)/(api-key)`.

bmzig:
> > I see you are still adding more networks? I believe this could also work for DRPC, although, we admittedly don't use them as much. I think you also need to bump the version for [across-protocol/relayer#1758](https://github.com/across-protocol/relayer/pull/1758)?
> 
> Yeah; DRPC should be compatible with this. We can definitely consider adding them. @bmzig Do you want to just push that to this pr?

Sounds good. I'll branch off of this one.

PR #706, opened 2024-08-20
add drpc

bmzig:


src/providers/drpc.ts
```diff
@@ -0,0 +1,22 @@
+import { CHAIN_IDs, MAINNET_CHAIN_IDs as _MAINNET_CHAIN_IDs, PUBLIC_NETWORKS } from "../constants";
+import { RPCTransport } from "./types";
+
+const MAINNET_CHAIN_IDs = Object.values(_MAINNET_CHAIN_IDs);
```
james-a-morris, line 4: I think this is unused

bmzig: Yeah that's right. My mistake. https://github.com/across-protocol/sdk/pull/706/commits/bc0213e7f71b59713540d0cc1481deb3f531ad7a

src/providers/drpc.ts
```diff
@@ -0,0 +1,20 @@
+import { CHAIN_IDs, PUBLIC_NETWORKS } from "../constants";
+import { RPCTransport } from "./types";
+
+// Chain-specific overrides for when the Alchemy endpoint does not match the canonical chain name.
+const endpoints: { [chainId: string]: string } = {
+  [CHAIN_IDs.ARBITRUM]: "arbitrum",
+  [CHAIN_IDs.MAINNET]: "ethereum",
+  [CHAIN_IDs.SEPOLIA]: "sepolia",
```
pxrl, line 8: I think you're OK to remove the SEPOLIA definition. The endpoint names are inherited from the network definitions in https://github.com/across-protocol/constants/blob/master/src/networks.ts, so we only need to add entries here in case where the name differs from the endpoint (i.e. `Arbitrum One` vs `Arbitrum` and `Mainnet` vs `Ethereum`).

src/providers/drpc.ts
```diff
@@ -0,0 +1,20 @@
+import { CHAIN_IDs, PUBLIC_NETWORKS } from "../constants";
+import { RPCTransport } from "./types";
+
+// Chain-specific overrides for when the Alchemy endpoint does not match the canonical chain name.
```
pxrl, line 4:
```suggestion
// Chain-specific overrides for when the endpoint name does not match the canonical chain name.
```

src/providers/types.ts
```diff
@@ -1,2 +1,2 @@
-export type RPCProvider = "INFURA" | "ALCHEMY";
+export type RPCProvider = "INFURA" | "ALCHEMY" | "DRPC";
```
pxrl, line 1: Is this missing `INFURA_DIN` here?

bmzig: Yeah, I think so: https://github.com/across-protocol/sdk/pull/706/commits/5aadadf9c0a04a7cc349a53d1be42aeb40b48cd5

src/providers/drpc.ts
bmzig: https://github.com/across-protocol/sdk/pull/706/commits/5aadadf9c0a04a7cc349a53d1be42aeb40b48cd5

src/providers/drpc.ts
bmzig: True. https://github.com/across-protocol/sdk/pull/706/commits/5aadadf9c0a04a7cc349a53d1be42aeb40b48cd5

src/providers/types.ts
pxrl: My mistake on that fwiw - I added INFURA_DIN to the type predicate but not the actual type :facepalm:

src/providers/types.ts
```diff
@@ -1,2 +1,2 @@
-export type RPCProvider = "INFURA" | "ALCHEMY";
+export type RPCProvider = "INFURA" | "ALCHEMY" | "DRPC" | "INFURA_DIN";
```
pxrl, line 1:
```suggestion
export type RPCProvider = "ALCHEMY" | "DRPC" | "INFURA" | "INFURA_DIN";
```

PR #713, opened 2024-08-26
feat(SpokePoolClient): getter function for speedUps

melisaguevara:
In order to index 'RequestedSpeedUpV3' events, the indexer needs to get the speedUps map from the SDK.

### DIFF
--- a/src/interfaces/SpokePool.ts
+++ b/src/interfaces/SpokePool.ts
@@ -94,6 +94,7 @@ export interface RootBundleRelay {
 export interface RootBundleRelayWithBlock extends RootBundleRelay, SortableEvent {}
 
 export interface RelayerRefundExecution extends RelayerRefundLeaf {
+  caller: string;
   rootBundleId: number;
 }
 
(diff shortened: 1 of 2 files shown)


src/interfaces/SpokePool.ts
```diff
@@ -111,6 +112,7 @@ export interface RunningBalances {
 
 export interface TokensBridged extends SortableEvent {
   amountToReturn: BigNumber;
+  caller: string;
```
james-a-morris, line 115: Will this impact any downstream events in the relayer?

amateima (approved):

daywiss (approved):

james-a-morris (approved):

src/interfaces/SpokePool.ts
melisaguevara: Based on the tests we performed (linking the relayer with this version) it shouldn't. The unit tests of the relayer are passing and both the relayer and the finalizer are working ok in simulation mode.

PR #717, opened 2024-09-02
feat: allow usage of cached `gasUsage` + `SpeedProvider`

dohaki:
This PR enables two performance tweaks in the API:
- Using a cached `gasUnits` value for a fill - this saves an RPC request in the fee calculator
- Using the `SpeedProvider` which makes use of `Promise.any`

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -242,7 +242,6 @@ export type TransactionCostEstimate = {
  * @param provider A valid ethers provider - will be used to reason the gas price.
  * @param gasMarkup Markup on the estimated gas cost. For example, 0.2 will increase this resulting value 1.2x.
  * @param gasPrice A manually provided gas price - if set, this function will not resolve the current gas price.
- * @param gasLimit A manually provided gas limit - if set, this function will not estimate the gas limit.
  * @param gasUnits A manually provided gas units - if set, this function will not estimate the gas units.
  * @returns Estimated cost in units of gas and the underlying gas token (gasPrice * estimatedGasUnits).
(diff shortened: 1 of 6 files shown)


src/utils/common.ts
```diff
@@ -262,11 +264,13 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
-    gasPrice,
-    gasLimit,
-  });
+  let nativeGasCost = gasUnits
+    ? BigNumber.from(gasUnits)
+    : await voidSigner.estimateGas({
+        ...unsignedTx,
+        gasPrice,
+        gasLimit,
+      });
```
pxrl, line 273: ooc, do you think it's a hard requirement to be able to specify `gasLimit` and `gasUnits` independently?

(I think the historical reason for needing to override gasLimit was probably debunked and we fortunately haven't hit another issue...at least as far as I am aware)

dohaki: Yea, that's a good point. We don't use that anywhere. I can remove that parameter then

dohaki: https://github.com/across-protocol/sdk/pull/717/commits/e662f595ec43410d79cf4c81be8414e6fa474cd5

gsteenkamp89 (approved):
Looks great! 🚀

james-a-morris (approved):
Is this a migration?

dohaki:
> Is this a migration?

Can you elaborate a bit? Migration from what?

The context of these features introduced here are primarily for improving the performance of the API

PR #719, opened 2024-09-03
chore: migrate test function

james-a-morris:
`findFillBlock` is only used in test cases so we can move it there


pxrl:
@james-a-morris ooc, is there a concrete need/motivation for migrating this?

I recall that I had a use case for this as part of the RPC resiliency work (RE detecting fills where there was no FilledV3Relay event returned by the RPC provider). I have verified that it's not used at the moment, but I do still see this as a genuinely useful function that we'd ultimately want to return back to src/utils.

PR #723, opened 2024-09-04
improve: reduce remove direct contract/rpc calls in bundle client

james-a-morris:
This PR aims to reduce the amount of direct RPC calls / direct use of contracts. The context is that by encapsulating this logic in our contract clients we can overload these functions in the indexer to be used with our persistence layer.

### DIFF
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -244,32 +244,6 @@ export function getRelayHashFromEvent(e: Deposit | Fill | SlowFillRequest): stri
   return getRelayDataHash(e, e.destinationChainId);
 }
 
-/**
- * Find the amount filled for a deposit at a particular block.
- * @param spokePool SpokePool contract instance.
- * @param relayData Deposit information that is used to complete a fill.
- * @param blockTag Block tag (numeric or "latest") to query at.
- * @returns The amount filled for the specified deposit at the requested block (or latest).
- */
-export async function relayFillStatus(
-  spokePool: Contract,
-  relayData: RelayData,
-  blockTag?: number | "latest",
-  destinationChainId?: number
-): Promise<FillStatus> {
-  destinationChainId ??= await spokePool.chainId();
-  const hash = getRelayDataHash(relayData, destinationChainId!);
-  const _fillStatus = await spokePool.fillStatuses(hash, { blockTag });
-  const fillStatus = Number(_fillStatus);
-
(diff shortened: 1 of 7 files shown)


src/utils/SpokeUtils.ts
```diff
@@ -316,11 +290,12 @@ export async function fillStatusArray(
  * @returns The block number at which the relay was completed, or undefined.
  */
 export async function findFillBlock(
-  spokePool: Contract,
+  spokePoolClient: SpokePoolClient,
```
james-a-morris, line 293: This function currently isn't used anymore - I think we have future plans for it though so we can keep it around. cc @pxrl

amateima (approved):
Looks good from my perspective

src/utils/SpokeUtils.ts
```diff
@@ -244,32 +244,6 @@ export function getRelayHashFromEvent(e: Deposit | Fill | SlowFillRequest): stri
...
-  const fillStatus = Number(_fillStatus);
-
-  if (![FillStatus.Unfilled, FillStatus.RequestedSlowFill, FillStatus.Filled].includes(fillStatus)) {
-    const { originChainId, depositId } = relayData;
-    throw new Error(`relayFillStatus: Unexpected fillStatus for ${originChainId} deposit ${depositId} (${fillStatus})`);
-  }
-
-  return fillStatus;
-}
-
```
pxrl, line 272: Can we retain this function here and just call down into it from within the SpokePool?

It's a really good initiative to start to centralise the random RPC calls that we initiate throughout the bots, but bundling all of the implementations directly into the SpokePoolClient reduces useability for a lot of use cases. It's generally useful to have this method available for use (also third-parties who want to do this), but making them instantiate the complete SpokePoolClient with all its overheads and dependencies is going to be a showstopper for most third-parties.

The pattern that I strongly prefer is to maintain small, self-contained utility functions that can be used independently of other functionality, and to then collect them into larger groupings that provide some e2e function (i.e. the various clients).

src/utils/SpokeUtils.ts
pxrl: Can we also revert this change? I'd really like to avoid forcing instantiation of a SpokePoolClient instance just to query the the fill status at a given spoke.

src/utils/SpokeUtils.ts
james-a-morris: Fair call +1

pxrl (approved):

PR #725, opened 2024-09-04
refactor: Centralise BigNumber imports

pxrl:
Avoid importing BigNumber from ethers directly throughout the repository, and instead just re-export the standalone ethers BigNumber variant from src/utils/BigNumberUtils. This should make it easier to migrate away from ethers v5.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,8 +29,11 @@ module.exports = {
     "no-restricted-imports": [
       "error",
       {
+        "patterns": [
+          { group: ["@ethersproject/bignumber"] },
+        ],
         "paths": [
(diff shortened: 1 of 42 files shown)


nicholaspai (approved):
This is awesome work! So now we only have a single Bn import from ethers?

pxrl:
> This is awesome work! So now we only have a single Bn import from ethers?

Yeah, hopefully this achieves that. It's a bit tricky to be 100% sure but I've eliminated as many cases of importing directly from ethers as I could identify. If we implement a BigInt -> BigNumber shim then we might be able to play with v6.

dohaki (approved):
Nice work! I think this shouldn't break anything on the FE AFAICT. Also really like @gsteenkamp89 suggestion

gsteenkamp89 (approved):
nice one! this will help for sure :)

PR #728, opened 2024-09-10
improve(RetryProvider): improve logs for quorum mismatch or failures

melisaguevara:
In cases where we threw the QuorumError early, we were not logging any details of the mismatch.

### DIFF
--- a/src/providers/retryProvider.ts
+++ b/src/providers/retryProvider.ts
@@ -125,15 +125,38 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
       return values[0][1];
     }
 
-    const throwQuorumError = (fallbackValues?: [ethers.providers.StaticJsonRpcProvider, unknown][]) => {
-      const errorTexts = errors.map(([provider, errorText]) => formatProviderError(provider, errorText));
-      const successfulProviderUrls = values.map(([provider]) => provider.connection.url);
-      const mismatchedProviders = Object.fromEntries(
-        [...values, ...(fallbackValues || [])]
+    const getMismatchedProviders = (values: [ethers.providers.StaticJsonRpcProvider, unknown][]) => {
+      return Object.fromEntries(
+        values
           .filter(([, result]) => !compareRpcResults(method, result, quorumResult))
           .map(([provider, result]) => [provider.connection.url, result])
       );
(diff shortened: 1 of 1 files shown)


src/providers/retryProvider.ts
```diff
@@ -199,21 +205,31 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+      message: "Some providers mismatched with the quorum result or failed 🚸",
+      notificationPath: "across-warn",
+      method,
+      params: JSON.stringify(params),
+      quorumProviders,
+      mismatchedProviders: JSON.stringify(mismatchedProviders),
+      erroringProviders: errors.map(([provider, errorText]) => formatProviderError(provider, errorText)),
+    });
+  }
+
```
pxrl, line 232: IMO this one could probably also be a closure, since as far as I can see, we don't really expect to call it from any other methods.

pxrl (commented):
@melisaguevara Nice change! This will make it easier to debug in future :+1: I have a couple of minor comments, but overall it looks good.

src/providers/retryProvider.ts
melisaguevara: Yes, you're right. I think we'll be calling it only from this method. I modified it here [3165d0d](https://github.com/across-protocol/sdk/pull/728/commits/3165d0d36a2964c1382bca6af4838458cba3cb2f)

src/providers/retryProvider.ts
```diff
@@ -125,9 +125,15 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
 
-    const throwQuorumError = () => {
+    const throwQuorumError = (fallbackValues?: [ethers.providers.StaticJsonRpcProvider, unknown][]) => {
       const errorTexts = errors.map(([provider, errorText]) => formatProviderError(provider, errorText));
       const successfulProviderUrls = values.map(([provider]) => provider.connection.url);
+      const mismatchedProviders = Object.fromEntries(
+        [...values, ...(fallbackValues || [])]
+          .filter(([, result]) => !compareRpcResults(method, result, quorumResult))
+          .map(([provider, result]) => [provider.connection.url, result])
+      );
```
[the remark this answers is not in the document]
melisaguevara: Modified also here [3165d0d](https://github.com/across-protocol/sdk/pull/728/commits/3165d0d36a2964c1382bca6af4838458cba3cb2f)

pxrl (approved):

PR #735, opened 2024-09-18
fix(package): Ignore typechain outputs in watcher

pxrl:
These files are touched unconditionally by the build process, so if they are not ignored it triggers an infinite loop in the watcher.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "node": ">=16.18.0"
   },
   "scripts": {
-    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build'",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "node": ">=16.18.0"
   },
   "scripts": {
-    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build'",
+    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --ignore src/utils/abi/typechain --exec 'yarn build'",
```
pxrl, line 15: @gsteenkamp89 This is a quick fix to stop the infinite loop. I'm not sure whether I prefer this or moving the typechain outputs to be outside of src. The latter is obviously more invasive. wdyt?

pxrl:
Not relevant after G's PR.

PR #736, opened 2024-09-18
improve(SpokePoolClient): Skip unnecessary LP fee computation

pxrl:
The SpokePoolClient currently piggybacks on the HubPoolClient's batch LP fee computation in order to optimise resolving quoteTimestamp to a mainnet block number. This was historically necessary prior to the v3 transition, but it's no longer needed now that the LP fee is computed by upper-layer code. Dropping the LP fee computation will reduce RPC queries on update.

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -291,9 +291,31 @@ export class HubPoolClient extends BaseAbstractClient {
     return getL1TokenInfo(tokenAddress, chain);
   }
 
-  getBlockNumber(timestamp: number): Promise<number | undefined> {
+  /**
+   * Resolve a given timestamp to a block number on the HubPool chain.
+   * @param timestamp A single timestamp to be resolved to a block number on the HubPool chain.
+   * @returns The block number corresponding to the supplied timestamp.
+   */
+  getBlockNumber(timestamp: number): Promise<number> {
     const hints = { lowBlock: this.deploymentBlock };
     return getCachedBlockForTimestamp(this.chainId, timestamp, this.blockFinder, this.cachingMechanism, hints);
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -291,11 +291,33 @@ export class HubPoolClient extends BaseAbstractClient {
...
+   * @dev Querying block numbers sequentially also improves BlockFinder efficiency.
+   * @param timestamps Array of timestamps to be resolved to a block number on the HubPool chain.
+   * @returns A mapping of quoteTimestamp -> HubPool block number.
+   */
+  async getBlockNumbers(timestamps: number[]): Promise<{ [quoteTimestamp: number]: number }> {
+    const sortedTimestamps = dedupArray(timestamps).sort((x, y) => x - y);
+    const blockNumbers: { [quoteTimestamp: number]: number } = {};
+    for (const timestamp of sortedTimestamps) {
+      blockNumbers[timestamp] = await this.getBlockNumber(timestamp);
+    }
```
dohaki, line 316: NIT: Can this be wrapped in a `Promise.all`?

dohaki (approved):
Looking good 👍 Left one nit / q

bmzig (approved):
I think this makes sense.

src/clients/HubPoolClient.ts
pxrl: You'd think so, but it's deliberately done this way to allow the BlockFinder to handle these one by one. If we hit the BlockFinder in parallel then it doesn't get any of the benefits of previous cached blocks, so we risk kicking off lots of parallel RPC queries. If we hit it sequentially with the oldest timestamp first then it should be able to infer the correct block number on subsequent requests very accurately based on the fairly stable block times on mainnet.

PR #737, opened 2024-09-18
mount watcher after typings are generated

gsteenkamp89:


package.json
```diff
@@ -12,8 +12,9 @@
     "node": ">=16.18.0"
   },
   "scripts": {
-    "start": "nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn build'",
+    "start": "yarn typechain && nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn dev'",
     "build": "yarn run clean && yarn typechain && yarn run build:cjs & yarn run build:esm & yarn run build:types; wait",
```
gsteenkamp89, line 16: i did not want to alter the current build/ci workflow by changing this script. we also still want to run typechain when we build and publish so instead created a new `dev` script that excludes the typechain command

package.json
```diff
@@ -12,8 +12,9 @@
...
+    "dev": "yarn run clean && yarn run build:cjs & yarn run build:esm & yarn run build:types; wait",
```
pxrl, line 17: wdyt about dropping the `clean` target here? It _should_ be redundant in almost all scenarios I expect.

```suggestion
    "dev": "yarn run build:cjs & yarn run build:esm & yarn run build:types; wait",
```

gsteenkamp89: Yeah makes sense!

PR #738, opened 2024-09-19
refactor: Replace ethers Event with Log

pxrl:
The ethers Event type is an extension of the underlying Log type, which is actually the type that contains almost all of the event data that the SDK needs for Across. Since Event extends Log, it's safe to narrow the scope of the RPC data down to a Log type. This significantly reduces the delta to viem's Log type and is a key stepping stone for viem and ethers to be used interchangeably for RPC interfacing & transport.

It additionally seems like this simplification _might_ also support Solana events (...with some eventual coercion).

It should be noted that this change touches some code that makes a lot of assumptions about the event types that are being handled, and there are some warts. I've tried not to make things any less safe than they already are, but this code is generally in need of a refactoring to improve type safety.

### DIFF
--- a/src/clients/mocks/MockConfigStoreClient.ts
+++ b/src/clients/mocks/MockConfigStoreClient.ts
@@ -1,7 +1,8 @@
 import assert from "assert";
 import winston from "winston";
-import { Contract, Event, ethers } from "ethers";
-import { EventSearchConfig, MakeOptional, isDefined, utf8ToHex } from "../../utils";
+import { Contract, ethers } from "ethers";
+import { Log } from "../../interfaces";
+import { getCurrentTime, EventSearchConfig, MakeOptional, isDefined, utf8ToHex } from "../../utils";
 import {
   AcrossConfigStoreClient,
   ConfigStoreUpdate,
@@ -60,7 +61,7 @@ export class MockConfigStoreClient extends AcrossConfigStoreClient {
     this.configStoreVersion = version;
   }
 
-  async _update(): Promise<ConfigStoreUpdate> {
+  _update(): Promise<ConfigStoreUpdate> {
     // Backwards compatibility for pre-existing MockConfigStoreClient users.
     if (this.eventManager === null) {
       return super._update();
@@ -72,7 +73,7 @@ export class MockConfigStoreClient extends AcrossConfigStoreClient {
     // Ensure an array for every requested event exists, in the requested order.
     // All requested event types must be populated in the array (even if empty).
     const globalConfigUpdateTimes: number[] = [];
-    const _events: Event[][] = eventNames.map(() => []);
+    const _events: Log[][] = eventNames.map(() => []);
     for (const event of this.eventManager.getEvents().flat()) {
       const idx = eventNames.indexOf(event.event as string);
       if (idx !== -1) {
@@ -80,15 +81,14 @@ export class MockConfigStoreClient extends AcrossConfigStoreClient {
       }
 
       if (event.event === "UpdatedGlobalConfig") {
-        const block = await event.getBlock();
-        globalConfigUpdateTimes.push(block.timestamp);
+        globalConfigUpdateTimes.push(getCurrentTime());
       }
     }
 
     // Transform 2d-events array into a record.
     const events = Object.fromEntries(eventNames.map((eventName, idx) => [eventName, _events[idx]]));
 
-    return {
+    return Promise.resolve({
       success: true,
       chainId: this.chainId as number,
       searchEndBlock: this.eventSearchConfig.toBlock || latestBlockSearched,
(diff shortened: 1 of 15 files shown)


pxrl:
Also for reference, the relevant type definitions are...

ethers v5 Event: https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/contracts/src.ts/index.ts#L75
ethers v5 Log: https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/abstract-provider/src.ts/index.ts#L90

src/interfaces/Common.ts
```diff
@@ -1,5 +1,12 @@
+import { Log as _Log } from "@ethersproject/abstract-provider";
 import { BigNumber } from "../utils";
 
+export type Log = _Log & {
```
pxrl, line 4: It might be necessary to omit `topics` from `_Log` because it seems to be only present in ethers v5 Log, not viem Log. I think this should be OK but will need to test it.

src/utils/EventUtils.ts
```diff
@@ -1,17 +1,18 @@
...
 import { Contract, Event, EventFilter } from "ethers";
+import { Log, SortableEvent } from "../interfaces";
+import { delay } from "./common";
 
 const maxRetries = 3;
 const retrySleepTime = 10;
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function spreadEvent(args: Result = {} as Result): any {
+export function spreadEvent(args: Result | Record<string, any>): { [key: string]: any } {
```
james-a-morris, line 11: nit: we should probably use `unknown` in lieu of `any`

src/utils/EventUtils.ts
```diff
@@ -1,17 +1,18 @@
...
   const keys = Object.keys(args).filter((key: string) => isNaN(+key)); // Extract non-numeric keys.
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const returnedObject: any = {};
+  const returnedObject: { [key: string]: any } = {};
```
james-a-morris, line 15: nit: same as above

src/contracts/utils.ts
```diff
@@ -1,5 +1,6 @@
 import assert from "assert";
-import { Event } from "ethers";
+// eslint-disable-next-line no-restricted-imports
```
james-a-morris, line 2: Is this a required import?

james-a-morris (approved):
lgtm - imo this should be a minor bump because it may break the FE + relayer + indexer.

src/contracts/utils.ts
pxrl: I'm a little unsure on this specifically - it does omit most of the fields that we don't care about within `Event`, but the downstream `poolClient` subsequently converts up to an `Event-ish` type.

As background, this is basically legacy code that was imported from the UMA repo a while ago. We depend on it in the FE, but not in the bots. This makes it a little trickier for me to test, so I took the conservative route here (...admittedly, last time I touched something in the `poolClient` it subtly broke the `Pool` page in the FE).

src/utils/EventUtils.ts
pxrl: Good call: https://github.com/across-protocol/sdk/pull/738/commits/6df63d0c688dd7f5563175df6e0046439fbc7c3a

src/utils/EventUtils.ts
pxrl: Good call: https://github.com/across-protocol/sdk/pull/738/commits/6df63d0c688dd7f5563175df6e0046439fbc7c3a

src/clients/mocks/MockConfigStoreClient.ts
```diff
@@ -72,23 +73,22 @@ export class MockConfigStoreClient extends AcrossConfigStoreClient {
...
     for (const event of this.eventManager.getEvents().flat()) {
       const idx = eventNames.indexOf(event.event as string);
       if (idx !== -1) {
         _events[idx].push(event);
       }
 
       if (event.event === "UpdatedGlobalConfig") {
-        const block = await event.getBlock();
-        globalConfigUpdateTimes.push(block.timestamp);
+        globalConfigUpdateTimes.push(getCurrentTime());
```
bmzig, line 84: This doesn't look equivalent to what was deleted. Just want to verify that this is correct.

src/contracts/utils.ts
pxrl: @james-a-morris Also on the above, the fact that this is legacy-leaning code and is only used by the FE makes me wonder if it could be a candidate for relocation to the FE. wdyt?

src/clients/mocks/MockConfigStoreClient.ts
bmzig: I guess it doesn't matter since it is in mocks

melisaguevara (approved):
lgtm!
I've tested the indexer linking this version and there are no issues.

src/clients/mocks/MockConfigStoreClient.ts
pxrl: Good catch; the behaviour _should_ be ~equivalent because the old `event.getBlock()` just resolved to `Date.now() / 1000`.

getBlock

PR #742, opened 2024-09-26
fix: do not pass gasPrice to estimateGas

dohaki:
When passing cached gas prices into the relay fee calculator in the API, `estimateGas` was erroring. I think this is due to the reason that we provided the value into the estimate method as well.


pxrl (approved):
Nice catch.

PR #745, opened 2024-09-30
experimental: gasPriceOracle overhaul

pxrl:
nb. This is an experimental PR that's thrown up for some troubleshooting, and to debate whether it's worth going down this path with viem.

This change updates the gasPriceOracle to leverage viem's gas pricing strategies. This is beneficial because viem has an active community that are actively adding support for new chains and improving support for existing chains. An example of this is Linea, where linea_estimateGas is supported out of the box after a recent update.

In order to maintain some degree of compatibility with existing users of the gasPriceOracle, the external interface still accepts an ethers Provider instance. It then internally instantiates a viem PublicClient instance and uses that instead. There is still a residual issue of how to resolve the relevant provider URLs, since it's not necessarily reliable to pull from the provider due to the uncertain type (i.e. StaticJsonRpcProvider or one of the extended variants that wraps multiple providers).

### DIFF
--- a/e2e/oracle.e2e.ts
+++ b/e2e/oracle.e2e.ts
@@ -1,10 +1,11 @@
 // @note: This test is _not_ run automatically as part of git hooks or CI.
 import dotenv from "dotenv";
 import winston from "winston";
-import { providers, utils as ethersUtils } from "ethers";
+import { custom } from "viem";
+import { providers } from "ethers";
 import { getGasPriceEstimate } from "../src/gasPriceOracle";
-import { BigNumber } from "../src/utils";
-import { assertPromiseError, expect } from "../test/utils";
+import { BigNumber, bnZero, parseUnits } from "../src/utils";
+import { expect } from "../test/utils";
 dotenv.config({ path: ".env" });
 
 const dummyLogger = winston.createLogger({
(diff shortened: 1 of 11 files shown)


e2e/oracle.e2e.ts
pxrl, line 1: nb. This file has some breakage and currently doesn't pass. I also nuked some types of test here; it's not clear to me whether that was the right call or not, so it might be worth restoring them.

In general, I am however wondering whether it makes sense to continue to support legacy pricing, since it seems almost all chains have some form of EIP-1559 support now.

src/gasPriceOracle/adapters/polygon.ts
pxrl, line 1: I think this file could ultimately be transformed into a custom viem Transport that redirects `eth_getGasPrice()` and `eth_getMaximumPriorityFee()` and redirects that query to the GasStation. That would be a nice way of providing sensible gas estimates directly via the provider, rather than via this separate utility.

src/gasPriceOracle/oracle.ts
```diff
@@ -14,35 +14,36 @@ import * as polygon from "./adapters/polygon";
...
-  legacyFallback = true
-): Promise<GasPriceEstimate> {
-  if (chainId === undefined) {
-    ({ chainId } = await provider.getNetwork());
-  }
+export async function getGasPriceEstimate(provider: providers.Provider, transport?: Transport): Promise<GasPriceEstimate> {
+  const { chainId } = await provider.getNetwork();
+
+  transport ??= http(); // @todo: Inherit URL from provider.
+  const chain = extractChain({ chains: Object.values(chains), id: chainId as any });
```
pxrl, line 21: @gsteenkamp89 @dohaki Headache #1 - `extractChain()` apparently requires `id` as an explicit union type, rather than as a standard number. I bodged it here by casting to `any`, but that sucks.

src/gasPriceOracle/oracle.ts
```diff
@@ -14,35 +14,36 @@ import * as polygon from "./adapters/polygon";
...
-      throw new Error(`No suitable gas price oracle for Chain ID ${chainId}`);
-    }
-    gasPriceFeed = chainIsOPStack(chainId) ? ethereum.eip1559 : ethereum.legacy;
+  let maxFeePerGas: bigint;
+  let maxPriorityFeePerGas: bigint;
+  if (gasPriceFeeds[chainId]) {
+		// declare function action<chain extends Chain | undefined>(client: Client<Transport, chain>): void
+		// action(viemProvider);
+		// https://github.com/wevm/viem/issues/2763
+    ({ maxFeePerGas, maxPriorityFeePerGas } = await gasPriceFeeds[chainId](viemProvider as any, chainId));
```
pxrl, line 35: @gsteenkamp89 @dohaki Example #2 where I can't get anything working without as typecast to `any`. The `Chain` input to `createPublicClient()` results in a very narrow inferred data type that I am struggling to know how to specify. This makes it really hard to pass that `PublicClient` instance around to other functions. The referenced github issue seems to acknowledge this and provides a solution. I don't understand what's going on there, though. Any clues?

pxrl (commented):
Just dropping some initial breadcrumbs.

src/gasPriceOracle/oracle.ts
gsteenkamp89: @pxrl You can opt out of viem's narrow type inference when some of your values are dynamic. Like here, your `chainId` is not strictly known at compile time so it is just a `number`. 

Don't use viem's `extractChain` utility. It is just `Array.find` with static param checking
```ts
const chain: Chain | undefined = Object.values(chains).find((chain) => chain.id === chainId);
```

But it is useful to note that the `PublicClient` type exported by viem doesn't have to be super narrow to access the necessary methods you would likely use. It doesn' have to know which chain you're on to fetch a txReceipt for example, it just has to know that _a_ chain has been configured in this client. 

So you can build this type:

```ts
export type ConfiguredPublicClient = PublicClient<Transport, Chain>;
```

So what I would do is this:

```ts
export type ConfiguredPublicClient = PublicClient<Transport, Chain>;

function getViemPublicClient(chainId: number, transport?: Transport): ConfiguredPublicClient {
  // don't use viem's extractChain utility when using dynamic chainIds
  const chain: Chain | undefined = Object.values(chains).find((chain) => chain.id === chainId);
  if (!chain) {
    throw new Error("Chain config not found");
  }
  // default to public http endpoint
  transport ??= http();
  return createPublicClient({ chain, transport });
}
```

gsteenkamp89: Another option would be to write a type guard that will narrow the type of the chainId such that it is in fact a chainId of one of the members of the `chains` union. This way you can get nicer type inference, but this is probably not needed, but here is how I would do it

```ts
function isSupportedChainId<Chains extends Chain[], Id extends Chains[number]["id"]>(
  chains: Chains,
  chainId: number
): chainId is Id {
  return chains.some((chain) => chain.id === chainId);
}

function getViemPublicClient(chainId: number, transport?: Transport) {
  const supportedChains = Object.values(chains);

  if (!isSupportedChainId(supportedChains, chainId)) {
    throw new Error("Chain not supported");
  }

  const chain = extractChain({ chains: supportedChains, id: chainId });
  if (!chain) {
    throw new Error("Chain config not found");
  }
  // default to public http endpoint
  transport ??= http();
  return createPublicClient({ chain, transport });
}
```

pxrl: > @pxrl You can opt out of viem's narrow type inference when some of your values are dynamic. Like here, your `chainId` is not strictly known at compile time so it is just a `number`.
> 
> Don't use viem's `extractChain` utility. It is just `Array.find` with static param checking
> 
> ```ts
> const chain: Chain | undefined = Object.values(chains).find((chain) => chain.id === chainId);
> ```

This works better and seems to build without problem. My editor (running tsserver in the background) opens the gates of hell on this one though. I can probably live with it.

err

src/gasPriceOracle/util.ts
```diff
@@ -1,6 +1,15 @@
-import { providers } from "ethers";
-import { BigNumber } from "../utils";
+import assert from "assert";
+import { type Chain, type PublicClient, createPublicClient, http, Transport } from "viem";
+import * as chains from "viem/chains";
```
pxrl, line 3: fwiw it seems like this would be a non-starter for the FE/API, so I'd probably import the specific chains we support as a follow-up change.

dohaki: I think this should be fine because the API will be using the `gasPriceOracle` and we don't care too much about the bundle size in that context

src/gasPriceOracle/oracle.ts
```diff
@@ -16,33 +16,34 @@ import * as polygon from "./adapters/polygon";
...
+  let maxPriorityFeePerGas: bigint;
+  if (gasPriceFeeds[chainId]) {
+    // declare function action<chain extends Chain | undefined>(client: Client<Transport, chain>): void
+    // action(viemProvider);
+    // https://github.com/wevm/viem/issues/2763
+    ({ maxFeePerGas, maxPriorityFeePerGas } = await gasPriceFeeds[chainId](viemProvider, chainId));
+  } else {
+    let gasPrice: bigint | undefined;
+    ({ maxFeePerGas, maxPriorityFeePerGas, gasPrice } = await viemProvider.estimateFeesPerGas());
+    // console.log(`Got maxFeePerGas: ${maxFeePerGas}, maxPriorityFeePerGas: ${maxPriorityFeePerGas}, gasPrice: ${gasPrice}.`);
```
dohaki, line 39: Do we still need this?

src/gasPriceOracle/oracle.ts
```diff
@@ -16,33 +16,34 @@ import * as polygon from "./adapters/polygon";
...
-    if (!legacyFallback) {
-      throw new Error(`No suitable gas price oracle for Chain ID ${chainId}`);
-    }
-    gasPriceFeed = chainIsOPStack(chainId) ? ethereum.eip1559 : ethereum.legacy;
+  let maxFeePerGas: bigint;
+  let maxPriorityFeePerGas: bigint;
+  if (gasPriceFeeds[chainId]) {
+    // declare function action<chain extends Chain | undefined>(client: Client<Transport, chain>): void
+    // action(viemProvider);
+    // https://github.com/wevm/viem/issues/2763
```
dohaki, line 34: Is this still required?

src/gasPriceOracle/adapters/ethereum.ts
```diff
@@ -1,28 +1,19 @@
...
-  [feeData.lastBaseFeePerGas, feeData.maxPriorityFeePerGas].forEach((field: BigNumber | null) => {
-    if (!BigNumber.isBigNumber(field) || field.lt(bnZero)) gasPriceError("getFeeData()", chainId, feeData);
-  });
-
-  const maxPriorityFeePerGas = feeData.maxPriorityFeePerGas as BigNumber;
-  const maxFeePerGas = maxPriorityFeePerGas.add(feeData.lastBaseFeePerGas as BigNumber);
-
-  return { maxPriorityFeePerGas, maxFeePerGas };
+export function eip1559(provider: PublicClient, _chainId: number): Promise<InternalGasPriceEstimate> {
+  return provider.estimateFeesPerGas();
```
dohaki, line 5: The linter is erroring here. But I assume the reason why we keep the same function signature is for backwards-compatibility? In that case, we can add something like 
```
{ "argsIgnorePattern": "^_" }
```
to eslint config. See https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern

dohaki (commented):
This looks good 👍 I like the approach to gradually try viem in the SDK. Left some minor comments and there are still a bunch of linting errors. I think we should add https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern to the eslint config

dohaki (approved):
Looking good 👍 Left one nit

src/gasPriceOracle/oracle.ts
```diff
@@ -17,24 +21,46 @@ import * as polygon from "./adapters/polygon";
...
   chainId?: number,
+  transport?: Transport,
   legacyFallback = true
 ): Promise<GasPriceEstimate> {
   if (chainId === undefined) {
     ({ chainId } = await provider.getNetwork());
   }
 
-  const gasPriceFeeds: { [chainId: number]: GasPriceFeed } = {
+  const useViem = process.env[`NEW_GAS_PRICE_ORACLE_${chainId}`] === "true";
```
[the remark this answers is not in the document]
pxrl: I generally think reading env-vars directly in application code is an anti-pattern and I would normally propose config options to be implemented as part of the interface. I embedded it here because this config is only really intended as a short-term handle to support proving of viem, and I think it'd be removed as part of a cleanup task to remove ethers altogether. In that case I figured it made more sense to embed it directly to avoid imposing config burden on the calling code. wdyt?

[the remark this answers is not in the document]
dohaki: ok yea in that case it makes sense to me like that

PR #750, opened 2024-10-14
improve(SpokePoolClient): Multicall latestBlock updates

pxrl:
The SpokePoolClient is currently throwing semi-regular errors on update. This seems to be particularly occurring against Polygon. The underlying reason for this is that the RPC quroum system is performing an eth_call against multiple providers, tagged at the latest block number as resolved by only one of them. Some providers may not have resolved that block number yet, in which case the query fails.

The instance of this can presumably be reduced by bundling the two calls where this occurs into a single SpokePool multicall. As a side-benefit, it will reduce RPC consumption.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -499,20 +499,28 @@ export class SpokePoolClient extends BaseAbstractClient {
       };
     });
 
+    const { spokePool } = this;
     this.log("debug", `Updating SpokePool client for chain ${this.chainId}`, {
       eventsToQuery,
       searchConfig,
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient.ts
pxrl, line 1: Reminder to bump package.json version before merging this.

pxrl: Done: https://github.com/across-protocol/sdk/pull/750/commits/7eda9b90bc8f5c20bddfa09824436b5e7cc4d6d9

PR #752, opened 2024-10-16
refactor: Remove SpokePoolClient oldestTime lookup

pxrl:
The `oldestTime` attribute is only needed for a very specific check when validating block ranges in the dataworker. Removing it from the standard SpokePoolClient update flow is a simplification for non-dataworker use-cases - i.e. relayer, monitor, finalizer. The dataworker can instead resolve the timestamp of the searchConfig fromBlock on demand.

The underlying motivation for this change is to simplify the interface between the relayer and the external SpokePool listener processes in order to make way for additional updates.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -66,6 +66,7 @@ export type SpokePoolUpdate = SpokePoolUpdateSuccess | SpokePoolUpdateFailure;
  */
 export class SpokePoolClient extends BaseAbstractClient {
   protected currentTime = 0;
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient.ts
pxrl, line 1: Reminder to bump package version before merging this.

pxrl: https://github.com/across-protocol/sdk/pull/752/commits/8960cb1bef76736ee4086c5bb493bd9dd7ba6719

dohaki (approved):
LGTM 👍

PR #753, opened 2024-10-23
feat: return raw `gasUnits` without padding

dohaki:
Additionally return raw unpadded `gasUnits` in `getGasCosts` method. This allows us to correctly make use of a cached value by caching the unpadded `gasUnits` value and passing that in.

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -230,6 +230,7 @@ export function retry<T>(call: () => Promise<T>, times: number, delayS: number):
 }
 
 export type TransactionCostEstimate = {
+  gasUnits: BigNumber; // Units: gas
   nativeGasCost: BigNumber; // Units: gas
   tokenGasCost: BigNumber; // Units: wei (nativeGasCost * wei/gas)
 };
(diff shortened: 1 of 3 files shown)


gsteenkamp89 (approved):

src/utils/common.ts
```diff
@@ -261,37 +262,38 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
       const gasPriceEstimate = await getGasPriceEstimate(provider, chainId);
       gasPrice = gasPriceEstimate.maxFeePerGas;
     }
-    tokenGasCost = nativeGasCost.mul(gasPrice);
+    tokenGasCost = gasUnits.mul(gasPrice);
   }
 
   // Scale the results by the computed multiplier.
-  nativeGasCost = nativeGasCost.mul(gasTotalMultiplier).div(fixedPointAdjustment);
+  const nativeGasCost = gasUnits.mul(gasTotalMultiplier).div(fixedPointAdjustment);
```
pxrl, line 292: @dohaki Wdyt about applying the gas multiplier back in the `relayFeeCalculator`? That's the only place this function is used anyway, and I think it'd probably minimise confusion around `gasUnits` and `nativeGasCost`.

dohaki: yea that is prob cleaner. but then for both values right? `tokenGasCost` and `nativeGasCost`

PR #755, opened 2024-10-24
test: fix with custom transport

dohaki:


test/relayFeeCalculator.test.ts
```diff
@@ -74,6 +75,7 @@ class ExampleQueries implements QueryInterface {
   constructor(private defaultGas = "305572") {}
 
   getGasCosts(): Promise<TransactionCostEstimate> {
+    console.log("#####");
```
pxrl, line 78: fwiw you can use the logger in test by running `LOG_IN_TEST=yes yarn test ...`.

dohaki: ah ok good to know

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -61,17 +62,25 @@ export class QueryBase implements QueryInterface {
...
     relayer = DEFAULT_SIMULATED_RELAYER_ADDRESS,
-    gasPrice = this.fixedGasPrice,
-    gasUnits?: BigNumberish,
-    omitMarkup?: boolean
+    options: Partial<{
+      gasPrice: BigNumberish;
+      gasUnits: BigNumberish;
+      omitMarkup: boolean;
+      transport: Transport;
+    }> = {}
```
pxrl, line 80: +1 for this approach - I really think it improves usability to eliminate the need for long convoluted sets of parameters.

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -17,14 +17,14 @@ import {
...
+import { Transport } from "viem";
 
 // This needs to be implemented for every chain and passed into RelayFeeCalculator
 export interface QueryInterface {
   getGasCosts: (
     deposit: Deposit,
     relayer: string,
-    gasPrice?: BigNumberish,
-    gasLimit?: BigNumberish
+    options?: Partial<{ gasPrice: BigNumberish; gasUnits: BigNumberish; omitMarkup: boolean; transport: Transport }>
```
pxrl, line 27: I think it'd be nice to avoid having `Transport` bleed out into the `QueryInterface` definitions. I'm not sure if that's acheivable atm and it may not be actionable right now. But in general I think it would be good to keep the implementation choices of the gas price oracle defined internally - especially since it's only really used for test (...it might hint at the need to define a special class for test? idk).

pxrl (approved):

src/relayFeeCalculator/relayFeeCalculator.ts
dohaki: So currently we use `getGasCosts` directly in the API. That's why I had to expose it here 🤔

dohaki: But yea agree it would be nicer to keep it internal

PR #757, opened 2024-10-25
improve(retryProvider): Avoid retry on eth_call & eth_estimateGas reverts

pxrl:
eth_call requests where the provider returns a response indicating revert have a near-zero chance of succeeding against another RPC provider. Instead of wasting time rotating through the list of RPC providers, just throw immediately and let the upper layers handle it.

This will potentially save quite a bit of time in the relayer in cases where it unsuccessfully performs a static call to determine whether a fill would complete and should also notably reduce RPC consumption across secondary and tertiary providers.


src/providers/retryProvider.ts
```diff
@@ -88,9 +89,27 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+            const {
+              error: { code, message },
+            } = err;
+            // [-32768, -32100] is reserved by the JSON-RPC spec.
+            // [-32099, -32000] is allocated for implementation-defined responses.
+            // Everything else is available for use by the application space.
+            // Most node implementations return 3 for an eth_call revert, but some return -32000.
+            // See also https://www.jsonrpc.org/specification
+            if (code < -32768 || code > -32100) {
+              if (method === "eth_call" && message.toLowerCase().includes("revert")) {
```
pxrl, line 103: This could maybe also be conditional on `quorumThreshold === 1`.

src/providers/retryProvider.ts
pxrl, line 1: Still todo after discussion w/ @mrice32: update `_trySend()` such that it can identify a revert on an eth_call and short-circuit the retry loop that is inevitably occurring. This would shave off at least 3+ seconds per call.

src/providers/retryProvider.ts
pxrl: https://github.com/across-protocol/sdk/pull/757/commits/90374cb8545867b7c70d2cf031c80a4d58fbb006

src/providers/retryProvider.ts
```diff
@@ -260,6 +268,27 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
     return response;
   }
 
+  // For an error emitted in response to an eth_call request, determine whether the response body indicates that the
+  // call reverted during execution. The exact RPC responses returned can vary, but `error.body` has consistently
```
pxrl, line 272:
```suggestion
  // call reverted during execution. The exact RPC responses returned can vary, but `error.body` has reliably
```

src/providers/types.ts
```diff
@@ -1,2 +1,22 @@
+import { literal, nullable, number, string, type, union } from "superstruct";
+
 export type RPCProvider = "ALCHEMY" | "DRPC" | "INFURA" | "INFURA_DIN";
 export type RPCTransport = "https" | "wss";
+
+// JSON-RPC 2.0 Error object
+// See 5.1 Error Object https://www.jsonrpc.org/specification
```
pxrl, line 7:
```suggestion
// See 5 Response Object https://www.jsonrpc.org/specification
```

pxrl: https://github.com/across-protocol/sdk/pull/757/commits/ddc8ec069cc0ed20de74993b0b2780a48ee5e6d9

gsteenkamp89 (approved):
Awesome work, this should save a ton of time!

src/providers/retryProvider.ts
```diff
@@ -271,6 +300,9 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
         return await this._sendAndValidate(provider, method, params);
       } catch (_err: unknown) {
         error = _err;
+        if (method === "eth_call" && this.callReverted(method, error)) {
+          break;
+        }
```
gsteenkamp89, line 305: bail out early, nice 👍

src/providers/retryProvider.ts
```diff
@@ -271,6 +300,9 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
         return await this._sendAndValidate(provider, method, params);
       } catch (_err: unknown) {
         error = _err;
+        if (method === "eth_call" && this.callReverted(method, error)) {
```
mrice32, line 303: Not a strong opinion, but it seems like our way of modeling responses is a little off here and the solution should be just to fix that modeling issue.

Retries are intended to deal with _invalid responses_ (think 500s). A revert is a _valid response_ that happens to generate an application level error (maybe 400 is the right mental model here). _Valid responses_ do not benefit from retries.

Should we be detecting these at the very bottom level (in _trySend), not throwing, and returning special type that indicates a valid response that we convert into an error in the top-level send method?

I think this will avoid us having to do a ton of special

pxrl: It's a good call; we should improve this. Implementing repeated catching and inspection of errors is clunky, and detecting this at the lowest level and returning a meaningful response upwards is more architecturally pure. I also think there are other cases where we'll want custom behaviour - i.e. I think we could probably also tackle the "header not found" error down here, which would ideally help to reduce a fairly significant source of RPC errors (i.e. as we query multiple providers at a blockTag sourced from only a single provider).

I do admit that at this point I'm mostly just interested to get some feedback on how this affects performance at scale in production; the relative ease that we can surgically insert this current band-aid makes me more comfortable in proceeding with a less-than-ideal solution for now, because I'm pretty confident we'll be back here shortly to make further improvements if this change delivers the anticipated performance boost.

mrice32 (approved):

PR #758, opened 2024-10-30
feat: support Aleph Zero

melisaguevara:
To support Aleph Zero, this PR:
- Bumps contracts and constants packages to get the new chain definitions.
- Defines an override query class for `relayFeeCalculator` as Aleph Zero has a custom gas token.
- Specifies the `gasPriceOracle` implementation to use the same as Arbitrum.

We should wait to merge this until we have the final version of the contracts package but in the meantime it can be reviewed.


bmzig (approved):

james-a-morris (approved):

james-a-morris (changes requested):
(hold for resolution on contracts)

dohaki (commented):
Just flagging that we need to add the multicall3 address here https://github.com/across-protocol/sdk/blob/b43acb896801647dce6bdea82dfad90b908f7224/src/utils/Multicall.ts#L19

This step might be missing in the linear issue

melisaguevara:
> Just flagging that we need to add the multicall3 address here

Added here [0c90608](https://github.com/across-protocol/sdk/pull/758/commits/0c90608aa9d933696f80df7c517df52637a118d1)

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "3.2.11",
+  "version": "3.2.11-beta.1",
```
pxrl, line 4: The only thing I think we should be careful about is this version; I assume this PR will be held until a proper release can be made from the contracts repo.

melisaguevara: Yes, exactly @pxrl

james-a-morris (approved):

PR #761, opened 2024-10-30
fix(RetryProvider): Avoid unnecessary promise reassignment

pxrl:
Debug logging has shown that the existing logic unconditionally completes the for() loop `this.retries` times for each RPC query. This doesn't seem to submit `this.retries` queries but does result in a lot of unnecessary promise reassignment, which _might_ (speculation) be wasting CPU and RAM in a tight loop.

The structure of the change here is made in anticipation of a follow-up commit to introduce some intelligence into the retry loop to avoid retries when the failure was actually "legitimate" (i.e. eth_call revert). Without this change it was proving very difficult to bail out of the retry loop due to the promise chaining that implies the use of closures.

### DIFF
--- a/src/providers/retryProvider.ts
+++ b/src/providers/retryProvider.ts
@@ -265,25 +265,17 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
     method: string,
     params: Array<unknown>
   ): Promise<unknown> {
-    const loop = true;
-    let promise: Promise<unknown>;
-    let i = 0;
-
-    do {
-      promise = this._sendAndValidate(provider, method, params);
+    let error: unknown;
+    for (let i = 0; i < this.retries; ++i) {
       try {
-        await promise;
-        break;
-      } catch (err: unknown) {
-        if (++i >= this.retries) {
-          throw err;
-        }
-
+        return await this._sendAndValidate(provider, method, params);
+      } catch (_err: unknown) {
+        error = _err;
         await delay(this.delay);
       }
-    } while (loop);
+    }
 
-    return promise;
+    throw error;
   }
 
   _getQuorum(method: string, params: Array<unknown>): number {
--- a/src/providers/retryProvider.ts
+++ b/src/providers/retryProvider.ts
@@ -260,11 +260,30 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
     return response;
   }
 
-  _trySend(provider: ethers.providers.StaticJsonRpcProvider, method: string, params: Array<unknown>): Promise<unknown> {
+  async _trySend(
+    provider: ethers.providers.StaticJsonRpcProvider,
+    method: string,
+    params: Array<unknown>
+  ): Promise<unknown> {
+    const loop = true;
+
     let promise = this._sendAndValidate(provider, method, params);
-    for (let i = 0; i < this.retries; i++) {
-      promise = promise.catch(() => delay(this.delay).then(() => this._sendAndValidate(provider, method, params)));
-    }
+    let i = 0;
+
+    do {
+      try {
+        await promise;
+        break;
+      } catch (err: unknown) {
+        if (++i >= this.retries) {
+          throw err;
+        }
+
+        await delay(this.delay);
+        promise = this._sendAndValidate(provider, method, params);
+      }
+    } while (loop);
+
     return promise;
   }
 
--- a/src/providers/retryProvider.ts
+++ b/src/providers/retryProvider.ts
@@ -265,17 +265,17 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
     method: string,
     params: Array<unknown>
   ): Promise<unknown> {
-    let error: unknown;
-    for (let i = 0; i < this.retries; ++i) {
-      try {
-        return await this._sendAndValidate(provider, method, params);
-      } catch (_err: unknown) {
-        error = _err;
-        await delay(this.delay);
+    let { retries } = this;
+    while(true) {
+      const [settled] = await Promise.allSettled([this._sendAndValidate(provider, method, params)]);
+      if (settled.status === "fulfilled") {
+        return settled.value;
       }
+      if (retries-- <= 0) {
+        throw settled.reason;
+      }
+      await delay(this.delay);
     }
-
-    throw error;
   }
 
   _getQuorum(method: string, params: Array<unknown>): number {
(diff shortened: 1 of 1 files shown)


src/providers/retryProvider.ts
```diff
@@ -260,11 +260,30 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+        if (++i >= this.retries) {
+          throw err;
+        }
+
+        await delay(this.delay);
+        promise = this._sendAndValidate(provider, method, params);
+      }
+    } while (loop);
+
     return promise;
```
pxrl, line 287: I wanted to drop this and return directly on like 275, but the linter was unhappy that there was a theoretical risk of returning undefined in that case. I don't believe there's any practical risk of that because the only alternative exit point is to throw an exception.

src/providers/retryProvider.ts
```diff
@@ -260,11 +260,30 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
   }
 
-  _trySend(provider: ethers.providers.StaticJsonRpcProvider, method: string, params: Array<unknown>): Promise<unknown> {
+  async _trySend(
+    provider: ethers.providers.StaticJsonRpcProvider,
+    method: string,
+    params: Array<unknown>
+  ): Promise<unknown> {
+    const loop = true;
+
```
pxrl, line 269:
```suggestion
```

pxrl:
Here's the follow-up PR that leverages this refactor: https://github.com/across-protocol/sdk/pull/757

src/providers/retryProvider.ts
```diff
@@ -260,11 +260,29 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+      try {
+        await promise;
+        break;
+      } catch (err: unknown) {
+        if (++i >= this.retries) {
+          throw err;
+        }
+
+        await delay(this.delay);
+        promise = this._sendAndValidate(provider, method, params);
```
dohaki, line 282: Is this re-assignment necessary?

dohaki (approved):
Looks good! Left one question

src/providers/retryProvider.ts
pxrl: It was given the surrounding code, but I was able to consolidate it in this commit: https://github.com/across-protocol/sdk/pull/761/commits/25516ed445a8246e621883efc40a32f43c4fe559

gsteenkamp89 (approved):

src/providers/retryProvider.ts
```diff
@@ -260,11 +260,29 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
-    }
+    let i = 0;
+
+    do {
+      try {
+        await promise;
+        break;
+      } catch (err: unknown) {
+        if (++i >= this.retries) {
+          throw err;
```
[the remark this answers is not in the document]
pxrl: I like the suggestion - it's cleaner and more concise than what I had. At this point I think we only care about the final error, but we might iterate on this a bit and start exposing the underlying errors. It might be noisy, though.

https://github.com/across-protocol/sdk/pull/761/commits/d0a877df2808af58d5218f277dcfcc08b2da88d4

src/providers/retryProvider.ts
pxrl: Subsequently negated by this commit anyway: https://github.com/across-protocol/sdk/commit/d0a877df2808af58d5218f277dcfcc08b2da88d4

src/providers/retryProvider.ts
```diff
@@ -260,12 +260,22 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
-    let promise = this._sendAndValidate(provider, method, params);
-    for (let i = 0; i < this.retries; i++) {
-      promise = promise.catch(() => delay(this.delay).then(() => this._sendAndValidate(provider, method, params)));
+  async _trySend(
+    provider: ethers.providers.StaticJsonRpcProvider,
+    method: string,
+    params: Array<unknown>
+  ): Promise<unknown> {
+    let error: unknown;
+    for (let i = 0; i < this.retries; ++i) {
```
mrice32, line 269: This is subtle, but I think with the way the loop is designed now, we're using this.retries as the number of total tries, not retries.

Put differently, in the previous loop, 2 retries would imply 3 total attempts (first with 2 retries). In the current loop, 2 retries would imply to total attempts.

I know the language here is a little confusing. Just want to make sure we're not unintentionally changing behavior.

src/providers/retryProvider.ts
```diff
@@ -260,12 +260,22 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+      try {
```
mrice32, line 270: My best attempt at this loop with 0 promise chaining and the clearest control flow would be:
```ts
const retriesLeft = this.retries;
while(true) {
  const [settled] = await Promise.allSettled([this.sendAndValidate(provider, method, params)]);
  if (settled.status === "fulfilled") {
    return settled.value;
  }
  if (retriesLeft-- <= 0) {
    throw settled.reason;
  }
  await delay(this.delay);
}

File: src/providers/retryProvider.ts [hunk 7]
```diff
@@ -260,12 +260,22 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+        return await this._sendAndValidate(provider, method, params);
+      } catch (_err: unknown) {
+        error = _err;
+        await delay(this.delay);
```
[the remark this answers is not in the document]
pxrl: Good observation. Fixed per your proposal: https://github.com/across-protocol/sdk/pull/761/commits/bbaa33ac1ba3b9d8c2d22560a3d382e2fba67308

src/providers/retryProvider.ts
pxrl: It was unintentional - good catch. Should also be fixed per your proposal: https://github.com/across-protocol/sdk/pull/761/commits/bbaa33ac1ba3b9d8c2d22560a3d382e2fba67308

src/providers/retryProvider.ts
pxrl: Implemented: https://github.com/across-protocol/sdk/pull/761/commits/bbaa33ac1ba3b9d8c2d22560a3d382e2fba67308

mrice32 (approved):
LGTM!

PR #765, opened 2024-11-05
fix(TokenUtils): Support identifying Zora USDzC

pxrl:
This PR adds missing support for Zora USDC to a TokenUtils utility function. I spotted the need for this because the relayer was refusing to fill a USDzC -> Arbitrum USDC.e deposit and complained that it implied a token swap.

I don't feel great about this function; we generally need:
- A new implementation that does not rely on unique token addresses.
- A consolidation of various "is this bridged USDC?" functionalities that are scattered around the various repos.


bmzig (approved):
:raised_hands:

PR #771, opened 2024-11-15
feat: new CoinGecko fn getCurrentPriceById

melisaguevara:


src/coingecko/Coingecko.ts
```diff
@@ -150,6 +151,18 @@ export class Coingecko {
     assert(tokenPrice !== undefined);
     return [tokenPrice.timestamp.toString(), tokenPrice.price];
   }
+
+  async getCurrentPriceById(contractAddress: string, currency = "usd"): Promise<number> {
+    const coingeckoId = getCoingeckoTokenIdByAddress(contractAddress);
```
bmzig, line 156: Can/Should we do caching of prices here similar to [L127](https://github.com/across-protocol/sdk/blob/16e5906dbd4d58dc9d5e947bc28c0b0b8a3dc2a8/src/coingecko/Coingecko.ts#L127)?

james-a-morris (approved):
we should bump sdk version

src/coingecko/Coingecko.ts
nicholaspai: I think we should yeah

james-a-morris: We can/should - but I think this should be a follow-on. This function follows similar logic to the code below. My recommendation is that we should get this out and fast-follow after the AZERO launch. 

https://github.com/across-protocol/sdk/blob/16e5906dbd4d58dc9d5e947bc28c0b0b8a3dc2a8/src/coingecko/Coingecko.ts#L93-L116 and is

nicholaspai: aren't we going to make this request every single run though? Isn't that going to result in a lot of CG reqs?

james-a-morris: Ah shoot you're right this will be used in the relayer.

src/coingecko/Coingecko.ts
```diff
@@ -275,6 +270,58 @@ export class Coingecko {
...
+    const now: number = msToS(Date.now());
+    const tokenPrice: CoinGeckoPrice | undefined = priceCache[contractAddress];
+    if (tokenPrice === undefined || tokenPrice.timestamp + this.maxPriceAge <= now) {
+      if (this.maxPriceAge > 0) {
+        this.logger.debug({
+          at: "Coingecko#getCurrentPriceByContract",
+          message: `Cache miss on ${platform_id}/${currency} for ${contractAddress}`,
+          maxPriceAge: this.maxPriceAge,
+          tokenPrice: tokenPrice,
+        });
```
james-a-morris, line 288: Would we not always want to display this log?

src/coingecko/Coingecko.ts
```diff
@@ -275,6 +270,58 @@ export class Coingecko {
...
+      priceCache[contractAddress] = {
+        address: contractAddress,
+        price: cgPrice[currency],
+        timestamp: cgPrice.last_updated_at,
+      };
+      this.logger.debug({
+        at: "Coingecko#updatePriceCache",
+        message: `Updated ${platform_id}/${currency}/${contractAddress} token price cache.`,
+      });
+    } else if (cgPrice.last_updated_at === priceCache[contractAddress].timestamp) {
```
james-a-morris, line 316: OOC would we want to relax this constraint? If we make it `else` then we'd get the case where we pass an earlier value
```suggestion
    } else {
```

src/coingecko/Coingecko.ts
```diff
@@ -275,6 +270,58 @@ export class Coingecko {
...
+    currency: string,
+    platform_id: string
+  ): CoinGeckoPrice | undefined {
+    const priceCache = this.getPriceCache(currency, platform_id);
+    const now: number = msToS(Date.now());
+    const tokenPrice: CoinGeckoPrice | undefined = priceCache[contractAddress];
+    if (tokenPrice === undefined || tokenPrice.timestamp + this.maxPriceAge <= now) {
+      if (this.maxPriceAge > 0) {
+        this.logger.debug({
+          at: "Coingecko#getCurrentPriceByContract",
```
bmzig, line 284: I think you should change the markers on this and the other (3?) logs from `#getCurrentPriceByContract`  or `#getContractPrices` to `#getCachedAddressPrice` or `#updatePriceCache` (depending on the function they are in).

melisaguevara: Def, updated here: [3c4ba03](https://github.com/across-protocol/sdk/pull/771/commits/3c4ba03f51d271e34c98042f5d02c2e19d48e141)

src/coingecko/Coingecko.ts
melisaguevara: Yes, we can do that. I would expect that case to be uncommon but can def use just else here.

src/coingecko/Coingecko.ts
melisaguevara: You mean we should also have a log when tokenPrice is undefined? We already have a log for when it is defined, I'm not sure this is needed.

bmzig (approved):

PR #773, opened 2024-11-18
extend client to resolve non across supported tokens

gsteenkamp89:
related to ACX-3283

We need to be able to find current and historic prices for tokens that are NOT supported by Across, ie not in the TOKEN_SYMBOLS_MAP constant.

Necessary Updates:
- For current prices we need to be able to find the platformId with a chainId. 
- For historic prices we need to find the tokenId for any token on a chain. First attempt to find the id with TOKEN_SYMBOLS_MAP and if not supported, use a discovery endpoint from CG to build a map of tokenId => {platform: address } we can use to resolve tokenIds


gsteenkamp89:
Looks like this might be duplicating work done in this PR https://github.com/across-protocol/sdk/pull/771 🤔

PR #774, opened 2024-11-18
improve(PriceClient): Add default price adapter

pxrl:
This adapter can be instantiated by the caller to permit the PriceClient to default to a price of 0 for any token that it was unable to resolve via the preceding adapters. It can additionally be automatically instantiated by the PriceClient based on an input option being set. This therefore provides the caller with a simple way of controlling whether a token price resolution failure should throw or not. It additionally provides a pathway for the caller to be able to specify fixed prices for one or more tokens.

### DIFF
--- a/src/priceClient/adapters/default.ts
+++ b/src/priceClient/adapters/default.ts
@@ -20,7 +20,7 @@ export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
     return price[0];
   }
 
-  async getPricesByAddress(addresses: string[], _currency: string): Promise<TokenPrice[]> {
-    return addresses.map((address) => ({ address, price: 0, timestamp: 0 }));
+  getPricesByAddress(addresses: string[], _currency: string): Promise<TokenPrice[]> {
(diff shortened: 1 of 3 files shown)


james-a-morris (approved):

src/priceClient/adapters/default.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This adapter can be used as a last-resort to default to a token price of 0
+ * in case all other adapters fail to resolve a token price.
+ *
+ * Note: This adapter could be augmented to permit returning a custom price on a per-token bsis.
+ * This would allow the caller to peg token prices, which has been a use cases in the Across API.
+ * That's left as an open possibility for now.
+ */
+export class PriceFeed extends BaseHTTPAdapter implements PriceFeedAdapter {
+  constructor() {
+    super("Zero Adapter", "127.0.0.1", {});
```
pxrl, line 15:
```suggestion
    super("Default Adapter", "127.0.0.1", {});
```

pxrl:
This has tested OK locally when the DefaultAdapter is specified explicitly, as well as implicitly via toggling throwOnError.

PR #781, opened 2024-11-22
fix: hotfix to remove bad config update

mrice32:

PR #783, opened 2024-11-22
feat(AcrossConfigStoreClient): Handle invalid TokenConfig updates better

nicholaspai:
Follow up to #781

This PR includes a few code cleanups to the AcrossConfigStoreClient class:

1. Improves on logic to skip one specific transaction hash by protecting better against future invalid TokenConfig updates. Adds a single rule, for now, to check that a rate model's `UBar` parameter is between 0% and 100% (exclusive). We can build off this PR in the future to add other rules.
2. Removes reliance on `@umaprotocol/rateModel` class
3. Removes UBA logic


src/clients/AcrossConfigStoreClient/AcrossConfigStoreClient.ts
```diff
@@ -89,37 +105,43 @@ export class AcrossConfigStoreClient extends BaseAbstractClient {
...
+  ): RateModel {
     // Use route-rate model if available, otherwise use default rate model for l1Token.
     const route = `${originChainId}-${destinationChainId}`;
     const routeRateModel = this.getRouteRateModelForBlockNumber(l1Token, route, blockNumber);
-    return routeRateModel ?? this.rateModelDictionary.getRateModelForBlockNumber(l1Token, blockNumber);
+    if (routeRateModel) {
+      return routeRateModel;
+    }
+
+    const defaultRateModelUpdate = sortEventsDescending(this.cumulativeRateModelUpdates).find(
```
nicholaspai, line 123: The cleanup here is to read rate model events directly from this class rather than rely on this client's `update()` method to store the events into a `RateModelDictionary` class and then read it from there.

src/contracts/acrossConfigStore.ts
```diff
@@ -14,21 +14,21 @@ const L1TokenConfigSs = object({
   rateModel: RateModelSs,
   routeRateModel: optional(record(string(), RateModelSs)),
 });
-export type RateModel = Infer<typeof RateModelSs>;
-export type L1TokenConfig = Infer<typeof L1TokenConfigSs>;
+type RateModelSS = Infer<typeof RateModelSs>;
```
nicholaspai, line 17: Renamed these types and removed the `export` so that other classes don't confuse this type with the `RateModel` type in `lpFeeCalculator/rateModel.ts`

bmzig (approved):

PR #786, opened 2024-11-26
feat: Custom gas price oracle adapter for Linea

pxrl:


gsteenkamp89 (approved):
Looks good!
Have we tested this is actually cheaper?

pxrl:
> Looks good! Have we tested this is actually cheaper?

Yep - in my testing it looked to be about 0.5 Gwei less to query via linea_estimateGas (even with a dummy transaction) over eth_gasPrice().

PR #789, opened 2024-11-29
Widen logger type for profiler

gsteenkamp89:
Currently this profiler does not accept a logger that doesn't exactly implement Winston's Logger interface, blocking us from using it in the API/frontend.

This allows any logger that has `warn` and `debug` methods.

### DIFF
--- a/src/utils/Profiler.ts
+++ b/src/utils/Profiler.ts
@@ -1,7 +1,11 @@
 import { performance } from "node:perf_hooks";
-import { Logger } from "winston";
 import crypto from "crypto";
 import { DefaultLogLevels } from "./LogUtils";
+
+type Logger = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  log: (level: string, ...meta: any[]) => void;
+};
 
 type Detail = {
   message?: string;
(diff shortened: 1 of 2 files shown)


src/utils/Profiler.ts
```diff
@@ -1,8 +1,14 @@
 import { performance } from "node:perf_hooks";
-import { Logger } from "winston";
 import crypto from "crypto";
 import { DefaultLogLevels } from "./LogUtils";
 
+type Logger = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  debug: (...args: any[]) => void;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  warn: (...args: any[]) => void;
```
pxrl, line 9: Could we use `unknown` here instead of `any` ?

pxrl (approved):
One q, otherwise OK :+1:

src/utils/Profiler.ts
gsteenkamp89: i want to keep it as close to most implementations which is usually types as `any` (winston ans global `console` in node and browser contexts type it this way).
Also using unknown is actually a bit too narrow, stopping us from using the spread operator

PR #796, opened 2024-12-17
improve(BundleDataClient): Add comment about not running binary search on "unsafe" deposit ID's

nicholaspai:
I'm combing through the SDK for any functions that will need to change once this contracts  [PR](https://github.com/across-protocol/contracts/pull/583) is merged and here are some changes:
- Added comments to BundleDataClient
- Export useful constant `MAX_SAFE_DEPOSIT_ID`
- Export useful utility `isUnsafeDepositId()`
- Delete unused functions that wouldn't work post spoke pool upgrade because the `V3RelayData.depositId` type will change to uint256

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -4,7 +4,6 @@ import assert from "assert";
 import Decimal from "decimal.js";
 import { ethers, PopulatedTransaction, providers, VoidSigner } from "ethers";
 import { getGasPriceEstimate } from "../gasPriceOracle";
-import { TypedMessage } from "../interfaces/TypedData";
 import { BigNumber, BigNumberish, BN, formatUnits, parseUnits, toBN } from "./BigNumberUtils";
 import { ConvertDecimals } from "./FormattingUtils";
 import { chainIsOPStack } from "./NetworkUtils";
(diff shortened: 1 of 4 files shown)


src/utils/common.ts
```diff
@@ -323,106 +322,6 @@ async function getLineaGasFees(chainId: number, transport: Transport | undefined
...
-      depositId,
-      originChainId,
-      updatedRelayerFeePct,
-      updatedRecipient,
-      updatedMessage,
-    },
-  };
-}
-
-export function getUpdateV3DepositTypedData(
```
bmzig, line 393: We use this [here](https://github.com/across-protocol/frontend/blob/7a57b128a9730b8b3009ecc6317523bc3c99988a/src/views/Transactions/hooks/useSpeedUp.tsx#L71) in the FE, so do we want to remove this now?

bmzig (approved):

src/utils/common.ts
nicholaspai: The FE has its own definition of this we use: https://github.com/across-protocol/frontend/blob/7a57b128a9730b8b3009ecc6317523bc3c99988a/src/utils/sdk.ts#L17

bmzig: Ah I see.

bmzig (approved):

PR #798, opened 2024-12-17
refactor(SpokePoolClient): Remove _getDepositIdAtBlock

nicholaspai:
This function only seems to be used once in a `SpokeUtils.getBlockRangeForDepositId()` function but it can just as easily be replaced with a call to `getDepositIdAtBlock`

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -35,9 +35,8 @@ import {
   SpeedUpWithBlock,
   TokensBridged,
 } from "../interfaces";
-import { SpokePool } from "../typechain";
 import { getNetworkName } from "../utils/NetworkUtils";
-import { getBlockRangeForDepositId, getDepositIdAtBlock, relayFillStatus } from "../utils/SpokeUtils";
+import { getBlockRangeForDepositId, relayFillStatus } from "../utils/SpokeUtils";
 import { BaseAbstractClient, isUpdateFailureReason, UpdateFailureReason } from "./BaseAbstractClient";
 import { HubPoolClient } from "./HubPoolClient";
 import { AcrossConfigStoreClient } from "./AcrossConfigStoreClient";
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -401,15 +401,6 @@ export class SpokePoolClient extends BaseAbstractClient {
   }
 
   /**
-   * Finds the deposit id at a specific block number.
-   * @param blockTag The block number to search for the deposit ID at.
-   * @returns The deposit ID.
-   */
-  public _getDepositIdAtBlock(blockTag: number): Promise<number> {
-    return getDepositIdAtBlock(this.spokePool as SpokePool, blockTag);
-  }
-
-  /**
    * @notice Return maximum of fill deadline buffer at start and end of block range. This is a contract
    * immutable state variable so we can't query other events to find its updates.
    * @dev V3 deposits have a fill deadline which can be set to a maximum of fillDeadlineBuffer + deposit.block.timestamp.
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -401,33 +401,6 @@ export class SpokePoolClient extends BaseAbstractClient {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


pxrl (approved):

pxrl:
Only one consideration from me - for the evm/svm split I'd anticipated that we'd force most RPC calls to go via the client layer. This protects the caller from having to do a lot of convoluted branching based on what the chain ID is. So ideally the SpokePoolClient would configure itself for evm/svm at startup, and then the caller doesn't need to think about it because the underlying evm/svm implementations are called correctly.

nicholaspai:
> Only one consideration from me - for the evm/svm split I'd anticipated that we'd force most RPC calls to go via the client layer. This protects the caller from having to do a lot of convoluted branching based on what the chain ID is. So ideally the SpokePoolClient would configure itself for evm/svm at startup, and then the caller doesn't need to think about it because the underlying evm/svm implementations are called correctly.

So are you saying that the svm/evm SpokePoolClient's getDepositIdAtBlock implementations would look different? If so, then you'd want to avoid having an `if (typeof SVMClient) {} else {}` branch in a `getDepositIdAtBlock` call?

nicholaspai:
Not needed if we merge in https://github.com/across-protocol/sdk/pull/799

pxrl:
> So are you saying that the svm/evm SpokePoolClient's getDepositIdAtBlock implementations would look different? If so, then you'd want to avoid having an `if (typeof SVMClient) {} else {}` branch in a `getDepositIdAtBlock` call?

Yeah - the existing `getDepositIdAtBlock()` already specifies a Contract object as part of the interface, so this seems inherently incompatible. Ideally we'd avoid having to make this a union type and select on the results - IMO the best result would be to have an evm and svm compatibility layer, and have the SpokePoolClient instantiate the correct compat layer based on the chain ID.

```
/**
 * Finds the deposit id at a specific block number.
 * @param blockTag The block number to search for the deposit ID at.
 * @returns The deposit ID.
 */
export async function getDepositIdAtBlock(contract: Contract, blockTag: number): Promise<number> {
  const depositIdAtBlock = await contract.numberOfDeposits({ blockTag });
  // Sanity check to ensure that the deposit ID is an integer and is greater than or equal to zero.
  if (!Number.isInteger(depositIdAtBlock) || depositIdAtBlock < 0) {
    throw new Error("Invalid deposit count");
  }
  return depositIdAtBlock;
}
```

nicholaspai:
> > So are you saying that the svm/evm SpokePoolClient's getDepositIdAtBlock implementations would look different? If so, then you'd want to avoid having an `if (typeof SVMClient) {} else {}` branch in a `getDepositIdAtBlock` call?
> 
> Yeah - the existing `getDepositIdAtBlock()` already specifies a Contract object as part of the interface, so this seems inherently incompatible. Ideally we'd avoid having to make this a union type and select on the results - IMO the best result would be to have an evm and svm compatibility layer, and have the SpokePoolClient instantiate the correct compat layer based on the chain ID.
> 
> ```
> /**
>  * Finds the deposit id at a specific block number.
>  * @param blockTag The block number to search for the deposit ID at.
>  * @returns The deposit ID.
>  */
> export async function getDepositIdAtBlock(contract: Contract, blockTag: number): Promise<number> {
>   const depositIdAtBlock = await contract.numberOfDeposits({ blockTag });
>   // Sanity check to ensure that the deposit ID is an integer and is greater than or equal to zero.
>   if (!Number.isInteger(depositIdAtBlock) || depositIdAtBlock < 0) {
>     throw new Error("Invalid deposit count");
>   }
>   return depositIdAtBlock;
> }
> ```

So doesn't this PR help out if we renamed `getDepositIdAtBlock` to `evm_getDepositIdAtBlock`? Then the evm_SpokePoolClient calls this one, and the solana one calls the other?

pxrl:
> So doesn't this PR help out if we renamed `getDepositIdAtBlock` to `evm_getDepositIdAtBlock`? Then the evm_SpokePoolClient calls this one, and the solana one calls the other?

I think it helps if the caller calls SpokePoolClient.getDepositIdAtBlock(), and the SpokePoolClient internally handles that translation :+1:. Ideally the top-level caller should not have to be conscious of whether they're checking an SVM or EVM chain.

PR #799, opened 2024-12-17
refactor(SpokePoolClient): Move findDeposit to helper function file

nicholaspai:
This function `findDeposit` is only used by `queryHistoricalDepositForFill`, a utility function. I propose moving this helper function closer to the helper function as it doesn't need to be a client level function. Though, it does make lots of spoke pool client calls but I see it just as intuitive to pass in the relevant spoke pool client into the function. The function effecitvely "finds a deposit" using a spoke pool client's memory.

### DIFF
--- a/src/utils/DepositUtils.ts
+++ b/src/utils/DepositUtils.ts
@@ -1,6 +1,6 @@
 import assert from "assert";
 import { SpokePoolClient } from "../clients";
-import { DEFAULT_CACHING_TTL, EMPTY_MESSAGE } from "../constants";
+import { DEFAULT_CACHING_TTL, EMPTY_MESSAGE, ZERO_ADDRESS } from "../constants";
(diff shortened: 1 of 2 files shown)


src/utils/DepositUtils.ts
```diff
@@ -125,6 +127,73 @@ export async function queryHistoricalDepositForFill(
   };
 }
 
+export async function findDeposit(
```
pxrl, line 130: Also RE https://github.com/across-protocol/sdk/pull/798#issuecomment-2549449045, I think this will make it more difficult to handle an SVM integration, because the caller (or this function) will need to consider whether the target chain is evm or svm.

So in general my preference would be to route calls via the SpokePoolClient, and have it implement pointers to the correct underlying interfaces depending on whether it's evm or svm.

nicholaspai: hm i'm ok with that

PR #801, opened 2024-12-19
feat(gasPriceOracle): Allow caller to pass in base fee multiplier

nicholaspai:
## Motivation:

We should have quoting API apply its GAS_MARKUP to the base fee rather than the estimated gas units.

## Features

- Apply a base fee multiplier to base fees rather than the gas cost estimate. This will provide more accurate gas price estimation to downstream users like the quoting API and the relayer
- Add Linea viem gas price oracle adapter
- Remove any gas price estimation that doesn't go through gasPriceOracle, such as the Linea logic in `estimateTotalGasRequiredByUnsignedTransaction`
- Add unit tests to CI
- Remove Arbitrum, Ethereum and Polygon viem gas price oracle adapters as the ethers adapter should work just as well
- Refactors gas price oracle getGasPriceEstimate() function and consolidates all optional parameters into a single GasPriceEstimateOptions object which will be easier to maintain

## Downstream PR's
- Relayer can submit transactions with more accurate gas prices: [PR](https://github.com/across-protocol/relayer/pull/1969) 
- Frontend API quotes gas costs using gas prices: [PR](https://github.com/across-protocol/frontend/pull/1339)


src/gasPriceOracle/adapters/polygon.ts
```diff
@@ -67,23 +67,28 @@ class PolygonGasStation extends BaseHTTPAdapter {
...
     // Per the GasStation docs, the minimum priority fee on Polygon is 30 Gwei.
     // https://docs.polygon.technology/tools/gas/polygon-gas-station/#interpretation
     const minPriorityFee = parseUnits("30", 9);
     if (maxPriorityFeePerGas.lt(minPriorityFee)) {
       const priorityDelta = minPriorityFee.sub(maxPriorityFeePerGas);
       maxPriorityFeePerGas = minPriorityFee;
-      maxFeePerGas = maxFeePerGas.add(priorityDelta);
+      maxFeePerGas = maxFeePerGas.mul(markup).add(priorityDelta);
     }
   }
```
pxrl, line 93: For Polygon I think it makes sense to defer the markup until after this try/catch block, and to skip a markup on the call to `eip1559()`. Then it can be applied once, and doesn't risk double-padding (as with lines 83 + 91).

src/gasPriceOracle/oracle.ts
```diff
@@ -17,22 +17,28 @@ import * as polygonViem from "./adapters/polygon-viem";
  * @param chainId The chain ID to query for gas prices.
  * @param provider A valid ethers provider.
  * @param legacyFallback In the case of an unrecognised chain, fall back to type 0 gas estimation.
+ * @parm baseFeeMarkup Multiplier applied to base fee for EIP1559 gas prices (or total fee for legacy).
  * @returns Am object of type GasPriceEstimate.
  */
 export async function getGasPriceEstimate(
   provider: providers.Provider,
   chainId?: number,
+  baseFeeMarkup = 1.0,
```
pxrl, line 26: Not sure about the name `baseFeeMarkup` since it's actually a multiplier, not a markup. Consider `baseFeeMultiplier` instead?

src/gasPriceOracle/adapters/arbitrum.ts
```diff
@@ -19,7 +23,7 @@ export async function eip1559(provider: providers.Provider, chainId: number): Pr
 
   // eip1559() sets maxFeePerGas = lastBaseFeePerGas + maxPriorityFeePerGas, so revert that.
   // The caller may apply scaling as they wish afterwards.
-  const maxFeePerGas = _maxFeePerGas.sub(maxPriorityFeePerGas).add(bnOne);
+  const maxFeePerGas = _maxFeePerGas.mul(markup).sub(maxPriorityFeePerGas).add(bnOne);
```
pxrl, line 26: Does this work for `0 < markup < 1` ? When I test it locally it causes an underflow locally because ethers BigNumber doesn't like numbers < 1.0.

```
> y = ethers.BigNumber.from("1000000000")
BigNumber { _hex: '0x3b9aca00', _isBigNumber: true }
> y.mul(0.9)
Uncaught:
Error: underflow [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT-underflow ] (fault="underflow", operation="BigNumber.from", value=0.9, code=NUMERIC_FAULT, version=bignumber/5.7.0)
```

src/gasPriceOracle/adapters/polygon.ts
nicholaspai: won't this multiple the base fee a bit more in the case where the `catch` is hit and `maxPriorityFeePerGas.lt(minPriorityFee)`, then the markup gets applied to the maxFeePerGas + priorityDelta?

It's not really a big deal but is this truly a base fee markup or do we need more logic?

nicholaspai: 4d76862ff39d4f8ad55a1a2c4d192a6a617f7976

src/gasPriceOracle/adapters/arbitrum.ts
nicholaspai: good point, floored it at 1.0 instead

src/gasPriceOracle/adapters/polygon.ts
pxrl: Hm; yeah you're right - I missed that. I think the change is OK and keeps the code simple, though if it was any chain other than Polygon it'd probably make sense to revert to your original implementation and set `eip1559(provider, chainId, 1.0)` instead.

nicholaspai: This might be more accurate actually: b4d4847989d5495442ffd6e92e2740032d01ec02

pxrl: Yep - looks good!

src/gasPriceOracle/oracle.ts
```diff
@@ -44,7 +51,8 @@ export async function getGasPriceEstimate(
 function getEthersGasPriceEstimate(
   provider: providers.Provider,
   chainId: number,
-  legacyFallback = true
+  legacyFallback = true,
+  baseFeeMultiplier = 1.0
```
pxrl, line 55:
```suggestion
  baseFeeMultiplier: number,
```

src/gasPriceOracle/oracle.ts
```diff
@@ -71,7 +79,8 @@ function getEthersGasPriceEstimate(
  */
 export async function getViemGasPriceEstimate(
   providerOrChainId: providers.Provider | number,
-  transport?: Transport
+  transport?: Transport,
+  baseFeeMultiplier = 1.0
```
pxrl, line 83:
```suggestion
  baseFeeMultiplier: number
```

src/utils/common.ts
```diff
@@ -265,9 +265,10 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
   const voidSigner = new VoidSigner(senderAddress, provider);
 
   // Estimate the Gas units required to submit this transaction.
+  const gasMultiplier = 1.0; // Don't apply any multiplier to the gas price.
   const queries = [
     gasUnits ? Promise.resolve(BigNumber.from(gasUnits)) : voidSigner.estimateGas(unsignedTx),
-    _gasPrice ? Promise.resolve({ maxFeePerGas: _gasPrice }) : getGasPriceEstimate(provider, chainId, transport),
+    _gasPrice ? Promise.resolve({ maxFeePerGas: _gasPrice }) : getGasPriceEstimate(provider, chainId, gasMultiplier, transport),
```
pxrl, line 271: @nicholaspai I updated this to fix a merge conflcit after merging my PR, but I'm curious about why we don't want to apply any buffer here. This is where the API would otherwise want to apply it, if I've understood correctly.

src/relayFeeCalculator/chain-queries/alephZero.ts
```diff
@@ -14,16 +14,16 @@ export class AlephZeroQueries extends QueryBase {
...
+    // gasMarkup = 0
   ) {
     assert(isDefined(spokePoolAddress));
     super(
       provider,
       symbolMapping,
       spokePoolAddress,
       simulatedRelayerAddress,
-      gasMarkup,
+      // gasMarkup,
```
pxrl, line 26: fwiw I think we need to retain these, otherwise the API will just inherit the default 1.0 multiplier.

src/utils/common.ts
nicholaspai: Oh you're right this is a mistake. I only applied it to the fresh query to the oracle but not to the input value if defined

nicholaspai: Is this what you were thinking? 4a0858a4b39738b494db70ae378e91275cc1800e

src/gasPriceOracle/adapters/ethereum.ts
```diff
@@ -29,7 +37,7 @@ export async function eip1559Raw(provider: providers.Provider, chainId: number):
   assert(BigNumber.isBigNumber(baseFeePerGas), `No baseFeePerGas received on ${getNetworkName(chainId)}`);
 
   return {
-    maxFeePerGas: maxPriorityFeePerGas.add(baseFeePerGas),
+    maxFeePerGas: maxPriorityFeePerGas.add(baseFeePerGas).mul(baseFeeMultiplier),
```
nicholaspai, line 40: @pxrl i fixed this in your PR where you were applying the `.mul(baseFeeMultiplier)` to the `maxPriorityFeePerGas` rather than the  `baseFeePerGas` which I believe was a mistake. LMK

src/relayFeeCalculator/chain-queries/alephZero.ts
nicholaspai: i'm confused by your comment here, what do you mean? It seems that we are removing the gasMarkup param from the relayFeeCalculator's constructor so why would we need to retain it?

nicholaspai: Also I think this might be relevant that we'll merge in [this pr](https://github.com/across-protocol/frontend/pull/1339) alongside an import of this sdk release into the FE

src/utils/common.ts
pxrl: @nicholaspai I'm not sure tbh. Do we want to apply the multiplier if the caller specifies the gasPrice? There is a risk that we revert to the recursive gas price multiplication that we had a while ago (i.e. the API sources a gas price including multiplier, then supplies it back to the gasPriceOracle and it gets multiplied _again_).

But I was also looking at line 268, where we have `const gasMultiplier = 1.0`. I think we want that to be user-supplied, rather than hardcoded in this file, right?

src/relayFeeCalculator/chain-queries/alephZero.ts
pxrl: Ah, I understand now. So the gas multiplier would effectively only apply in the API?

src/gasPriceOracle/adapters/ethereum.ts
pxrl: Agree - good catch, thanks!

src/relayFeeCalculator/chain-queries/alephZero.ts
nicholaspai: That's how we currently use the multiplier so yeah

src/utils/common.ts
nicholaspai: You're right this is a mistake. We should allow caller to pass in gas price multiplier and NOT apply it if the gas price is passed in.

src/utils/common.ts
```diff
@@ -256,18 +256,21 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
   options: Partial<{
     gasPrice: BigNumberish;
     gasUnits: BigNumberish;
+    baseFeeMultiplier: number;
```
nicholaspai, line 259: @pxrl I added in an optional multiplier arg here. I don't think we want to use this anywhere in the SDK since all calls to this function only care about the `gasCost`. In fact, we might want to change this function soon to only return the `gasCost` and simplify the interface

src/utils/common.ts
```diff
@@ -256,20 +252,24 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
...
 
   const { chainId } = await provider.getNetwork();
   const voidSigner = new VoidSigner(senderAddress, provider);
 
   // Estimate the Gas units required to submit this transaction.
   const queries = [
     gasUnits ? Promise.resolve(BigNumber.from(gasUnits)) : voidSigner.estimateGas(unsignedTx),
-    _gasPrice ? Promise.resolve({ maxFeePerGas: _gasPrice }) : getGasPriceEstimate(provider, chainId, transport),
+    _gasPrice
+      ? Promise.resolve({ maxFeePerGas: _gasPrice })
```
nicholaspai, line 269: Note we don't multiply the gas price if its passed in

src/gasPriceOracle/adapters/linea-viem.ts
```diff
@@ -1,14 +1,36 @@
...
+import { PopulatedTransaction } from "ethers";
 
-export async function eip1559(provider: PublicClient, _chainId?: number): Promise<InternalGasPriceEstimate> {
-  const { baseFeePerGas, priorityFeePerGas } = await estimateGas(provider, {
-    account,
-    to: account,
-    value: BigInt(1),
+/**
+ * @notice The Linea viem provider calls the linea_estimateGas RPC endpoint to estimate gas. Linea is unique
+ * in that the recommended fee per gas is hardcoded to 7 gwei while the priority fee is dynamic based on the
```
pxrl, line 9:
```suggestion
 * in that the recommended fee per gas is hardcoded to 7 wei while the priority fee is dynamic based on the
```

src/gasPriceOracle/oracle.ts
```diff
@@ -1,51 +1,93 @@
...
+  };
+  assert(
+    baseFeeMultiplier >= 1.0 && baseFeeMultiplier <= 5,
+    `Require 1.0 < base fee multiplier (${baseFeeMultiplier}) <= 5.0 for a total gas multiplier within [+1.0, +5.0]`
+  );
+
+  const chainId = _chainId ?? (await provider.getNetwork()).chainId;
 
+  // We only use the unsignedTx in the viem flow.
   const useViem = process.env[`NEW_GAS_PRICE_ORACLE_${chainId}`] === "true";
```
pxrl, line 68: We only have the linea-viem.ts adapter remaining so we'd better constrain this.

```suggestion
  const useViem = chainId === CHAIN_IDs.LINEA && process.env[`NEW_GAS_PRICE_ORACLE_${chainId}`] === "true";
```

src/utils/common.ts
```diff
@@ -256,20 +252,24 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
   options: Partial<{
     gasPrice: BigNumberish;
     gasUnits: BigNumberish;
+    baseFeeMultiplier: number;
     transport: Transport;
   }> = {}
 ): Promise<TransactionCostEstimate> {
-  const { gasPrice: _gasPrice, gasUnits, transport } = options || {};
+  const { gasPrice: _gasPrice, gasUnits, baseFeeMultiplier: _baseFeeMultiplier, transport } = options || {};
+  const baseFeeMultiplier = _baseFeeMultiplier ?? 1.0; // A default 1.0x multiplier means no multiplier by default.
```
pxrl, line 260: We can get away with one line here :+1: 

```suggestion
  const { gasPrice: _gasPrice, gasUnits, baseFeeMultiplier = 1.0, transport } = options || {};
```

pxrl (approved):
I am OK w/ this, subject to a couple of minor comments.

@dohaki Would be good if you could also take a look, since this will require changes in the API.

src/gasPriceOracle/oracle.ts
```diff
@@ -1,51 +1,93 @@
...
   provider: providers.Provider,
-  chainId: number,
-  legacyFallback = true
+  opts: EthersGasPriceEstimateOptions
 ): Promise<GasPriceEstimate> {
+  const { baseFeeMultiplier, chainId, legacyFallback } = opts;
+
   const gasPriceFeeds = {
     [CHAIN_IDs.ALEPH_ZERO]: arbitrum.eip1559,
     [CHAIN_IDs.ARBITRUM]: arbitrum.eip1559,
```
nicholaspai, line 93: @pxrl should Linea still be legacy in the ethers provider case? Currently the linea.eip1559 implementation is just a passthrough to the ethereum.legacy call

src/gasPriceOracle/adapters/linea.ts
```diff
@@ -6,6 +6,10 @@ import { providers } from "ethers";
...
 import * as ethereum from "./ethereum";
 
-export function eip1559(provider: providers.Provider, chainId: number): Promise<GasPriceEstimate> {
-  return ethereum.legacy(provider, chainId);
+export function eip1559(
+  provider: providers.Provider,
+  chainId: number,
+  baseFeeMultiplier: number
+): Promise<GasPriceEstimate> {
+  return ethereum.legacy(provider, chainId, baseFeeMultiplier);
```
nicholaspai, line 14: @pxrl should the Linea ethers provider still passthrough to the legacy call?

nicholaspai:
> I am OK w/ this, subject to a couple of minor comments.
> 
> @dohaki Would be good if you could also take a look, since this will require changes in the API.

You should re-review. I refactored the code a bit:
- Passed in `GasPriceEstimateOptions` as a param to all individual gas price feeds. This makes it easier to unit test certain gas price feeds like the Polygon GasStation, where for example we can pass in a mocked GasStation object.
- Add unit tests for all feeds.

The only outstanding question I have on implementation is how we should treat the Linea Ethers flow (i.e. when Viem provider is not toggled on via Env var).

Question for @dohaki and @pxrl separately is whether the new interface works for our Relayer and Quoting API use case

src/gasPriceOracle/adapters/ethereum.ts
```diff
@@ -3,63 +3,87 @@ import { providers } from "ethers";
...
+ * getFeeData() method which adds eth_getBlock("latest").baseFee to a hardcoded priority fee of 1.5 gwei.
  * @param provider ethers RPC provider instance.
- * @param chainId Chain ID of provider instance.
  * @returns Promise of gas price estimate object.
  */
-export function eip1559(provider: providers.Provider, chainId: number): Promise<GasPriceEstimate> {
-  const useRaw = process.env[`GAS_PRICE_EIP1559_RAW_${chainId}`] === "true";
-  return useRaw ? eip1559Raw(provider, chainId) : eip1559Bad(provider, chainId);
+export function eip1559(provider: providers.Provider, opts: GasPriceEstimateOptions): Promise<GasPriceEstimate> {
+  const useRaw = process.env[`GAS_PRICE_EIP1559_RAW_${opts.chainId}`] === "true";
```
nicholaspai, line 16: Should we not make `useRaw` the default for all chains...?

src/utils/common.ts
dohaki: The Quotes API currently caches the gas price via `getGasCosts -> nativeGasCost`. So with this change is it expected that we should apply the base fee multiplier before we cache it? Or at the time we calculate the fee, i.e. when we call `RelayFeeCalculator.relayerFeeDetails`?

dohaki (commented):
This looks really good to me! I only left one question. And I think these changes should be easy to integrate into the Quotes API

src/utils/common.ts
nicholaspai: You can do either. I think its probably cleaner if you cache the multiplied gas price that way all the API methods can use the same gas price that takes into account the GAS_MARKUP.

But it depends if there are other use cases that read from this cache that do NOT want the base fee multiplied, in that case then you'd only want to cache the base fee without multiplication

src/gasPriceOracle/adapters/ethereum.ts
nicholaspai: Talked with Paul offline about this and the thinking is, we want to make this "raw" `eip1559()` function the default and only option eventually, but we also want to slowly roll it out and test it in production. So gate-keeping it with an environment variable is the easiest way to slowly roll it out. 35a0420bd4c5f1b1782d6cdbf29fa12bcf5820d0

pxrl:
@gsteenkamp89 Since Dong-Ha is out this week, are you interested to take a look at updating the api to handle this change?

gsteenkamp89 (approved):

gsteenkamp89:
> @gsteenkamp89 Since Dong-Ha is out this week, are you interested to take a look at updating the api to handle this change?

Yeah I can tackle this 👍

nicholaspai:
> > @gsteenkamp89 Since Dong-Ha is out this week, are you interested to take a look at updating the api to handle this change?
> 
> Yeah I can tackle this 👍

G this might be relevant: https://github.com/across-protocol/frontend/pull/1339

src/gasPriceOracle/oracle.ts
```diff
@@ -1,51 +1,79 @@
...
+  // legacyFallback In the case of an unrecognized chain, fall back to type 0 gas estimation.
+  legacyFallback: boolean;
+  // chainId The chain ID to query for gas prices. If omitted can be inferred by provider.
+  chainId: number;
+  // unsignedTx The unsigned transaction used for simulation by Linea's Viem provider to produce the priority gas fee.
+  unsignedTx?: PopulatedTransaction;
+  // transport Viem Transport object to use for querying gas fees used for testing.
+  transport?: Transport;
+  // polygonGasStation Custom Polygon GasStation class used for testing.
+  polygonGasStation?: PolygonGasStation;
```
pxrl, line 27: I'm not super keen on exposing this type in the general options type. It's intended to be hidden behind the Polygon adapter and is really an internal type. Can we maintain that while still permitting this to be testable?

src/gasPriceOracle/oracle.ts
```diff
@@ -60,39 +88,42 @@ function getEthersGasPriceEstimate(
...
     typeof providerOrChainId === "number" ? providerOrChainId : (await providerOrChainId.getNetwork()).chainId;
   const viemProvider = getPublicClient(chainId, transport);
 
-  const gasPriceFeeds = {
-    [CHAIN_IDs.ALEPH_ZERO]: arbitrumViem.eip1559,
-    [CHAIN_IDs.ARBITRUM]: arbitrumViem.eip1559,
+  const gasPriceFeeds: Record<
+    number,
+    (provider: PublicClient, opts: GasPriceEstimateOptions) => Promise<InternalGasPriceEstimate>
+  > = {
```
pxrl, line 114: ooc, do we need to be explicit on this typing? It should have previously been inferred OK.

nicholaspai: you mean removing the `as const`? We can do that yes 20461a6ae6dea975870334b34dd66e1af99d8abb

src/gasPriceOracle/oracle.ts
```diff
@@ -1,51 +1,79 @@
...
-  chainId ?? ({ chainId } = await provider.getNetwork());
+  const baseFeeMultiplier = opts.baseFeeMultiplier ?? GAS_PRICE_ESTIMATE_DEFAULTS.baseFeeMultiplier;
+  assert(
+    baseFeeMultiplier >= 1.0 && baseFeeMultiplier <= 5,
+    `Require 1.0 < base fee multiplier (${baseFeeMultiplier}) <= 5.0 for a total gas multiplier within [+1.0, +5.0]`
+  );
+  const chainId = opts.chainId ?? (await provider.getNetwork()).chainId;
+  const optsWithDefaults: GasPriceEstimateOptions = {
+    ...GAS_PRICE_ESTIMATE_DEFAULTS,
+    ...opts,
```
pxrl, line 53: What's the rule in case of a conflict between `GAS_PRICE_ESTIMATE_DEFAULTS` and `opts` ? IS it the ordering that decides?

src/gasPriceOracle/oracle.ts
pxrl: Ah, actually I meant the type definitions that were added here. Were they necessary due to some tsc issue? It seems like they should be redundant as long as the members of the object are consistent.

```
 Record<
    number,
    (provider: PublicClient, opts: GasPriceEstimateOptions) => Promise<InternalGasPriceEstimate>
  > = {
```

src/gasPriceOracle/oracle.ts
nicholaspai: I agree with that, best I could think of is using an env var to toggle it: https://github.com/across-protocol/sdk/pull/801/commits/576dac68ab79f46b6e12d7ac1e1404b79390a522

src/gasPriceOracle/oracle.ts
nicholaspai: Yeah that's how I would understand the syntax. So fill in defaults first, but allow user to override them. The test cases where I set the `baseFeeMultiplier = 2` seem to confirm this behavior

src/gasPriceOracle/oracle.ts
nicholaspai: oh true a232f056d7261f428c75208c167e1889cc66dbb1

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -75,37 +75,90 @@ export class QueryBase implements QueryInterface {
...
-    const gasTotalMultiplier = toBNWei(1.0 + gasMarkup);
+    const { gasPrice = this.fixedGasPrice, gasUnits, baseFeeMultiplier, transport } = options;
 
     const tx = await populateV3Relay(this.spokePool, deposit, relayer);
     const {
       nativeGasCost,
       tokenGasCost,
       gasPrice: impliedGasPrice,
-    } = await estimateTotalGasRequiredByUnsignedTransaction(tx, relayer, this.provider, {
+    } = await this.estimateTotalGasRequiredByUnsignedTransaction(tx, relayer, this.provider, {
```
pxrl, line 89: Given that we're relocating this I'll optimistically propose a rename:

```suggestion
    } = await this.estimateGas(tx, relayer, this.provider, {
```

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -75,37 +75,90 @@ export class QueryBase implements QueryInterface {
...
+   * @param unsignedTx The unsigned transaction that this function will estimate.
+   * @param senderAddress The address that the transaction will be submitted from.
+   * @param provider A valid ethers provider - will be used to reason the gas price.
+   * @param options
+   * @param options.gasPrice A manually provided gas price - if set, this function will not resolve the current gas price.
+   * @param options.gasUnits A manually provided gas units - if set, this function will not estimate the gas units.
+   * @param options.transport A custom transport object for custom gas price retrieval.
+   * @returns Estimated cost in units of gas and the underlying gas token (gasPrice * estimatedGasUnits).
+   */
+  async estimateTotalGasRequiredByUnsignedTransaction(
```
pxrl, line 114:
```suggestion
  async estimateGas(
```

test/relayFeeCalculator.test.ts
```diff
@@ -454,3 +457,63 @@ describe("RelayFeeCalculator: Composable Bridging", function () {
     });
   });
 });
+
+describe("QueryBase", function () {
+  describe("estimateTotalGasRequiredByUnsignedTransaction", function () {
```
pxrl, line 462:
```suggestion
  describe("estimateGas", function () {
```

test/relayFeeCalculator.test.ts
```diff
@@ -454,3 +457,63 @@ describe("RelayFeeCalculator: Composable Bridging", function () {
...
+        undefined,
+        this.logger
+      );
+    });
+    it("Uses passed in options", async function () {
+      const options = {
+        gasUnits: BigNumber.from(300_000),
+        gasPrice: toGWei("1.5"),
+      };
+      const result = await queryBase.estimateTotalGasRequiredByUnsignedTransaction(
```
pxrl, line 480:
```suggestion
      const result = await queryBase.estimateGas(
```

test/relayFeeCalculator.test.ts
```diff
@@ -454,3 +457,63 @@ describe("RelayFeeCalculator: Composable Bridging", function () {
...
+        gasPrice: undefined,
+        baseFeeMultiplier: toBNWei("2"),
+      };
+      // Mocked provider gets queried to compute gas price.
+      const stdLastBaseFeePerGas = toGWei("12");
+      const stdMaxPriorityFeePerGas = toGWei("1");
+      const chainId = 1; // get gas price from GasPriceOracle.ethereum.eip1559()
+      const mockedProvider = new MockedProvider(stdLastBaseFeePerGas, stdMaxPriorityFeePerGas, chainId);
+
+      const result = await queryBase.estimateTotalGasRequiredByUnsignedTransaction(
```
pxrl, line 502:
```suggestion
      const result = await queryBase.estimateGas(
```

pxrl (approved):

src/gasPriceOracle/oracle.ts
pxrl: I think I have a solution but it's low priority - let's just get this in as-is to unblock the API.

src/gasPriceOracle/oracle.ts
pxrl: This seems to test OK in the REPL at least:

```
> y = { 1: 1, 2: 2, 3: 3 }
{ '1': 1, '2': 2, '3': 3 }

> a = { 1: "a" }
{ '1': 'a' }

> b = { 1: "b" }
{ '1': 'b' }

> { ...y, ...a, ...b }
{ '1': 'b', '2': 2, '3': 3 }

> { ...y, ...b, ...a }
{ '1': 'a', '2': 2, '3': 3 }
 
> { ...y, ...b, ...a, 1: "x" }
{ '1': 'x', '2': 2, '3': 3 }
```

PR #804, opened 2024-12-20
feat(ink): add deployment changes to support Ink

james-a-morris:


src/utils/NetworkUtils.ts
```diff
@@ -87,6 +87,15 @@ export function chainIsAlephZero(chainId: number): boolean {
   return [CHAIN_IDs.ALEPH_ZERO].includes(chainId);
 }
 
+/**
+ * Determines whether a chain ID is an Ink implementation
+ * @param chainId Chain ID to evaluate
+ * @returns True if chainId is an Ink chain, otherwise false
+ */
+export function chainIsInk(chainId: number): boolean {
+  return [CHAIN_IDs.INK, CHAIN_IDs.INK_SEPOLIA].includes(chainId);
```
nicholaspai, line 96: Is there a way we can consolidate all these chainIsX calls by adding a dictionary of paired networks?

nicholaspai (approved):

src/utils/NetworkUtils.ts
james-a-morris: There was work to do this originally but we scrapped it due to the composability of these individual fuctions. I agree a table where we associate similar chains (mainnet + testnet) with a single function, way more optimized like:

```
const pairings = [ [1, 5, 11155111, ...], ...]

const chainIs = (chainA: number, chainB: number) => pairings.find(() => pairings.includes(chainA) && pairings.includes(chainB)) !== undefined;

```

pxrl: Why do we need `chainIsInk()` at all? Is there something special about Ink that means we can't just group it as an OP stack chain?

james-a-morris: This is purely a requirement being ticked from the NCA guide.  I'm happy to remove it

PR #805, opened 2024-12-21
refactor(utils): Redirect all gas price queries to gasPriceOracle

pxrl:
Rearrange this utility function to force gas price queries via the gasPriceOracle. It has been verified that it's OK to bypass the OP sdk in this case, since it only queries provider.gasPrice() anyway.

nb. This is an interim refactor and is not the end result. Ultimately the gasPriceOracle will need to accept a transaction object and will be able to give an estimate back. This is necessary for Linea, but can also be (ab)used to simplify the OP stack workflow for estimating gas consumption.


src/utils/common.ts
```diff
@@ -265,8 +265,12 @@ export async function estimateTotalGasRequiredByUnsignedTransaction(
   const voidSigner = new VoidSigner(senderAddress, provider);
 
   // Estimate the Gas units required to submit this transaction.
-  const nativeGasCost = gasUnits ? BigNumber.from(gasUnits) : await voidSigner.estimateGas(unsignedTx);
-  assert(nativeGasCost.gt(0), "Gas cost should not be 0");
+  const queries = [
+    gasUnits ? Promise.resolve(BigNumber.from(gasUnits)) : voidSigner.estimateGas(unsignedTx),
+    _gasPrice ? Promise.resolve({ maxFeePerGas: _gasPrice }) : getGasPriceEstimate(provider, chainId, transport)
+  ] as const;
+  let [nativeGasCost, { maxFeePerGas: gasPrice }] = await Promise.all(queries);
```
pxrl, line 272: This feels a bit like a bodge, but it seems to work OK. I'm open to another formulation if we can find something simpler, but I also anticipate that this will be removed entirely before too long (pending the gasPriceOracle being able to estimate gas internally - coming real soon now).

nicholaspai (approved):

PR #807, opened 2024-12-24
improve(GasPriceOracle): Standardize gas price oracle usage

nicholaspai:
- Removes special case in `src/utils/common.ts#estimateTotalGasRequiredByUnsignedTransaction` that handles Linea gas price estimates by avoiding the gasPriceOracle call
- Implements `eip1559()` function in `linea-viem.ts` gas price oracle adapter file so that gas price oracle users now get access to Linea EIP1559 gas price estimates via viem. This requires callers to pass in an optional `unsignedTx` object
- Refactors gas price oracle `getGasPriceEstimate()` function and consolidates all optional parameters into a single `GasPriceEstimateOptions` object which will be easier to maintain


src/gasPriceOracle/adapters/arbitrum-viem.ts
pxrl, line 1: I think we could also just delete this file for now; I doubt we'll be using it in the immediate future.

src/gasPriceOracle/adapters/ethereum-viem.ts
```diff
@@ -1,8 +1,16 @@
...
-  return provider.estimateFeesPerGas();
+export async function eip1559(
+  provider: PublicClient,
+  _chainId: number,
+  baseFeeMultiplier: number
+): Promise<InternalGasPriceEstimate> {
+  const { maxFeePerGas: _maxFeePerGas, maxPriorityFeePerGas } = await provider.estimateFeesPerGas();
+  const maxFeePerGasScaled = BigNumber.from(_maxFeePerGas.toString()).mul(baseFeeMultiplier);
+  const maxFeePerGas = BigInt(maxFeePerGasScaled.toString()) + maxPriorityFeePerGas;
+  return { maxFeePerGas, maxPriorityFeePerGas };
```
pxrl, line 13: These should be BigInt types, not BigNumber. `getViemGasPriceEstimate()` handles BigInt -> BigNumber conversions centrally. That will make it easier for us to eventually transition BigNumber -> BigInt.

One additional thought: since these adapters return both `maxFeePerGas` and `maxPriorityFeePerGas`, we can actually just apply the base fee padding centrally in `getViemGasPriceEstimate()` anyway. That should mean we don't need to touch any of these adapters.

Example:

```
({ maxFeePerGas, maxPriorityFeePerGas } = await gasPriceFeeds[chainId](...));
maxFeePerGas = (maxFeePerGas - maxPriorityFeePerGas) * baseMultiplier + maxPriorityFeePerGas;
```

src/gasPriceOracle/oracle.ts
```diff
@@ -96,18 +142,22 @@ export async function getViemGasPriceEstimate(
...
+      unsignedTx
+    ));
   } else {
     let gasPrice: bigint | undefined;
     ({ maxFeePerGas, maxPriorityFeePerGas, gasPrice } = await viemProvider.estimateFeesPerGas());
 
-    maxFeePerGas ??= gasPrice!;
+    maxFeePerGas ??= gasPrice! * BigInt(baseFeeMultiplier);
     maxPriorityFeePerGas ??= BigInt(0);
   }
```
pxrl, line 158: This is fwiw where I think we could unconditionally apply the baseFeeMultiplier (as mentioned above):

`maxFeePerGas = (maxFeePerGas - maxPriorityFeePerGas) * baseMultiplier + maxPriorityFeePerGas;`

I think applying it here should simplify and declutter all of the adapters, since they won't need to be concerned with the baseFeeMultiplier at all.

src/gasPriceOracle/adapters/arbitrum-viem.ts
nicholaspai: You think we'll just use the ethers version for arbitrum, you mean?

src/gasPriceOracle/adapters/ethereum-viem.ts
nicholaspai: I don't love applying the multiplier in the getViemPriceEstimate function because it means backing out the priority fee. And some chains might not always add the priority fee to the base fee. This also would diverge from how the ethers adapters work

src/gasPriceOracle/oracle.ts
nicholaspai: But doesn't that mean that we're assuming all chains add priority fee to base fee to get full base fee? Is that always right?

pxrl: Correct; and that's not valid on Linea :headstone:. That's pretty annoying...

src/gasPriceOracle/adapters/ethereum-viem.ts
pxrl: Yeah, per subsequent comment it would ideally have been done in`getGasEstimate()`. Problem is that it won't work properly for Linea :\

nicholaspai: Hm why wouldn't it work for linea? In the linea adapter we'd return the priority fee and the base fee and then in the top level function getGasPriceEstimate we could back out the priority fee?

src/gasPriceOracle/adapters/arbitrum-viem.ts
pxrl: Yeah; I think this new general eip1559 approach using ethers should probably be effective on mainnet, OP and Orbit chains (via the Arbitrum custom logic that sets the priority fee to 1). So we'll be left with zkSync, Linea and Scroll as the outliers. We definitely have a use for viem for Linea; Scroll and zkSync is tbd.

src/gasPriceOracle/oracle.ts
nicholaspai: Sorry I don't understand why is that not valid on linea?

src/gasPriceOracle/adapters/linea-viem.ts
```diff
@@ -1,17 +1,23 @@
...
-    to: account,
-    value: BigInt(1),
+    account: (_unsignedTx?.from as Address) ?? account,
+    to: (_unsignedTx?.to as Address) ?? account,
+    value: BigInt(_unsignedTx?.value?.toString() || "1"),
   });
 
   return {
-    maxFeePerGas: baseFeePerGas + priorityFeePerGas,
+    maxFeePerGas: baseFeePerGas * BigInt(baseFeeMultiplier) + priorityFeePerGas,
```
nicholaspai, line 20: So @pxrl are you suggesting we should apply the multiplier on the priority fee here for linea?

src/gasPriceOracle/adapters/linea-viem.ts
```diff
@@ -1,17 +1,23 @@
...
+  provider: PublicClient,
+  _chainId: number,
+  baseFeeMultiplier: number,
+  _unsignedTx?: PopulatedTransaction
+): Promise<InternalGasPriceEstimate> {
   const { baseFeePerGas, priorityFeePerGas } = await estimateGas(provider, {
-    account,
-    to: account,
-    value: BigInt(1),
+    account: (_unsignedTx?.from as Address) ?? account,
```
nicholaspai, line 14: Also @pxrl dont you find it weird we don't need to pass the transactions calldata here? If so then why does the unsigned tx even matter for priority fee pricing?

src/gasPriceOracle/adapters/arbitrum-viem.ts
nicholaspai: what about polygon?

PR #808, opened 2024-12-27
refactor(QueryBase): Remove constructors in derived classes 

nicholaspai:
The constructor calls in AlephZeroQueries and PolygonQueries are not useful since the `getTokenPrice()` class function is the main changed logic from the parent `QueryBase` class. We can reduce LOC in the repo by removing these constructors. Additionally, we should be able to just use the parent `QueryBase`'s `constructor` directly.

This was partially motivated by work on https://github.com/across-protocol/sdk/pull/801 where I had to modify these constructors because the `QueryBase` constructor changed

### DIFF
--- a/src/relayFeeCalculator/chain-queries/alephZero.ts
+++ b/src/relayFeeCalculator/chain-queries/alephZero.ts
@@ -1,36 +1,10 @@
-import assert from "assert";
-import { getDeployedAddress } from "../../utils/DeploymentUtils";
-import { DEFAULT_LOGGER, Logger } from "../relayFeeCalculator";
-import { providers } from "ethers";
-import { CHAIN_IDs, DEFAULT_SIMULATED_RELAYER_ADDRESS, TOKEN_SYMBOLS_MAP } from "../../constants";
+import { CHAIN_IDs } from "../../constants";
 import { Coingecko } from "../../coingecko/Coingecko";
-import { isDefined } from "../../utils";
 import { QueryBase } from "./baseQuery";
 
 // @dev This class only exists because querying the native token price for this network is not the standard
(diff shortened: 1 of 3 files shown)


james-a-morris (approved):
Fair call since we're routing all instantiations through the factory class.

src/relayFeeCalculator/chain-queries/alephZero.ts
```diff
@@ -1,36 +1,10 @@
...
-    assert(isDefined(spokePoolAddress));
-    super(
-      provider,
-      symbolMapping,
-      spokePoolAddress,
-      simulatedRelayerAddress,
-      logger,
-      coingeckoProApiKey,
-      undefined,
-      "usd"
```
nicholaspai, line 30: This `coingeckoBaseCurrency` class variable is only used in the `QueryBase`'s `getTokenPrice()` function which gets overridden by this class anyways, so its not useful to pass through. The only argument I see to keeping this code here is to prevent a future developer from accidentally using `this.coingeckoBaseCurrency` within this class and using an unexpected currency

src/relayFeeCalculator/chain-queries/factory.ts
```diff
@@ -37,8 +37,10 @@ export class QueryBase__factory {
         symbolMapping,
         spokePoolAddress,
         simulatedRelayerAddress,
+        logger,
         coingeckoProApiKey,
-        logger
+        fixedGasPrice[chainId],
+        "usd"
```
nicholaspai, line 43: I'm pretty sure we can just pass through and use the parent class's constructor here and therefore delete the derived construcotr in PolygonQueries and AlephZeroQueries

pxrl (approved):

PR #817, opened 2025-01-06
feat(GasPriceOracle): Allow caller to set a priority fee multiplier

nicholaspai:
This acts similarly to the baseFeeMultiplier. It is unused for `legacy()` pricing which returns a single gas price so we use the `baseFeeMultiplier` to scale the gas price. It's also unused for the Linea and Arbitrum adapters which only have one component of the gas price that is variable (the priority fee and base fee, respectively) and we use the `baseFeeMultiplier` in both cases.

### DIFF
--- a/src/gasPriceOracle/adapters/arbitrum.ts
+++ b/src/gasPriceOracle/adapters/arbitrum.ts
@@ -10,7 +10,7 @@ import { GasPriceEstimateOptions } from "../oracle";
  * Reference: https://docs.arbitrum.io/how-arbitrum-works/gas-fees so we hardcode the priority fee
  * to 1 wei.
  * @param provider Ethers Provider
- * @param {GasPriceEstimateOptions} opts See notes below on specific parameters.
+ * @param opts See notes below on specific parameters.
  * @param baseFeeMultiplier Amount to multiply base fee.
  * @param priorityFeeMultiplier Unused in this function because arbitrum priority fee is hardcoded to 1 wei by this
  * function.
--- a/src/gasPriceOracle/adapters/arbitrum.ts
+++ b/src/gasPriceOracle/adapters/arbitrum.ts
@@ -10,6 +10,10 @@ import { GasPriceEstimateOptions } from "../oracle";
  * Reference: https://docs.arbitrum.io/how-arbitrum-works/gas-fees so we hardcode the priority fee
  * to 1 wei.
  * @param provider Ethers Provider
+ * @param {GasPriceEstimateOptions} opts See notes below on specific parameters.
+ * @param baseFeeMultiplier Amount to multiply base fee.
+ * @param priorityFeeMultiplier Unused in this function because arbitrum priority fee is hardcoded to 1 wei by this
+ * function.
  * @returns GasPriceEstimate
  */
 export async function eip1559(provider: providers.Provider, opts: GasPriceEstimateOptions): Promise<GasPriceEstimate> {
--- a/src/gasPriceOracle/adapters/ethereum.ts
+++ b/src/gasPriceOracle/adapters/ethereum.ts
@@ -14,13 +14,16 @@ import { GasPriceEstimateOptions } from "../oracle";
  * eth_getBlock("pending").baseFee to eth_maxPriorityFeePerGas, otherwise calls the ethers provider's
  * getFeeData() method which adds eth_getBlock("latest").baseFee to a hardcoded priority fee of 1.5 gwei.
  * @param provider ethers RPC provider instance.
+ * @param {GasPriceEstimateOptions} opts See notes below on specific parameters.
+ * @param baseFeeMultiplier Amount to multiply base fee or total fee for legacy gas pricing.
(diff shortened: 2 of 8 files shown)


src/gasPriceOracle/adapters/linea-viem.ts
```diff
@@ -15,10 +15,12 @@ import { fixedPointAdjustment } from "../../utils";
...
  * @param baseFeeMultiplier Amount to multiply priority fee, since Linea's base fee is hardcoded while its priority
- * fee is dynamic
- * @param _unsignedTx Should contain any params passed to linea_estimateGas, which are listed
+ * fee is dynamic.
+ * @param priorityFeeMultiplier Unused in this function because the baseFeeMultiplier is applied to the dynamic
+ * Linea priority fee while the base fee is hardcoded.
+ * @param unsignedTx Should contain any params passed to linea_estimateGas, which are listed
  * here: https://docs.linea.build/api/reference/linea-estimategas#parameters
  * @returns
  */
```
bmzig, line 26: Do we intentionally not use `opts.priorityFeeMultiplier` for this?

bmzig (approved):
No comments from me.

src/gasPriceOracle/adapters/linea-viem.ts
bmzig: I should have read the top level description first. Looks good.

src/gasPriceOracle/adapters/arbitrum.ts
```diff
@@ -10,6 +10,10 @@ import { GasPriceEstimateOptions } from "../oracle";
  * Reference: https://docs.arbitrum.io/how-arbitrum-works/gas-fees so we hardcode the priority fee
  * to 1 wei.
  * @param provider Ethers Provider
+ * @param {GasPriceEstimateOptions} opts See notes below on specific parameters.
```
james-a-morris, line 13: We shouldn't need to type params in TSDoc formatting
```suggestion
 * @param opts See notes below on specific parameters.
```

src/gasPriceOracle/adapters/ethereum.ts
```diff
@@ -42,10 +46,11 @@ export async function eip1559Raw(
   const maxPriorityFeePerGas = BigNumber.from(_maxPriorityFeePerGas);
   assert(BigNumber.isBigNumber(baseFeePerGas), `No baseFeePerGas received on ${getNetworkName(chainId)}`);
 
+  const scaledPriorityFee = maxPriorityFeePerGas.mul(priorityFeeMultiplier).div(fixedPointAdjustment);
   const scaledBaseFee = baseFeePerGas.mul(baseFeeMultiplier).div(fixedPointAdjustment);
   return {
-    maxFeePerGas: maxPriorityFeePerGas.add(scaledBaseFee),
-    maxPriorityFeePerGas,
+    maxFeePerGas: scaledPriorityFee.add(scaledBaseFee),
+    maxPriorityFeePerGas: scaledPriorityFee,
```
james-a-morris, line 53: We should probably include a comment in the function def that indicates that `maxPriorityFeePerGas` is a scaled value

src/gasPriceOracle/adapters/linea-viem.ts
```diff
@@ -15,10 +15,12 @@ import { fixedPointAdjustment } from "../../utils";
  * @dev Because the Linea priority fee is more volatile than the base fee, the base fee multiplier will be applied
  * to the priority fee.
  * @param provider Viem PublicClient
- * @param _chainId Unused in this adapter
+ * @param {GasPriceEstimateOptions} opts Relevant options for Linea are baseFeeMultiplier and unsignedTx.
```
james-a-morris, line 18: Should be similar to above
```suggestion
 * @param opts Relevant options for Linea are baseFeeMultiplier and unsignedTx.
```

src/gasPriceOracle/adapters/linea.ts
```diff
@@ -7,6 +7,14 @@ import { GasPriceEstimate } from "../types";
 import * as ethereum from "./ethereum";
 import { GasPriceEstimateOptions } from "../oracle";
 
+/**
+ * Returns Linea gas price using the legacy eth_gasPrice RPC call.
+ * @param provider
+ * @param {GasPriceEstimateOptions} opts See notes below on specific parameters.
```
james-a-morris, line 13: Similar to above

```suggestion
 * @param opts See notes below on specific parameters.
```

src/gasPriceOracle/adapters/ethereum.ts
james-a-morris, line 53: No specific location is coming to mind, but are sure we don't reference `maxPriorityFeePerGas` anywhere that would break because we changed its definition here?

nicholaspai: I don't understand, we're using the multiplied value here so of course its going to change..?

james-a-morris: Maybe I'm thinking too in the weeds on this. I was just ensuring this doesn't have unintended side-effects. But this will shake itself out.

PR #818, opened 2025-01-06
improve(RelayFeeCalculator): Add parameters to getGasCosts method

nicholaspai:
Adds `priorityFeeMultiplier` and `opStackL1GasCostMultiplier` parameters to allow caller to further customize queries

### DIFF
--- a/src/relayFeeCalculator/chain-queries/baseQuery.ts
+++ b/src/relayFeeCalculator/chain-queries/baseQuery.ts
@@ -15,6 +15,7 @@ import {
   bnZero,
   assert,
   chainIsOPStack,
+  fixedPointAdjustment,
 } from "../../utils";
 import { Logger, QueryInterface } from "../relayFeeCalculator";
 import { Transport } from "viem";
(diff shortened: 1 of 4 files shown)


src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -146,8 +167,9 @@ export class QueryBase implements QueryInterface {
         gasLimit: nativeGasCost, // prevents additional gas estimation call
       });
       const l1GasCost = await (provider as L2Provider<providers.Provider>).estimateL1GasCost(populatedTransaction);
+      const scaledL1GasCost = l1GasCost.mul(opStackL1GasCostMultiplier).div(fixedPointAdjustment);
```
bmzig, line 170: Does this mean we scale twice for OP stack chains? Once in `getGasPriceEstimate` and once here?

bmzig (approved):

src/relayFeeCalculator/chain-queries/baseQuery.ts
james-a-morris: I don't believe so but it could be good to have @nicholaspai confirm - I will defer the approval to this answer

nicholaspai: we won't scale twice. The l1 gas cost is being scaled independently of the base fee or priotiy fee, which is already computed on line 156

james-a-morris (approved):

PR #819, opened 2025-01-07
improve(GasPriceOracle): Remove eip1559Bad and lineaEthers adapters

nicholaspai:
These are the default adapters today but in production we use the more accurate eip1559Raw and lineaViem variants. These were always designed to be set as the defaults following testing.

The benefit is we can remove the `GAS_PRICE_EIP1559_RAW_` and `NEW_GAS_PRICE_ORACLE_` environment variables


src/gasPriceOracle/adapters/ethereum.ts
```diff
@@ -58,37 +49,6 @@ export async function eip1559Raw(
...
-/**
- * @notice Returns fee data using provider's getFeeData() method.
- * @note Resolves priority gas pricing poorly, because the priority fee is hardcoded to 1.5 Gwei in ethers v5's
- * getFeeData() method
- * @dev TODO: Remove this function soon. See note above about slowly rolling out eip1559Raw.
- * @param provider ethers RPC provider instance.
- * @param chainId Chain ID of the provider instance.
- * @returns Promise of gas price estimate object.
- */
-export async function eip1559Bad(
```
mrice32, line 70: Maybe a dumb question, but we don't currently use this _anywhere_? No chain relies on it?

Totally fine with reversing the default, but want to make sure we're not breaking anything.

nicholaspai: in production no, we do not use it at all

nicholaspai:
Note to self: before releasing this into the relayer repo we should update the default scaler values in order to acknowledge that users of the relayer might not have set these "new" or "raw" oracles and they might see unexpected fees as a result of the bump

PR #821, opened 2025-01-08
feat(GasPriceAdapter): Allow caller to set a minimum priority fee for eip1559Raw adapter

nicholaspai:
Can floor the priority fee returned by eth_maxPriorityFeePerGas by using a new environment variable `MIN_PRIORITY_FEE_PER_GAS_`

### DIFF
--- a/src/gasPriceOracle/adapters/ethereum.ts
+++ b/src/gasPriceOracle/adapters/ethereum.ts
@@ -5,6 +5,10 @@ import { GasPriceEstimate } from "../types";
 import { gasPriceError } from "../util";
 import { GasPriceEstimateOptions } from "../oracle";
 
+const DEFAULT_MIN_PRIORITY_FEE: { [chainId: number]: string } = {
+  1: "0.5",
+};
+
 // TODO: We intend to remove `eip1559Bad()` as an option and make eip1559Raw the only option eventually. The reason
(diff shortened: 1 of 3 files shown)


src/gasPriceOracle/adapters/ethereum.ts
```diff
@@ -43,7 +43,11 @@ export async function eip1559Raw(
     provider.getBlock("pending"),
     (provider as providers.JsonRpcProvider).send("eth_maxPriorityFeePerGas", []),
   ]);
-  const maxPriorityFeePerGas = BigNumber.from(_maxPriorityFeePerGas);
+  let maxPriorityFeePerGas = BigNumber.from(_maxPriorityFeePerGas);
+  const flooredPriorityFeePerGas = parseUnits(process.env[`MIN_PRIORITY_FEE_PER_GAS_${chainId}`] || "0", 9);
```
nicholaspai, line 47: Question for reviewers: do you think its worth adding a default value for some chains here (e.g. "1": 0.5) to make it easier for API and relayer to be aligned? I kind of think it might be useful to keep relayer's floor value higher than the API and we can leave the API floor at 0 while we set the relayer's at 0.5 gwei

nicholaspai: no this could be set per-chain, so we could have a 0.5 gwei one for Ethereum

PR #823, opened 2025-01-09
improve(BundleDataClient): Support pre-fills

nicholaspai:
## Motivation

Support use case that will become frequent once deterministic relay hashes go live where a fill precedes a deposit in real world time.

### Update #1:

I'm not even sure there is a problem with refunding pre fills that occur near the end of a bundle (that precede a deposit at the beginning of the following bundle) because bundle end blocks are currently set to lag HEAD on all chains. Additionally, the bundle data client takes into account all deposits, even those right near HEAD, when validating fills. So unless the fill precedes the deposit by more than the [bundle end block buffer](https://github.com/across-protocol/relayer/blob/master/src/common/Constants.ts#L135) then there should be no issue. 

## Problem

This opens up the possibility that even with a [SpokePoolClient.query lag](https://github.com/across-protocol/sdk/blob/40888c405db19f1860f0d3f22e26cc5dc541306f/src/clients/BundleDataClient/utils/PoolRebalanceUtils.ts#L28) applied on all chains, there is a very unlucky situation where a fill is sent at the end of one bundle block range and a deposit is one of the first events in the next bundle block range. Currently, the dataworker would never issue a refund for the fill.

There are two ways around this I think and one is very complex to implement while the other is simple but increases relayer repayment latency and puts responsibility of how early to-prefill on the relayer:

1. BundleDataClient will need to "remember" which refunds it has already issued. It can do this easily using the arweave client though this increases dependencies on Arweave. Additionally, it will be quite complex to implement the most efficient data structure. I don't love this approach because of the tradeoffs of the following approach.
2. Increase the buffer that a fill must follow its destination chain HEAD block before the dataworker attempts to repay it. This additional buffer converted to seconds becomes the maximum amount of time that a fill can precede a deposit. I think this can be to something reasonable like 10 minutes assuming that pre-fills are only used by fillers who can guarantee somehow that a deposit lands on-chain when they want it to. Therefore the filler should be able to send a deposit after a fill in under 10 minutes in all cases.

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -849,13 +849,22 @@ export class BundleDataClient {
         // Keep track of fast fills that replaced slow fills, which we'll use to create "unexecutable" slow fills
         // if the slow fill request was sent in a prior bundle.
         const fastFillsReplacingSlowFills: string[] = [];
+
+        // Returns block number when we should start evaluating this fill for a refund. The lag builds in space for
+        // a filler to send a fill before a deposit in cases where the depositor is delegating transaction submission
+        // to the filler and giving them a signed transaction object.
+        const getFillBlockWithLag = (fill: V3FillWithBlock): number => {
+          const fillBuffer = /* FILL_BUFFER[fill.destinationChainId] ?? */ 0;
+          return fill.blockNumber - fillBuffer;
+        };
+
         await forEachAsync(
           destinationClient
             .getFillsForOriginChain(originChainId)
             // We can remove fills for deposits with input amount equal to zero because these will result in 0 refunded
             // tokens to the filler. We can't remove non-empty message deposit here in case there is a slow fill
             // request for the deposit, we'd want to see the fill took place.
-            .filter((fill) => fill.blockNumber <= destinationChainBlockRange[1] && !isZeroValueDeposit(fill)),
+            .filter((fill) => getFillBlockWithLag(fill) <= destinationChainBlockRange[1] && !isZeroValueDeposit(fill)),
           async (fill) => {
             const relayDataHash = this.getRelayHashFromEvent(fill);
             fillCounter++;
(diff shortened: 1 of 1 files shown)


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -849,13 +849,22 @@ export class BundleDataClient {
         // Keep track of fast fills that replaced slow fills, which we'll use to create "unexecutable" slow fills
         // if the slow fill request was sent in a prior bundle.
         const fastFillsReplacingSlowFills: string[] = [];
+
+        // Returns block number when we should start evaluating this fill for a refund. The lag builds in space for
+        // a filler to send a fill before a deposit in cases where the depositor is delegating transaction submission
+        // to the filler and giving them a signed transaction object.
+        const getFillBlockWithLag = (fill: V3FillWithBlock): number => {
+          const fillBuffer = /* FILL_BUFFER[fill.destinationChainId] ?? */ 0;
```
nicholaspai, line 857: I imagine to start we'd define a hardcoded value for this fill buffer. Technically it would need to be in the UMIP so it might even have to be a config store variable...

pxrl:
@nicholaspai I've been working on the UMIP and I'm wondering if we can get away with an even smaller change for this. So far I _think_ it's feasible to just resolve these old fills via binary search, assuming that we won't be flooded with them. For this we'd also need to ensure that the proposer only ever considers deposits that _also_ occur within its current proposal block range (i.e. so it should _not_ consider deposits that come after).

I'm hopeful that it'll work OK but it needs some pressure testing.

https://github.com/UMAprotocol/UMIPs/pull/611#discussion_r1922502298

nicholaspai:
> @nicholaspai I've been working on the UMIP and I'm wondering if we can get away with an even smaller change for this. So far I _think_ it's feasible to just resolve these old fills via binary search, assuming that we won't be flooded with them. For this we'd also need to ensure that the proposer only ever considers deposits that _also_ occur within its current proposal block range (i.e. so it should _not_ consider deposits that come after).
> 
> I'm hopeful that it'll work OK but it needs some pressure testing.
> 
> [UMAprotocol/UMIPs#611 (comment)](https://github.com/UMAprotocol/UMIPs/pull/611#discussion_r1922502298)

Directionally I agree with you. I don't think we even need binary searching for these deposits, since pre-fills will happen on deterministic deposits which have finite fill deadlines so the deposits will always be within the validator's lookback window. Moreover, I don't think the proposer needs to change its behavior at all...

See my Update here where I comment that we might need zero changes:

> > ### Update #1:
>> I'm not even sure there is a problem with refunding pre fills that occur near the end of a bundle (that precede a deposit at the beginning of the following bundle) because bundle end blocks are currently set to lag HEAD on all chains. Additionally, the bundle data client takes into account all deposits, even those right near HEAD, when validating fills. So unless the fill precedes the deposit by more than the [bundle end block buffer](https://github.com/across-protocol/relayer/blob/master/src/common/Constants.ts#L135) then there should be no issue.

PR #824, opened 2025-01-10
feat: lens testnet support

dohaki:
Closes ACX-3611 and ACX-3609

- Adds support for Lens Sepolia
- Refactors custom gas token queries into generic class


src/relayFeeCalculator/chain-queries/factory.ts
```diff
@@ -31,30 +30,21 @@ export class QueryBase__factory {
...
+          symbolMapping,
+          spokePoolAddress,
+          simulatedRelayerAddress,
+          logger,
+          coingeckoProApiKey,
+          fixedGasPrice[chainId],
+          "usd",
+        ],
+        customGasTokenSymbol,
+      });
```
gsteenkamp89, line 47: nice one, very neat

src/constants.ts
```diff
@@ -60,3 +60,12 @@ export const BRIDGED_USDC_SYMBOLS = [
...
   TOKEN_SYMBOLS_MAP.USDzC.symbol,
 ];
+
+export const CUSTOM_GAS_TOKENS = {
+  [CHAIN_IDs.POLYGON]: "MATIC",
+  [CHAIN_IDs.POLYGON_AMOY]: "MATIC",
+  [CHAIN_IDs.ALEPH_ZERO]: "AZERO",
+  // FIXME: Replace with GRASS price once listed on Coingecko.
+  // For testing purposes, we use ETH price instead.
+  [CHAIN_IDs.LENS_SEPOLIA]: "ETH",
```
gsteenkamp89, line 70: What are the implications of using eth here other than fee misleading fee calculation?

gsteenkamp89 (approved):
Approved, just a question or 2, but looks good!

src/constants.ts
dohaki: I don't think there should be any other implications other than incorrect relayer fee calculation. Because this is the only place where we use this constant (currently at least)

PR #828, opened 2025-01-14
feat(svm): add svm clients

md0x:


scripts/generate-clients.ts
```diff
@@ -0,0 +1,27 @@
...
+codama.accept(
+    renderJavaScriptVisitor(path.join(clientsPath, "SvmSpoke"))
+);
+
+codama = createFromRoot(rootNodeFromAnchor(MulticallHandlerIdl as AnchorIdl));
+codama.accept(
+    renderJavaScriptVisitor(path.join(clientsPath, "MulticallHandler"))
+);
+
+// codama = createFromRoot(rootNodeFromAnchor(MessageTransmitterIdl as AnchorIdl));
```
Reinis-FRP, line 19: we will also need MessageTransmitter client to finalize CCTP messages. It would be better if we could get IDL from upstream, but last time I checked they were not, so we ended up fetching old anchor version IDL from on-chain and converting in our contracts repo. Need to check if this has changed since the latest CCTP upgrade.

package.json
```diff
@@ -12,12 +12,14 @@
...
   },
   "scripts": {
+    "postinstall": "yarn generate-clients",
     "start": "yarn typechain && nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn dev'",
     "build": "yarn run clean && yarn typechain && yarn run build:cjs & yarn run build:esm & yarn run build:types; wait",
     "dev": "yarn run build:cjs & yarn run build:esm & yarn run build:types; wait",
     "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --removeComments --verbatimModuleSyntax false && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
     "build:esm": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm && echo > ./dist/esm/package.json '{\"type\":\"module\",\"sideEffects\":false}'",
     "build:types": "tsc --project tsconfig.build.json --module esnext --declarationDir ./dist/types --emitDeclarationOnly --declaration --declarationMap",
+    "generate-clients": "yarn ts-node ./scripts/generate-clients.ts",
```
Reinis-FRP, line 22: maybe better rename so its clear these are solana clients?

PR #829, opened 2025-01-17
feat: spoke pool client queries old and new event names

bmzig:
This refactors the spoke pool client so that it may query events for the upgraded spoke pool. We must be able to query and manage both events since the spoke pool upgrade will occur asynchronously. 

I believe the goal of this PR is to simply be able to index the new events into our spoke pool clients. It is then up to the function/client using the spoke pool clients to handle these new event structures. 

We need to handle:
- ExecutedRelayerRefundRoot - No change. Addresses/bytes32 are stored as strings in `RelayerRefundLeaf` (see [here](https://github.com/across-protocol/sdk/blob/db244fbaec15dc049452cd104ca7bec5856ad3da/src/interfaces/HubPool.ts#L19))
- EmergencyDeleteRootBundle - No change. We do not query this event in the hub pool client.
- TokensBridged - No change. The `l2TokenAddress` is [kept as a string](https://github.com/across-protocol/sdk/blob/db244fbaec15dc049452cd104ca7bec5856ad3da/src/interfaces/SpokePool.ts#L118) by us. 

For the next four signatures, both events are queried using the same logic as before, and the event data is assigned to a single state variable. For example, when we query both `V3FundsDeposited` and `FundsDeposited`, we create the internal `DepositWithBlock` struct out of the event data and store all these events into `this.depositHashes` (so `this.depositHashes` will have deposits with addresses as `inputToken` and `outputToken` if it was processed from a `V3FundsDeposited` event, and will also have deposits with `inputToken` and `outputToken` as bytes32 values if it was processed from a `FundsDeposited` event.
- V3FundsDeposited -> FundsDeposited
- RequestedSpeedUpV3Deposit -> RequestedSpeedUpDeposit
- FilledV3Relay -> FilledRelay
- RequestedV3SlowFill -> RequestedSlowFill


src/clients/SpokePoolClient.ts
```diff
@@ -570,7 +581,7 @@ export class SpokePoolClient extends BaseAbstractClient {
         deposit.fromLiteChain = this.isOriginLiteChain(deposit);
         deposit.toLiteChain = this.isDestinationLiteChain(deposit);
 
-        if (deposit.outputToken === ZERO_ADDRESS) {
+        if (deposit.outputToken === ZERO_ADDRESS || deposit.outputToken === ZERO_BYTES) {
           deposit.outputToken = this.getDestinationTokenForDeposit(deposit);
```
bmzig, line 585: This has some assumptions about how tokens are going to be stored in the hub pool client, namely, whether they are going to be stored all as bytes32 or as a mix of bytes32s and addresses. I'd probably prefer the latter.

bmzig:
We should probably split this into multiple PRs since changes to `getRelayDataHash` now makes this breaking. In general, the three main things to look out for are the change from `message` to `messageHash` for the two `FilledRelay` events, the change in calculating the relay hash depending on whether it was a V3 deposit or not, and the changes of parsing the deposit/filled relay events since `depositId` is now a u256.

src/clients/SpokePoolClient.ts
pxrl: The HubPool is constrained to work with the `address` type, so I believe these will be "virtual" addresses, and then we would map those to non-EVM addresses separately.

src/clients/SpokePoolClient.ts
```diff
@@ -570,7 +581,7 @@ export class SpokePoolClient extends BaseAbstractClient {
         deposit.fromLiteChain = this.isOriginLiteChain(deposit);
         deposit.toLiteChain = this.isDestinationLiteChain(deposit);
 
-        if (deposit.outputToken === ZERO_ADDRESS) {
+        if (deposit.outputToken === ZERO_ADDRESS || deposit.outputToken === ZERO_BYTES) {
```
pxrl, line 584: @bmzig I think we'd benefit from a utility function here, a bit as we have with `isMessageEmpty()`.

Something like `isOutputTokenZero(deposit)`. wdyt?

src/clients/SpokePoolClient.ts
```diff
@@ -796,18 +843,33 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+            toBlock: searchBounds.high,
+            maxBlockLookBack: this.eventSearchConfig.maxBlockLookBack,
+          }
+        ),
+      ])
+    ).flat();
     const tStop = Date.now();
 
-    const event = query.find(({ args }) => args["depositId"] === depositId);
+    const event = query.find(({ args }) => args["depositId"].eq(depositId));
```
pxrl, line 872: @bmzig This sticks out as one we'd want to defer. I guess we can leave it here for the sake of reviewing, but should remember to split it out into a follow-on PR.

src/clients/SpokePoolClient.ts
bmzig: Done here https://github.com/across-protocol/sdk/pull/829/commits/2dd1bb0dec27fa8fd3f60e3f760cb99cffd5513f. I also reverted the `isMessageEmpty` change since we can probably get something better than what was done. Changing the `getRelayDataHash` is still pending the contract changes.

james-a-morris:
> > This will break the FE/Relayer/Indexer/Scraper. Should we assign tasks to resolve these or will this be a fast follow?
> 
> Ideally this PR shouldn't break anything. Which part do you think is breaking? Edit: Ah I think you mean the package bump to contracts. Yeah, we will have to probably have to set up using both ABIs in parallel or something.

This migration will break the FE/relayer/indexer/scraper*

src/clients/SpokePoolClient.ts
```diff
@@ -110,21 +111,33 @@ export class SpokePoolClient extends BaseAbstractClient {
     this.configStoreClient = hubPoolClient?.configStoreClient;
   }
 
+  public eventSignatures: Record<string, string> = {
+    FundsDeposited: "FundsDeposited(bytes32,bytes32,uint256,uint256,uint256,uint256,uint32,uint32,uint32,bytes32,bytes32,bytes32,bytes)",
+    FilledRelay: "FilledRelay(bytes32,bytes32,uint256,uint256,uint256,uint256,uint256,uint32,uint32,bytes32,bytes32,bytes32,bytes32,bytes32,(bytes32,bytes32,uint256,uint8))",
+    RequestedSpeedUpDeposit: "RequestedSpeedUpDeposit(uint256,uint256,bytes32,bytes32,bytes,bytes)",
+    RequestedSlowFill: "RequestedSlowFill(bytes32,bytes32,uint256,uint256,uint256,uint256,uint32,uint32,bytes32,bytes32,bytes32,bytes32)",
+  };
```
pxrl, line 119: This is temporary and should be deleted once a new contracts release is available.

src/utils/AddressUtils.ts
```diff
@@ -39,10 +39,14 @@ export function compareAddressesSimple(addressA?: string, addressB?: string): bo
...
 
-// Converts an input (assumed to be) bytes32 string into a bytes20 string.
-// If the input is not a bytes32 but is less than type(uint160).max, then this function
-// will still succeed.
-// Throws an error if the string as an unsigned integer is greater than type(uint160).max.
+// Converts an input hex data string into a bytes32 string. Note that the output bytes will be lowercase
+// so that it naturally matches with ethers event data.
+// Throws an error if the input string is already greater than 32 bytes.
+export function toBytes32(address: string): string {
+  return utils.hexZeroPad(address, 32).toLowerCase();
```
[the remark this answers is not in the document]
pxrl: I was wrong on this - when extending to bytes32, checksum case doesn't make sense. I see that `toAddress()` correctly handles re-formatting to a checksummed address.

src/clients/SpokePoolClient.ts
```diff
@@ -898,15 +902,24 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    const [fromBlock, toBlock] = [searchBounds.low, searchBounds.high];
+    const { maxBlockLookBack } = this.eventSearchConfig;
+    const query = (
+      await Promise.all([
+        paginatedEventQuery(
+          this.spokePool,
+          this.spokePool.filters.V3FundsDeposited(null, null, null, null, null, depositId),
+          { fromBlock, toBlock, maxBlockLookBack }
+        ),
+
```
pxrl, line 915:
```suggestion
```

test/SpokePoolClient.SpeedUp.ts
```diff
@@ -65,6 +65,7 @@ describe("SpokePoolClient: SpeedUp", function () {
       outputToken,
       outputAmount
     );
+
```
pxrl, line 68:
```suggestion
```

test/SpokePoolClient.SpeedUp.ts
```diff
@@ -104,6 +106,7 @@ describe("SpokePoolClient: SpeedUp", function () {
       updatedRecipient,
     };
     const updatedDeposit = spokePoolClient.appendMaxSpeedUpSignatureToDeposit(deposit);
+
```
pxrl, line 109:
```suggestion
```

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -274,13 +275,15 @@ describe("SpokePoolClient: Fill Validation", function () {
...
-    const fill = await fillV3Relay(spokePool_2, deposit, relayer);
+    const fill = await fillV3Relay(spokePool_2, _deposit, relayer);
     expect(spokePoolClient2.getDepositForFill(fill)).to.equal(undefined);
     await spokePoolClient1.update();
 
-    // @todo: Drop `messageHash` exclusion once messageHash is reliably part of fills.
+    const deposit = spokePoolClient1.getDepositForFill(fill);
+    expect(deposit).to.exist;
+
     expect(spokePoolClient1.getDepositForFill(fill))
```
pxrl, line 285:
```suggestion
    deposit = deposit!;
    
    expect(deposit)
```

test/SpokePoolClient.v3Events.ts
```diff
@@ -738,36 +738,5 @@ describe("SpokePoolClient: Event Filtering", function () {
         expect(relayExecutionInfo.updatedMessageHash).to.equal(getMessageHash(deposit.message));
       }
     });
-
-    it("Correctly appends FilledV3Relay messageHash", async function () {
```
pxrl, line 742: This test was duplicated in this file; see lines 698 -> 740 above.

test/relayFeeCalculator.test.ts
```diff
@@ -452,7 +452,7 @@ describe("RelayFeeCalculator: Composable Bridging", function () {
       updatedOutputAmount: spreadFill.relayExecutionInfo.updatedOutputAmount.toString(),
     }).to.deep.eq({
       updatedRecipient: testContract.address,
-      updatedMessage: "0xabcdef",
+      updatedMessageHash: ethers.utils.keccak256("0xabcdef"),
```
pxrl, line 455:
```suggestion
      updatedMessageHash: getMessageHash("0xabcdef"),
```

mrice32 (approved):
LGTM!

package.json
```diff
@@ -100,7 +100,7 @@
   "dependencies": {
     "@across-protocol/across-token": "^1.0.0",
     "@across-protocol/constants": "^3.1.32",
-    "@across-protocol/contracts": "3.0.26-beta.1",
+    "@across-protocol/contracts": "4.0.0-beta.3",
```
bmzig, line 103: Bump to 4.0.0 now?

pxrl: https://github.com/across-protocol/sdk/pull/829/commits/11cf3beb9af2330fa1e84f46dee7cc8df62c4531

PR #830, opened 2025-01-20
improve: interpret depositIds as BigNumbers

bmzig:
This should change all instances of `depositId` from a `number` type to a `BigNumber` type. This is done in anticipation of `depositIds` becoming `uint256`s. 

Also, this adds some (currently unused) utility functions to be used in follow-up PRs.

### DIFF
--- a/src/clients/BundleDataClient/utils/SuperstructUtils.ts
+++ b/src/clients/BundleDataClient/utils/SuperstructUtils.ts
@@ -10,6 +10,7 @@ import {
   pattern,
   boolean,
   defaulted,
+  union,
   type,
 } from "superstruct";
 import { BigNumber } from "../../../utils";
@@ -17,7 +18,7 @@ import { BigNumber } from "../../../utils";
 const PositiveIntegerStringSS = pattern(string(), /\d+/);
 const Web3AddressSS = pattern(string(), /^0x[a-fA-F0-9]{40}$/);
 
-const BigNumberType = coerce(instance(BigNumber), string(), (value) => {
+const BigNumberType = coerce(instance(BigNumber), union([string(), number()]), (value) => {
   try {
     // Attempt to convert the string to a BigNumber
     return BigNumber.from(value);
--- a/src/clients/BundleDataClient/utils/SuperstructUtils.ts
+++ b/src/clients/BundleDataClient/utils/SuperstructUtils.ts
@@ -92,6 +92,7 @@ const V3FillSS = {
   repaymentChainId: number(),
   relayExecutionInfo: V3RelayExecutionEventInfoSS,
   quoteTimestamp: number(),
+  messageHash: optional(string()),
 };
 
 const V3FillWithBlockSS = {
--- a/src/clients/BundleDataClient/utils/SuperstructUtils.ts
+++ b/src/clients/BundleDataClient/utils/SuperstructUtils.ts
@@ -41,7 +41,7 @@ const V3RelayDataSS = {
   originChainId: number(),
   depositor: string(),
   recipient: string(),
-  depositId: number(),
+  depositId: BigNumberType,
   message: string(),
 };
 
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -314,7 +314,7 @@ export class SpokePoolClient extends BaseAbstractClient {
 
     this.logger.debug({
       at: "SpokePoolClient::getDepositForFill",
-      message: `Rejected fill for ${getNetworkName(fill.originChainId)} deposit ${fill.depositId.toString()}.`,
+      message: `Rejected fill for ${getNetworkName(fill.originChainId)} deposit ${fill.depositId}.`,
       reason: match.reason,
       deposit,
       fill,
@@ -815,7 +815,7 @@ export class SpokePoolClient extends BaseAbstractClient {
       const srcChain = getNetworkName(this.chainId);
       const dstChain = getNetworkName(destinationChainId);
       throw new Error(
-        `Could not find deposit ${depositId.toString()} for ${dstChain} fill` +
+        `Could not find deposit ${depositId} for ${dstChain} fill` +
           ` between ${srcChain} blocks [${searchBounds.low}, ${searchBounds.high}]`
       );
     }
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -314,7 +314,7 @@ export class SpokePoolClient extends BaseAbstractClient {
 
     this.logger.debug({
       at: "SpokePoolClient::getDepositForFill",
-      message: `Rejected fill for ${getNetworkName(fill.originChainId)} deposit ${fill.depositId}.`,
+      message: `Rejected fill for ${getNetworkName(fill.originChainId)} deposit ${fill.depositId.toString()}.`,
       reason: match.reason,
       deposit,
       fill,
@@ -815,7 +815,7 @@ export class SpokePoolClient extends BaseAbstractClient {
       const srcChain = getNetworkName(this.chainId);
       const dstChain = getNetworkName(destinationChainId);
       throw new Error(
-        `Could not find deposit ${depositId} for ${dstChain} fill` +
+        `Could not find deposit ${depositId.toString()} for ${dstChain} fill` +
           ` between ${srcChain} blocks [${searchBounds.low}, ${searchBounds.high}]`
       );
     }
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -399,7 +399,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @note This hash takes the form of: `${depositId}-${originChainId}`.
    */
   public getDepositHash(event: { depositId: BigNumber; originChainId: number }): string {
-    return `${event.depositId.toString()}-${event.originChainId}`;
+    return `${event.depositId}-${event.originChainId}`;
   }
 
   /**
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -513,7 +513,7 @@ export class SpokePoolClient extends BaseAbstractClient {
       currentTime: currentTime.toNumber(), // uint32
       oldestTime: oldestTime.toNumber(),
       firstDepositId,
-      latestDepositId: _latestDepositId > bnZero ? _latestDepositId : bnZero,
+      latestDepositId: _latestDepositId.gt(bnZero) ? _latestDepositId : bnZero,
       searchEndBlock: searchConfig.toBlock,
       events,
     };
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -13,6 +13,7 @@ import {
   isDefined,
   toBN,
   bnOne,
+  isUnsafeDepositId,
 } from "../utils";
 import {
   paginatedEventQuery,
@@ -582,10 +583,10 @@ export class SpokePoolClient extends BaseAbstractClient {
         }
         assign(this.depositHashes, [this.getDepositHash(deposit)], deposit);
 
-        if (deposit.depositId.lt(this.earliestDepositIdQueried)) {
+        if (deposit.depositId.lt(this.earliestDepositIdQueried) && !isUnsafeDepositId(deposit.depositId)) {
           this.earliestDepositIdQueried = deposit.depositId;
         }
-        if (deposit.depositId.gt(this.latestDepositIdQueried)) {
+        if (deposit.depositId.gt(this.latestDepositIdQueried) && !isUnsafeDepositId(deposit.depositId)) {
           this.latestDepositIdQueried = deposit.depositId;
         }
       }
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -12,6 +12,7 @@ import {
   getRelayDataHash,
   isDefined,
   toBN,
+  bnOne,
 } from "../utils";
 import {
   paginatedEventQuery,
@@ -46,8 +47,8 @@ type SpokePoolUpdateSuccess = {
   success: true;
   currentTime: number;
   oldestTime: number;
-  firstDepositId: number;
-  latestDepositId: number;
+  firstDepositId: BigNumber;
+  latestDepositId: BigNumber;
   events: Log[][];
   searchEndBlock: number;
 };
@@ -66,7 +67,7 @@ export class SpokePoolClient extends BaseAbstractClient {
   protected oldestTime = 0;
   protected depositHashes: { [depositHash: string]: DepositWithBlock } = {};
   protected depositHashesToFills: { [depositHash: string]: FillWithBlock[] } = {};
-  protected speedUps: { [depositorAddress: string]: { [depositId: number]: SpeedUpWithBlock[] } } = {};
+  protected speedUps: { [depositorAddress: string]: { [depositId: string]: SpeedUpWithBlock[] } } = {};
   protected slowFillRequests: { [relayDataHash: string]: SlowFillRequestWithBlock } = {};
   protected depositRoutes: { [originToken: string]: { [DestinationChainId: number]: boolean } } = {};
   protected tokensBridged: TokensBridged[] = [];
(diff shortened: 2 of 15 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -510,7 +513,7 @@ export class SpokePoolClient extends BaseAbstractClient {
       currentTime: currentTime.toNumber(), // uint32
       oldestTime: oldestTime.toNumber(),
       firstDepositId,
-      latestDepositId: Math.max(numberOfDeposits - 1, 0),
+      latestDepositId: _latestDepositId > bnZero ? _latestDepositId : bnZero,
```
dohaki, line 516: Shouldn't this be
```suggestion
      latestDepositId: _latestDepositId.gt(bnZero) ? _latestDepositId : bnZero,
```

dohaki (commented):
Looks good 👍 Left one remark

src/clients/SpokePoolClient.ts
bmzig: Ah yes, absolutely. Nice catch.

dohaki (approved):

src/utils/EventUtils.ts
```diff
@@ -58,6 +59,13 @@ export function spreadEvent(args: Result | Record<string, unknown>): { [key: str
   if (returnedObject.rootBundleId) {
     returnedObject.rootBundleId = Number(returnedObject.rootBundleId);
   }
+  // If depositId is included in the event, cast it to a BigNumber. Need to check if it is defined since the deposit ID can
+  // be 0, which would still make this evaluate as false.
+  if (isDefined(returnedObject.depositId)) {
+    // Assuming a numeric output, we can safely cast the unknown to BigNumberish since the depositId will either be a uint32 (and therefore a TypeScript `number`),
+    // or a uint256 (and therefore an ethers BigNumber).
+    returnedObject.depositId = toBN(returnedObject.depositId as BigNumberish);
+  }
```
pxrl, line 68: Were you able to confirm what is natively returned here? I guess it looks like a `number` by default, since we don't already check for it here. With the new ABI then I'd assume ethers is storing it as a BigNumber instead. In that case we might subsequently be able to drop this check again, once we're over the upgrade hurdle.

src/utils/SpokeUtils.ts
```diff
@@ -70,7 +72,7 @@ export function populateV3Relay(
  *        // where the deposit with deposit ID = targetDepositId was created.
  */
 export async function getBlockRangeForDepositId(
```
pxrl, line 74: In general we should probably mark this as a function that only works for deposits with a sequential depositId.

We might also want a utility that accepts a deposit instance and returns true/false depending on whether the depositId is > 2³²-1. Then we could use it in this function to apply special handling if it ever gets requested to resolve a non-sequential depositId.

It'd be nice to assert() to catch bugs, but since this function would also deal with dynamic input, that would pose a potential availability problem. So it might instead be nice if we could warn loudly via the logger.

wdyt?

src/utils/SpokeUtils.ts
```diff
@@ -380,3 +384,37 @@ export async function findFillBlock(
...
+    : {
+        ...relayData,
+        depositor: toBytes32(relayData.depositor),
+        recipient: toBytes32(relayData.recipient),
+        exclusiveRelayer: toBytes32(relayData.exclusiveRelayer),
+        inputToken: toBytes32(relayData.inputToken),
+        outputToken: toBytes32(relayData.outputToken),
+        message: isMessageEmpty(relayData.message) ? ZERO_BYTES : ethersUtils.keccak256(relayData.message),
+      };
+}
```
pxrl, line 415: I'm wondering if we instead want to have everything in its natural representation? If we promote everything to bytes32-formatting (effectively just adding leading 0s to EVM addresses) then our logging will suddenly look awful; long and unreadable.

I'm wondering if it could instead be preferable to normalise all the data on ingestion into the SpokePool and convert EVM addresses as bytes32 back into normal 40-char addresses. The fact that we're working with strings makes this seem more achievable than in the contracts, where we're forced to go to bytes32 for a universal type.

pxrl (commented):
Nice work! I have a few queries to flag for discussion. The BigNumber modifications seem 100% ready, though I think we should ensure that we have the downstream modifications ready to go before merging any of this because we might still need to get an SDK fix through quickly in the mean time.

src/utils/EventUtils.ts
bmzig: Yes. A `number` is returned here. We can/should delete this part once the upgrade rolls out.

src/clients/SpokePoolClient.ts
```diff
@@ -313,7 +314,7 @@ export class SpokePoolClient extends BaseAbstractClient {
 
     this.logger.debug({
       at: "SpokePoolClient::getDepositForFill",
-      message: `Rejected fill for ${getNetworkName(fill.originChainId)} deposit ${fill.depositId}.`,
+      message: `Rejected fill for ${getNetworkName(fill.originChainId)} deposit ${fill.depositId.toString()}.`,
```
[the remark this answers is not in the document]
bmzig: I see. Reverted here: https://github.com/across-protocol/sdk/pull/830/commits/8194b366af4878fccb549e2afd9919bdf76a61b1

src/utils/DepositUtils.ts
```diff
@@ -143,7 +143,7 @@ export function isZeroValueDeposit(deposit: Pick<Deposit, "inputAmount" | "messa
  * @returns True if the message is empty, false otherwise.
  */
 export function isMessageEmpty(message = EMPTY_MESSAGE): boolean {
-  return message === "" || message === "0x";
+  return message === "" || message === "0x" || message === ZERO_BYTES;
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/sdk/pull/830/commits/83c4c935391c7bb21fb403b5a9ceb1cef55ba23a

src/utils/SpokeUtils.ts
bmzig: The only issue with that is we can't convert from new to old if the new came from a `FilledRelay` event, since then we only have the message hash. We may not need this function, though. I had it for calculating the new relay hash type, which is now not necessary.

bmzig:
When using this code in the relayer repo, I've noticed that while running the dataworker, we will sometimes error out due to type errors when loading data in the BundleDataClient. I believe that this is due to the fact that the data we persist to Arweave contains depositIds as `number` types, so when we try to deserialize them into `BigNumber` types, things break (for reference, I can execute the proposer fine when I do not load data from arweave, but I error out when I do). Essentially, when we bump the relayer, we will probably need to run without loading data from Arweave for a bundle or so.

pxrl:
> I believe that this is because the data we persist to Arweave contains depositIds as `number` types, so when we try to deserialize them into `BigNumber` types, things break (for reference, I can execute the proposer fine when I do not load data from arweave, but I error out when I do). Essentially, when we bump the relayer, we will probably need to run without loading data from Arweave for a bundle or so.

This also implies that we'll have to change the format of the arweave payload to handle `depositId` as a string instead of a number, otherwise we'll end up posting bad data (or we simply won't be able to post anything, which is probably more likely).

If we modify the arweave schema to accept `depositId` as a string in advance of any other changes, and handle unpacking it into a number when we read it back, then I think we could probably make a subsequent transition of number -> BigNumber without this issue. wdyt?

src/clients/BundleDataClient/utils/SuperstructUtils.ts
```diff
@@ -41,7 +41,7 @@ const V3RelayDataSS = {
   originChainId: number(),
   depositor: string(),
   recipient: string(),
-  depositId: number(),
+  depositId: BigNumberType,
```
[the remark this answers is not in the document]
pxrl: This might be it: https://docs.superstructjs.org/api-reference/types#union

`union([string(), number()])`

src/clients/SpokePoolClient.ts
```diff
@@ -66,18 +67,18 @@ export class SpokePoolClient extends BaseAbstractClient {
...
   protected relayerRefundExecutions: RelayerRefundExecutionWithBlock[] = [];
   protected queryableEventNames: string[] = [];
   protected configStoreClient: AcrossConfigStoreClient | undefined;
-  public earliestDepositIdQueried = Number.MAX_SAFE_INTEGER;
-  public latestDepositIdQueried = 0;
-  public firstDepositIdForSpokePool = Number.MAX_SAFE_INTEGER;
-  public lastDepositIdForSpokePool = Number.MAX_SAFE_INTEGER;
+  public earliestDepositIdQueried = MAX_BIG_INT;
+  public latestDepositIdQueried = bnZero;
+  public firstDepositIdForSpokePool = MAX_BIG_INT;
```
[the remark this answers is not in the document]
nicholaspai: These are useful for the `queryHistoricalDepositForFill` function which we need to maintain as long as infinite lookbacks are supported. This means we can mark these variables for deletion following removal of the `depositLegacy()` function.

src/clients/SpokePoolClient.ts
```diff
@@ -438,10 +439,11 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-    if (firstDepositId === Number.MAX_SAFE_INTEGER) {
+    let firstDepositId = this.firstDepositIdForSpokePool;
+    if (firstDepositId.eq(MAX_BIG_INT)) {
       firstDepositId = await this.spokePool.numberOfDeposits({ blockTag: this.deploymentBlock });
-      if (isNaN(firstDepositId) || firstDepositId < 0) {
+      firstDepositId = BigNumber.from(firstDepositId); // Cast input to a big number.
+      if (!BigNumber.isBigNumber(firstDepositId) || firstDepositId.lt(bnZero)) {
         throw new Error(`SpokePoolClient::update: Invalid first deposit id (${firstDepositId})`);
       }
     }
```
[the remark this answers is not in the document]
nicholaspai: But what about all the deposit ID's that are not deterministic?

src/utils/SpokeUtils.ts
nicholaspai: I agree with everything you said above but I'm not sure why we can't use `assert(depositId <  2³²-1)`?

src/clients/BundleDataClient/utils/SuperstructUtils.ts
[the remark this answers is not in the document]
bmzig: Updated here. https://github.com/across-protocol/sdk/pull/830/commits/846bc86b3a1ed7be6f1813695f36e80332570ca4 Would be good to get @james-a-morris's opinion about this Arweave change.

src/utils/SpokeUtils.ts
bmzig: I removed it since we likely won't need it post contract changes. https://github.com/across-protocol/sdk/pull/830/commits/59582b35d542cf00c4fb60c71c0025b0e5e4e3fd

src/utils/SpokeUtils.ts
bmzig: For now, I've opted to just throw errors if we have an unsafe deposit ID (since we don't have a logger in utils) so that we can handle it upstream.

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
bmzig: I think what we should do is keep all this functionality and then filter its use based on whether the depositId is unsafe or not.

src/utils/SpokeUtils.ts
pxrl: I'm still a bit nervous that this might cause an outage in production under some unforseen circumstance. But in the short term it will also help us to clearly identify that, so OK :+1:

src/utils/SpokeUtils.ts
```diff
@@ -200,10 +205,11 @@ export async function getBlockRangeForDepositId(
...
  * @returns The deposit ID.
  */
-export async function getDepositIdAtBlock(contract: Contract, blockTag: number): Promise<number> {
-  const depositIdAtBlock = await contract.numberOfDeposits({ blockTag });
+export async function getDepositIdAtBlock(contract: Contract, blockTag: number): Promise<BigNumber> {
+  const _depositIdAtBlock = await contract.numberOfDeposits({ blockTag });
+  const depositIdAtBlock = toBN(_depositIdAtBlock);
   // Sanity check to ensure that the deposit ID is an integer and is greater than or equal to zero.
-  if (!Number.isInteger(depositIdAtBlock) || depositIdAtBlock < 0) {
+  if (!BigNumber.isBigNumber(depositIdAtBlock) || depositIdAtBlock.lt(bnZero)) {
```
pxrl, line 212: @bmzig Instead of checking `BigNumber.isBigNumber()`, I think we'd want to wrap `toBN()` in a try/catch clause. This handles the case where `depositIdAtBlock` is some unexpected type that can't be converted to a BN.

src/utils/SpokeUtils.ts
```diff
@@ -380,3 +386,8 @@ export async function findFillBlock(
 
   return lowBlockNumber;
 }
+
+// Determines if the input address (either a bytes32 or bytes20) is the zero address.
+export function isZeroAddress(address: string): boolean {
+  return address === ZERO_ADDRESS || address === ZERO_BYTES;
+}
```
pxrl, line 393: @bmzig This isn't related to the depositId changes; just want to make sure it's a conscious decision to have it included in this PR.

src/utils/AddressUtils.ts
pxrl, line 1: These changes seem unrelated; just want to confirm that they're supposed to be here :+1:

src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -92,17 +92,20 @@ export class MockSpokePoolClient extends SpokePoolClient {
           events[idx].push(event);
         }
       });
+    const bnMax = (val: BigNumber, cmp: BigNumber) => {
+      return val.gt(cmp) ? val : cmp;
+    };
```
pxrl, line 97: I thought we had this somewhere in `BNUtils` or something - but I can't find it anywhere. It'd be good to put it somewhere common because I'm sure it's not the last time we'll use it.

pxrl (commented):
@bmzig Looking really good. I have a couple of minor points, but otherwise I'm happy for this to go in as long as we have the relayer changes ready to go.

src/clients/SpokePoolClient.ts
```diff
@@ -396,16 +398,16 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @note This hash is used to match deposits and fills together.
    * @note This hash takes the form of: `${depositId}-${originChainId}`.
    */
-  public getDepositHash(event: { depositId: number; originChainId: number }): string {
-    return `${event.depositId}-${event.originChainId}`;
+  public getDepositHash(event: { depositId: BigNumber; originChainId: number }): string {
+    return `${event.depositId.toString()}-${event.originChainId}`;
```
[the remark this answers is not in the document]
bmzig: Done here: https://github.com/across-protocol/sdk/pull/830/commits/20d698c8a610ed4b37e6976a1dd971816338ff7e

src/utils/SpokeUtils.ts
bmzig: I removed it and didn't add it in a try/catch block since it looks like this function throws errors anyways. Do you think that's fine? https://github.com/across-protocol/sdk/pull/830/commits/20d698c8a610ed4b37e6976a1dd971816338ff7e

src/utils/SpokeUtils.ts
bmzig: I kept this since other PRs point to this one, and this is used in those PRs.

src/utils/AddressUtils.ts
bmzig: Removed: https://github.com/across-protocol/sdk/pull/830/commits/20d698c8a610ed4b37e6976a1dd971816338ff7e

src/clients/mocks/MockSpokePoolClient.ts
bmzig: Moved: https://github.com/across-protocol/sdk/pull/830/commits/20d698c8a610ed4b37e6976a1dd971816338ff7e

src/clients/BundleDataClient/utils/SuperstructUtils.ts
```diff
@@ -10,14 +10,15 @@ import {
...
+  union,
   type,
 } from "superstruct";
 import { BigNumber } from "../../../utils";
 
 const PositiveIntegerStringSS = pattern(string(), /\d+/);
 const Web3AddressSS = pattern(string(), /^0x[a-fA-F0-9]{40}$/);
 
-const BigNumberType = coerce(instance(BigNumber), string(), (value) => {
+const BigNumberType = coerce(instance(BigNumber), union([string(), number()]), (value) => {
```
james-a-morris, line 21: I think we have a BigNumber SS type elsewhere in the org. Either in the relayer or the SDK.

src/clients/BundleDataClient/utils/SuperstructUtils.ts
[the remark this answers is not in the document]
james-a-morris: @bmzig From above we do this elsewhere with BigNumbers in SuperStruct. My recommendation would be to reference the existing code.

src/utils/SpokeUtils.ts
pxrl: Yeah it has the same effect and the code is more concise than adding a try/catch clause. It might be harder to debug, but I doubt we'll ever see this in practice.

pxrl (approved):
OK from me, nice work @bmzig

PR #831, opened 2025-01-21
feat: add updated UMIP invalidity cases

bmzig:
A deposit refund is now "invalid" if the depositor specified cannot be repaid (since the origin chain is an EVM chain and the depositor address is not a bytes20 address).
A fill is invalid if the relayer supplies a 32 byte address and requests repayment on an EVM chain **and** the `msg.sender` is a 32-byte address when inspecting the destination chain fill.


src/utils/NetworkUtils.ts
```diff
@@ -123,6 +123,56 @@ export function chainIsL1(chainId: number): boolean {
   return [CHAIN_IDs.MAINNET, CHAIN_IDs.SEPOLIA].includes(chainId);
 }
 
+/**
+ * Determines whether a chain ID runs on an EVM-like execution layer.
+ * @param chainId Chain ID to evaluate.
+ * @returns True if chain corresponding to chainId has an EVM-like execution layer.
+ */
+export function chainIsEvm(chainId: number): boolean {
```
bmzig, line 131: Thoughts on moving this up to constants? So like `ChainFamily`, add another parameter to the `PublicNetwork` interface like `executionLayer` and specify evm/svm/etc.

src/utils/AddressUtils.ts
```diff
@@ -55,3 +55,12 @@ export function toAddress(bytes32: string): string {
   const rawAddress = utils.hexZeroPad(utils.hexStripZeros(bytes32), 20);
   return utils.getAddress(rawAddress);
 }
+
+// Checks if an input address is a 32-byte address or not.
+export function isAddressBytes32(address: string): boolean {
```
bmzig, line 60: It might be good to do something like what viem does here and define a type `Address` and enforce them to be `0x${string}`, maybe even of certain lengths if we can do that in TypeScript.

src/utils/NetworkUtils.ts
pxrl: Yeah I think we need some central way of inferring the execution environment. We can infer most of it based on the ChainFamily alone, though that doesn't help with Ethereum, Linea and Scroll. It'd be nice if we could avoid adding another property, but we might be forced to.

src/utils/NetworkUtils.ts
```diff
@@ -123,6 +123,56 @@ export function chainIsL1(chainId: number): boolean {
...
+    POLYGON,
+    REDSTONE,
+    SCROLL,
+    SONEIUM,
+    SUPERSEED,
+    WORLD_CHAIN,
+    ZK_SYNC,
+    ZORA,
+  } = CHAIN_IDs;
+  return [
```
pxrl, line 153: This could be shortened by just checking for ChainFamily OP_STACK, ZK_STACK or ORBIT, and then additionall checking for MAINNET, SCROLL or LINEA. That'd be a little bit more maintainable than an explicit list, but I still think we can do better.

src/utils/DepositUtils.ts
```diff
@@ -40,6 +41,9 @@ export async function queryHistoricalDepositForFill(
   fill: Fill | SlowFillRequest,
   cache?: CachingMechanismInterface
 ): Promise<DepositSearchResult> {
+  if (isUnsafeDepositId(fill.depositId)) {
```
pxrl, line 44: I think this should probably instead return an instance of `DepositSearchResult`:

```
    return {
      found: false,
      code: InvalidFill.DepositIdOutOfRange,
      reason: `Deposit ID ${depositId} is outside of the sequential deposit ID range [0, 2^32 - 1] .`,
    };
```

bmzig: Good call. Done here: https://github.com/across-protocol/sdk/pull/831/commits/aeb9a95db71a669e9d2fe3e77a17e174979c0c6c

src/utils/NetworkUtils.ts
bmzig: Fair enough. Alternatively, we can just make a chain family for Solana (and other non evm chains in the future) and check that it isn't in those set. https://github.com/across-protocol/sdk/pull/831/commits/bdaa603fb53c95641090caaf99d9172a03f3c616

nicholaspai (commented):
I think the implementation of the Bundle Data Client changes can be refactored to make sure we're capturing all cases where we're validating deposits and refunds

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -805,8 +808,15 @@ export class BundleDataClient {
             return;
           }
 
+          // A deposit refund for a deposit is invalid if the depositor has a bytes32 address input for an EVM chain. It is valid otherwise.
```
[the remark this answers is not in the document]
bmzig: Fair enough. Updated here: https://github.com/across-protocol/sdk/pull/831/commits/59315b8a3694f2eb6f2f6c1de6b88d9bd47b9005

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -914,11 +924,32 @@ export class BundleDataClient {
                 bundleInvalidFillsV3.push(fill);
                 return;
               }
+              // If the fill's repayment address is not a valid EVM address and the repayment chain is an EVM chain, the fill is invalid.
+              if (chainIsEvm(fill.repaymentChainId) && isAddressBytes32(fill.relayer)) {
```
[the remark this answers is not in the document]
bmzig: Never mind. Updated here: https://github.com/across-protocol/sdk/pull/831/commits/76de69a24a709b48b3b8efa71acb3bdf2704d898

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -914,9 +926,20 @@ export class BundleDataClient {
                 bundleInvalidFillsV3.push(fill);
                 return;
               }
+              // If the fill's repayment address is not a valid EVM address and the repayment chain is an EVM chain, the fill is invalid.
+              if (chainIsEvm(fill.repaymentChainId) && !utils.isAddress(fill.relayer)) {
+                const fillTransaction = await originClient.spokePool.provider.getTransaction(fill.transactionHash);
+                const originRelayer = fillTransaction.from;
+                // Repayment chain is still an EVM chain, but the msg.sender is a bytes32 address, so the fill is invalid.
+                if (!utils.isAddress(originRelayer)) {
```
nicholaspai, line 934: How could the repayment chain be EVM but the sender is not a valid address? I understand including this here for completeness, though maybe it should be an assert instead of an if, since it seems impossible to happen in reality

src/utils/NetworkUtils.ts
```diff
@@ -123,6 +123,49 @@ export function chainIsL1(chainId: number): boolean {
...
+      ARBITRUM_SEPOLIA,
+      LINEA,
+      POLYGON,
+      POLYGON_AMOY,
+      SCROLL,
+      SCROLL_SEPOLIA,
+      ZK_SYNC,
+      ZK_SYNC_SEPOLIA,
+    } = CHAIN_IDs;
+    return [
```
nicholaspai, line 147: Should this just return true unless the chain ID is solana for now? In other words, it should always return true currently?

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: The check is on the origin chain. There could technically be a fill on a non-evm chain with repayment on an evm chain.

src/utils/NetworkUtils.ts
bmzig: Yes. I didn't do that since we don't have Solana in constants yet.

src/clients/BundleDataClient/BundleDataClient.ts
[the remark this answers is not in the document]
bmzig: I've realized that we probably do _not_ want to use the ethers check, since once the new events start coming in, and depositor/recipient/etc are bytes32 values, the ethers check will always evaluate to false. So we will need to make a new function.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: yeah i guess this will never be possible while all chains are evm. This is a fine placeholder for now i guess

src/utils/NetworkUtils.ts
nicholaspai: maybe just return true for now then and add a note that we can add solana chain in here once we have it avaialble

nicholaspai (approved):

PR #835, opened 2025-01-23
feat(BundleDataClient): Support refunds for pre-fills/slow-fill-requests and duplicate deposits

nicholaspai:
### Definitions

- "pre-fill": any fill that is in a bundle that precedes the bundle containing the matched deposit
- "pre-slow-fill-request": a mouthful, and any slow fill request in a bundle that precedes the bundle containing the matched deposit
- "duplicate deposit": a deposit with the same relay hash as an existing one

## Core logic for refunding pre-fills:

Evaluate all deposits in the current bundle that were not filled in the current bundle. Figure out whether the deposit was filled, by either:
- Finding the fill in the spoke pool client's memory, or
- Querying the fill status on-chain

If there is a fill for this deposit then we need to issue a refund for this fill because the fill occurred in a previous bundle where it might not have been refunded.

We need to use a similar algorithm for creating slow fill leaves for pre-slow-fill-requests with the caveat that we cannot create slow fill leaves for deposits that have expired

## Non-determinism for refunding pre-fills

We don't deterministically know whether this pre-fill was refunded in the prior bundle because we don't know for certain what kind of event search window the proposer of the prior bundle used when constructing the bundle.

To illustrate this problem, imagine if a fill and a deposit are sent 10 minutes apart such that the fill is at the end of the current bundle and the deposit is at the beginning of the next. The prior bundle proposer probably would have issued a refund for this fill, but we don't know for certain.

So, one way around this non-determinism is to introduce a new rule to the UMIP:

### Do not issue refunds for fills or slow fill leaves where the matched deposit is later than the current bundle block range

This rule makes it always apparent which bundle should include a refund or slow fill leaf for a pre-fill or pre-slow-fill-request.

This will require a UMIP change to [this PR](https://github.com/UMAprotocol/UMIPs/pull/611)

### findFillEvents

When we need to query `fillStatuses()` on-chain  to detect whether a deposit corresponds to a very old pre-fill, then we also need to locate that old fill. We assume this is rare and will use the relatively expensive `findFillEvents()` function which is built off of the existing `findFillBlock()` function to locate that event.

## Duplicate deposits

This PR also adds support for duplicate deposits that are possible to create with unsafeDeposit. Duplicate deposits are added to `bundleDeposits` for they should be included in `runningBalances`. Duplicate deposits can be refunded upon expiry but for all deposits, expiry refunds cannot occur once the deposit has been filled. 

Due to the presence of pre-fills, we can now no longer instantly include that a deposit submitted in the current bundle that has expired is due a refund. We now need to check its relay status in the case where we don't see a matching fill in our memory.

## Unit tests

Can be found here: https://github.com/across-protocol/relayer/pull/2010

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -145,6 +145,9 @@ function updateBundleExcessSlowFills(
 }
 
 function updateBundleSlowFills(dict: BundleSlowFills, deposit: V3DepositWithBlock & { lpFeePct: BigNumber }): void {
+  if (chainIsEvm(deposit.destinationChainId) && !isValidEvmAddress(deposit.recipient)) {
+    return;
+  }
   const { destinationChainId, outputToken } = deposit;
   if (!dict?.[destinationChainId]?.[outputToken]) {
     assign(dict, [destinationChainId, outputToken], []);
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -1062,10 +1062,7 @@ export class BundleDataClient {
 
         // @todo Only start refunding pre-fills and slow fill requests after a config store version is activated. We
         // should remove this check once we've advanced far beyond the version bump block.
-        if (
-          versionAtProposalBlock >= PRE_FILL_MIN_CONFIG_STORE_VERSION ||
-          process.env.FORCE_PRE_FILL_REFUNDS === "true"
-        ) {
+        if (versionAtProposalBlock >= PRE_FILL_MIN_CONFIG_STORE_VERSION) {
           await mapAsync(
             bundleDepositHashes.filter((depositHash) => {
               const { deposit } = v3RelayHashes[depositHash];
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -939,6 +939,16 @@ export class BundleDataClient {
                   ) {
                     fastFillsReplacingSlowFills.push(relayDataHash);
                   }
+
+                  // Now that know this deposit has been filled on-chain, identify any duplicate deposits sent for this fill and refund
+                  // them, because they would not be refunded otherwise. These deposits can no longer expire and get
+                  // refunded as an expired deposit, and they won't trigger a pre-fill refund because the fill is
+                  // in this bundle. Pre-fill refunds only happen when deposits are sent in this bundle and the
+                  // fill is from a prior bundle.
+                  const duplicateDeposits = v3RelayHashes[relayDataHash].deposits!.slice(1);
+                  duplicateDeposits.forEach((duplicateDeposit) => {
+                    updateExpiredDepositsV3(expiredDepositsToRefundV3, duplicateDeposit);
+                  });
                 }
               } else {
                 throw new Error("Duplicate fill detected");
@@ -1016,6 +1026,10 @@ export class BundleDataClient {
                 ) {
                   fastFillsReplacingSlowFills.push(relayDataHash);
                 }
+
+                // No need to check for duplicate deposits here since we would have seen them in memory if they
+                // had a non-infinite fill deadline, and duplicate deposits with infinite deadlines are impossible
+                // to send.
               }
             }
           }
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -50,6 +50,7 @@ import {
   getRefundsFromBundle,
   getWidestPossibleExpectedBlockRange,
   isChainDisabled,
+  isEvmRepaymentValid,
   PoolRebalanceRoot,
   prettyPrintV3SpokePoolEvents,
   V3DepositWithBlock,
@@ -92,10 +93,11 @@ function updateBundleFillsV3(
   repaymentToken: string,
   repaymentAddress: string
 ): void {
-  // It is impossible to refund a deposit if the repayment chain is EVM and the relayer is a non-evm address.
-  if (chainIsEvm(repaymentChainId) && !isValidEvmAddress(repaymentAddress)) {
-    return;
-  }
+  // We shouldn't pass any unrepayable fills into this function, so we perform an extra safety check.
+  assert(
+    chainIsEvm(repaymentChainId) && isEvmRepaymentValid(fill, repaymentChainId),
+    "validatedBundleV3Fills dictionary should only contain fills with valid repayment information"
+  );
   if (!dict?.[repaymentChainId]?.[repaymentToken]) {
     assign(dict, [repaymentChainId, repaymentToken], {
       fills: [],
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -927,6 +927,27 @@ export class BundleDataClient {
                       ...fillToRefund,
                       quoteTimestamp: v3RelayHashes[relayDataHash].deposits![0].quoteTimestamp, // ! due to assert above
                     });
+
+                    // Now that we know this deposit has been filled on-chain, identify any duplicate deposits
+                    // sent for this fill and refund them to the filler, because this value would not be paid out
+                    // otherwise. These deposits can no longer expire and get refunded as an expired deposit,
+                    // and they won't trigger a pre-fill refund because the fill is in this bundle.
+                    // Pre-fill refunds only happen when deposits are sent in this bundle and the
+                    // fill is from a prior bundle. Paying out the filler keeps the behavior consistent for how
+                    // we deal with duplicate deposits regardless if the deposit is matched with a pre-fill or
+                    // a current bundle fill. If the fill is a slow fill,
+                    const duplicateDeposits = v3RelayHashes[relayDataHash].deposits!.slice(1);
+                    duplicateDeposits.forEach((duplicateDeposit) => {
+                      // If fill is a slow fill, refund deposit to depositor, otherwise refund to filler.
+                      if (isSlowFill(fill)) {
+                        updateExpiredDepositsV3(expiredDepositsToRefundV3, duplicateDeposit);
+                      } else {
+                        validatedBundleV3Fills.push({
+                          ...fillToRefund,
+                          quoteTimestamp: duplicateDeposit.quoteTimestamp,
+                        });
+                      }
+                    });
                   }
 
                   // If fill replaced a slow fill request, then mark it as one that might have created an
@@ -939,16 +960,6 @@ export class BundleDataClient {
                   ) {
                     fastFillsReplacingSlowFills.push(relayDataHash);
                   }
-
-                  // Now that know this deposit has been filled on-chain, identify any duplicate deposits sent for this fill and refund
-                  // them, because they would not be refunded otherwise. These deposits can no longer expire and get
-                  // refunded as an expired deposit, and they won't trigger a pre-fill refund because the fill is
-                  // in this bundle. Pre-fill refunds only happen when deposits are sent in this bundle and the
-                  // fill is from a prior bundle.
-                  const duplicateDeposits = v3RelayHashes[relayDataHash].deposits!.slice(1);
-                  duplicateDeposits.forEach((duplicateDeposit) => {
-                    updateExpiredDepositsV3(expiredDepositsToRefundV3, duplicateDeposit);
-                  });
                 }
               } else {
                 throw new Error("Duplicate fill detected");
@@ -1017,6 +1028,10 @@ export class BundleDataClient {
                     quoteTimestamp: matchedDeposit.quoteTimestamp,
                   });
                   v3RelayHashes[relayDataHash].fill = fillToRefund;
+
+                  // No need to check for duplicate deposits here since we would have seen them in memory if they
+                  // had a non-infinite fill deadline, and duplicate deposits with infinite deadlines are impossible
+                  // to send.
                 }
 
                 // slow fill requests for deposits from or to lite chains are considered invalid
@@ -1026,10 +1041,6 @@ export class BundleDataClient {
                 ) {
                   fastFillsReplacingSlowFills.push(relayDataHash);
                 }
-
-                // No need to check for duplicate deposits here since we would have seen them in memory if they
-                // had a non-infinite fill deadline, and duplicate deposits with infinite deadlines are impossible
-                // to send.
               }
             }
           }
@@ -1218,7 +1229,10 @@ export class BundleDataClient {
           // in the same bundle.
           if (fillStatus === FillStatus.Filled) {
             // We need to find the fill event to issue a refund to the right relayer and repayment chain,
-            // or msg.sender if relayer address is invalid for the repayment chain.
+            // or msg.sender if relayer address is invalid for the repayment chain. We don't need to
+            // verify the fill block is before the bundle end block on the destination chain because
+            // we queried the fillStatus at the end block. Therefore, if the fill took place after the end block,
+            // then we wouldn't be in this branch of the code.
             const prefill = await this.findMatchingFillEvent(deposit, destinationClient);
             assert(isDefined(prefill), `findFillEvent# Cannot find prefill: ${relayDataHash}`);
             assert(this.getRelayHashFromEvent(prefill!) === relayDataHash, "Relay hashes should match.");
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -697,6 +697,7 @@ export class BundleDataClient {
     const bundleFillsV3: BundleFillsV3 = {}; // Fills to refund in bundle block range.
     const bundleInvalidFillsV3: V3FillWithBlock[] = []; // Fills that are not valid in this bundle.
     const bundleUnrepayableFillsV3: V3FillWithBlock[] = []; // Fills that are not repayable in this bundle.
+    const bundleInvalidSlowFillRequests: SlowFillRequestWithBlock[] = []; // Slow fill requests that are not valid in this bundle.
     const bundleSlowFillsV3: BundleSlowFills = {}; // Deposits that we need to send slow fills
     // for in this bundle.
     const expiredDepositsToRefundV3: ExpiredDepositsToRefundV3 = {};
@@ -1111,18 +1112,21 @@ export class BundleDataClient {
             // want to perform a binary search lookup for it because the deposit ID is "unsafe" and cannot be
             // found using such a method) because infinite fill deadlines cannot be produced from the unsafeDepositV3()
             // function.
-            if (
-              INFINITE_FILL_DEADLINE.eq(slowFillRequest.fillDeadline) &&
-              slowFillRequest.blockNumber >= destinationChainBlockRange[0]
-            ) {
+            if (slowFillRequest.blockNumber >= destinationChainBlockRange[0]) {
+              if (!INFINITE_FILL_DEADLINE.eq(slowFillRequest.fillDeadline)) {
+                bundleInvalidSlowFillRequests.push(slowFillRequest);
+                return;
+              }
               const historicalDeposit = await queryHistoricalDepositForFill(originClient, slowFillRequest);
               if (!historicalDeposit.found) {
+                bundleInvalidSlowFillRequests.push(slowFillRequest);
                 return;
               }
               const matchedDeposit: V3DepositWithBlock = historicalDeposit.deposit;
               // If deposit is in a following bundle, then this slow fill request will have to be created
               // once that deposit is in the current bundle.
               if (matchedDeposit.blockNumber > originChainBlockRange[1]) {
+                bundleInvalidSlowFillRequests.push(slowFillRequest);
                 return;
               }
               // @dev Since queryHistoricalDepositForFill validates the slow fill request by checking individual
@@ -1443,7 +1447,7 @@ export class BundleDataClient {
     if (bundleInvalidFillsV3.length > 0) {
       this.logger.debug({
         at: "BundleDataClient#loadData",
-        message: "Finished loading V3 spoke pool data and found some invalid V3 fills in range",
+        message: "Finished loading V3 spoke pool data and found some invalid fills in range",
         blockRangesForChains,
         bundleInvalidFillsV3,
       });
@@ -1452,9 +1456,18 @@ export class BundleDataClient {
     if (bundleUnrepayableFillsV3.length > 0) {
       this.logger.debug({
         at: "BundleDataClient#loadData",
-        message: "Finished loading V3 spoke pool data and found some unrepayable V3 fills in range",
+        message: "Finished loading V3 spoke pool data and found some unrepayable fills in range",
         blockRangesForChains,
         bundleUnrepayableFillsV3,
+      });
+    }
+
+    if (bundleInvalidSlowFillRequests.length > 0) {
+      this.logger.debug({
+        at: "BundleDataClient#loadData",
+        message: "Finished loading V3 spoke pool data and found some invalid slow fill requests in range",
+        blockRangesForChains,
+        bundleInvalidSlowFillRequests,
       });
     }
 
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -1104,6 +1104,8 @@ export class BundleDataClient {
             );
           }),
           async (depositHash) => {
+            // We don't need to call verifyFillRepayment() here to replace the fill.relayer because this value should already
+            // be overwritten because the deposit and fill both exist.
             const { deposit, fill, slowFillRequest } = v3RelayHashes[depositHash];
             if (!deposit) throw new Error("Deposit should exist in relay hash dictionary.");
 
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -994,7 +994,7 @@ export class BundleDataClient {
                   // historical deposit query is not working as expected.
                   assert(this.getRelayHashFromEvent(matchedDeposit) === relayDataHash, "Relay hashes should match.");
                   validatedBundleV3Fills.push({
-                    ...fill,
+                    ...fillToRefund,
                     quoteTimestamp: matchedDeposit.quoteTimestamp,
                   });
                   v3RelayHashes[relayDataHash].fill = fillToRefund;
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -1163,7 +1163,13 @@ export class BundleDataClient {
                 destinationClient.deploymentBlock,
                 destinationClient.latestBlockSearched
               )) as unknown as FillWithBlock;
-              if (canRefundPrefills && !isSlowFill(prefill)) {
+              const verifiedFill = await verifyFillRepayment(
+                prefill,
+                destinationClient.spokePool.provider,
+                deposit,
+                allChainIds
+              );
+              if (canRefundPrefills && isDefined(verifiedFill) && !isSlowFill(prefill)) {
                 validatedBundleV3Fills.push({
                   ...prefill,
                   quoteTimestamp: deposit.quoteTimestamp,
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -918,6 +918,8 @@ export class BundleDataClient {
                     allChainIds
                   );
                   if (!isDefined(fillToRefund)) {
+                    // We won't repay the fill but the depositor has received funds so we don't need to make a
+                    // payment.
                     bundleUnrepayableFillsV3.push(fill);
                     // We don't return here yet because we still need to mark unexecutable slow fill leaves
                     // or duplicate deposits. However, we won't issue a fast fill refund.
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -1184,10 +1184,12 @@ export class BundleDataClient {
                 // If fill is in the current bundle then we can assume there is already a refund for it, so only
                 // include this pre fill if the fill is in an older bundle. If fill is after this current bundle, then
                 // we won't consider it, following the previous treatment of fills after the bundle block range.
-                validatedBundleV3Fills.push({
-                  ...fill,
-                  quoteTimestamp: deposit.quoteTimestamp,
-                });
+                if (!isSlowFill(fill)) {
+                  validatedBundleV3Fills.push({
+                    ...fill,
+                    quoteTimestamp: deposit.quoteTimestamp,
+                  });
+                }
               }
             }
             return;
@@ -1239,7 +1241,7 @@ export class BundleDataClient {
               );
               if (!isDefined(verifiedFill)) {
                 bundleUnrepayableFillsV3.push(prefill!);
-              } else if (canRefundPrefills) {
+              } else if (canRefundPrefills && !isSlowFill(verifiedFill)) {
                 validatedBundleV3Fills.push({
                   ...verifiedFill!,
                   quoteTimestamp: deposit.quoteTimestamp,
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -320,6 +320,12 @@ export class BundleDataClient {
         continue;
       }
       const chainIndex = chainIds.indexOf(chainId);
+
+      // @todo This function does not account for pre-fill refunds as it is optimized for speed. The way to detect
+      // pre-fill refunds is to load all deposits that are unmatched by fills in the spoke pool client's memory
+      // and then query the FillStatus on-chain, but that might slow this function down too much. For now, we
+      // will live with this expected inaccuracy as it should be small. The pre-fill would have to precede the deposit
+      // by more than the caller's event lookback window which is expected to be unlikely.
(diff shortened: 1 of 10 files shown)


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1043,37 +1059,109 @@ export class BundleDataClient {
...
+              return;
+            }
+
+            // If fill does not exist in memory but there is a slow fill request in memory, then we need to issue a
+            // slow fill leaf for the deposit. We can assume there was no fill preceding the slow fill request because
+            // slow fill requests cannot follow fills. If there were a fill following this request, we would have
+            // entered the above case.
+            const slowFillRequest = v3RelayHashes[relayDataHash].slowFillRequest;
+            if (slowFillRequest) {
+              if (!_canCreateSlowFillLeaf(deposit)) {
```
bmzig, line 1123:
```suggestion
              if (_canCreateSlowFillLeaf(deposit)) {
```

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1043,37 +1059,106 @@ export class BundleDataClient {
                 this.getRelayHashFromEvent(matchedDeposit) === relayDataHash,
                 "Deposit relay hashes should match."
               );
+              v3RelayHashes[relayDataHash].deposit = matchedDeposit;
```
nicholaspai, line 1062: Moved this higher up since we're now exiting below with the `_canCreateSlowFillLeaf` check. There is some change here where a deposit is not involved with a lite chain, gets saved into the `v3Relayhashes` dictionary before the code exits because the deposit is a token swap or the deposit has expired. The new behavior is to save the deposit into `v3Relayhashes` dictionary and then exit if any of those checks fail. I don't see any issues with this small change

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1043,37 +1059,106 @@ export class BundleDataClient {
...
+            // to the protocol.
+
+            // If fill exists in memory, then the only case in which we need to create a refund is if the
+            // the fill occurred in a previous bundle.
+            if (fill) {
+              if (!isSlowFill(fill) && fill.blockNumber < destinationChainBlockRange[0]) {
+                // If fill is in the current bundle then we can assume there is already a refund for it, so only
+                // include this pre fill if the fill is in an older bundle. If fill is after this current bundle, then
+                // we won't consider it, following the previous treatment of fills after the bundle block range.
+                validatedBundleV3Fills.push({
```
bmzig, line 1107: I think this algorithm makes sense. Do you think it is worthwhile to add a log here, at least temporarily, so that we can verify that we aren't double counting deposits (as a sanity check). We would also be able to do a pre-fill and make sure it is included by this code block.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1043,37 +1059,106 @@ export class BundleDataClient {
...
+            if (fillStatus === FillStatus.Filled) {
+              // We need to find the fill event to issue a refund to the right relayer and repayment chain,
+              // or msg.sender if relayer address is invalid for the repayment chain.
+              const prefill = (await findFillEvent(
+                destinationClient.spokePool,
+                deposit,
+                destinationClient.deploymentBlock,
+                destinationClient.latestBlockSearched
+              )) as unknown as FillWithBlock;
+              if (!isSlowFill(prefill)) {
```
bmzig, line 1144: Maybe a log here as well. We'd definitely have to remove them once pre-fills become common, though.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: I added unit tests here to check against this case: https://github.com/across-protocol/relayer/pull/2010 which I prefer over logs. Do you think duplicate refunds is a big risk based on the implementation?

bmzig: No not really. I can't think of a case where we would duplicate refund, but I just suggested it in case I am missing one. We can just alternatively run this locally in parallel with the "older" dataworker implementation and verify that the roots match.

nicholaspai: I was able to add a unit test here: b099cf0e6f910152b60c7dffd0618eae25a039fa

note that this test depends now on these SDK changes to support duplicate deposits https://github.com/across-protocol/sdk/pull/838

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: again i'm worried about adding a lot of noise. Imagine we enter here in a bundle, won't it get very noisy for that whole bundle?

bmzig: Yeah, it would have to be a debug log, and it would also have to be temporary so that once we actually do the rollout and pre-fills start happening, we can remove it. The idea was have it and make sure it never logged, but there are better ways to verify that this conditional never triggers now that I'm thinking about it.

nicholaspai: what kind of log are you proposing? or what are the better ways of verifying?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1119,23 +1204,11 @@ export class BundleDataClient {
...
-    });
-
-    // For all deposits older than this bundle, we need to check if they expired in this bundle and if they did,
-    // whether there was a slow fill created for it in a previous bundle that is now unexecutable and replaced
-    // by a new expired deposit refund.
-    await forEachAsync(Array.from(olderDepositHashes), async (relayDataHash) => {
+    // Add any newly expired deposits to the list of expired deposits to refund.
+    // For these refunds, we need to check whether there was a slow fill created for it in a previous bundle
+    // that is now unexecutable and replaced by a new expired deposit refund.
+    const possibleExpiredDeposits = Array.from(olderDepositHashes).concat(Array.from(expiredBundleDepositHashes));
```
nicholaspai, line 1210: We now need to treat `expiredBundleDepositHashes` (deposits in this bundle that expired in this bundle) the same as `olderDepositHashes` (deposits from older bundles that expired in this bundle) due to the presence of pre-fills.  This is because a deposit can now be sent so much after a pre-fill that we can never be sure if a fill was sent for a deposit without checking the fillStatuses on-chain mapping. And we never want to send an expired deposit refund for a deposit that was already filled.

This is kind of nice consolidation actually.

nicholaspai: Actually I think this is wrong because this would mean for a newly expired deposit submitted in the current bundle, we can create an `unexecutableSlowFill` for it without creating a slow fill leaf

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1119,23 +1158,10 @@ export class BundleDataClient {
     });
     start = performance.now();
 
-    // Go through expired deposits in this bundle and now prune those that we have seen a fill for to construct
-    // the list of expired deposits we need to refund in this bundle.
-    expiredBundleDepositHashes.forEach((relayDataHash) => {
```
nicholaspai, line 1124: We can no longer handle expired deposits here because we need to check that the fill status for these deposits is not Filled already

src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -122,7 +122,6 @@ export class MockSpokePoolClient extends SpokePoolClient {
     const { blockNumber, transactionIndex } = deposit;
     let { depositId, depositor, destinationChainId, inputToken, inputAmount, outputToken, outputAmount } = deposit;
     depositId ??= this.numberOfDeposits;
-    assert(depositId >= this.numberOfDeposits, `${depositId} < ${this.numberOfDeposits}`);
```
nicholaspai, line 125: Needed to remove this so I can create dupe deposits

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: Scratch the log. The better way to verify would be to run the this new dataworker in parallel with our current dataworker and verify that they produce the same bundles.

src/clients/BundleDataClient/BundleDataClient.ts
bmzig, line 1: Tangential but I believe the filter here https://github.com/across-protocol/sdk/blob/d8a1881440930536850f16ffd28844837cd56828/src/clients/BundleDataClient/BundleDataClient.ts#L823 to only look at fills which are not zero value won't work properly upon upgrading, since zero value fill events won't have a message, and the messageHash will look different than "" or "0x". We probably need another function like `isZeroValueFill` or a modification to `isZeroValueDeposit`

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1043,37 +992,127 @@ export class BundleDataClient {
...
+            // If a slow fill request exists in memory, then we know the deposit has not been filled because fills
+            // must follow slow fill requests and we would have seen the fill already if it existed. Therefore,
+            // we can conclude that either the deposit has expired and we need to create a deposit expiry refund, or
+            // we need to create a slow fill leaf for the deposit. The latter should only happen if the slow fill request
+            // took place in a prior bundle otherwise we would have already created a slow fill leaf for it.
+            if (slowFillRequest) {
+              if (deposit.fillDeadline < bundleBlockTimestamps[destinationChainId][1]) {
+                updateExpiredDepositsV3(expiredDepositsToRefundV3, deposit);
+              } else if (
+                _canCreateSlowFillLeaf(deposit) &&
```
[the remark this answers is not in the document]
nicholaspai: > Suppose I made a pre flow fill at block 90.

Did you mean to write pre slow fill request? If so, then yes I agree with this statement:

> This means we add two slow fills to validatedBundleSlowFills. This should be fine, right?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1154,16 +1180,7 @@ export class BundleDataClient {
...
-          // hasn't queried. This is because this function will usually be called
-          // in production with block ranges that were validated by
-          // DataworkerUtils.blockRangesAreInvalidForSpokeClients
-          Math.min(destinationBlockRange[1], spokePoolClients[destinationChainId].latestBlockSearched),
-          destinationChainId
-        );
+        const fillStatus = await _getFillStatusForDeposit(deposit, destinationBlockRange[1]);
 
         // If there is no matching fill and the deposit expired in this bundle and the fill status on-chain is not
         // Filled, then we can to refund it as an expired deposit.
```
bmzig, line 1186: It looks like if I have two normal duplicate deposits which end up in `olderDepositHashes` and expire, I will get refunded for both deposits. Does that sound right?

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: what would the hash be? If the message is 0x, the message hash will be keccak256(0x) which we can probably hardcode?

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: yep! that's my intention at least.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: also, are message hashes or messages emitted in RequestedSlowFill events?

bmzig: Ah the message hash is emitted there as well. https://github.com/across-protocol/contracts/blob/dd88c3c4af6396c83759e0f3f9063ead8d3fd699/contracts/SpokePool.sol#L1112

nicholaspai: yeah so if we leave this code in today, it means that zero value fills and slow fill requests could pass through this filter, but there won't be any ill effects. I also think something like this could fix it, wdyt? 719c29c67591a3f0d043f5e27a3389afcda27a40

src/clients/BundleDataClient/BundleDataClient.ts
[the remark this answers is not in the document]
nicholaspai: yeah I think that's the intention. If we want to prevent these I think we'd have to query for all historical duplicate deposits to be certain this is the first time we're creating a leaf. But that seems expensive

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: I agree it has no effect today. I did do this exact change in a previous PR, but the issue is we can mistakenly mark a (malformed) message as an empty message. Technically somebody can make a zero-value deposit with 0x0000...000 as their message, and then we will ignore it if somebody fills it. Maybe this is just something we change on the contract bump, though.

nicholaspai: so if a deposit has `inputAmount=0` and `message=0x0000...000`, then what's the harm in ignoring it in the bundle data client only? Maybe this commit might be safer so we don't change any relayer code unintentionally b4d85d756efb073eccb01dc13acd0760b414f3f6

bmzig: I think that's good. And then for backwards compatibility, the fill check would also need to do something like
```
export function isFillOrSlowFillRequestMessageEmpty(message: string): boolean {
  return isMessageEmpty(message) || message === EMPTY_MESSAGE_HASH;
}
```

nicholaspai: > Technically somebody can make a zero-value deposit with 0x0000...000 as their message, and then we will ignore it if somebody fills it.

I think this is fine as far as the BundleDataClient is concerned, because it only cares about issuing fill refunds (for which there will be none because inputAmount=0) and slow fill leaves (this will also be useless since the message will do nothing except for some very weird edge case that I think we can ignore because this will go away once we make the switch to emitting message hashes in slow fill requests)

nicholaspai: @bmzig wdyt about this change? https://github.com/across-protocol/sdk/pull/835/commits/b4d85d756efb073eccb01dc13acd0760b414f3f6 This should leave relayer behavior unchanged

bmzig: You need to also check if the message is "" or "0x", right? Otherwise you would have to time this merge with the contract upgrade?

nicholaspai: oh true! e6f499615b2301f6f34a008a385bf2e21736fb4c

nicholaspai (commented):
I'm not sure the best way to merge this PR. Any bundles produced by this SDK will likely disagree with any existing bundles and therefore could lead to a self-dispute. This is because each bundle has a decent chance empirically of having a deposit slightly after a bundle block range matching a fill at the end of the range. One way to do this is to make sure to merge this in when there are no bundles pending and propose the next bundle with this code.

src/utils/SpokeUtils.ts
```diff
@@ -389,6 +390,40 @@ export async function findFillBlock(
...
+      maxBlockLookBack: 0, // We can hardcode this to 0 to instruct paginatedEventQuery to make a single request
+      // for the same block number.
+    }
+  );
+  if (query.length === 0) throw new Error(`Failed to find fill event at block ${blockNumber}`);
+  const event = query[0];
+  // In production the chainId returned from the provider matches 1:1 with the actual chainId. Querying the provider
+  // object saves an RPC query becasue the chainId is cached by StaticJsonRpcProvider instances. In hre, the SpokePool
+  // may be configured with a different chainId than what is returned by the provider.
+  // @todo Sub out actual chain IDs w/ CHAIN_IDs constants
```
bmzig, line 416: You can probably remove this todo comment.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1043,37 +1002,127 @@ export class BundleDataClient {
...
+        const originBlockRange = getBlockRangeForChain(blockRangesForChains, originChainId, chainIds);
+
+        await mapAsync(
+          bundleDepositHashes.filter((depositHash) => {
+            const { deposit } = v3RelayHashes[depositHash];
+            return (
+              deposit &&
+              deposit.originChainId === originChainId &&
+              deposit.destinationChainId === destinationChainId &&
+              deposit.blockNumber >= originBlockRange[0] &&
```
bmzig, line 1035: These last three checks about the block ranges and zero value deposits are redundant since we check them when adding to `bundleDepositHashes`. Do we want to keep them for the sake of being extra safe (e.g. if we change how we add stuff to `bundleDepositHashes` in the future)?

bmzig (approved):
Do we want to activate this before or after this PR: https://github.com/across-protocol/sdk/pull/831/files? I'd prefer activating this before, since this is significantly more complex.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: good call will remove

bmzig:
> > Do we want to activate this before or after this PR: https://github.com/across-protocol/sdk/pull/831/files? I'd prefer activating this before, since this is significantly more complex.
> 
> ~yeah definitely before IMO~ Actually, I think there's merit to merging in the less complex code, letting it run and making sure no regressions, so that when the more complex code goes in, we can isolate any issues with high probability to the more complex code

This makes sense, but the invalid fills PR should only be a pass through for right now. That is, none of the new conditionals will be triggered since all addresses are bytes20 addresses and all chains are EVM.

nicholaspai:
i've verified that the bundles produced by this PR match exactly the existing bundles except that there are a handful of invalid fills very close to the destination chain bundle end blocks in this PR's bundle. These are fills that are technically now "pre-fills" that should be refunded once we bump the VERSION>=5 and propose a following bundle

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1067,37 +1047,124 @@ export class BundleDataClient {
...
+        // should remove this check once we've advanced far beyond the version bump block.
+        await mapAsync(
+          bundleDepositHashes.filter((depositHash) => {
+            const { deposit } = v3RelayHashes[depositHash];
+            return (
+              deposit && deposit.originChainId === originChainId && deposit.destinationChainId === destinationChainId
+            );
+          }),
+          async (depositHash) => {
+            const { deposit, fill, slowFillRequest } = v3RelayHashes[depositHash];
```
nicholaspai, line 1082: We don't 
```suggestion
            // We don't need to call verifyFillRepayment() here to replace the fill.relayer because this value should already
            // be overwritten because the deposit and fill both exist.
            const { deposit, fill, slowFillRequest } = v3RelayHashes[depositHash];
```

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -804,24 +809,24 @@ export class BundleDataClient {
...
-            // gives us bad data.
-            return;
+
+          // Duplicate deposits are treated like normal deposits.
+          if (!v3RelayHashes[relayDataHash]) {
+            v3RelayHashes[relayDataHash] = {
+              deposit: deposit,
+              fill: undefined,
+              slowFillRequest: undefined,
+            };
```
melisaguevara, line 828: should deposit key here be an array?
Or how are we storing duplicated deposits in v3RelayHashes?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -804,24 +809,24 @@ export class BundleDataClient {
...
           }
           depositCounter++;
           const relayDataHash = this.getRelayHashFromEvent(deposit);
-          if (v3RelayHashes[relayDataHash]) {
-            // If we've seen this deposit before, then skip this deposit. This can happen if our RPC provider
-            // gives us bad data.
-            return;
+
+          // Duplicate deposits are treated like normal deposits.
+          if (!v3RelayHashes[relayDataHash]) {
```
nicholaspai, line 823: we can probably just remove this if statement since its ok to overwrite the deposit. Currently it'll always get set to the oldest (the first one returned by getDepositsForDestinationChainWithDuplicates() and I also think that's ok

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: We're not storing duplicates in v3RelayHashes, but we are storing them in `bundleDepositsV3` via `updateBundleDepositsV3(bundleDepositsV3, deposit)` a bit lower down. The thinking is we don't actually need the duplicate deposits in the v3RelayHashes dictionary for any other logic. You can also see how we can push duplicate deposit hashes to `bundleDepositHashes` and `olderDepositHashes` to make sure we send duplicate expiry refunds. What do you think?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -831,24 +836,13 @@ export class BundleDataClient {
...
-            // bundle of refunded deposits.
-            if (deposit.fillDeadline < bundleBlockTimestamps[destinationChainId][1]) {
-              expiredBundleDepositHashes.add(relayDataHash);
-            }
           } else if (deposit.blockNumber < originChainBlockRange[0]) {
-            olderDepositHashes.add(relayDataHash);
+            olderDepositHashes.push(relayDataHash);
           }
         });
       }
```
bmzig, line 848: Line 880 needs to be 
```
.filter((fill) => fill.blockNumber <= destinationChainBlockRange[1] && !isZeroValueFillOrSlowFillRequest(fill)),
```
I believe.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1095,37 +1072,132 @@ export class BundleDataClient {
...
+              )) as unknown as FillWithBlock;
+              const verifiedFill = await verifyFillRepayment(
+                prefill,
+                destinationClient.spokePool.provider,
+                deposit,
+                allChainIds
+              );
+              if (canRefundPrefills && isDefined(verifiedFill) && !isSlowFill(prefill)) {
+                validatedBundleV3Fills.push({
+                  ...prefill,
```
bmzig, line 1176: I think you need to push verifiedFill here in case the relayer repayment address changed.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: yep this got reverted due to a merge conflict, i've added back

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: yep realized this while adding a unit test

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1200,22 +1250,13 @@ export class BundleDataClient {
         // If there is a valid fill that we saw matching this deposit, then it does not need a refund.
         !fill &&
         isDefined(deposit) && // Needed for TSC - we check this above.
-        deposit.fillDeadline < bundleBlockTimestamps[destinationChainId][1] &&
+        _depositIsExpired(deposit) &&
```
bmzig, line 1256: I think we might miss some deposits here. Suppose I deposit w/ origin blockNumber 20 and fillDeadline on destination 2003. A bundle is proposed and it looks like:
```
origin: [0, 100] => (timestamps) [0, 2000]
destination: [0, 100] => (timestamps) [0, 2000]
```
We enter the pre-fill loop with this deposit since it is in between the origin block ranges, but `_depositIsExpired` is false, since the timestamp for the end block range on destination is less than the fill deadline.
On the next bundle iteration, we propose:
```
origin: [101, 200] => (timestamps) [2010, 4000]
destination: [101, 200] => (timestamps) [2010, 4000]
```
The destination chain timestamps are > 2003 since the user can choose a fillDeadline, while the timestamps here are taken from the block timestamp.
The deposit is not evaluated in `olderDepositHashes` here, it is expired according to `_depositIsExpired`, but it is _not_ greater than `bundleBlockTimestamps[destinationChainId][0]`, so it isn't refunded. 

I have to be wrong here somewhere since I can see that functionally, we do the same refund process as before, so what am I missing?

nicholaspai: I think its impossible (improbable?) that the timestamp would shift from 2000 at the end of bundle 1 to 2010 at the beginning of bundle 2. The timestamps are queried [here](https://github.com/across-protocol/sdk/blob/57ce0138c77721ccde3c65ad3be99b13fe69e295/src/clients/BundleDataClient/BundleDataClient.ts#L1401). Why would the timestamp jump so much from block 100 to 101?

bmzig: It's probably improbably but more likely on chains like ethereum with longer block times. Definitely would be much more unlikely to happen on most L2s

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1362,14 +1423,26 @@ export class BundleDataClient {
   // keccak256 hash of the relay data, which can be used as input into the on-chain `fillStatuses()` function in the
   // spoke pool contract. However, this internal function is used to uniquely identify a bridging event
   // for speed since its easier to build a string from the event data than to hash it.
-  private getRelayHashFromEvent(event: V3DepositWithBlock | V3FillWithBlock | SlowFillRequestWithBlock): string {
+  protected getRelayHashFromEvent(event: V3DepositWithBlock | V3FillWithBlock | SlowFillRequestWithBlock): string {
```
bmzig, line 1426: Calls to this will break once `V3FillWithBlock` has a `messageHash` instead of a message. I think you should hash the message if it is a deposit, and then just swap `event.message ` with `event.messageHash` here.

nicholaspai: Can we do this in a follow up? Otherwise this will break until we upgrade the contracts on-chain

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1106,72 +1114,186 @@ export class BundleDataClient {
...
+          if (fill) {
+            if (!isDuplicateDepositInBundle && fill.blockNumber < destinationChainBlockRange[0]) {
+              duplicateDepositsInBundle.forEach((duplicateDeposit) => {
+                updateExpiredDepositsV3(expiredDepositsToRefundV3, duplicateDeposit);
+              });
+              if (canRefundPrefills) {
+                // If fill is in the current bundle then we can assume there is already a refund for it, so only
+                // include this pre fill if the fill is in an older bundle. If fill is after this current bundle, then
+                // we won't consider it, following the previous treatment of fills after the bundle block range.
+                if (!isSlowFill(fill)) {
```
bmzig, line 1187: Is this check necessary/Is it just to not bloat validatedBundleV3Fills? I notice that for the normal fill case, we do not check do this check before pushing to validatedBundleV3Fills since when we are modifying the refund dictionary, we exit early if the fill was a slow fill. (same question for this check on line 1244).

nicholaspai: Good catch! I agree with you

nicholaspai: https://github.com/across-protocol/sdk/pull/835/commits/dc6d4826d0fa465501113d022b73ba2727f5e9d8

nicholaspai: This was actually a crucial bug fix, because even slow fill executions add to the LP fees. This is how LP's get paid when slow leaves get executed and the LP fee is subtracted from the recipient's inputAmount. https://github.com/across-protocol/sdk/blob/13920ee180aa5946502cc27bbd2beeaf0046d1e3/src/clients/BundleDataClient/BundleDataClient.ts#L107

nicholaspai: Hm actually I'm thinking about this more we may need to keep these checks in. When a fill is a slow fill execution normally we'd want to charge lp fees on them. However if a slow fill leaf is being treated as a pre fill execution then we'd be charging lp fees a second time. This is because when the leaf execution originally happened, lp fees must have been charged. Because that fill would always be able to be matched with a deposit.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: https://github.com/across-protocol/sdk/pull/850/files should handle this

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -911,7 +924,7 @@ export class BundleDataClient {
                     v3RelayHashes[relayDataHash].fill = fillToRefund;
```
bmzig, line 924: What if I use unsafeDeposit to make a two, duplicate, normal deposits. I then fill one of them. It looks like I will be refunded once here, and then my second deposit is absorbed into the LP fund. Do we need to basically mimic a duplicate fill and push multiple, identical fills into `validatedBundleV3Fills`?

nicholaspai: hmm good point. We probably should add a check here to refund all duplicate deposits in the bundle once we have seen the fill. I think we can add back this logic here: https://github.com/across-protocol/sdk/pull/835/commits/e5cd4189c7172309b36cf608489ef981f9be4283

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -883,35 +895,60 @@ export class BundleDataClient {
...
+                    // a current bundle fill. If the fill is a slow fill,
+                    const duplicateDeposits = v3RelayHashes[relayDataHash].deposits!.slice(1);
+                    duplicateDeposits.forEach((duplicateDeposit) => {
+                      // If fill is a slow fill, refund deposit to depositor, otherwise refund to filler.
+                      if (isSlowFill(fill)) {
+                        updateExpiredDepositsV3(expiredDepositsToRefundV3, duplicateDeposit);
+                      } else {
+                        validatedBundleV3Fills.push({
+                          ...fillToRefund,
+                          quoteTimestamp: duplicateDeposit.quoteTimestamp,
```
nicholaspai, line 949: Something I thought about that might not be a big deal: the LP fee charged on these duplicate fill refunds is based on the duplicate deposit's quote timestamp. Quote timestamps can differ between duplicate deposits because they are not part of the relay data hash. I think this is OK but want to flag it for careful review

mrice32: Yeah, I think this works.

We just need to make sure we are consistent about _which_ deposit is used to determine the slow fill payout amount (I think it would have to be the first).

nicholaspai: yep. so do you agree then:
- lp fees charged on fill refunds arising from duplicate deposits are based on the duplicate deposit's quote timestamp
- lp fees for slow fills match only the first deposit

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1109,72 +1144,174 @@ export class BundleDataClient {
...
 
+        // Deposits can be submitted an arbitrary amount of time after matching fills and slow fill requests.
+        // Therefore, let's go through each deposit in this bundle again and check a few things in order:
+        // - Has the deposit been filled ? If so, then we need to issue a relayer refund for
+        //   this "pre-fill" if the fill took place in a previous bundle.
+        // - Or, has the deposit expired in this bundle? If so, then we need to issue an expiry refund.
+        // - And finally, has the deposit been slow filled? If so, then we need to issue a slow fill leaf
+        //   for this "pre-slow-fill-request" if this request took place in a previous bundle.
+        await mapAsync(bundleDepositHashes, async (depositHash) => {
+          // We don't need to call verifyFillRepayment() here to replace the fill.relayer because this value should already
```
bmzig, line 1169: I don't think this comment is correct. I think you do need to call `verifyFillRepayment` here since I can have an invalid pre-fill. E.g., I pre-fill in bundle 1 with invalid parameters, I deposit in bundle 2. The pre-fill is never overwritten to the correct refund parameters since it is now outside the bundle block ranges. https://github.com/across-protocol/sdk/blob/f853323fd684b11f4e9bc5efefad2fb5fa77fb9b/src/clients/BundleDataClient/BundleDataClient.ts#L918

bmzig (approved):
One comment.

bmzig (approved):

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: This is a good point. What do you think of this: 87cb051462af1f0ed36d04f5f6338ccbb7bc8c41 (unit tests are added [here](https://github.com/across-protocol/relayer/pull/2010/commits/65791812f6b407c402aaf3b14c18a43e14f18fa6))

bmzig (approved):

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -139,9 +147,10 @@ function updateBundleExcessSlowFills(
 }
 
 function updateBundleSlowFills(dict: BundleSlowFills, deposit: V3DepositWithBlock & { lpFeePct: BigNumber }): void {
-  if (chainIsEvm(deposit.destinationChainId) && !isValidEvmAddress(deposit.recipient)) {
-    return;
-  }
+  assert(
+    chainIsEvm(deposit.destinationChainId) && isValidEvmAddress(deposit.recipient),
```
bmzig, line 151: If you are adding this assert, I believe you will also need to change `_canCreateSlowFillLeaf` to check for a valid recipient since somebody can supply garbage to a deposit and request a slow fill for that.

nicholaspai: This is a good point I'm going to revert 5c32bd80eafa3faf05788d2718a09766dddbc7a9

nicholaspai:
@bmzig I added bundleInvalidSlowFIllRequests logging here https://github.com/across-protocol/sdk/pull/835/commits/cc9ebfcfc70b3bc221ee3c0b59c62f17aa1372a2

bmzig (approved):

PR #836, opened 2025-01-23
feat: Add findFillEvent utility function

nicholaspai:
Builds off of the existing `findFillBlock()` function. We'll need this function in the BundleDataClient to create refunds for pre-fills

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -1088,7 +1088,7 @@ export class BundleDataClient {
               deposit &&
               deposit.originChainId === originChainId &&
               deposit.destinationChainId === destinationChainId &&
-              deposit.blockNumber <= originBlockRange[0] &&
+              deposit.blockNumber >= originBlockRange[0] &&
               !isZeroValueDeposit(deposit)
           ),
           async ({ deposit, fill, slowFillRequest }) => {
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -1117,9 +1117,10 @@ export class BundleDataClient {
             // If fill does not exist in memory but there is a slow fill request in memory, then we need to issue a
             // slow fill leaf for the deposit. We can assume there was no fill preceding the slow fill request because
             // slow fill requests cannot follow fills. If there were a fill following this request, we would have
-            // entered the above case.
+            // entered the above case. Again as with pre-fills, we should only consider slow fill requests that were
+            // in previous bundles.
             if (slowFillRequest) {
-              if (_canCreateSlowFillLeaf(deposit)) {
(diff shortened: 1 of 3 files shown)


bmzig (approved):

src/utils/SpokeUtils.ts
```diff
@@ -380,3 +381,36 @@ export async function findFillBlock(
...
+  if (!blockNumber) return undefined;
+  const query = await paginatedEventQuery(
+    spokePool,
+    spokePool.filters.FilledV3Relay(null, null, null, null, null, relayData.originChainId, relayData.depositId),
+    {
+      fromBlock: blockNumber,
+      toBlock: blockNumber,
+      maxBlockLookBack: 1, // We should be able to hardcode this since we're only looking on a single block.
+    }
+  );
```
pxrl, line 401: @nicholaspai We'll soon need to have a parallel instance of this that queries for `FilledRelay`. However, we could alternatively use [eth_getBlockReceipts()](https://docs.alchemy.com/reference/eth-getblockreceipts) and pull down all the transaction receipts for the block. We'd then iterate over that and pull out our event.

The RPC cost if `eth_getBlockReceipts` is 20 CUs w/ Alchemy, whereas `getLogs` is currently 60. It's probably not that important in the scheme of things because we shouldn't be using this very often, but the advantage is that we don't need 2 RPC calls with `eth_getBlockReceipts`. Just something to think about for a possible improvement.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1088,7 +1088,7 @@ export class BundleDataClient {
               deposit &&
               deposit.originChainId === originChainId &&
               deposit.destinationChainId === destinationChainId &&
-              deposit.blockNumber <= originBlockRange[0] &&
+              deposit.blockNumber >= originBlockRange[0] &&
```
pxrl, line 1091: Is there a brief tldr on why this is needed btw?

nicholaspai: Oh sorry. I should move this commit to parent branch

pxrl (approved):

src/utils/SpokeUtils.ts
nicholaspai: I tried looking into this and unfortunately ethers/hardhat doesn't yet support this endpoint: https://github.com/ethers-io/ethers.js/pull/4769/files

I also explored calling eth_getLogs directly but its a bit convoluted to construct topics the right way and then to parse the logs, which ethers [abstracts](https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/contracts/src.ts/index.ts#L1054) away for us. I'm inclined to leave this be for now out of code simplicity even though we might incur one extra RPC call per situation.

The good thing is for queries on older fills, this result will be cached after one request, and this should be rare in general assuming the prefills don't precede deposits by more than an hour that frequently. I could see this changing though...

PR #838, opened 2025-01-23
feat(BundleDataClient): Support duplicate expired deposit refunds

nicholaspai:
I think these are the changes needed but I need to consider more the impact if two deposits expire and they both create slow fill excesses, for example, or other weird edge cases.


nicholaspai:
Hm actually, there are more fundamental changes required because the SpokePoolClient assumes deposit hashes are unique: https://github.com/across-protocol/sdk/blob/db244fbaec15dc049452cd104ca7bec5856ad3da/src/clients/SpokePoolClient.ts#L577

PR #841, opened 2025-01-27
improve: More BN .toString()'s

nicholaspai:
Should always log BN's as strings. Not sure if this happens implicitly or not

### DIFF
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -140,7 +140,7 @@ export async function getBlockRangeForDepositId(
     //                                     // targetId = 10  <- fail (triggers this error)          // 10 <= 10
     //                                     // targetId = 09  <- pass (does not trigger this error)  // 10 <= 09
     throw new Error(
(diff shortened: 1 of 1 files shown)


nicholaspai:
> @nicholaspai It does happen implicitly when it's logged as a variable in a string, so we should be fine without these changes. The only place where we need `BigNumber.toString()` should be when it gets passed directly or indirectly to `JSON.stringify()`, because `JSON.stringify()` will also dump the raw JSON object and not its BigNumber string representation.
> 
> ```
> > y = ethers.BigNumber.from(100)
> BigNumber { _hex: '0x64', _isBigNumber: true }
> 
> # Dump the raw object.
> > y
> BigNumber { _hex: '0x64', _isBigNumber: true }
> 
> # Dump the raw object via console.log(...).
> > console.log(y)
> BigNumber { _hex: '0x64', _isBigNumber: true }
> 
> # Stringify the BigNumber.
> > console.log(`${y}`)
> 100
> ```

i still think it's easier to just be safe and explicitly stringify this, no?

PR #842, opened 2025-01-27
fix: check for valid relayer address in all cases

bmzig:
The last PR did not account for the normal case: when there is a deposit and fill in the block range, but the fill's relayer is invalid.

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -896,14 +896,15 @@ export class BundleDataClient {
                   isDefined(v3RelayHashes[relayDataHash].deposit),
                   "Deposit should exist in relay hash dictionary."
                 );
+                // At this point, the v3RelayHashes entry already existed meaning that there is a matching deposit,
+                // so this fill is validated.
+                v3RelayHashes[relayDataHash].fill = fill;
+
                 // If the fill is invalid due to relayer repayment addresses, return early.
                 const fillHasValidRepaymentAddress = await verifyFill(fill, destinationClient);
                 if (!fillHasValidRepaymentAddress) {
                   return;
                 }
-                // At this point, the v3RelayHashes entry already existed meaning that there is a matching deposit,
-                // so this fill is validated.
-                v3RelayHashes[relayDataHash].fill = fill;
                 if (fill.blockNumber >= destinationChainBlockRange[0]) {
                   validatedBundleV3Fills.push({
                     ...fill,
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -14,6 +14,7 @@ import {
   ExpiredDepositsToRefundV3,
   Clients,
   CombinedRefunds,
+  FillWithBlock,
 } from "../../interfaces";
 import { AcrossConfigStoreClient, SpokePoolClient } from "..";
 import {
@@ -686,6 +687,23 @@ export class BundleDataClient {
       }
     });
 
+    // Verify that a fill sent to an EVM chain has a 20 byte address. If the fill does not, then attempt
+    // to repay the `msg.sender` of the relay transaction. Otherwise, add it to invalid fills.
+    const verifyFill = async (fill: FillWithBlock, spokePoolClient: SpokePoolClient) => {
+      if (chainIsEvm(fill.repaymentChainId) && !isValidEvmAddress(fill.relayer)) {
+        const fillTransaction = await spokePoolClient.spokePool.provider.getTransaction(fill.transactionHash);
+        const originRelayer = fillTransaction.from;
+        // Repayment chain is still an EVM chain, but the msg.sender is a bytes32 address, so the fill is invalid.
+        if (!isValidEvmAddress(originRelayer)) {
+          bundleInvalidFillsV3.push(fill);
+          return false;
+        }
+        // Otherwise, assume the relayer to be repaid is the msg.sender.
+        fill.relayer = originRelayer;
+      }
+      return true;
+    };
+
     // If spoke pools are V3 contracts, then we need to compute start and end timestamps for block ranges to
     // determine whether fillDeadlines have expired.
     // @dev Going to leave this in so we can see impact on run-time in prod. This makes (allChainIds.length * 2) RPC
@@ -878,17 +896,9 @@ export class BundleDataClient {
                   isDefined(v3RelayHashes[relayDataHash].deposit),
                   "Deposit should exist in relay hash dictionary."
                 );
-                // If the fill's repayment address is not a valid EVM address and the repayment chain is an EVM chain, the fill is invalid.
-                if (chainIsEvm(fill.repaymentChainId) && !isValidEvmAddress(fill.relayer)) {
-                  const fillTransaction = await originClient.spokePool.provider.getTransaction(fill.transactionHash);
-                  const originRelayer = fillTransaction.from;
-                  // Repayment chain is still an EVM chain, but the msg.sender is a bytes32 address, so the fill is invalid.
-                  if (!isValidEvmAddress(originRelayer)) {
-                    bundleInvalidFillsV3.push(fill);
-                    return;
-                  }
-                  // Otherwise, assume the relayer to be repaid is the msg.sender.
-                  fill.relayer = originRelayer;
+                // If the fill is invalid due to relayer repayment addresses, return early.
+                if (!verifyFill(fill, originClient)) {
+                  return;
                 }
                 // At this point, the v3RelayHashes entry already existed meaning that there is a matching deposit,
                 // so this fill is validated.
@@ -939,16 +949,8 @@ export class BundleDataClient {
                 return;
               }
               // If the fill's repayment address is not a valid EVM address and the repayment chain is an EVM chain, the fill is invalid.
-              if (chainIsEvm(fill.repaymentChainId) && !isValidEvmAddress(fill.relayer)) {
-                const fillTransaction = await originClient.spokePool.provider.getTransaction(fill.transactionHash);
-                const originRelayer = fillTransaction.from;
-                // Repayment chain is still an EVM chain, but the msg.sender is a bytes32 address, so the fill is invalid.
-                if (!isValidEvmAddress(originRelayer)) {
-                  bundleInvalidFillsV3.push(fill);
-                  return;
-                }
-                // Otherwise, assume the relayer to be repaid is the msg.sender.
-                fill.relayer = originRelayer;
+              if (!verifyFill(fill, originClient)) {
+                return;
               }
               // If deposit is using the deterministic relay hash feature, then the following binary search-based
               // algorithm will not work. However, it is impossible to emit an infinite fill deadline using
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -897,7 +897,8 @@ export class BundleDataClient {
                   "Deposit should exist in relay hash dictionary."
                 );
                 // If the fill is invalid due to relayer repayment addresses, return early.
-                if (!(await verifyFill(fill, destinationClient))) {
+                const fillHasValidRepaymentAddress = await verifyFill(fill, destinationClient);
+                if (!fillHasValidRepaymentAddress) {
                   return;
                 }
                 // At this point, the v3RelayHashes entry already existed meaning that there is a matching deposit,
@@ -949,7 +950,8 @@ export class BundleDataClient {
                 return;
               }
               // If the fill's repayment address is not a valid EVM address and the repayment chain is an EVM chain, the fill is invalid.
-              if (!(await verifyFill(fill, destinationClient))) {
+              const fillHasValidRepaymentAddress = await verifyFill(fill, destinationClient);
+              if (!fillHasValidRepaymentAddress) {
                 return;
               }
               // If deposit is using the deterministic relay hash feature, then the following binary search-based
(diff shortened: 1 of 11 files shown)


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -878,6 +878,18 @@ export class BundleDataClient {
                   isDefined(v3RelayHashes[relayDataHash].deposit),
                   "Deposit should exist in relay hash dictionary."
                 );
+                // If the fill's repayment address is not a valid EVM address and the repayment chain is an EVM chain, the fill is invalid.
+                if (chainIsEvm(fill.repaymentChainId) && !isValidEvmAddress(fill.relayer)) {
```
nicholaspai, line 882: Can we refactor this check into a helper function since its code is basically re-used in the historical deposit case?

nicholaspai: Also, do you think we can move this code into `updateBundleFillsV3`? We can still update invalid fills from there

bmzig: Refactored here: https://github.com/across-protocol/sdk/pull/842/commits/f648d61d360d0c80b0d3e4c3771fb1e15fc1a987

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -927,16 +949,8 @@ export class BundleDataClient {
...
-                const fillTransaction = await originClient.spokePool.provider.getTransaction(fill.transactionHash);
-                const originRelayer = fillTransaction.from;
-                // Repayment chain is still an EVM chain, but the msg.sender is a bytes32 address, so the fill is invalid.
-                if (!isValidEvmAddress(originRelayer)) {
-                  bundleInvalidFillsV3.push(fill);
-                  return;
-                }
-                // Otherwise, assume the relayer to be repaid is the msg.sender.
-                fill.relayer = originRelayer;
+              if (!verifyFill(fill, originClient)) {
```
nicholaspai, line 952: same needs to be await'd

nicholaspai (approved):

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -878,6 +896,10 @@ export class BundleDataClient {
                   isDefined(v3RelayHashes[relayDataHash].deposit),
                   "Deposit should exist in relay hash dictionary."
                 );
+                // If the fill is invalid due to relayer repayment addresses, return early.
+                if (!verifyFill(fill, originClient)) {
```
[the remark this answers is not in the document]
bmzig: Thanks. Good catch. https://github.com/across-protocol/sdk/pull/842/commits/a7e77d2a9ac6e47b73b08f1e8c5c56c001b3fe13

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: https://github.com/across-protocol/sdk/pull/842/commits/a7e77d2a9ac6e47b73b08f1e8c5c56c001b3fe13

nicholaspai (approved):

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -686,6 +687,23 @@ export class BundleDataClient {
       }
     });
 
+    // Verify that a fill sent to an EVM chain has a 20 byte address. If the fill does not, then attempt
+    // to repay the `msg.sender` of the relay transaction. Otherwise, add it to invalid fills.
+    const verifyFill = async (fill: FillWithBlock, spokePoolClient: SpokePoolClient) => {
```
pxrl, line 692: We should also verify here that `repaymentChainId` is valid - i.e. is it a currently supported chain?

nicholaspai: Could also move down here https://github.com/across-protocol/sdk/blob/master/src/clients/BundleDataClient/BundleDataClient.ts#L1282

nicholaspai: Alternatively, we should just make sure to call getRefundInformationFromFill in this function to load the repayment chain rather than just reading it directly from the event

nicholaspai: now that I'm looking at the code again, I'm thinking we might want to put this `verifyFill` logic inside `getRefundInformationFromFill`. This is because anytime we call `getRefundInformationFromFill`, for example in `getApproximateRefundsForBlockRange` we also would want to know if the repayment address is being swapped

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -882,6 +900,12 @@ export class BundleDataClient {
                 // At this point, the v3RelayHashes entry already existed meaning that there is a matching deposit,
                 // so this fill is validated.
                 v3RelayHashes[relayDataHash].fill = fill;
+
+                // If the fill is invalid due to relayer repayment addresses, return early.
+                const fillHasValidRepaymentAddress = await verifyFill(fill, destinationClient);
```
nicholaspai, line 905: I might move this check as early as possible into this async () call, because I think its strange that we would save this fill into `v3RelayHashes` if we subsequently invalidate it here. That way also if we overwrite the fill, we can do so as early as possible. For example, we could change to something like `async (_fill) => const fill = fillHasValidRepayment(_fill)` to minimize the diff

bmzig: https://github.com/across-protocol/sdk/pull/842/commits/48aecfd51aa3d43c36757b1942456099238973cc

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -899,24 +894,18 @@ export class BundleDataClient {
                 );
                 // At this point, the v3RelayHashes entry already existed meaning that there is a matching deposit,
                 // so this fill is validated.
-                v3RelayHashes[relayDataHash].fill = fill;
```
nicholaspai, line 884: This diff might be removable if we change `updatedFill` to `fill` above and change the `async (fill)` to `async (_fill)`. This way we again only need to deal with `fill` after the `verifyFillRepayment` call

src/clients/HubPoolClient.ts
```diff
@@ -805,6 +805,17 @@ export class HubPoolClient extends BaseAbstractClient {
     return { runningBalance, incentiveBalance };
   }
 
+  public isValidChainId(_chainId: number, mainnetBlockToCheck: number): boolean {
+    // Get the earliest cross chain contract set event.
+    const earliestCrossChainContractEvent = this.crossChainContracts[_chainId]?.reduce((prev, curr) => {
```
nicholaspai, line 810: Why use `reduce` here? I was thinking we could use very similar logic to [this](https://github.com/across-protocol/sdk/blob/479f2992e4ae4301fff01da0d986b4d87229f8ec/src/clients/HubPoolClient.ts#L166)

bmzig: Yeah, good point. https://github.com/across-protocol/sdk/pull/842/commits/b57ac22cf3bff05267008a4460b0b981e46ff510

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: https://github.com/across-protocol/sdk/pull/842/commits/b57ac22cf3bff05267008a4460b0b981e46ff510

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -309,50 +311,58 @@ export class BundleDataClient {
...
-            return false;
-          }
-          const matchingDeposit = this.spokePoolClients[fill.originChainId].getDeposit(fill.depositId);
-          const hasMatchingDeposit =
-            matchingDeposit !== undefined &&
-            this.getRelayHashFromEvent(fill) === this.getRelayHashFromEvent(matchingDeposit);
-          return hasMatchingDeposit;
-        })
-        .forEach((fill) => {
-          const matchingDeposit = this.spokePoolClients[fill.originChainId].getDeposit(fill.depositId);
```
bmzig, line 337: I didn't edit the `forEach` block. I just changed the `.filter` to a `filterAsync` so that I can call `verifyFillRepayment` in the filter.

src/clients/HubPoolClient.ts
```diff
@@ -805,6 +805,14 @@ export class HubPoolClient extends BaseAbstractClient {
     return { runningBalance, incentiveBalance };
   }
 
+  public isValidChainId(_chainId: number, mainnetBlockToCheck: number): boolean {
+    // Get the earliest cross chain contract set event.
+    const validCrossChainContractEvent = this.crossChainContracts[_chainId]?.find(
```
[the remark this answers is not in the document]
bmzig: We would need to wrap this in a try/catch in case the chain id was unsupported, which is fine, but would increase diff.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -309,50 +311,58 @@ export class BundleDataClient {
...
+          fill,
+          this.spokePoolClients[fill.destinationChainId].spokePool.provider,
+          this.clients.hubPoolClient,
+          blockRanges,
+          this.chainIdListForBundleEvaluationBlockNumbers
+        );
+        if (!isDefined(validRepayment)) {
+          return false;
+        }
+        const matchingDeposit = this.spokePoolClients[fill.originChainId].getDeposit(fill.depositId);
```
pxrl, line 340: We'll ultimately need to use more than just `depositId` here because we can now see multiple identical `depositIds` within a single bundle. I think it's fine to just add that as a todo to be handled in a follow-up PR.

pxrl: (nb. I see now that this is not for the dataworker, but is instead for relayer repayment approximation. It's fine to disregard the comment because it's not related to bundles)

src/clients/HubPoolClient.ts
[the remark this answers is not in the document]
nicholaspai: oh true, its ok nvm my comment

src/clients/HubPoolClient.ts
```diff
@@ -805,6 +805,14 @@ export class HubPoolClient extends BaseAbstractClient {
     return { runningBalance, incentiveBalance };
   }
 
+  public isValidChainId(_chainId: number, mainnetBlockToCheck: number): boolean {
```
pxrl, line 808: ooc, why is this `_chainId` ? Normally the underscore is to hint the linter that the variable is intentionally unused, but it is used here.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: yeah this function name has "approximately" in it :)

pxrl (approved):

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: you raise a good point about duplicate deposits though, where technically we should also be counting in pre-fill refunds into this function. Which means we should count all deposits that are not matched by a fill and have FillStatus == Filled, however I think this should be rare and not something we need to handle now

src/clients/HubPoolClient.ts
bmzig: Just so it isn't confused with `this.chainId`.

src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -65,8 +65,17 @@ export async function verifyFillRepayment(
     hubPoolClient.chainId,
     chainIdListForBundleEvaluationBlockNumbers
   )[1];
+  const fromLiteChain = hubPoolClient.configStoreClient
+    .getLiteChainIdIndicesForBlock(endBlockForMainnet)
```
nicholaspai, line 69: Can we use `spokePoolClient.isOriginLiteChain`?

src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -44,3 +46,51 @@ export function getRefundInformationFromFill(
...
+    endBlockForMainnet
+  );
+  const repaymentChainId = fromLiteChain ? fill.originChainId : fill.repaymentChainId;
+  // Return undefined if the requested repayment chain ID is not recognized by the hub pool.
+  if (!hubPoolClient.isValidChainId(repaymentChainId, endBlockForMainnet)) {
+    return undefined;
+  }
+  // If the fill was from a lite chain, the origin chain is an EVM chain, and the relayer address is invalid
+  // for EVM chains, then we cannot refund the relayer, so mark the fill as invalid.
+  if (fromLiteChain && chainIsEvm(fill.originChainId) && !isValidEvmAddress(fill.relayer)) {
```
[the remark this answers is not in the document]
bmzig: Yeah that's better. https://github.com/across-protocol/sdk/pull/842/commits/c270c97cd42609b65d8bd776885f4ec9269b7b22

nicholaspai (approved):

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -919,34 +946,34 @@ export class BundleDataClient {
...
+                  _fill,
+                  destinationClient.spokePool.provider,
+                  matchedDeposit,
+                  allChainIds
+                );
+                if (!isDefined(fill)) {
+                  bundleInvalidFillsV3.push(_fill);
+                  return;
+                }
+                v3RelayHashes[relayDataHash].fill = fill;
```
bmzig, line 975: I don't think we need to update fill again here.

PR #844, opened 2025-01-29
feat(SpokePoolClient): Prepare for new SlowFillRequest

pxrl:
`getRelayDataHash()` can no longer be used for hashing a SlowFillRequest because the new event types emit `messageHash` instead of `message`. This is OK in the SpokePoolClient because it only needs a very simple method of indexing SlowFillRequest events.

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -67,6 +67,7 @@ export class SpokePoolClient extends BaseAbstractClient {
   protected currentTime = 0;
   protected oldestTime = 0;
   protected depositHashes: { [depositHash: string]: DepositWithBlock } = {};
+  protected duplicateDepositHashes: { [depositHash: string]: DepositWithBlock[] } = {};
   protected depositHashesToFills: { [depositHash: string]: FillWithBlock[] } = {};
   protected speedUps: { [depositorAddress: string]: { [depositId: string]: SpeedUpWithBlock[] } } = {};
   protected slowFillRequests: { [relayDataHash: string]: SlowFillRequestWithBlock } = {};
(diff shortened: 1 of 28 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -278,7 +278,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @returns The corresponding SlowFIllRequest event if found, otherwise undefined.
    */
   public getSlowFillRequest(relayData: RelayData): SlowFillRequestWithBlock | undefined {
-    const hash = getRelayDataHash(relayData, this.chainId);
+    const hash = this.getDepositHash(relayData);
```
nicholaspai, line 281: Don't we lose resiliency to events that have the same deposit id now? Since getRelayDataHash hashes the entire struct. 

On the other hand, it shouldn't be possible to send duplicate slow fill requests

src/utils/SpokeUtils.ts
```diff
@@ -393,3 +395,7 @@ export async function findFillBlock(
 export function isZeroAddress(address: string): boolean {
   return address === ZERO_ADDRESS || address === ZERO_BYTES;
 }
+
+export function getMessageHash(message: string): string {
+  return isMessageEmpty(message) ? ZERO_BYTES : keccak256(message);
```
nicholaspai, line 400: Instead of ZERO_BYTES should this be ethers.constants.HashZero like I use here? https://github.com/across-protocol/sdk/blob/pre-fills/src/constants.ts#L60 

I think both are ok but maybe better to be consistent

src/clients/SpokePoolClient.ts
pxrl: It's not possible to make multiple slow fill requests for the same relayData hash, but it is still possible for the same depositId with different relayData hash to collide here. I agree that this change introduces a lack of collision-resistance that was not there before. However, we anyway have to solve this for our method of storing deposits within the SpokePoolClient, so this lack of resiliency was already there to begin with. My thinking was that we'd revisit `SpokePoolClient.getDepositHash()` and make it more resilient.

src/utils/SpokeUtils.ts
pxrl: Ah - we're exporting the same constant multiple times. We're also exporting `HashZero` as `ZERO_BYTES`, and `ZERO_BYTES` is what's used in `isZeroAddress()` (see line 396 in the diff above).

https://github.com/across-protocol/sdk/blob/2349b73a03d9ba62a8c52328255da6bf7ab48f78/src/constants.ts#L13

`ZERO_BYTES` might be more "portable" because `bytes32(0)` is also used in the context of the zero address within `isZeroAddress()`. wdyt?

src/clients/SpokePoolClient.ts
nicholaspai: could we not just hash the entire relay data but replacing the message with a message hash for deposits?

src/utils/SpokeUtils.ts
nicholaspai: ah ok i'll change my PR to use ZERO_BYTES then

nicholaspai (approved):

src/clients/SpokePoolClient.ts
pxrl: Possibly, but we could also just borrow your approach from the BundleDataClient and concatenate all the fields (including messageHash instead of message). We probably want to move away from calling it a hash; maybe `getDepositKey()` or something?

We fwiw have a similar issue in the `queryHistoricalDeposit...()` function, where we're also relying on real relayData hashes, but in practice we only need a unique key.

nicholaspai (approved):

src/clients/SpokePoolClient.ts
pxrl: It's hard to prove, but my gut feel on the hash speed issue is that it's actually the abi encoding that's expensive, not the actual hashing. The ABI has to be instantiated on each each call to the ABICoder and it's supplied as stringified JSON. This is converted into an ABI instance on the fly. Doing this once per relayData seems to add a lot of overhead.

viem seems to support instantiating the ABI instance separately from encoding and hashing the data. I've got an old branch where I switched over to using it but I hit a snag on that. I don't recall what it was but I'd like to pick it up again because unlocking hashing inefficiencies has historically been a source of big speedups in the bots.

bmzig (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -648,14 +648,12 @@ export class SpokePoolClient extends BaseAbstractClient {
       for (const event of slowFillRequests) {
         const slowFillRequest = {
           ...spreadEventWithBlockNumber(event),
+          messageHash: getMessageHash(event.args["message"]),
```
nicholaspai, line 651: Can we add a simple unit test for this addition?

pxrl: https://github.com/across-protocol/sdk/pull/844/commits/477d52147e70395d4ecdf9d098d431d31501a013

PR #845, opened 2025-01-29
fix(BundleDataClient): Make sure bundle block timestamps have no gaps

nicholaspai:
Currently there is a small chance that a fill deadline can fall between block ranges

inspired by @bmzig 's comment here https://github.com/across-protocol/sdk/pull/835#discussion_r1934742099

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -1401,10 +1401,10 @@ export class BundleDataClient {
           // end block is exactly equal to the bundle block timestamp for the next bundle's start block. This way
           // there are no gaps in block timestamps between bundles.
           const endBlockForChain = Math.min(_endBlockForChain + 1, spokePoolClient.latestBlockSearched);
-          const [startTime, _endTime] = await Promise.all([
-            spokePoolClient.getTimestampForBlock(startBlockForChain),
-            spokePoolClient.getTimestampForBlock(endBlockForChain),
-          ]);
+          const [startTime, _endTime] = [
+            await spokePoolClient.getTimestampForBlock(startBlockForChain),
+            await spokePoolClient.getTimestampForBlock(endBlockForChain),
+          ];
           // @dev similar to reasoning above to ensure no gaps between bundle block range timestamps and also
           // no overlap, subtract 1 from the end time.
           const endBlockDelta = endBlockForChain > startBlockForChain ? 1 : 0;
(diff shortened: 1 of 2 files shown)


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1395,7 +1395,12 @@ export class BundleDataClient {
           // will usually be called in production with block ranges that were validated by
           // DataworkerUtils.blockRangesAreInvalidForSpokeClients.
           const startBlockForChain = Math.min(_startBlockForChain, spokePoolClient.latestBlockSearched);
-          const endBlockForChain = Math.min(_endBlockForChain, spokePoolClient.latestBlockSearched);
+          // @dev Add 1 to the bundle end block. The thinking here is that there can be a gap between
+          // block timestamps in subsequent blocks. The bundle data client assumes that fill deadlines expire
+          // in exactly one bundle, therefore we must make sure that the bundle block timestamp for one bundle's
+          // end block is exactly equal to the bundle block timestamp for the next bundle's start block. This way
+          // there are no gaps in block timestamps between bundles.
+          const endBlockForChain = Math.min(_endBlockForChain + 1, spokePoolClient.latestBlockSearched);
```
bmzig, line 1403: The end timestamp of a bundle now overlaps with the start block of the next bundle. Maybe it is just safer to do this in addition with `endTime = await spokePoolClient.getTimestampForBlock(endBlockForChain) - 1`?

nicholaspai: Hmm well I think that's ok because of the way we check bundle block timestamps [here](https://github.com/across-protocol/sdk/blob/57ce0138c77721ccde3c65ad3be99b13fe69e295/src/clients/BundleDataClient/BundleDataClient.ts#L847) and [here](https://github.com/across-protocol/sdk/blob/57ce0138c77721ccde3c65ad3be99b13fe69e295/src/clients/BundleDataClient/BundleDataClient.ts#L1051) for example.

But I think you're right. By subtracting 1 from it, we'll be strictly safer and also guarantee no gaps

nicholaspai: 6e90289edb73ac4ce62c3fe26006171417527a36

bmzig (approved):

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1401,15 +1401,20 @@ export class BundleDataClient {
...
+            spokePoolClient.getTimestampForBlock(endBlockForChain),
+          ]);
           // @dev similar to reasoning above to ensure no gaps between bundle block range timestamps and also
           // no overlap, subtract 1 from the end time.
           const endBlockDelta = endBlockForChain > startBlockForChain ? 1 : 0;
-          const [startTime, endTime] = [
-            await spokePoolClient.getTimestampForBlock(startBlockForChain),
-            (await spokePoolClient.getTimestampForBlock(endBlockForChain)) - endBlockDelta,
-          ];
+          const endTime = Math.max(0, _endTime - endBlockDelta);
```
nicholaspai, line 1411: For some test cases, the end and start block are the same due to the `latestBlockSearched` in the test spoke pool client being lower than the bundle block range. Also in some tests, the timestamps are 0. This fix handles these cases.

bmzig (approved):

PR #846, opened 2025-01-30
feat(SpokePoolClient): Append messageHash to deposit events

pxrl:
In anticipation of needing to use the deposit messageHash when matching fills to deposits, precompute the messageHash when deposit events are ingested.


src/clients/BundleDataClient/utils/SuperstructUtils.ts
```diff
@@ -54,6 +56,7 @@ const SortableEventSS = {
 };
 
 const V3DepositSS = {
+  messageHash: defaulted(string(), UNDEFINED_MESSAGE_HASH),
```
pxrl, line 59: `defaulted()` will establish an instance of `messageHash` with the value `""`. I looked at superstruct for ways of dynamically assigning this (i.e. computing on the fly based on the rest of the object). It kinda seems like superstruct can support this, but we have a lot of nested structures in our object validations and I lost patience with it.

In any case, this is backwards compatible because we can populate messageHash values for existing deposits that were not posted with the field. I'd also consider dropping it from data posted to arweave, since there is some odd redundancy between `message` and `messageHash`, but we don't expect to verify one from the other when we pull them down simultaneously.

test/SpokePoolClient.ValidateFill.ts
```diff
@@ -353,7 +359,7 @@ describe("SpokePoolClient: Fill Validation", function () {
     // Searching for deposit ID -1 that doesn't exist should throw.
     await assertPromiseError(
       getBlockRangeForDepositId(
-        -1,
+        toBN(-1),
```
pxrl, line 362: It seems like this was probably left out of the recent BigNumber depositId migration.

bmzig (approved):

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -239,6 +241,19 @@ export class BundleDataClient {
       );
 
     const data = persistedData[0].data;
+
+    // Post-populate any missing message hashes.
+    // @todo This can be removed once the legacy types hurdle is cleared (earliest 7 days post migration).
+    Object.values(data.bundleDepositsV3).forEach((x) =>
```
nicholaspai, line 247: Can we unit test this?

pxrl: I've added a test for it in the subsequent PR here: https://github.com/across-protocol/sdk/pull/850/commits/1a6311b2fb6a34e72bdfef4beb924fa2c8adc5b3

bmzig (approved):

src/utils/SpokeUtils.ts
```diff
@@ -21,7 +21,7 @@ type BlockTag = providers.BlockTag;
  */
 export function populateV3Relay(
```
nicholaspai, line 22: Is there a way we currently test this function or is it used only in prod?

test/SpokePoolClient.v3Events.ts
```diff
@@ -536,6 +536,18 @@ describe("SpokePoolClient: Event Filtering", function () {
   });
 
   describe("SpokePoolClient: messageHash Handling", function () {
+    it("Correctly appends FundsDepositedV3 messageHash", async function () {
+      const _deposit = generateDeposit(originSpokePoolClient);
+      expect(_deposit?.args?.messageHash).to.equal(undefined);
+      await originSpokePoolClient.update(fundsDepositedEvents);
+
+      let deposit = originSpokePoolClient.getDeposit(_deposit.args.depositId);
+      expect(deposit).to.exist;
```
nicholaspai, line 545: Does exist equal not.undefined? I've considered using this syntax in the past but chose not to for consistency. Admittedly I didn't look into it deeply

nicholaspai (approved):
Two questions. Lgtm!

test/SpokePoolClient.v3Events.ts
pxrl: Docs say that `exist` checks for both `undefined` and `null`: https://www.chaijs.com/api/bdd/#method_exist

I typically only use this as a check before later re-assigning with the ! operator to force it to stop complaining about the object possibly being undefined.

src/utils/SpokeUtils.ts
pxrl: Seems like it's only used in prod - we use it in the relayFeeCalculator.

PR #848, opened 2025-01-30
feat: truncate select bytes32 values in events to bytes20 addresses

bmzig:
For now, select input fields on events should be automatically truncated to bytes20 evm addresses. This is so that all clients which use the spoke pool client to get deposit/relay/etc. information do not need to handle bytes32 addresses.

Draft until unit tests are made


pxrl (approved):
Nice! Really happy to see this.

PR #849, opened 2025-01-30
fix(BundleDataClient): If fill cannot be repaid, still account for unexecutable slow fills

nicholaspai:
If a fill cannot be repaid due to `relayerAddress` or `repaymentChainId` being invalid, it still cannot be filled on-chain and therefore has potential to create an unexecutable slow fill.

We currently are early exiting if the fill cannot be repaid and that means a filler could grief the system with invalid repayment info and getting slow fill funds stuck on the destination chain. This could be remedied via admin action but its undesirable.

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -243,6 +243,7 @@ export class BundleDataClient {
         bundleData.bundleDepositsV3,
         bundleData.bundleFillsV3,
         [], // Invalid fills are not persisted to Arweave.
+        [], // Unrepayable fills are not persisted to Arweave.
         bundleData.bundleSlowFillsV3,
         bundleData.expiredDepositsToRefundV3,
         bundleData.unexecutableSlowFills
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -676,6 +676,7 @@ export class BundleDataClient {
     const bundleDepositsV3: BundleDepositsV3 = {}; // Deposits in bundle block range.
     const bundleFillsV3: BundleFillsV3 = {}; // Fills to refund in bundle block range.
     const bundleInvalidFillsV3: V3FillWithBlock[] = []; // Fills that are not valid in this bundle.
+    const bundleUnrepayableFillsV3: V3FillWithBlock[] = []; // Fills that are not repayable in this bundle.
     const bundleSlowFillsV3: BundleSlowFills = {}; // Deposits that we need to send slow fills
     // for in this bundle.
     const expiredDepositsToRefundV3: ExpiredDepositsToRefundV3 = {};
@@ -904,7 +905,7 @@ export class BundleDataClient {
                     allChainIds
                   );
                   if (!isDefined(fillToRefund)) {
-                    bundleInvalidFillsV3.push(fill);
+                    bundleUnrepayableFillsV3.push(fill);
                     // We don't return here yet because we still need to mark unexecutable slow fill leaves
                     // or duplicate deposits. However, we won't issue a fast fill refund.
                   } else {
@@ -974,7 +975,7 @@ export class BundleDataClient {
                   allChainIds
                 );
                 if (!isDefined(fillToRefund)) {
-                  bundleInvalidFillsV3.push(fill);
+                  bundleUnrepayableFillsV3.push(fill);
                   // Don't return yet as we still need to mark down any unexecutable slow fill leaves
                   // in case this fast fill replaced a slow fill request.
                 } else {
@@ -1340,6 +1341,7 @@ export class BundleDataClient {
       bundleDepositsV3,
       bundleFillsV3,
       bundleInvalidFillsV3,
+      bundleUnrepayableFillsV3,
       bundleSlowFillsV3,
       expiredDepositsToRefundV3,
       unexecutableSlowFills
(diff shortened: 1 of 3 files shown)


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -946,44 +949,47 @@ export class BundleDataClient {
...
+                  // Don't return yet as we still need to mark down any unexecutable slow fill leaves
+                  // in case this fast fill replaced a slow fill request.
+                } else {
+                  // @dev Since queryHistoricalDepositForFill validates the fill by checking individual
+                  // object property values against the deposit's, we
+                  // sanity check it here by comparing the full relay hashes. If there's an error here then the
+                  // historical deposit query is not working as expected.
+                  assert(this.getRelayHashFromEvent(matchedDeposit) === relayDataHash, "Relay hashes should match.");
+                  validatedBundleV3Fills.push({
+                    ...fill,
```
bmzig, line 987:
```suggestion
                    ...fillToRefund,
```

nicholaspai: https://github.com/across-protocol/sdk/pull/849/commits/076b30c503cdbde30a61850a31f90661da57d76d

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -884,36 +884,37 @@ export class BundleDataClient {
...
+            const relayDataHash = this.getRelayHashFromEvent(fill);
             if (v3RelayHashes[relayDataHash]) {
               if (!v3RelayHashes[relayDataHash].fill) {
                 assert(
                   isDefined(v3RelayHashes[relayDataHash].deposit),
                   "Deposit should exist in relay hash dictionary."
                 );
-                // `fill` will only possibly differ from `_fill` in the `relayer` field, which does not affect the
-                // relay hash, so it is safe to modify.
-                const fill = await verifyFillRepayment(
```
bmzig, line 898: If I'm understanding this correctly, this issue was right here, where we would return before setting `v3RelayHashes` (since we did set `v3RelayHashes` before returning on incorrect repayment in the historical deposit case)? Also, the flow would be somebody fills with invalid relay data, we do not refund them here, but in the slow fill block, we don't see the invalid fill, so we create a leaf and send it to destination?

bmzig: In other words, returning early after finding that `fillToRefund` was undefined wasn't necessarily the issue (this would still be valid)
```
if (!isDefined(fillToRefund)) {
  bundleInvalidFillsV3.push(fill);
  return;
}
```
The issue was doing that before modifying `v3RelayHashes`?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -884,36 +884,37 @@ export class BundleDataClient {
...
+                  const fillToRefund = await verifyFillRepayment(
+                    fill,
+                    destinationClient.spokePool.provider,
+                    v3RelayHashes[relayDataHash].deposit!,
+                    allChainIds
+                  );
+                  if (!isDefined(fillToRefund)) {
+                    bundleInvalidFillsV3.push(fill);
+                    // We don't return here yet because we still need to mark unexecutable slow fill leaves
+                    // or duplicate deposits. However, we won't issue a fast fill refund.
```
pxrl, line 909: wdyt about classifying these separately from invalid fills? To me there's a clear distinction between an invalid fill and a valid fill with invalid repayment information. I think it's useful to record them separately; at least for logging purposes.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -926,6 +927,8 @@ export class BundleDataClient {
                     fastFillsReplacingSlowFills.push(relayDataHash);
                   }
                 }
+              } else {
+                throw new Error("Duplicate fill detected");
```
pxrl, line 931: wdym by this?

nicholaspai: this is just an assert. We should never see duplicate fills in memory

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: Ah I see.

bmzig (approved):

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: https://github.com/across-protocol/sdk/pull/849/commits/0fa9a9292cd169aadfa2c468bee9c8a2547101a6

bmzig (approved):

src/clients/BundleDataClient/utils/DataworkerUtils.ts
```diff
@@ -99,6 +100,11 @@ export function prettyPrintV3SpokePoolEvents(
       "destinationChainId",
       (fill) => `${fill.relayer}`
     ),
+    allUnrepayableFillsInRangeByDestinationChainAndRelayer: groupObjectCountsByTwoProps(
+      bundleUnrepayableFillsV3,
+      "repaymentChainId",
+      (fill) => `${fill.relayer}`
```
pxrl, line 106: This might be a bit weird post migration, where `relayer` is the address specified in the fill, which is apparently wrong for the resolved repayment chain. So idk how useful this log is going to be vs. the other log that dumps the actual unrepayable fills.

pxrl (approved):

src/clients/BundleDataClient/utils/DataworkerUtils.ts
nicholaspai: good point, wdyt about just deleting invalid fills and unrepayable fills from this log in favor of the subsequent log that dumps all those fills?

PR #850, opened 2025-01-30
feat(SpokePoolClient): Support fills with messageHash field

pxrl:
This change also restores the ability to directly compare a deposit with a fill. Instead of comparing messages, their message hashes are compared.

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -242,6 +242,15 @@ export class BundleDataClient {
         })
       )
     );
+    Object.values(data.bundleFillsV3).forEach((x) =>
+      Object.values(x).forEach((fills) =>
+        fills.fills.forEach((fill) => {
+          if (fill.messageHash === UNDEFINED_MESSAGE_HASH) {
+            fill.messageHash = getMessageHash(fill.message);
+          }
+        })
+      )
+    );
 
     const bundleData = {
       bundleFillsV3: convertTypedStringRecordIntoNumericRecord(data.bundleFillsV3),
--- a/src/interfaces/SpokePool.ts
+++ b/src/interfaces/SpokePool.ts
@@ -57,6 +57,7 @@ export interface RelayExecutionEventInfo {
 }
 
 export interface Fill extends RelayData {
+  messageHash: string;
   destinationChainId: number;
   relayer: string;
   repaymentChainId: number;
--- a/src/utils/DepositUtils.ts
+++ b/src/utils/DepositUtils.ts
@@ -5,7 +5,7 @@ import { CachingMechanismInterface, Deposit, DepositWithBlock, Fill, SlowFillReq
 import { getNetworkName } from "./NetworkUtils";
 import { getDepositInCache, getDepositKey, setDepositInCache } from "./CachingUtils";
 import { validateFillForDeposit } from "./FlowUtils";
-import { isUnsafeDepositId } from "./SpokeUtils";
+import { getMessageHash, isUnsafeDepositId } from "./SpokeUtils";
 import { getCurrentTime } from "./TimeUtils";
 import { isDefined } from "./TypeGuards";
 import { isDepositFormedCorrectly } from "./ValidatorUtils";
@@ -121,6 +121,8 @@ export async function queryHistoricalDepositForFill(
       await setDepositInCache(deposit, getCurrentTime(), cache, DEFAULT_CACHING_TTL);
     }
   }
+
+  deposit.messageHash ??= getMessageHash(deposit.message);
 
   const match = validateFillForDeposit(fill, deposit);
   if (match.valid) {
--- a/src/utils/FlowUtils.ts
+++ b/src/utils/FlowUtils.ts
@@ -14,13 +14,13 @@ export const RELAYDATA_KEYS = [
   "fillDeadline",
   "exclusivityDeadline",
   "exclusiveRelayer",
-  "message",
+  "messageHash",
 ] as const;
 
 // Ensure that each deposit element is included with the same value in the fill. This includes all elements defined
 // by the depositor as well as destinationToken, which are pulled from other clients.
 export function validateFillForDeposit(
-  relayData: RelayData & { destinationChainId: number },
+  relayData: Omit<RelayData, "message"> & { messageHash: string, destinationChainId: number },
   deposit?: Deposit
 ): { valid: true } | { valid: false; reason: string } {
   if (deposit === undefined) {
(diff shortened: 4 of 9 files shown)


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -242,6 +242,15 @@ export class BundleDataClient {
...
     );
+    Object.values(data.bundleFillsV3).forEach((x) =>
+      Object.values(x).forEach((fills) =>
+        fills.fills.forEach((fill) => {
+          if (fill.messageHash === UNDEFINED_MESSAGE_HASH) {
+            fill.messageHash = getMessageHash(fill.message);
+          }
+        })
+      )
+    );
```
pxrl, line 253: This is tbh clunky and annoying, but it should also be temporary. Once we pass the 7 day deprecation hurdle we can start to strip out the legacy events and can require that the `messageHash` field is present in arweave data retrieved. This would also involve a version bump.

test/utils/utils.ts
```diff
@@ -313,27 +313,28 @@ export async function depositV3(
...
+    inputToken: args.inputToken,
+    inputAmount: args.inputAmount,
+    outputToken: args.outputToken,
+    outputAmount: args.outputAmount,
+    quoteTimestamp: args.quoteTimestamp,
+    message: args.message,
+    messageHash: getMessageHash(args.message),
+    fillDeadline: args.fillDeadline,
+    exclusivityDeadline: args.exclusivityDeadline,
+    exclusiveRelayer: args.exclusiveRelayer,
```
pxrl, line 337: This is just a driveby improvement that I spotted when fixing the neighbouring code for this change.

src/interfaces/SpokePool.ts
```diff
@@ -57,6 +57,7 @@ export interface RelayExecutionEventInfo {
 }
 
 export interface Fill extends RelayData {
```
pxrl, line 59: todo: Should also drop `message` from this type, and from `SlowFillRequest`. That will flush out a few extra incompatibilities.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1401,7 +1410,7 @@ export class BundleDataClient {
       event.inputAmount
     }-${event.outputAmount}-${event.originChainId}-${event.depositId.toString()}-${event.fillDeadline}-${
       event.exclusivityDeadline
-    }-${event.message}-${event.destinationChainId}`;
+    }-${event.messageHash}-${event.destinationChainId}`;
```
nicholaspai, line 1413: Can probably eventually swap out this function into a utility function and use it in the SpokePoolClient in lieu of `getDepositHash`

nicholaspai: We can also use this util in SpokePoolClient [here](https://github.com/across-protocol/sdk/pull/844#discussion_r1936966152) so that we no longer use the `getRelayDataHash()` function  anywhere except when calling `relayFillStatuses()`

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: I'd like to add a unit test for this because it'll be hard to verify in prod if something goes wrong

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: Can we do this in this PR? That way we can unit test the function as well

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: I added tests for this and `BundleFillsV3` in the relayer in anticipation of the sdk bump https://github.com/UMAprotocol/across-configs/pull/34.

pxrl: I also have some here: https://github.com/across-protocol/sdk/pull/850/commits/3b683328a1004e5336972006dbe6c08e2d4a34cd

src/clients/BundleDataClient/BundleDataClient.ts
pxrl: I ended up pulling it out, but it's already been merged here: https://github.com/across-protocol/sdk/pull/860.

src/utils/DepositUtils.ts
```diff
@@ -122,6 +122,8 @@ export async function queryHistoricalDepositForFill(
     }
   }
 
+  deposit.messageHash ??= getMessageHash(deposit.message);
```
nicholaspai, line 125: I think you'll need something similar in `https://github.com/across-protocol/sdk/blob/f9a1e9c7753e83bdd666e89960daf76f6ff7cd67/src/utils/SpokeUtils.ts#L429` because it also makes a fresh event query.

nicholaspai (commented):
This looks great, just one change required i think https://github.com/across-protocol/sdk/pull/850/files#r1943460894

src/utils/DepositUtils.ts
pxrl: Good catch - https://github.com/across-protocol/sdk/pull/850/commits/7796919007fc7fe8237b7eef701de7d2adfa1e62

src/utils/SpokeUtils.ts
```diff
@@ -455,6 +455,7 @@ export async function findFillEvent(
   const fill = {
     ...spreadEventWithBlockNumber(event),
     destinationChainId,
+    messageHash: getMessageHash(event.args.message),
   } as FillWithBlock;
```
nicholaspai, line 459: How does this not get caught by typescript compile time?? ugh

pxrl: It's because of the nasty type assertion on line 459. It's a real anti-pattern to do this and we only get away with it because it's in a low-level function that rarely changes; otherwise it'd blow up regularly.

src/utils/FlowUtils.ts
```diff
@@ -30,7 +31,13 @@ export function validateFillForDeposit(
   // Note: this short circuits when a key is found where the comparison doesn't match.
   // TODO: if we turn on "strict" in the tsconfig, the elements of FILL_DEPOSIT_COMPARISON_KEYS will be automatically
   // validated against the fields in Fill and Deposit, generating an error if there is a discrepency.
-  const invalidKey = RELAYDATA_KEYS.find((key) => relayData[key].toString() !== deposit[key].toString());
+  let invalidKey = RELAYDATA_KEYS.find((key) => relayData[key].toString() !== deposit[key].toString());
+
+  // There should be no paths for `messageHash` to be unset, but mask it off anyway.
+  // @todo Add test.
```
pxrl, line 37: Can drop this todo now: https://github.com/across-protocol/sdk/pull/850/commits/db670655dd3918ebcb630d004ee8d9143d9a5f1b

```suggestion
```

PR #851, opened 2025-01-30
feat: Add getMessageHash utility

pxrl:


src/utils/SpokeUtils.ts
```diff
@@ -393,3 +395,7 @@ export async function findFillBlock(
 export function isZeroAddress(address: string): boolean {
   return address === ZERO_ADDRESS || address === ZERO_BYTES;
 }
+
+export function getMessageHash(message: string): string {
```
pxrl, line 399: nb should add a couple of tests for this.

src/utils/SpokeUtils.ts
```diff
@@ -393,3 +395,7 @@ export async function findFillBlock(
...
+  return isMessageEmpty(message) ? ZERO_BYTES : keccak256(message);
```
nicholaspai, line 400: I think instead of ZERO_BYTES this should be EMPTY_MESSAGE_HASH

```
export const EMPTY_MESSAGE_HASH = ethersConstants.HashZero;
```

pxrl: Ah, we have an unresolved thread in the previous PR where this change existed. tldr we are exporting HashZero twice from ethers - once as ZERO_BYTES and once as EMPTY_MESSAGE_HASH. I probably prefer only one export called `ZERO_BYTES`, otherwise we need to keep two (because it doesn't make sense to check for `bytes32(address(0))` under the name `EMPTY_MESSAGE_HASH`.

Or do we just want two exports?

https://github.com/across-protocol/sdk/pull/844#discussion_r1935540043

src/utils/SpokeUtils.ts
pxrl: Added: https://github.com/across-protocol/sdk/pull/851/commits/d051eaf006b67d24a1571dfc243c1dc382da0dee

nicholaspai (approved):

PR #852, opened 2025-01-30
feat: Forwards-compatible getRelayDataHash()

pxrl:
Promoting address (bytes20) -> bytes32 and uint32 -> uint256 still produces the same RelayData hashes. This is verified by the existing tests passing (they are verified against the existing SpokePool deployments with address and uint32 in RelayData).

### DIFF
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -226,11 +226,11 @@ export function getRelayDataHash(relayData: RelayData, destinationChainId: numbe
     ethersUtils.defaultAbiCoder.encode(
       [
         "tuple(" +
-          "address depositor," +
-          "address recipient," +
-          "address exclusiveRelayer," +
-          "address inputToken," +
-          "address outputToken," +
+          "bytes32 depositor," +
+          "bytes32 recipient," +
+          "bytes32 exclusiveRelayer," +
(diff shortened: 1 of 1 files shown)


src/utils/SpokeUtils.ts
```diff
@@ -222,26 +222,34 @@ export async function getDepositIdAtBlock(contract: Contract, blockTag: number):
...
+          "bytes32 inputToken," +
+          "bytes32 outputToken," +
           "uint256 inputAmount," +
           "uint256 outputAmount," +
           "uint256 originChainId," +
-          "uint32 depositId," +
+          "uint256 depositId," +
           "uint32 fillDeadline," +
           "uint32 exclusivityDeadline," +
           "bytes message" +
```
nicholaspai, line 248: What's the plan for message hash? Just curious

nicholaspai (approved):

src/utils/SpokeUtils.ts
pxrl: Per https://github.com/across-protocol/contracts/pull/867, `messageHash` won't be part of `RelayData`, so there's nothing to do for that. Unfortunately it reduces the utility of this function though, since it means we can't pass a raw `FilledRelay` in. We need to find alternatives for the few cases where we rely on that.

PR #853, opened 2025-01-31
improve(Relayer): Log unrepayable fills

nicholaspai:
Will be useful with new contracts if someone sets an invalid repayment chain or address, though its probably impossible to set an invalid address


src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -55,17 +55,18 @@ export async function verifyFillRepayment(
   matchedDeposit: DepositWithBlock,
   possibleRepaymentChainIds: number[]
 ): Promise<FillWithBlock | undefined> {
+  const updatedFill = _.cloneDeep(fill);
```
nicholaspai, line 58: Just refactored here. We should always be returning a clone of the fill, even if we exit early

src/clients/SpokePoolClient.ts
```diff
@@ -334,6 +336,21 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+      return true;
+    }
+    // Lite chain deposits force repayment on origin chain.
+    const repaymentChainId = matchedDeposit.fromLiteChain ? fill.originChainId : fill.repaymentChainId;
+    // Return undefined if the requested repayment chain ID is not recognized by the hub pool.
+    const possibleRepaymentChainIds = this.configStoreClient?.getEnabledChains();
+    if (isDefined(possibleRepaymentChainIds) && !possibleRepaymentChainIds.includes(repaymentChainId)) {
+      return false;
+    }
+    return chainIsEvm(repaymentChainId) && !isValidEvmAddress(fill.relayer);
```
bmzig, line 351: For forwards compatibility, you should probably do `return !chainIsEvm(repaymentChainId) || (chainIsEvm(repaymentChainId) && isValidEvmAddress(fill.relayer))`. Note that it should be `isValidEvmAddress(fill.relayer)`.

src/clients/SpokePoolClient.ts
```diff
@@ -354,16 +371,21 @@ export class SpokePoolClient extends BaseAbstractClient {
       return { unfilledAmount: outputAmount, fillCount: 0, invalidFills: [] };
     }
 
-    const { validFills, invalidFills } = fillsForDeposit.reduce(
-      (groupedFills: { validFills: Fill[]; invalidFills: Fill[] }, fill: Fill) => {
+    const { validFills, invalidFills, unrepayableFills } = fillsForDeposit.reduce(
+      (groupedFills: { validFills: Fill[]; invalidFills: Fill[]; unrepayableFills: Fill[] }, fill: Fill) => {
         if (validateFillForDeposit(fill, deposit).valid) {
+          if (this.isEvmRepaymentValid(fill, deposit)) {
```
bmzig, line 377: Should this be `if (!this.isEvmRepaymentValid(fill, deposit))`?

src/clients/SpokePoolClient.ts
nicholaspai: https://github.com/across-protocol/sdk/pull/853/commits/6a7d527829360d11554faf52d5f1d1285795c84c

src/clients/SpokePoolClient.ts
nicholaspai: https://github.com/across-protocol/sdk/pull/853/commits/6a7d527829360d11554faf52d5f1d1285795c84c

src/clients/SpokePoolClient.ts
```diff
@@ -334,6 +336,21 @@ export class SpokePoolClient extends BaseAbstractClient {
...
 
+  private isEvmRepaymentValid(fill: Fill, matchedDeposit: Deposit): boolean {
+    // Slow fills don't result in repayments so they're always valid.
+    if (isSlowFill(fill)) {
+      return true;
+    }
+    // Lite chain deposits force repayment on origin chain.
+    const repaymentChainId = matchedDeposit.fromLiteChain ? fill.originChainId : fill.repaymentChainId;
+    // Return undefined if the requested repayment chain ID is not recognized by the hub pool.
+    const possibleRepaymentChainIds = this.configStoreClient?.getEnabledChains();
```
nicholaspai, line 347: We could alternatively assert(isDefined(this.configStoreClient)) but I'm worried this will make unit tests more cumbersome.

src/clients/SpokePoolClient.ts
nicholaspai: I think this is equivalent and also less verbose: 3f821ad3eedef8252dc6b8cb6aa15810a6754afa wdyt

src/clients/SpokePoolClient.ts
```diff
@@ -334,6 +336,24 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    if (isDefined(possibleRepaymentChainIds) && !possibleRepaymentChainIds.includes(repaymentChainId)) {
+      return false;
+    }
+    if (!chainIsEvm(repaymentChainId)) {
```
bmzig, line 351: Should this be `false` or `true`. Assuming we support solana, this would just mean we can never count a repayment in solana as valid, so I'd be more inclined to return true here.

src/clients/SpokePoolClient.ts
```diff
@@ -334,6 +336,24 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+  private isEvmRepaymentValid(fill: Fill, matchedDeposit: Deposit): boolean {
+    // Slow fills don't result in repayments so they're always valid.
+    if (isSlowFill(fill)) {
+      return true;
+    }
+    // Lite chain deposits force repayment on origin chain.
+    const repaymentChainId = matchedDeposit.fromLiteChain ? fill.originChainId : fill.repaymentChainId;
+    // Return undefined if the requested repayment chain ID is not recognized by the hub pool.
+    const possibleRepaymentChainIds = this.configStoreClient?.getEnabledChains();
+    if (isDefined(possibleRepaymentChainIds) && !possibleRepaymentChainIds.includes(repaymentChainId)) {
```
pxrl, line 348: You might be able to consolidate this a little bit by defaulting to an empty array, though I'm unclear what is intended to happen here during test (i.e. when no ConfigStoreClient is available, so `possibleRepaymentChainIds` will revert to `undefined`. It looks like we just assume repayment to any chain is OK in that case?

```suggestion
    const possibleRepaymentChainIds = this.configStoreClient?.getEnabledChains() ?? [];
    if (!possibleRepaymentChainIds.includes(repaymentChainId)) {
```

src/clients/SpokePoolClient.ts
pxrl: The function is named `isEvmRepaymentValid()`, so the existing implementation seems correct to me. We'd obviously need to revise this for a non-EVM chain though.

src/clients/SpokePoolClient.ts
```diff
@@ -334,6 +336,24 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+      return false;
+    }
+    if (!chainIsEvm(repaymentChainId)) {
+      return false;
+    }
+    return isValidEvmAddress(fill.relayer);
```
pxrl, line 354: Small nit - can maybe squash this a bit. I think this helps readability.

```suggestion
    return chainIsEvm(repaymentChainId)) && isValidEvmAddress(fill.relayer);
```

src/clients/SpokePoolClient.ts
```diff
@@ -334,6 +336,24 @@ export class SpokePoolClient extends BaseAbstractClient {
     return fills?.find((fill) => validateFillForDeposit(fill, deposit));
   }
 
+  private isEvmRepaymentValid(fill: Fill, matchedDeposit: Deposit): boolean {
+    // Slow fills don't result in repayments so they're always valid.
+    if (isSlowFill(fill)) {
+      return true;
+    }
+    // Lite chain deposits force repayment on origin chain.
+    const repaymentChainId = matchedDeposit.fromLiteChain ? fill.originChainId : fill.repaymentChainId;
```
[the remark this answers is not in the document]
nicholaspai: yeah i tried to do this a bit. thought about it some more--what do you think of this? 6e6488cc10e86b43f25707058d13966a43742897 I can migrate the changes to this PR if we like something like this.

bmzig (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -407,6 +421,17 @@ export class SpokePoolClient extends BaseAbstractClient {
...
     }
+    const unrepayableFillsForDeposit = unrepayableFills.filter((x) => x.depositId.eq(deposit.depositId));
+    if (unrepayableFillsForDeposit.length > 0) {
+      this.logger.warn({
+        at: "SpokePoolClient",
+        chainId: this.chainId,
+        message: "Unrepayable fills found.",
+        deposit,
+        unrepayableFills: Object.fromEntries(unrepayableFillsForDeposit.map((x) => [x.relayer, x])),
+        notificationPath: "across-invalid-fills",
```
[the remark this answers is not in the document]
bmzig: Yeah good call to do that.

bmzig (approved):

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
nicholaspai: updated to across-unrepayable-fills

PR #856, opened 2025-01-31
improve: do not force address coercion for spoke events

bmzig:
While right now, we know that spoke pool events will only emit bytes20 addresses, once the contracts are upgraded, anybody can supply bytes32 addresses to deposit/fill/speedups/etc. functions. This means we should not force these addresses to be bytes20.

### DIFF
--- a/src/utils/EventUtils.ts
+++ b/src/utils/EventUtils.ts
@@ -90,6 +90,7 @@ export function spreadEvent(args: Result | Record<string, unknown>): { [key: str
       try {
         address = toAddress(address);
       } catch (_e) {
+        // Lint appeasement
         _e;
       }
       returnedObject[field] = address;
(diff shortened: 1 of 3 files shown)


src/utils/EventUtils.ts
```diff
@@ -86,7 +86,13 @@ export function spreadEvent(args: Result | Record<string, unknown>): { [key: str
   // Truncate all fields which may be bytes32 into a bytes20 string.
   for (const field of knownExtendedAddressFields) {
     if (isDefined(returnedObject[field])) {
-      returnedObject[field] = toAddress(String(returnedObject[field]));
+      let address = String(returnedObject[field]);
+      try {
+        address = toAddress(address);
+      } catch (_e) {
+        _e;
```
nicholaspai, line 93: are we throwing e? logging it?

nicholaspai (commented):
One q

src/utils/EventUtils.ts
bmzig: This is just lint appeasement. Added comment about that here: https://github.com/across-protocol/sdk/pull/856/commits/82154e41da9b8c0f3191d3a0ae5e7fb2fcc6cfac

nicholaspai (approved):

PR #859, opened 2025-02-04
Doctor who

gsteenkamp89:


src/utils/NetworkUtils.ts
```diff
@@ -114,6 +114,15 @@ export function chainIsLinea(chainId: number): boolean {
...
 }
 
+/**
+ * Determines whether a chain ID is a DoctorWho implementation.
+ * @param chainId Chain ID to evaluate.
+ * @returns True if chainId is a DoctorWho chain, otherwise false.
+ */
+export function chainIsDoctorWho(chainId: number): boolean {
+  return [CHAIN_IDs.DOCTOR_WHO].includes(chainId);
+}
```
james-a-morris, line 124: We should probably clean up all these specific on-shot functions (for the future)

james-a-morris (approved):

src/utils/NetworkUtils.ts
gsteenkamp89: I'm not really sure what we use these for, or even if they're needed tbh, but it was in the ticket. We only have them for a couple of chains anyway

src/utils/Multicall.ts
```diff
@@ -26,6 +26,7 @@ const NON_DETERMINISTIC_MULTICALL_ADDRESSES = {
 const DETERMINISTIC_MULTICALL_CHAINS = [
   CHAIN_IDs.ALEPH_ZERO,
   CHAIN_IDs.ARBITRUM,
+  CHAIN_IDs.DOCTOR_WHO,
```
pxrl, line 29: @gsteenkamp89 This isn't necessary for OP stack chains (see line 41 below).

src/utils/NetworkUtils.ts
pxrl: We don't need this :+1: 
```suggestion
```

pxrl (commented):
@gsteenkamp89 Were you able to confirm the block time for this chain? You may want to add a manual entry to src/utils/BlockUtils.ts.

src/utils/Multicall.ts
gsteenkamp89: ah its a preinstall, got it 👍

pxrl (approved):

PR #860, opened 2025-02-04
refactor: Relocate event-to-string utility function

pxrl:
This functionality can be used in more places than just the BundleDataClient. Refactor it out and add a simple test.


nicholaspai:
Can we add this to the spoke pool client and replace any instances of getDepositHash?

pxrl:
> Can we add this to the spoke pool client and replace any instances of getDepositHash?

https://github.com/across-protocol/sdk/pull/860/commits/1b60cf7bec88129121ae30487fa434c1e11d2c5c

nicholaspai:
Cool so now the only call to getRelayDataHash is in SpokeUtils now, right?

src/utils/SpokeUtils.ts
```diff
@@ -57,6 +57,33 @@ export function populateV3Relay(
...
 }
 
+/**
+ * Concatenate all fields from a Deposit, Fill or SlowFillRequest into a single string.
+ * This can be used to identify a bridge event in a mapping. This is used instead of the actual keccak256 hash
+ * (getRelayDataHash()) for two reasons: performance and the fact that only Deposit includes the `message` field, which
+ * is required to compute a complete RelayData hash.
+ * note: This function should _not_ be used to query the SpokePool.fillStatuses mapping.
+ */
+export function getRelayEventKey(data: RelayData & { destinationChainId: number }): string {
```
nicholaspai, line 67: I like this name so it's not as easily confused with the keccak256 hash used by the contracts

src/utils/SpokeUtils.ts
```diff
@@ -57,6 +57,33 @@ export function populateV3Relay(
...
+    data.inputToken,
+    data.outputToken,
+    data.inputAmount,
+    data.outputAmount,
+    data.originChainId,
+    data.destinationChainId,
+    data.depositId,
+    data.fillDeadline,
+    data.exclusivityDeadline,
+    data.message,
```
nicholaspai, line 81: So all we need to do is replace this with messageHash once that's set in the spoke pool clients right?

test/SpokeUtils.ts
```diff
@@ -1,8 +1,47 @@
...
+    const data = {
+      originChainId: random(),
+      destinationChainId: random(),
+      depositor: randomAddress(),
+      recipient: randomAddress(),
+      inputToken: randomAddress(),
+      inputAmount: toBN(random()),
+      outputToken: randomAddress(),
+      outputAmount: toBN(random()),
+      message: "0x",
```
nicholaspai, line 18: Nit can use randomBytes32 here to add entropy

nicholaspai (approved):

bmzig (approved):

test/SpokeUtils.ts
pxrl: https://github.com/across-protocol/sdk/pull/860/commits/fd27acbcf7b39863277892558a17fe1c7d2ccdd9

src/utils/SpokeUtils.ts
pxrl: Yeah; that gets us over the event hurdle, after which we won't have `message` available in a subset of events.

src/clients/SpokePoolClient.ts
```diff
@@ -297,8 +298,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @returns The corresponding deposit if found, undefined otherwise.
    */
   public getDeposit(depositId: BigNumber): DepositWithBlock | undefined {
-    const depositHash = this.getDepositHash({ depositId, originChainId: this.chainId });
-    return this.depositHashes[depositHash];
+    return Object.values(this.depositHashes).find(({ depositId: _depositId }) => _depositId.eq(depositId));
```
pxrl, line 301: For `getDeposit()` we have to revert to a search because the only input available is the depositId. This is not robust to duplicate depositIds.

PR #862, opened 2025-02-05
refactor(test): Test tweaks in advance of pending event changes

pxrl:
This change makes some test adjustments to reduce the diff in a subsequent PR, relating to the introduction of new event types. It also takes advantages of some recent simplifications in the event mocking framework to remove some annoying type assertions.


bmzig (approved):

test/SpokePoolClient.v3Events.ts
```diff
@@ -362,7 +373,10 @@ describe("SpokePoolClient: Event Filtering", function () {
         exclusiveRelayer,
       } as DepositWithBlock);
       await originSpokePoolClient.update(["V3FundsDeposited"]);
-      const deposit = originSpokePoolClient.getDeposits().at(-1);
+      let deposit = originSpokePoolClient.getDeposits().at(-1);
```
nicholaspai, line 376: This looks great. Do we have any tests where we mock "old" events (not V2, but V3 pre-bytes32 upgrade)? Would be nice to see spoke pool client query both in the same test and produce the same internal data structures via getDeposits/getFills

pxrl: Yeah, the branch pxrl/filledRelayMessageHash does this in its tests.

PR #863, opened 2025-02-05
fix(FillUtils): verifyFillRepayment should check that repayment chain has pool rebalance route

nicholaspai:
We cannot refund fills whose repayment chain doesn't exist for an input token.


src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -75,12 +66,24 @@ export async function verifyFillRepayment(
...
+  try {
+    const l1TokenCounterpart = hubPoolClient.getL1TokenForL2TokenAtBlock(
+      fill.inputToken,
+      fill.originChainId,
+      matchedDeposit.quoteBlockNumber
+    );
+    hubPoolClient.getL2TokenForL1TokenAtBlock(l1TokenCounterpart, repaymentChainId, matchedDeposit.quoteBlockNumber);
+    // Repayment token could be found, this is a valid repayment chain.
+  } catch {
+    // Repayment token doesn't exist on repayment chain via PoolRebalanceRoutes, impossible to repay filler.
```
pxrl, line 84: Do we want to log this via unpayable-refunds?

bmzig (approved):
Fix looks correct

src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -75,12 +66,26 @@ export async function verifyFillRepayment(
...
   matchedDeposit: DepositWithBlock,
-  possibleRepaymentChainIds: number[] = []
+  hubPoolClient: HubPoolClient
 ): Promise<FillWithBlock | undefined> {
   const fill = _.cloneDeep(_fill);
 
   const repaymentChainId = getRepaymentChainId(fill, matchedDeposit);
-  const validEvmRepayment = isEvmRepaymentValid(fill, repaymentChainId, possibleRepaymentChainIds);
+  const validEvmRepayment = isEvmRepaymentValid(fill, repaymentChainId);
+  if (!isSlowFill(fill)) {
```
pxrl, line 75: Should we be calling `verifyFillRepayment()` on slow fills?

src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -75,12 +66,26 @@ export async function verifyFillRepayment(
...
+    try {
+      const l1TokenCounterpart = hubPoolClient.getL1TokenForL2TokenAtBlock(
+        fill.inputToken,
+        fill.originChainId,
+        matchedDeposit.quoteBlockNumber
+      );
+      hubPoolClient.getL2TokenForL1TokenAtBlock(l1TokenCounterpart, repaymentChainId, matchedDeposit.quoteBlockNumber);
+      // Repayment token could be found, this is a valid repayment chain.
+    } catch {
```
mrice32, line 86: This universal try catch is quite dangerous. Should we fix this now or in a follow up.

One way to mostly fix it is to make the hub error a specific type and catch only that werror type.

src/clients/BundleDataClient/utils/FillUtils.ts
nicholaspai: I think it depends where you want to make the `isSlowFill` check. Currently we call verifyFillRepayment before `isSlowFill` so this is best for reducing diff in the BundleDataClient for now

src/clients/BundleDataClient/utils/FillUtils.ts
nicholaspai: changed behavior to switch to destination chain

bmzig (approved):

src/clients/BundleDataClient/utils/FillUtils.ts
mrice32: Still dangerous as a bug will silently change repayment location, causing invalid bundles.

We can resolve in follow up.

nicholaspai: but wouldn't that indicate the bug that causes the invalid bundles is an invalid code change?

mrice32 (approved):

src/clients/BundleDataClient/utils/FillUtils.ts
mrice32: We want an error in an inner function to bubble up, not silently create bad bundles.

My point is that we're assuming the error is X when we could be catching error Y without realizing it.

src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -49,24 +50,31 @@ export function getRefundInformationFromFill(
 
 export function getRepaymentChainId(fill: Fill, matchedDeposit: Deposit): number {
   // Lite chain deposits force repayment on origin chain.
-  return matchedDeposit.fromLiteChain ? fill.originChainId : fill.repaymentChainId;
+  return matchedDeposit.fromLiteChain ? matchedDeposit.originChainId : fill.repaymentChainId;
 }
 
-export function isEvmRepaymentValid(
-  fill: Fill,
+export function willOverwriteRepaymentChain(
```
bmzig, line 56: nit: I'd probably prefer something like `forceDestinationRepayment` since `willOverwriteRepaymentChain` seems a bit ambiguous.

bmzig (approved):
One (potentially non-actionable) comment.

src/clients/BundleDataClient/utils/FillUtils.ts
nicholaspai: i like that its more precise

src/clients/BundleDataClient/utils/FillUtils.ts
pxrl: > This universal try catch is quite dangerous. Should we fix this now or in a follow up?
> 
> One way to mostly fix it is to make the hub error a specific type and catch only that werror type.

We started on this a while ago and should pick it up again: https://github.com/across-protocol/sdk/pull/643

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -365,26 +363,20 @@ export class BundleDataClient {
...
       });
-      fillsToCount.forEach((fill) => {
-        const matchingDeposit = this.spokePoolClients[fill.originChainId].getDeposit(fill.depositId);
+      await forEachAsync(fillsToCount, async (_fill) => {
+        const matchingDeposit = this.spokePoolClients[_fill.originChainId].getDeposit(_fill.depositId);
         assert(isDefined(matchingDeposit), "Deposit not found for fill.");
+        const fill = await verifyFillRepayment(
+          _fill,
+          this.spokePoolClients[_fill.destinationChainId].spokePool.provider,
+          matchingDeposit!,
```
pxrl, line 374: Driveby question - any idea why it's necessary to assert the existence of `matchingDeposit` here when the check above should guarantee it?

src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -75,20 +83,24 @@ export async function verifyFillRepayment(
...
   }
-  // Case 2: Repayment chain is EVM but repayment address is not a valid EVM address. Attempt to switch repayment
-  // address to msg.sender of relay transaction.
-  else if (chainIsEvm(repaymentChainId) && !isValidEvmAddress(fill.relayer)) {
+
+  let repaymentChainId = getRepaymentChainId(fill, matchedDeposit);
+
+  // If repayment chain doesn't have a Pool Rebalance Route for the input token, then change the repayment
+  // chain to the destination chain.
+  if (!isSlowFill(fill) && forceDestinationRepayment(repaymentChainId, matchedDeposit, hubPoolClient)) {
```
pxrl, line 99: We already return early for a slow fill so the initial check of `isSlowFill()` should be redundnat.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: I think because TSC doesn't interpret asserts the same way it does if statements. We could change the above assert to an if statement to avoid this !

src/clients/BundleDataClient/utils/FillUtils.ts
nicholaspai: ah yes good catch

nicholaspai: https://github.com/across-protocol/sdk/pull/863/commits/68b52fce4f3cb32350a288de1aa64e9cbd718bae

src/clients/BundleDataClient/BundleDataClient.ts
pxrl: I figured it out - it's because we have some locally-defined `assert` that instead throws an exception.

This diff fixes it so that tsc can infer things from the use of asserts:

```
diff --git a/src/clients/BundleDataClient/BundleDataClient.ts b/src/clients/BundleDataClient/BundleDataClient.ts
index cb0f4332..d9dc2469 100644
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -1,3 +1,4 @@
+import assert from "assert";
 import _ from "lodash";
 import {
   ProposedRootBundle,
@@ -24,7 +25,6 @@ import {
   bnZero,
   queryHistoricalDepositForFill,
   assign,
-  assert,
   fixedPointAdjustment,
   isDefined,
   toBN,
@@ -371,7 +371,7 @@ export class BundleDataClient {
         const fill = await verifyFillRepayment(
           _fill,
           this.spokePoolClients[_fill.destinationChainId].spokePool.provider,
-          matchingDeposit!,
+          matchingDeposit,
           this.clients.hubPoolClient
         );
         if (!isDefined(fill)) {
~
```

PR #865, opened 2025-02-06
feat(SpokePoolClient): Initial new event support

pxrl:
These new events are supported in test, but not in production. A contract bump w/ new ABIs is still required for production support, but this commit at least frontloads a lot of the necessary changes.

nb. most of the SpokePoolClient changes here were pulled out of the existing branch bz/eventIngestion.


test/SpokePoolClient.ValidateFill.ts
```diff
@@ -751,7 +751,7 @@ describe("SpokePoolClient: Fill Validation", function () {
     expect((result as { reason: string }).reason.startsWith("outputAmount mismatch")).to.be.true;
 
     // Invalid depositId.
-    result = validateFillForDeposit({ ...fill, depositId: 1337 }, deposit_2);
+    result = validateFillForDeposit({ ...fill, depositId: toBN(1337) }, deposit_2);
```
pxrl, line 754: Driveby fix.

src/clients/SpokePoolClient.ts
```diff
@@ -648,11 +653,18 @@ export class SpokePoolClient extends BaseAbstractClient {
           this.latestDepositIdQueried = deposit.depositId;
         }
       }
+    };
+
+    // Query "legacy" V3FundsDeposited events.
+    for (const event of ["V3FundsDeposited", "FundsDeposited"]) {
```
nicholaspai, line 659: `FundsDeposited` are new right? Comment is confusing me

src/interfaces/SpokePool.ts
```diff
@@ -52,7 +52,7 @@ export enum FillType {
 export interface RelayExecutionEventInfo {
   updatedRecipient: string;
   updatedOutputAmount: BigNumber;
-  updatedMessage?: string;
+  updatedMessage: string;
```
nicholaspai, line 55: Why is this possible now?

src/clients/SpokePoolClient.ts
pxrl: Ah - I'll nuke the comment.

src/utils/AddressUtils.ts
```diff
@@ -39,10 +39,14 @@ export function compareAddressesSimple(addressA?: string, addressB?: string): bo
...
 }
 
-// Converts an input (assumed to be) bytes32 string into a bytes20 string.
-// If the input is not a bytes32 but is less than type(uint160).max, then this function
-// will still succeed.
-// Throws an error if the string as an unsigned integer is greater than type(uint160).max.
+// Converts an input hex data string into a bytes32 string. Note that the output bytes will be lowercase
+// so that it naturally matches with ethers event data.
+// Throws an error if the input string is already greater than 32 bytes.
+export function toBytes32(address: string): string {
```
nicholaspai, line 45: Should there be a check here that the address is a valid evm address (i.e. has 20 bytes)? Or does that not matter

nicholaspai (commented):
Couple nits

src/utils/AddressUtils.ts
pxrl: We could validate w/ ethers getAddress, but that introduces an issue of what to do if it doesn't validate. Ultimately we probably don't want the bot to throw here, but we also need to return a string. I think we need to sharpen this handling up e2e but idk whether we should make the upgrade conditional on that. wdyt?

src/clients/SpokePoolClient.ts
pxrl: https://github.com/across-protocol/sdk/pull/865/commits/c4d6936e07d558291f523f90aa57acd13f7a3e73

src/interfaces/SpokePool.ts
pxrl: I'd hit some type errors but I just took a 2nd pass on those and I can work around them OK. https://github.com/across-protocol/sdk/pull/865/commits/3ec724a4d233225ca0b0ddbea93ffd3746c2f9ce

src/utils/AddressUtils.ts
bmzig: I think a possible advantage of not checking is being able to put in either bytes20 or bytes32 addresses into here and getting back bytes32 addresses regardless.

nicholaspai (approved):

src/clients/SpokePoolClient.ts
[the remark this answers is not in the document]
pxrl: I don't want to do that until the ABI is in, because otherwise we rely on the ABI and it'd crash if it was run in prod.

PR #866, opened 2025-02-06
chore(SpokePoolClient): Drop fill (etc) message field

pxrl:
Hoist all utils over to use messageHash instead.


src/constants.ts
```diff
@@ -35,7 +35,11 @@ export const PRE_FILL_MIN_CONFIG_STORE_VERSION = 5;
 export const ARWEAVE_TAG_APP_NAME = "across-protocol";
 
 // A hardcoded version number used, by default, to tag all Arweave records.
-export const ARWEAVE_TAG_APP_VERSION = 3;
+// Version 3:
+// - Changed depositId number -> string (uint32 -> uint256).
+// Version 3:
```
nicholaspai, line 40: should this be 4?
```suggestion
// Version 4:
```

nicholaspai (approved):

bmzig (approved):

PR #868, opened 2025-02-06
fix: Promote addresses to bytes32 before comparison

pxrl:


src/utils/FlowUtils.ts
```diff
@@ -31,7 +31,17 @@ export function validateFillForDeposit(
...
   // TODO: if we turn on "strict" in the tsconfig, the elements of FILL_DEPOSIT_COMPARISON_KEYS will be automatically
   // validated against the fields in Fill and Deposit, generating an error if there is a discrepency.
-  let invalidKey = RELAYDATA_KEYS.find((key) => relayData[key].toString() !== deposit[key].toString());
+  let invalidKey = RELAYDATA_KEYS.find((key) => {
+    const [depositField, fillField] = [deposit[key].toString(), relayData[key].toString()];
+
+    // Promote address types to bytes32 to avoid mismatches.
+    // @todo: This must be hardened on the interfaces where these values are ingested.
+    if (["depositor", "recipient", "inputToken", "outputToken", "exclusiveRelayer"].includes(key)) {
+      return toBytes32(depositField) !== toBytes32(fillField);
```
pxrl, line 40: This could fwiw go bad if one of these addresses was incorrectly truncated where the upper bytes were not zeroed, because we would then backfill those bytes with zeroes. We have to be sure that we're not doing that anywhere.

src/utils/FlowUtils.ts
```diff
@@ -31,7 +31,17 @@ export function validateFillForDeposit(
   // Note: this short circuits when a key is found where the comparison doesn't match.
   // TODO: if we turn on "strict" in the tsconfig, the elements of FILL_DEPOSIT_COMPARISON_KEYS will be automatically
   // validated against the fields in Fill and Deposit, generating an error if there is a discrepency.
-  let invalidKey = RELAYDATA_KEYS.find((key) => relayData[key].toString() !== deposit[key].toString());
+  let invalidKey = RELAYDATA_KEYS.find((key) => {
+    const [depositField, fillField] = [deposit[key].toString(), relayData[key].toString()];
+
+    // Promote address types to bytes32 to avoid mismatches.
+    // @todo: This must be hardened on the interfaces where these values are ingested.
+    if (["depositor", "recipient", "inputToken", "outputToken", "exclusiveRelayer"].includes(key)) {
```
nicholaspai, line 39: These keys are also in `knownExtendedAddressFields` in /src/utils/EventUtils

PR #869, opened 2025-02-06
fix: allow collisions for deposit to fill mapping

bmzig:
This is needed so that we can find duplicate fills (i.e. fills with different relay data but with the same deposit ID). E.g. we need to have collisions here so that we can determine invalid fills https://github.com/across-protocol/sdk/blob/2a10e145d028f9c3c6f47ed0c65fc16256c5aa42/src/clients/SpokePoolClient.ts#L384

### DIFF
--- a/src/clients/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient.ts
@@ -364,7 +364,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @returns A valid fill for the deposit, or undefined.
    */
   public getFillForDeposit(deposit: Deposit): FillWithBlock | undefined {
-    const fills = this.depositHashesToFills[getRelayEventKey(deposit)];
+    const fills = this.depositHashesToFills[this.getDepositHash(deposit)];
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):

src/clients/SpokePoolClient.ts
```diff
@@ -364,7 +364,7 @@ export class SpokePoolClient extends BaseAbstractClient {
    * @returns A valid fill for the deposit, or undefined.
    */
   public getFillForDeposit(deposit: Deposit): FillWithBlock | undefined {
```
nicholaspai, line 366: If you have time can you add a very simple unit test here you create two fills for same `getDepositHash()` and test that it returns > 1 ? Just for regression prevention

nicholaspai: https://github.com/across-protocol/sdk/pull/869/commits/e87700f265e1f94288c251d7f91d7afe81e287fc

nicholaspai:
This relayer PR should be merged once we release this SDK: https://github.com/across-protocol/relayer/pull/2051

PR #873, opened 2025-02-07
fix(test): Correct testing of FilledRelay messageHash

pxrl:
This change fixes an issue in the way the MockSpokePoolClient produces FilledRelay events. It incorrectly bundles the deposit message field into the fill. When combined with a separate (already fixed) issue in the SpokePoolClient that was unintentionally overwriting the FilledRelay messageHash field based on the message field, this meant that the messageHash was unintentionally _correctly_ recomputed in test, but not in production.

### DIFF
--- a/src/clients/mocks/MockSpokePoolClient.ts
+++ b/src/clients/mocks/MockSpokePoolClient.ts
@@ -237,7 +237,6 @@ export class MockSpokePoolClient extends SpokePoolClient {
       relayer,
       depositor,
       recipient,
-      message,
       relayExecutionInfo: {
         updatedRecipient: fill.relayExecutionInfo?.updatedRecipient ?? recipient,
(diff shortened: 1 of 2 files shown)


src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -237,7 +237,6 @@ export class MockSpokePoolClient extends SpokePoolClient {
       relayer,
       depositor,
       recipient,
-      message,
       relayExecutionInfo: {
         updatedRecipient: fill.relayExecutionInfo?.updatedRecipient ?? recipient,
         updatedMessage,
```
pxrl, line 242:
```suggestion
```

bmzig (approved):

nicholaspai (approved):

PR #874, opened 2025-02-07
Invalid fill bundle logging

nicholaspai:


src/clients/SpokePoolClient.ts
```diff
@@ -354,16 +354,7 @@ export class SpokePoolClient extends BaseAbstractClient {
...
       return deposit;
     }
-
-    this.logger.debug({
-      at: "SpokePoolClient::getDepositForFill",
-      message: `Rejected fill for ${getNetworkName(fill.originChainId)} deposit ${fill.depositId.toString()}.`,
-      reason: match.reason,
-      deposit,
-      fill,
-    });
```
pxrl, line 364: I'm in two minds about dropping this, because we don't log the reason for the mismatch in the new code.

nicholaspai: well we don't call this function elsewhere so its only used in the bundle data client currently

PR #875, opened 2025-02-07
fix(BundleDataClient): Handle updated Arweave formatting

pxrl:


pxrl:
fwiw I can retrieve and parse arweave bundle data with this change. It might be a bit more than we strictly need, though.

bmzig (approved):
Can you also bump package version?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.1.2",
+  "version": "4.1.3",
```
nicholaspai, line 4: needs to be 4.1.4

src/clients/BundleDataClient/utils/SuperstructUtils.ts
```diff
@@ -85,12 +85,13 @@ const V3RelayExecutionEventInfoSS = object({
   updatedOutputAmount: BigNumberType,
   fillType: FillTypeSS,
   updatedRecipient: string(),
-  updatedMessage: string(),
+  updatedMessage: optional(string()),
```
nicholaspai, line 88: I thought we didn't want  optional types?

nicholaspai (approved):

src/clients/BundleDataClient/utils/SuperstructUtils.ts
pxrl: In this case I think optional() is preferred because it allows for the presence of message, but doesn't auto-populate it in the retrieved data. After 7 days we'll be OK to remove this.

PR #884, opened 2025-02-10
improve(BundleDataClient,SpokePoolClient): Log about duplicate events and delete `getLatestProposedBundleData`

nicholaspai:
To be paired with https://github.com/across-protocol/relayer/pull/2085 in order to speed up relayer marginally when pending bundle data is not yet available on Arweave. Also adds safety checks to BundleDataClient in case there are duplicate events

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -456,28 +456,21 @@ export class BundleDataClient {
     if (!isDefined(bundleDataOnArweave)) {
       return undefined;
     }
-    const pendingBundleData = await this.loadData(bundleBlockRanges, this.spokePoolClients, true);
-    // This should be defined since we've checked that the data exists on Arweave, but handle the unexpected
-    // case where its not.
-    if (!isDefined(pendingBundleData)) {
-      return undefined;
-    } else {
-      const hubPoolClient = this.clients.hubPoolClient;
-      const root = _buildPoolRebalanceRoot(
-        hubPoolClient.latestBlockSearched,
-        bundleBlockRanges[0][1],
-        pendingBundleData.bundleDepositsV3,
-        pendingBundleData.bundleFillsV3,
-        pendingBundleData.bundleSlowFillsV3,
-        pendingBundleData.unexecutableSlowFills,
-        pendingBundleData.expiredDepositsToRefundV3,
-        {
-          hubPoolClient,
-          configStoreClient: hubPoolClient.configStoreClient as AcrossConfigStoreClient,
-        }
-      );
-      return root;
-    }
+    const pendingBundleData = await this.loadArweaveData(bundleBlockRanges);
+    const root = _buildPoolRebalanceRoot(
+      hubPoolClient.latestBlockSearched,
+      bundleBlockRanges[0][1],
+      pendingBundleData.bundleDepositsV3,
+      pendingBundleData.bundleFillsV3,
+      pendingBundleData.bundleSlowFillsV3,
+      pendingBundleData.unexecutableSlowFills,
+      pendingBundleData.expiredDepositsToRefundV3,
+      {
+        hubPoolClient,
+        configStoreClient: hubPoolClient.configStoreClient as AcrossConfigStoreClient,
+      }
+    );
+    return root;
   }
 
   // @dev This function should probably be moved to the InventoryClient since it bypasses loadData completely now.
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -885,14 +885,7 @@ export class BundleDataClient {
             fillCounter++;
             const relayDataHash = getRelayEventKey(fill);
             if (v3RelayHashes[relayDataHash]) {
-              if (v3RelayHashes[relayDataHash].fill) {
-                this.logger.debug({
-                  at: "BundleDataClient#loadData",
-                  message: "Duplicate fill detected",
-                  fill,
-                });
-                throw new Error("Duplicate fill detected");
-              }
+              if (!v3RelayHashes[relayDataHash].fill) {
                 const { deposits } = v3RelayHashes[relayDataHash];
                 assert(isDefined(deposits) && deposits.length > 0, "Deposit should exist in relay hash dictionary.");
                 v3RelayHashes[relayDataHash].fill = fill;
@@ -944,63 +937,71 @@ export class BundleDataClient {
                   ) {
                     fastFillsReplacingSlowFills.push(relayDataHash);
                   }
-              }
-              assert(
-                isDefined(v3RelayHashes[relayDataHash].deposits) && v3RelayHashes[relayDataHash].deposits!.length > 0,
-                "Deposit should exist in relay hash dictionary."
-              );
-              v3RelayHashes[relayDataHash].fill = fill;
-              if (fill.blockNumber >= destinationChainBlockRange[0]) {
-                const fillToRefund = await verifyFillRepayment(
-                  fill,
-                  destinationClient.spokePool.provider,
-                  v3RelayHashes[relayDataHash].deposits![0],
-                  this.clients.hubPoolClient
+                }
+                assert(
+                  isDefined(v3RelayHashes[relayDataHash].deposits) && v3RelayHashes[relayDataHash].deposits!.length > 0,
+                  "Deposit should exist in relay hash dictionary."
                 );
-                if (!isDefined(fillToRefund)) {
-                  bundleUnrepayableFillsV3.push(fill);
-                  // We don't return here yet because we still need to mark unexecutable slow fill leaves
-                  // or duplicate deposits. However, we won't issue a fast fill refund.
-                } else {
-                  v3RelayHashes[relayDataHash].fill = fillToRefund;
-                  validatedBundleV3Fills.push({
-                    ...fillToRefund,
-                    quoteTimestamp: v3RelayHashes[relayDataHash].deposits![0].quoteTimestamp,
-                  });
+                v3RelayHashes[relayDataHash].fill = fill;
+                if (fill.blockNumber >= destinationChainBlockRange[0]) {
+                  const fillToRefund = await verifyFillRepayment(
+                    fill,
+                    destinationClient.spokePool.provider,
+                    v3RelayHashes[relayDataHash].deposits![0],
+                    this.clients.hubPoolClient
+                  );
+                  if (!isDefined(fillToRefund)) {
+                    bundleUnrepayableFillsV3.push(fill);
+                    // We don't return here yet because we still need to mark unexecutable slow fill leaves
+                    // or duplicate deposits. However, we won't issue a fast fill refund.
+                  } else {
+                    v3RelayHashes[relayDataHash].fill = fillToRefund;
+                    validatedBundleV3Fills.push({
+                      ...fillToRefund,
+                      quoteTimestamp: v3RelayHashes[relayDataHash].deposits![0].quoteTimestamp,
+                    });
 
-                  // Now that we know this deposit has been filled on-chain, identify any duplicate deposits
-                  // sent for this fill and refund them to the filler, because this value would not be paid out
-                  // otherwise. These deposits can no longer expire and get refunded as an expired deposit,
-                  // and they won't trigger a pre-fill refund because the fill is in this bundle.
-                  // Pre-fill refunds only happen when deposits are sent in this bundle and the
-                  // fill is from a prior bundle. Paying out the filler keeps the behavior consistent for how
-                  // we deal with duplicate deposits regardless if the deposit is matched with a pre-fill or
-                  // a current bundle fill.
-                  const duplicateDeposits = v3RelayHashes[relayDataHash].deposits!.slice(1);
-                  duplicateDeposits.forEach((duplicateDeposit) => {
-                    if (isSlowFill(fill)) {
-                      updateExpiredDepositsV3(expiredDepositsToRefundV3, duplicateDeposit);
-                    } else {
-                      validatedBundleV3Fills.push({
-                        ...fillToRefund,
-                        quoteTimestamp: duplicateDeposit.quoteTimestamp,
-                      });
-                    }
-                  });
+                    // Now that we know this deposit has been filled on-chain, identify any duplicate deposits
+                    // sent for this fill and refund them to the filler, because this value would not be paid out
+                    // otherwise. These deposits can no longer expire and get refunded as an expired deposit,
+                    // and they won't trigger a pre-fill refund because the fill is in this bundle.
+                    // Pre-fill refunds only happen when deposits are sent in this bundle and the
+                    // fill is from a prior bundle. Paying out the filler keeps the behavior consistent for how
+                    // we deal with duplicate deposits regardless if the deposit is matched with a pre-fill or
+                    // a current bundle fill.
+                    const duplicateDeposits = v3RelayHashes[relayDataHash].deposits!.slice(1);
+                    duplicateDeposits.forEach((duplicateDeposit) => {
+                      if (isSlowFill(fill)) {
+                        updateExpiredDepositsV3(expiredDepositsToRefundV3, duplicateDeposit);
+                      } else {
+                        validatedBundleV3Fills.push({
+                          ...fillToRefund,
+                          quoteTimestamp: duplicateDeposit.quoteTimestamp,
+                        });
+                      }
+                    });
+                  }
+
+                  // If fill replaced a slow fill request, then mark it as one that might have created an
+                  // unexecutable slow fill. We can't know for sure until we check the slow fill request
+                  // events.
+                  if (
+                    fill.relayExecutionInfo.fillType === FillType.ReplacedSlowFill &&
+                    _canCreateSlowFillLeaf(v3RelayHashes[relayDataHash].deposits![0])
+                  ) {
+                    fastFillsReplacingSlowFills.push(relayDataHash);
+                  }
                 }
 
-                // If fill replaced a slow fill request, then mark it as one that might have created an
-                // unexecutable slow fill. We can't know for sure until we check the slow fill request
-                // events.
-                if (
-                  fill.relayExecutionInfo.fillType === FillType.ReplacedSlowFill &&
-                  _canCreateSlowFillLeaf(v3RelayHashes[relayDataHash].deposits![0])
-                ) {
-                  fastFillsReplacingSlowFills.push(relayDataHash);
-                }
+                return;
               }
-
-              return;
+            } else {
+              this.logger.debug({
+                at: "BundleDataClient#loadData",
+                message: "Duplicate fill detected",
+                fill,
+              });
+              throw new Error("Duplicate fill detected");
             }
 
             // At this point, there is no relay hash dictionary entry for this fill, so we need to
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -88,7 +88,7 @@ function updateBundleDepositsV3(dict: BundleDepositsV3, deposit: V3DepositWithBl
       (d) => d.transactionHash === deposit.transactionHash && d.logIndex === deposit.logIndex
     )
   ) {
-    throw new Error("Duplicate deposit in bundleDeposits");
+    throw new Error(`Duplicate deposit in bundleDeposits: ${JSON.stringify(deposit)}`);
   }
   dict[originChainId][inputToken].push(deposit);
 }
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -464,8 +464,8 @@ export class BundleDataClient {
     // validated bundle, otherwise we'll grab the second most recently validated bundle.
     let n = hubPoolClient.hasPendingProposal() ? 1 : 2;
 
-    // eslint-disable-next-line no-constant-condition
-    while (true) {
+    // Try to find an older bundle that has arweave data but cut off after a few tries and just load data from scratch.
+    while (n < 4) {
       const bundleDataOnArweave = await this.getBundleDataFromArweave(bundleBlockRanges);
       if (!isDefined(bundleDataOnArweave)) {
         // Bundle data is not arweave, try the next most recently validated bundle.
@@ -475,14 +475,16 @@ export class BundleDataClient {
           hubPoolClient.getNthFullyExecutedRootBundle(-n)!
         );
       } else {
-        return {
-          blockRanges: bundleBlockRanges,
-          bundleData: await this.loadData(bundleBlockRanges, this.spokePoolClients, true),
-        };
+        break;
       }
 
       n++;
     }
+
+    return {
+      blockRanges: bundleBlockRanges,
+      bundleData: await this.loadData(bundleBlockRanges, this.spokePoolClients, true),
+    };
   }
 
   async getLatestPoolRebalanceRoot(): Promise<{ root: PoolRebalanceRoot; blockRanges: number[][] }> {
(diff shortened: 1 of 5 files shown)


src/clients/SpokePoolClient.ts
```diff
@@ -656,6 +656,21 @@ export class SpokePoolClient extends BaseAbstractClient {
         }
 
         if (this.depositHashes[getRelayEventKey(deposit)] !== undefined) {
+          // Sanity check that this event is not a duplicate, even though the relay data hash is a duplicate.
+          const allDeposits = this._getDuplicateDeposits(deposit).concat(this.depositHashes[getRelayEventKey(deposit)]);
```
bmzig, line 660: I believe only need to check deposit hashes here since this is the block which assigns values to duplicate deposits.

src/clients/SpokePoolClient.ts
```diff
@@ -656,6 +656,21 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+          if (
+            allDeposits.some((e) => {
+              return e.transactionHash === deposit.transactionHash && e.logIndex === deposit.logIndex;
+            })
+          ) {
+            this.logger.warn({
```
bmzig, line 666: If this is ever logged, this basically means that there is a bug somewhere. I think this should be an error.

nicholaspai: this might take down the fast relayer though, no?

src/clients/SpokePoolClient.ts
nicholaspai: I'm not sure what you mean here?

src/clients/SpokePoolClient.ts
bmzig: I think the relayers are the only bots which do not stop on errors, since they error a lot when txns revert in simulation. @pxrl could probably confirm that I'm not mistaken, though.

src/clients/SpokePoolClient.ts
nicholaspai: no its not. We only store a single deposit in `this.depositHashes` and any others get added to `this.duplicateDeposits`

src/clients/SpokePoolClient.ts
nicholaspai: I'm not sure we should crash here. I'd rather just ignore the event and try our best to  keep the events stored in memory correct by filtering out duplicates

src/clients/SpokePoolClient.ts
bmzig: Ok yeah, I see. My mistake.

src/clients/SpokePoolClient.ts
bmzig: Ideally, this should never trigger, though. I'd argue that if this is logged, then we introduced a bug somewhere and should look into recent commits. Right now, we know that this would occasionally log in the relayer, but if this happened somewhere else in the future, we'd probably want to look into it.

nicholaspai: hm so if we crashed here, this would bring down the fast relayer which seems to handlese duplicate events ok today. Are we ok with that? yes it'd help us debug, but also be disprutive to production service. WDYT?

bmzig: If the fast relayer crashes on errors, then I agree, we should not log an error and go with what you currently have. I don't _think_ the fast relayer crashes on logged errors, though. Am I wrong?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -83,6 +83,13 @@ function updateBundleDepositsV3(dict: BundleDepositsV3, deposit: V3DepositWithBl
   if (!dict?.[originChainId]?.[inputToken]) {
     assign(dict, [originChainId, inputToken], []);
   }
+  if (
+    dict[originChainId][inputToken].some(
+      (d) => d.transactionHash === deposit.transactionHash && d.logIndex === deposit.logIndex
+    )
+  ) {
+    throw new Error("Duplicate deposit in bundleDeposits");
```
pxrl, line 91: Can we add a log message here to identify the relevant deposits? That will speed up debugging time significantly if we're able to isolate them.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -435,24 +450,39 @@ export class BundleDataClient {
...
-        true // this bundle data should have been published to arweave
-      ),
-    };
+    // Check if bundle data exists on arweave, otherwise fallback to last published bundle data. If the
+    // first bundle block range we are trying is the pending proposal, then we'll grab the most recently
+    // validated bundle, otherwise we'll grab the second most recently validated bundle.
+    let n = hubPoolClient.hasPendingProposal() ? 1 : 2;
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
pxrl, line 468:
```suggestion
    while (n++ < 4) {
```

src/clients/SpokePoolClient.ts
```diff
@@ -751,6 +778,22 @@ export class SpokePoolClient extends BaseAbstractClient {
           fill.relayExecutionInfo.updatedMessageHash = getMessageHash(event.args.relayExecutionInfo.updatedMessage);
         }
 
+        // Sanity check that this event is not a duplicate.
+        if (
+          this.fills[fill.originChainId] !== undefined &&
+          this.fills[fill.originChainId].some(
+            (f) => f.transactionHash === fill.transactionHash && f.logIndex === fill.logIndex
+          )
```
pxrl, line 786: Prettier probably complains about this formatting, but it should be possible to drop the first check.

```suggestion
        if (
          this.fills[fill.originChainId]?
            .some((f) => f.transactionHash === fill.transactionHash && f.logIndex === fill.logIndex)
```

src/clients/SpokePoolClient.ts
```diff
@@ -751,6 +778,22 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+        ) {
```
pxrl, line 787: Prettier probably complains about this formatting, but it should be possible to drop the first check and simplify the statements.

```suggestion
        const duplicate = this.fills[fill.originChainId]?.find((f) => f.transactionHash === fill.transactionHash && f.logIndex === fill.logIndex);
        if (duplicate) {
```

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: bf3bae61529cd5f2f1adda0e398199eabc09a75d

src/clients/SpokePoolClient.ts
nicholaspai: ce0be1dae58f9d634803ea60161943ea236682ea

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: wdyt of faebdb8a8020235f85c295d254ea2dd32536c3b4 where we fallback to just loading from scratch.

src/clients/SpokePoolClient.ts
nicholaspai: 42fb574ba924ca723ead11c6c62048a4053fc918

pxrl (approved):

nicholaspai:
@bmzig @pxrl if we do experience duplicate events, I can imagine the logs would get very noisy and be distracting. WDYT about emitting one error level log per `update()` run? https://github.com/across-protocol/sdk/pull/884/commits/7545cf232e6167288f320e1371ebce833c15eca6

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -435,23 +450,46 @@ export class BundleDataClient {
...
       this.clients.configStoreClient,
       hubPoolClient.hasPendingProposal()
         ? hubPoolClient.getLatestProposedRootBundle()
-        : hubPoolClient.getLatestFullyExecutedRootBundle(hubPoolClient.latestBlockSearched)! // ! because we know there is a bundle
+        : hubPoolClient.getNthFullyExecutedRootBundle(-1)!
     );
+    // Check if bundle data exists on arweave, otherwise fallback to last published bundle data. If the
+    // first bundle block range we are trying is the pending proposal, then we'll grab the most recently
+    // validated bundle, otherwise we'll grab the second most recently validated bundle.
+    let n = hubPoolClient.hasPendingProposal() ? 1 : 2;
```
bmzig, line 465: Why is it OK to do this? This function now won't necessarily return the latest proposed bundle data, and will instead likely return an older bundle's arweave data. Wouldn't this mean that calculating the "latest" pool rebalance root isn't actually going to calculate the latest root, and instead some older one (unless the latest data has been published).

nicholaspai: maybe we should re-name this function to `getLatestProposedRootWithArweaveData` because if you look at how this function is used, its used by the InventoryClient [here](https://github.com/across-protocol/relayer/blob/7f46802348920b4b1fb3bcb7d551528b57466c64/src/clients/InventoryClient.ts#L627) to get the best approximation of the latest running balances in order to compute which chains are "over-allocated" [here](https://github.com/across-protocol/relayer/blob/7f46802348920b4b1fb3bcb7d551528b57466c64/src/clients/InventoryClient.ts#L762).

So, when a bundle is first proposed but doesn't have its bundle data published to arweave, there is really little negative consequence for the InventoryClient. The InventoryClient will use an outdated running balance until the arweave bundle data is published. If we assume the arweave data gets published once a challenge period, then at worst the "latest" pool rebalance root will be one bundle behind. In reality, it only lags by ~15 mins until the bundle data is very likely published to Arweave.

2b6a8ddfc42f63f697ac666c39d756bf74a93bcb

bmzig (approved):

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -992,6 +1038,11 @@ export class BundleDataClient {
                   }
                 }
               } else {
+                this.logger.debug({
+                  at: "BundleDataClient#loadData",
+                  message: "Duplicate fill detected",
+                  fill,
+                });
```
pxrl, line 1045: This and the subsequent `Duplicate slow fill request detected` log message are pretty far away from the if statement that can trigger them; it's hard to see the logic in github at least. wdyt about relocating them to the top?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -431,32 +439,58 @@ export class BundleDataClient {
...
+    // Try to find an older bundle that has arweave data but cut off after a few tries and just load data from scratch.
+    while (n < 4) {
+      const bundleDataOnArweave = await this.getBundleDataFromArweave(bundleBlockRanges);
+      if (!isDefined(bundleDataOnArweave)) {
+        this.logger.debug({
+          at: "BundleDataClient#getLatestProposedBundleDataFromArweave",
+          message: `No bundle data found on arweave for ${this.getArweaveBundleDataClientKey(
+            bundleBlockRanges
+          )}, trying previous bundle block range`,
+        });
```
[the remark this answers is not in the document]
nicholaspai: > so I'd imagine we could rip through all the loops ~instantly and just resolve the exact same Arweave response each time.

Why would we resolve the exact same arweave response each time? Do you mean, if we resolved undefined?

But let's say we got `undefined` each time. The end result is we're going to call `loadData` on the 4th most recent bundle and that may or may not be on arweave and if its not on arweave, we'll have to load from scratch. This would be very very rare.

But I think we want the fallback logic to be just loading data of the pending root bundle from scratch, so maybe we should lower the loop to n < 3, so worst case we're only looping twice and then falling back to loading the pending bundle data from scratch: fafbbb381172fc60a0e390093e78dad3b82f2bef

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: creates a big diff but good point: 9e3090bdb9d1926d0569b0a08393bb48362ef59a

src/clients/SpokePoolClient.ts
```diff
@@ -751,6 +769,15 @@ export class SpokePoolClient extends BaseAbstractClient {
           fill.relayExecutionInfo.updatedMessageHash = getMessageHash(event.args.relayExecutionInfo.updatedMessage);
         }
 
+        // Sanity check that this event is not a duplicate.
+        const duplicateFill = this.fills[fill.originChainId]?.find(
+          (f) => f.transactionHash === fill.transactionHash && f.logIndex === fill.logIndex
+        );
```
pxrl, line 775: We repeat this iterative searching enough that it could probably be factored out to a utility function - that might condense the code a bit. Something like:

```
const sameEvent = (a: SortableEvent, b: SortableEvent): boolean =>
  a.transactionHash === b.transactionHash && a.logIndex === b.logIndex;

// ...

const duplicateFill = this.fills[fill.originChainId]?.find(sameEvent);
```

src/clients/BundleDataClient/BundleDataClient.ts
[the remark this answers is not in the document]
pxrl: > Why would we resolve the exact same arweave response each time? Do you mean, if we resolved undefined?

My understanding (and I might be misinterpreting the logic) is that if we resolve `undefined`, it's likely because the arweave bot hasn't posted any data to arweave yet. So when we retry immediately, it'd be with the hope that the arweave data might have posted the data in the time since the previous query. However, this is only likely to be milliseconds, so it seems most likely to me that we'd just issue a few identical queries, get identical responses, and then move on. Assuming this is accurate, the retry logic doesn't seem to achieve anything unless it handles transient failures.

src/clients/SpokePoolClient.ts
nicholaspai: [duplicate event util](https://github.com/across-protocol/sdk/pull/884/commits/29b94649c4674a91b523f13f01989d1c9ff388b6)

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -431,50 +440,44 @@ export class BundleDataClient {
       }, toBN(0));
   }
 
-  private async getLatestProposedBundleData(): Promise<{ bundleData: LoadDataReturnValue; blockRanges: number[][] }> {
+  async getPendingPoolRebalanceLeavesFromArweave(): Promise<PoolRebalanceRoot | undefined> {
```
nicholaspai, line 443: I've changed this function to only query once from Arweave on the pending bundle data if it exists. Otherwise return undefined. The relayer's inventory client already handles this fallback well https://github.com/across-protocol/relayer/blob/7863335cce2ecf36e917ab7f81fce11e3a31f9cc/src/clients/InventoryClient.ts#L644

and falls back to using the ExecutedRootBundle events in the HubPoolClient

src/clients/BundleDataClient/BundleDataClient.ts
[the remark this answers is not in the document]
nicholaspai: We're not retrying the same bundle block range though, we're decrementing the range each time. I changed this logic however to only query for the latest bundle block range otherwise return undefined.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -920,64 +924,69 @@ export class BundleDataClient {
             fillCounter++;
             const relayDataHash = getRelayEventKey(fill);
             if (v3RelayHashes[relayDataHash]) {
-              if (!v3RelayHashes[relayDataHash].fill) {
-                assert(
-                  isDefined(v3RelayHashes[relayDataHash].deposits) && v3RelayHashes[relayDataHash].deposits!.length > 0,
-                  "Deposit should exist in relay hash dictionary."
+              if (v3RelayHashes[relayDataHash].fill) {
```
nicholaspai, line 927: The diff here is large because we're now reversing the if-else logic. First we check if the fill is a duplicate, and then throw, otherwise continue with the rest of the logic that was formerly in the `if` statement.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1077,30 +1086,34 @@ export class BundleDataClient {
...
-
-                if (
-                  slowFillRequest.blockNumber >= destinationChainBlockRange[0] &&
-                  _canCreateSlowFillLeaf(matchedDeposit) &&
-                  !_depositIsExpired(matchedDeposit)
-                ) {
-                  validatedBundleSlowFills.push(matchedDeposit);
-                }
-              } else {
+              if (v3RelayHashes[relayDataHash].slowFillRequest) {
```
nicholaspai, line 1089: Same logic for large diff as here https://github.com/across-protocol/sdk/pull/884/files#r1953050255

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -431,52 +438,6 @@ export class BundleDataClient {
       }, toBN(0));
   }
 
-  private async getLatestProposedBundleData(): Promise<{ bundleData: LoadDataReturnValue; blockRanges: number[][] }> {
```
nicholaspai, line 434: We no longer need these functions in https://github.com/across-protocol/relayer/pull/2085

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1077,30 +1043,34 @@ export class BundleDataClient {
...
               }
+              v3RelayHashes[relayDataHash].slowFillRequest = slowFillRequest;
+              if (v3RelayHashes[relayDataHash].fill) {
+                // Exiting here assumes that slow fill requests must precede fills, so if there was a fill
+                // following this slow fill request, then we would have already seen it. We don't need to check
+                // for a fill older than this slow fill request.
+                return;
+              }
+              assert(
+                isDefined(v3RelayHashes[relayDataHash].deposits) && v3RelayHashes[relayDataHash].deposits!.length > 0,
```
pxrl, line 1062: I just merged https://github.com/across-protocol/sdk/pull/867, so you should be able to drop the `!` assertion.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -920,64 +881,69 @@ export class BundleDataClient {
...
+
+                  // Now that we know this deposit has been filled on-chain, identify any duplicate deposits
+                  // sent for this fill and refund them to the filler, because this value would not be paid out
+                  // otherwise. These deposits can no longer expire and get refunded as an expired deposit,
+                  // and they won't trigger a pre-fill refund because the fill is in this bundle.
+                  // Pre-fill refunds only happen when deposits are sent in this bundle and the
+                  // fill is from a prior bundle. Paying out the filler keeps the behavior consistent for how
+                  // we deal with duplicate deposits regardless if the deposit is matched with a pre-fill or
+                  // a current bundle fill.
+                  const duplicateDeposits = v3RelayHashes[relayDataHash].deposits!.slice(1);
```
pxrl, line 923: Same for https://github.com/across-protocol/sdk/pull/884/files#r1954398158.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -920,64 +881,69 @@ export class BundleDataClient {
...
-                  }
+                if (!isDefined(fillToRefund)) {
+                  bundleUnrepayableFillsV3.push(fill);
+                  // We don't return here yet because we still need to mark unexecutable slow fill leaves
+                  // or duplicate deposits. However, we won't issue a fast fill refund.
+                } else {
+                  v3RelayHashes[relayDataHash].fill = fillToRefund;
+                  validatedBundleV3Fills.push({
+                    ...fillToRefund,
+                    quoteTimestamp: v3RelayHashes[relayDataHash].deposits![0].quoteTimestamp,
```
pxrl, line 912: Same for https://github.com/across-protocol/sdk/pull/884/files#r1954398158.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -1077,30 +1043,34 @@ export class BundleDataClient {
...
+              assert(
+                isDefined(v3RelayHashes[relayDataHash].deposits) && v3RelayHashes[relayDataHash].deposits!.length > 0,
+                "Deposit should exist in relay hash dictionary."
+              );
+              const matchedDeposit = v3RelayHashes[relayDataHash].deposits![0];
+
+              if (
+                slowFillRequest.blockNumber >= destinationChainBlockRange[0] &&
+                _canCreateSlowFillLeaf(matchedDeposit) &&
+                !_depositIsExpired(matchedDeposit)
```
pxrl, line 1070: Side note: For a future optimisation, I wonder if this could also exclude slow fills that _will_ be unfillable based on the fillDeadline and the challenge window required at the time of the proposal. Not sure if there are chicken and egg problems there though, since the challenge window could be modified up or down whilst the proposer is building a proposal. It would however be a way of returning deposit refunds to users faster, because the slow-fill that is ultimately unfillable is both an unfortunate use of capital as well as bad UX because the user has to wait for 2 bundles to get their repayment.

src/clients/HubPoolClient.ts
```diff
@@ -776,17 +776,19 @@ export class HubPoolClient extends BaseAbstractClient {
...
-          executedLeaf.chainId === chain &&
-          executedLeaf.l1Tokens.map((l1Token) => l1Token.toLowerCase()).includes(l1Token.toLowerCase())
-        );
-      }
-    ) as ExecutedRootBundle;
+    return sortEventsDescending(this.executedRootBundles).find((executedLeaf: ExecutedRootBundle) => {
+      return (
+        executedLeaf.blockNumber <= block &&
+        executedLeaf.chainId === chain &&
+        executedLeaf.l1Tokens.map((l1Token) => l1Token.toLowerCase()).includes(l1Token.toLowerCase())
```
pxrl, line 785: This comparison sticks out to me. `l1Token` as passed into the function is overridden within the `Array.call()` closure. Additionally, `l1Token` is a string in either case, and `String.includes()` reverts to checking for substrings, rather than a complete match-

```suggestion
        executedLeaf.l1Tokens.map((token) => token.toLowerCase() === l1Token.toLowerCase()
```

src/clients/HubPoolClient.ts
```diff
@@ -776,17 +776,19 @@ export class HubPoolClient extends BaseAbstractClient {
...
+      );
+    }) as ExecutedRootBundle;
```
pxrl, line 787: Is this to ward of an `...is potentially undefined` warning? Is it still necessary if the above change is made?

src/clients/SpokePoolClient.ts
```diff
@@ -656,6 +658,16 @@ export class SpokePoolClient extends BaseAbstractClient {
         }
 
         if (this.depositHashes[getRelayEventKey(deposit)] !== undefined) {
+          // Sanity check that this event is not a duplicate, even though the relay data hash is a duplicate.
+          const allDeposits = this._getDuplicateDeposits(deposit).concat(this.depositHashes[getRelayEventKey(deposit)]);
+          if (
+            allDeposits.some((e) => {
+              return duplicateEvent(deposit, e);
+            })
+          ) {
```
pxrl, line 667: Does this work?

```suggestion
          if (allDeposits.some((e) => duplicateEvent(deposit, e))) {
```

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: @pxrl I've decided to revert back to this implementation becaues the diff is so big and hard to verify.

src/clients/HubPoolClient.ts
nicholaspai: thats a good point, though I don't think the .map result matters since the map returns the modified array right? (which is then thrown away)

src/clients/HubPoolClient.ts
nicholaspai: Good call, this is an improvement: 329d44c2ad4e553a5bdb906b332aea411c0303db

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: I don't think we can do this because of this non-determinism:

> since the challenge window could be modified up or down whilst the proposer is building a proposal

We just don't know which block number this bundle will be proposed. Theoretically the mainnet bundle end block could be hours before the proposal lands

PR #886, opened 2025-02-12
refactor: Remove SpokePoolClient oldestTime lookup

nicholaspai:
Reopens PR #752 which was closed due to a breaking test in `relayer` repo, which I'll fix here https://github.com/across-protocol/relayer/pull/1865


pxrl (approved):
Great! Definitely happy to see that we can get this in.

PR #892, opened 2025-02-17
improve(retryProvider): Short-circuit on nonce errors

pxrl:
There's no point retrying a transaction with nonce-too-low, so just fail immediately. This mirrors the strategy with transaction simulation failures.

### DIFF
--- a/src/providers/retryProvider.ts
+++ b/src/providers/retryProvider.ts
@@ -276,7 +276,7 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
 
       return error.error;
     } catch {
-      return
+      return;
     } // do nothing
   }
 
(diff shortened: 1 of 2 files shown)


bmzig (approved):

src/providers/retryProvider.ts
```diff
@@ -258,36 +258,62 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
...
+      if (!JsonRpcError.is(error)) {
+        return;
+      }
+
+      return error.error;
     } catch {
-      return false;
-    }
+      return;
+    } // do nothing
```
pxrl, line 280:
```suggestion
    }
```

bmzig (approved):

PR #895, opened 2025-02-21
feat: Decide on Cache TTL for some RPC responses post-response

nicholaspai:
The CacheProvider decides pre-request which TTL to store an RPC response in the cache which makes sense when we were originally caching only `eth_getLogs` and `eth_call` which have `blockNumber` as part of the request parameters. The TTL decision is essentially based on how much the requested data is older than the HEAD network block. The intuitionhere is that events that are already-finalized we can cache with infinite TTL while those that have re-org risk we should not cache at all or have shorter TTL.

However, this logic prevents us from being able to cache results of RPC endpoints like `eth_getTransactionReceipt` where the `blockNumber` is not part of the request parameters.

This PR introduces a new `CacheType` = `DECIDE_TTL_POST_SEND` where we defer the TTL decision until after we receive the RPC response. This response should include a `blockNumber` which we can then use to set the TTl. This logic incurs no extra compute costs or RPC requests, as regardless when we set the TTL based on the blockNumber we need to call `getBlockNumberFromRpcResponse()`.

As a motivating example, the finalizer is one example of a bot that makes the `eth_getTransactionReceipt` request in order to associate a `TokensBridged` event with the L2-specific withdrawal event needed to create finalization calldata. This is done on every single OPStack `TokensBridged` event [here](https://github.com/across-protocol/relayer/blob/356c1bc270940109cc07a82d2d778b287f779c3c/src/finalizer/utils/opStack.ts#L240). The Finalizer will regularly process the same `TokensBridged` event repeatedly until it drifts out of the lookback window so this caching of `eth_getTransactionReceipt` should yield immediate results.

### DIFF
--- a/src/providers/cachedProvider.ts
+++ b/src/providers/cachedProvider.ts
@@ -5,9 +5,7 @@ import { RateLimitedProvider } from "./rateLimitedProvider";
 import { CacheType } from "./utils";
 
 export class CacheProvider extends RateLimitedProvider {
-  public readonly getBlockByNumberPrefix: string;
-  public readonly getLogsCachePrefix: string;
-  public readonly callCachePrefix: string;
+  public readonly cachePrefix: string;
   public readonly baseTTL: number;
 
   constructor(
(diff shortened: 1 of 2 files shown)


src/providers/cachedProvider.ts
```diff
@@ -78,16 +78,11 @@ export class CacheProvider extends RateLimitedProvider {
...
       case "eth_getBlockByNumber":
-        return this.getBlockByNumberPrefix + JSON.stringify(params);
-      case "eth_getLogs":
-        return this.getLogsCachePrefix + JSON.stringify(params);
-      case "eth_call":
-        return this.callCachePrefix + JSON.stringify(params);
+        return `${this.cachePrefix}:getBlockByNumber,` + JSON.stringify(params);
       default:
-        throw new Error(`CacheProvider::buildRedisKey: invalid JSON-RPC method ${method}`);
+        return `${this.cachePrefix}:${method},` + JSON.stringify(params);
```
pxrl, line 85: @mrice32 can confirm, but I think the existing implementation might have been an optimisation to avoid lots of repeated string operations in the hot path.

pxrl (approved):

src/providers/cachedProvider.ts
nicholaspai: reverted

pxrl (approved):

PR #899, opened 2025-02-25
feat(svm): svm spoke events client

md0x:
Changes proposed in this PR:
- Add SvmSpokeEventsClient, an events client that fetches CPI events from SvmSpoke by event type, with optional filtering by fromSlot and/or toSlot.

How to use it:
```ts
import { SvmSpokeClient } from "@across-protocol/contracts";
import { createSolanaRpc } from "@solana/kit";
import { SvmSpokeEventsClient } from "./src/svm";

async function main() {
    const rpc = createSolanaRpc("YOUR_RPC_URL");
    const eventsClient = await SvmSpokeEventsClient.create(rpc);
    const fromSlot = BigInt(362424536);
    const toSlot = BigInt(363326129);
    const events = await eventsClient.queryEvents<SvmSpokeClient.FilledRelay>("FilledRelay", fromSlot, toSlot);
    console.log(events);
}

main();
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.1.20",
+  "version": "4.1.21",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "files": [


package.json
```diff
@@ -12,6 +12,8 @@
     "node": ">=20.18.0"
   },
   "scripts": {
+    "build-bigint-buffer": "[ -d node_modules/bigint-buffer ] && command -v node-gyp > /dev/null && cd node_modules/bigint-buffer && node-gyp configure && node-gyp build || echo 'Skipping bigint-buffer build: folder or node-gyp not found'",
```
md0x, line 15: Same as [in relayer](https://github.com/across-protocol/relayer/blob/b082eb2150a4ae1bc3619eca98a8821ebfa6ed30/package.json#L52-L53) to get rid of the bigint-buffer warning

src/svm/utils/events.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  return eventData;
+}
+
+/**
+ * Gets the event name from a raw name.
+ */
+export function getEventName(rawName?: string): EventName {
+  if (!rawName) throw new Error("Raw name is undefined");
```
md0x, line 52: This should never happen to we throw an error

src/svm/utils/events.ts
```diff
@@ -0,0 +1,94 @@
...
+    case "EmergencyDeletedRootBundle":
+      return eventData as EmergencyDeletedRootBundleEvent;
+    case "RequestedSlowFill":
+      return eventData as RequestedSlowFillEvent;
+    case "ClaimedRelayerRefund":
+      return eventData as ClaimedRelayerRefundEvent;
+    case "TransferredOwnership":
+      return eventData as TransferredOwnershipEvent;
+    default:
+      throw new Error(`Unknown event name: ${name}`);
```
md0x, line 92: Same here

src/svm/eventsClient.ts
```diff
@@ -0,0 +1,214 @@
...
+   * @param options - Options for fetching signatures.
+   * @param finality - Commitment level.
+   * @returns A promise that resolves to an array of events matching the eventName.
+   */
+  public async queryEvents<T extends EventData>(
+    eventName: EventName,
+    fromSlot?: bigint,
+    toSlot?: bigint,
+    options: GetSignaturesForAddressConfig = { limit: 1000 },
+    finality: Commitment = "confirmed"
```
Reinis-FRP, line 61: does it make sense to pass separate commitment in the `finality` parameter (used in `getTransaction` calls) from `options.commitment` that gets used in `getSignaturesForAddress` calls?

src/svm/eventsClient.ts
```diff
@@ -0,0 +1,214 @@
...
+      return true;
+    });
+
+    // Fetch events for all signatures in parallel.
+    const eventsWithSlots = await Promise.all(
+      filteredSignatures.map(async (signatureTransaction) => {
+        const events = await this.readEventsFromSignature(signatureTransaction.signature, program, anchorIdl, finality);
+        return events.map((event) => ({
+          ...event,
+          confirmationStatus: signatureTransaction.confirmationStatus || "Unknown",
```
Reinis-FRP, line 123: maybe pass the `Commitment | null` type as is?

src/svm/eventsClient.ts
```diff
@@ -0,0 +1,214 @@
...
+          blockTime: signatureTransaction.blockTime || unixTimestamp(BigInt(0)),
```
Reinis-FRP, line 124: maybe pass `UnixTimestamp | null` type as is?

src/svm/utils/events.ts
Reinis-FRP: why accept optional param in the first place?

src/svm/eventsClient.ts
```diff
@@ -0,0 +1,214 @@
...
+  GetTransactionApi,
+  RpcTransport,
+  Signature,
+  unixTimestamp,
+} from "@solana/web3-v2.js";
+import { EventData, EventName, EventWithData } from "./types";
+import { getEventName, parseEventData } from "./utils/events";
+import { isDevnet } from "./utils/helpers";
+
+type GetTransactionReturnType = ReturnType<GetTransactionApi["getTransaction"]>;
```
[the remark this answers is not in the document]
md0x: Good call. I've added a type wrapper to extract only the "json" encoded return type.

src/svm/eventsClient.ts
```diff
@@ -0,0 +1,205 @@
...
+    toSlot?: bigint,
+    options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
+  ): Promise<EventWithData<EventData>[]> {
+    const allSignatures: GetSignaturesForAddressTransaction[] = [];
+    let hasMoreSignatures = true;
+    let currentOptions = options;
+
+    while (hasMoreSignatures) {
+      const signatures: GetSignaturesForAddressApiResponse = await this.rpc
+        .getSignaturesForAddress(this.svmSpokeAddress, currentOptions)
```
nicholaspai, line 102: Is a `signature` any transaction sent to the `svmSpokeAddress`?

src/svm/eventsClient.ts
```diff
@@ -0,0 +1,205 @@
...
+   * @returns A promise that resolves to an array of events matching the eventName.
+   */
+  public async queryEvents<T extends EventData>(
+    eventName: EventName,
+    fromSlot?: bigint,
+    toSlot?: bigint,
+    options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
+  ): Promise<EventWithData<T>[]> {
+    const events = await this.queryAllEvents(fromSlot, toSlot, options);
+    return events.filter((event) => event.name === eventName) as EventWithData<T>[];
```
nicholaspai, line 80: So there's no way to query the `rpc` for just the `eventName`? You always need to query all events and then filter from there?

Seems expensive but if we're smart about setting from/toSlot we can cache the results

nicholaspai (commented):
i just left some questions for my own knowledge

src/svm/eventsClient.ts
md0x: Yes, this signatures are any transactions that involve the svmSpokeAddress, they could be direct calls to the spoke or cross program invocations.

src/svm/eventsClient.ts
md0x: Correct, that's one of the trade-offs of using CPI Events.

src/svm/eventsClient.ts
```diff
@@ -116,12 +126,12 @@ export class SvmSpokeEventsClient {
 
     // Fetch events for all signatures in parallel.
     const eventsWithSlots = await Promise.all(
-      filteredSignatures.map(async (signatureTransaction) => {
-        const events = await this.readEventsFromSignature(signatureTransaction.signature, program, anchorIdl, finality);
+      filteredSignatures.flatMap(async (signatureTransaction) => {
```
Reinis-FRP, line 129: `flatMap` does not seem to work as expected in async function and we anyways return `eventsWithSlots.flat()`, so I think we can leave just `map` here.

src/svm/utils/helpers.ts
```diff
@@ -0,0 +1,9 @@
+import web3, { RpcTransport } from "@solana/web3-v2.js";
+
+/**
+ * Helper to determine if the current RPC network is devnet.
+ */
+export async function isDevnet(rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>): Promise<boolean> {
+  const genesisHash = await rpc.getGenesisHash().send();
+  return genesisHash === "EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG";
+}
```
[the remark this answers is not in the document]
md0x: Yes, they are case sensitive @james-a-morris.

src/svm/eventsClient.ts
md0x: Makes sense 👍
Updated.

src/svm/eventsClient.ts
```diff
@@ -0,0 +1,205 @@
...
+  private async queryAllEvents(
+    fromSlot?: bigint,
+    toSlot?: bigint,
+    options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
+  ): Promise<EventWithData<EventData>[]> {
+    const allSignatures: GetSignaturesForAddressTransaction[] = [];
+    let hasMoreSignatures = true;
+    let currentOptions = options;
+
+    while (hasMoreSignatures) {
```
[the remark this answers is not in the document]
md0x: We can’t determine how many there are, and each loop depends on the results of the previous one because we need the last signature found.

src/svm/eventsClient.ts
```diff
@@ -0,0 +1,205 @@
...
+      if (fromSlot && signatureTransaction.slot < fromSlot) return false;
+      if (toSlot && signatureTransaction.slot > toSlot) return false;
+      return true;
+    });
+
+    // Fetch events for all signatures in parallel.
+    const eventsWithSlots = await Promise.all(
+      filteredSignatures.map(async (signatureTransaction) => {
+        const events = await this.readEventsFromSignature(signatureTransaction.signature, options.commitment);
+        return events.map((event) => ({
```
chrismaree, line 131: you can add more info to these events such as the name of the event, which is useful, from `event.name` ([see](https://github.com/across-protocol/contracts/blob/4d89becb02ddd6277b1055dbe7561bd35f334dee/src/svm/web3-v1/solanaProgramUtils.ts#L168)). 

also do you think it's better to have `confirmationStatus` &`blockTime` defined as a string/UnixTimestamp or undefined or better to set them to hard limits so the types are always string & UnixTimestamp as done [here](https://github.com/across-protocol/contracts/blob/4d89becb02ddd6277b1055dbe7561bd35f334dee/src/svm/web3-v1/solanaProgramUtils.ts#L164)?

package.json
```diff
@@ -105,7 +107,9 @@
     "@eth-optimism/sdk": "^3.3.1",
     "@ethersproject/bignumber": "^5.7.0",
     "@pinata/sdk": "^2.1.0",
-    "@solana/web3.js": "^2.0.0",
+    "@solana/web3-v2.js": "npm:@solana/web3.js@2",
```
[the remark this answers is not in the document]
Reinis-FRP: shall we use `@solana/kit` now that anza renamed it?

src/svm/eventsClient.ts
md0x: Yep, the event name and parsed data, which match the Codama clients’ generated event types, all come from [this function](https://github.com/across-protocol/sdk/blob/f3ed1129ca2d3a56004c39d0a602c8a906a39e0b/src/svm/eventsClient.ts#L159-L201) and included when we destructure `...event`

md0x: Regarding the hard limits, I agreed with Reinis’ suggestion to use `null` as they are already defined in solana web3 ([link](https://github.com/across-protocol/sdk/pull/899#discussion_r1975259824)).
It feels more natural to keep the original outputs and let the consumer decide, but I’m not strongly opinionated on this.

PR #901, opened 2025-02-25
refactor(CacheProvider): Early exit if CacheType == NONE

nicholaspai:
We can remove a large if-statement indent if we early exit. There is no logical change in this PR


pxrl (approved):
+1 this is a nice fix.

PR #902, opened 2025-02-27
feat: Add class to handle address retrieval

pxrl:
This commit introduces a class that can be used to retrieve a set of addresses from one or more remote sources. Different backends can be implemented to support different response formats.

Effort is made to be flexible with the responses received. Addresses are not required to be checksummed; as long as they can be coerced into a checksummed address then they will be accepted. Invalid addresses are dropped without side-effects.


src/utils/addressList/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      .flat()
+      .map((address) => {
+        try {
+          return ethersUtils.getAddress(address.toLowerCase());
+        } catch {
+          return INVALID_ADDRESS;
+        }
+      })
+      .filter((address) => address !== INVALID_ADDRESS);
+    return { addresses: new Set(addresses), length: addresses.length };
```
dohaki, line 20: Is `length` the size of the array before removing duplicates via `Set` on purpose? Should probably be the size of the set, correct?

dohaki (commented):
LGTM! One small remark on the returned `length` of the addresses

src/utils/addressList/index.ts
pxrl: Good catch on the mismatch there; that's definitely not intended. Length should be the number of elements in the set. The set API is just frustrating in that it seems you have to revert back to an array in order to figure out the length; that's an annoyance I'd hoped to encapsulate here.

src/utils/addressList/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+      })
+      .filter((address) => address !== INVALID_ADDRESS);
+
+    // Dedup the aggregated, normalised, filtered set of addresses.
+    const addresses = new Set(allAddresses);
+
+    return {
+      addresses,
+      length: Array.from(addresses).length,
```
gsteenkamp89, line 26: NIT: You can try `addresses.size` on the Set itself to find the length.

pxrl: Damn; I missed that, but that's nice - fixed here: https://github.com/across-protocol/sdk/pull/902/commits/3963dcace03d37594937e499665e46a7aef26e2c

src/utils/addressList/index.ts
pxrl: Fixed here: https://github.com/across-protocol/sdk/pull/902#discussion_r1975751401

src/addressAggregator/adapters/env.ts
pxrl, line 1: For subsequent work, I'd propose to add a `file` adapter, which supports reading config directly from a file. Then the entire process of fetching addresses can be outsourced and the bots can just rely on a file being available in the local fs.

pxrl:
nb. Currently lacks tests; would want to add some simple cases  for each adapter to catch obvious errors.

pxrl:
nb. To get this into the bots today I will propose to merge this in as-is, but it will be removed later once the code has settled in an upstream repository.

dohaki (approved):
LGTM

PR #904, opened 2025-02-27
feat: start using a custom Address class

bmzig:
Still need to add a lot of unit tests. There are a few concerns with this PR, notably
- Aill allocating new `Address` classes meaningfully slow down the bots? This will need to be tested.
- We use addresses as keys for a variety of maps. Are these keys still consistent given the same address? (i.e. it would be a bug if we set it with `.toString` and access it with `.toAddress`). In general, if the token is *guaranteed* to be a bytes20 address, which is true when the token is an L1 token (but not true if the token is an L2 token), then we set/access maps with `toAddress`. Otherwise, we use `toString`.


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -67,18 +67,20 @@ type DataCache = Record<string, Promise<LoadDataReturnValue>>;
 // V3 dictionary helper functions
 function updateExpiredDepositsV3(dict: ExpiredDepositsToRefundV3, deposit: V3DepositWithBlock): void {
   // A deposit refund for a deposit is invalid if the depositor has a bytes32 address input for an EVM chain. It is valid otherwise.
-  if (chainIsEvm(deposit.originChainId) && !isValidEvmAddress(deposit.depositor)) {
+  if (chainIsEvm(deposit.originChainId) && !deposit.depositor.isValidEvmAddress()) {
```
pxrl, line 70: wdyt about internalising the chain -> address type mapping within the Address class, such that we could write: `!deposit.depositor.valid(deposit.originChainId)` ?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -67,18 +67,20 @@ type DataCache = Record<string, Promise<LoadDataReturnValue>>;
...
     return;
   }
-  const { originChainId, inputToken } = deposit;
+  const { originChainId, inputToken: _inputToken } = deposit;
+  const inputToken = _inputToken.toString();
```
pxrl, line 74: @dohaki @gsteenkamp89 Can you think of any way we can make the toString automatic, so we can rely on it happening in the background when `inputToken` is supplied as a key to an Object? Not sure if that's even possible, but if not, this usage scenario represents a sharp edge that we'll probably run into.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -410,21 +415,23 @@ export class BundleDataClient {
...
   }
 
-  getUpcomingDepositAmount(chainId: number, l2Token: string, latestBlockToSearch: number): BigNumber {
+  getUpcomingDepositAmount(chainId: number, l2Token: Address, latestBlockToSearch: number): BigNumber {
     if (this.spokePoolClients[chainId] === undefined) {
       return toBN(0);
     }
     return this.spokePoolClients[chainId]
       .getDeposits()
-      .filter((deposit) => deposit.blockNumber > latestBlockToSearch && deposit.inputToken === l2Token)
```
pxrl, line 427: @dohaki @gsteenkamp89 Are you aware of any eslint rule we could use to flag checking equality between Address instances? @bmzig has removed all known instances from this PR but it seems like something we'd trip on in future.

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -599,15 +606,15 @@ export class BundleDataClient {
...
-  getRefundsFor(bundleRefunds: CombinedRefunds, relayer: string, chainId: number, token: string): BigNumber {
-    if (!bundleRefunds[chainId] || !bundleRefunds[chainId][token]) {
+  getRefundsFor(bundleRefunds: CombinedRefunds, relayer: Address, chainId: number, token: Address): BigNumber {
+    if (!bundleRefunds[chainId] || !bundleRefunds[chainId][token.toString()]) {
       return BigNumber.from(0);
     }
-    const allRefunds = bundleRefunds[chainId][token];
-    return allRefunds && allRefunds[relayer] ? allRefunds[relayer] : BigNumber.from(0);
+    const allRefunds = bundleRefunds[chainId][token.toString()];
+    return allRefunds && allRefunds[relayer.toString()] ? allRefunds[relayer.toString()] : BigNumber.from(0);
```
pxrl, line 614: Side-loading a bnZero here:

```suggestion
    return allRefunds && allRefunds[relayer.toString()] ? allRefunds[relayer.toString()] : bnZero;
```

src/clients/BundleDataClient/utils/SuperstructUtils.ts
```diff
@@ -31,19 +31,30 @@ const BigNumberType = coerce(instance(BigNumber), union([string(), number()]), (
   }
 });
 
+const AddressType = coerce(instance(Address), string(), (value) => {
+  try {
+    // Attempt to convert the string to an Address
+    return Address.from(value);
+  } catch (error) {
+    // In case of any error during conversion, return the original value
+    // This will lead to a validation error, as the resulting value won't match the expected Address type
```
pxrl, line 40: Any thoughts on how to handle this safely?

src/clients/BundleDataClient/utils/SuperstructUtils.ts
```diff
@@ -31,19 +31,30 @@ const BigNumberType = coerce(instance(BigNumber), union([string(), number()]), (
...
+    return value;
+  }
+});
+
 const FillTypeSS = number();
 
 const V3RelayDataSS = {
-  inputToken: string(),
+  inputToken: AddressType,
```
pxrl, line 48: How does this manifest when posted to arweave? Do we encode as a stringified address or a stringified object?

src/clients/HubPoolClient.ts
```diff
@@ -516,30 +531,33 @@ export class HubPoolClient extends BaseAbstractClient {
     );
   }
 
-  getTokenInfo(chainId: number | string, tokenAddress: string): L1Token | undefined {
-    const deposit = { originChainId: parseInt(chainId.toString()), inputToken: tokenAddress } as Deposit;
+  getTokenInfo(chainId: number | string, tokenAddress: Address): L1Token | undefined {
+    const deposit = {
+      originChainId: parseInt(chainId.toString()),
```
pxrl, line 536: Driveby: Should be able to drop the `toString()` here:

```
> y = 1
1
> parseInt(y)
1
> parseInt(y.toString())
1
```

```suggestion
      originChainId: parseInt(chainId.toString()),
```

pxrl: nvm - this is a tsc constraint, not a js constraint.

src/clients/HubPoolClient.ts
```diff
@@ -779,33 +797,33 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    l1Token: EvmAddress,
     executedRootBundle: ExecutedRootBundle | undefined
   ): TokenRunningBalance {
     let runningBalance = toBN(0);
     if (executedRootBundle) {
       const indexOfL1Token = executedRootBundle.l1Tokens
-        .map((l1Token) => l1Token.toLowerCase())
-        .indexOf(l1Token.toLowerCase());
+        .map((_l1Token) => _l1Token.toAddress().toLowerCase())
+        .indexOf(l1Token.toAddress().toLowerCase());
```
pxrl, line 826: `Address.toAddress()` returns a checksummed address so it should be OK to drop the `.toLowerCase()` here.

src/clients/HubPoolClient.ts
```diff
@@ -516,30 +531,33 @@ export class HubPoolClient extends BaseAbstractClient {
     );
   }
 
-  getTokenInfo(chainId: number | string, tokenAddress: string): L1Token | undefined {
-    const deposit = { originChainId: parseInt(chainId.toString()), inputToken: tokenAddress } as Deposit;
+  getTokenInfo(chainId: number | string, tokenAddress: Address): L1Token | undefined {
+    const deposit = {
+      originChainId: parseInt(chainId.toString()),
+      inputToken: tokenAddress,
+    } as Deposit;
```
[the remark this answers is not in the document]
bmzig: Merged master https://github.com/across-protocol/sdk/pull/904/commits/f35aa7730e86c462df4e40a105f217d4475cdfdf

src/clients/HubPoolClient.ts
bmzig: True. https://github.com/across-protocol/sdk/pull/904/commits/16dbd7e71e3f07706b0b708ea5e94aba544c84f5

src/clients/HubPoolClient.ts
```diff
@@ -254,14 +258,18 @@ export class HubPoolClient extends BaseAbstractClient {
...
-  ): string {
+  ): Address {
     const l1Token = this.getL1TokenForDeposit(deposit);
     // Use the latest hub block number to find the L2 token counterpart.
     return this.getL2TokenForL1TokenAtBlock(l1Token, l2ChainId, deposit.quoteBlockNumber);
   }
 
-  l2TokenEnabledForL1Token(l1Token: string, destinationChainId: number): boolean {
-    return this.l1TokensToDestinationTokens?.[l1Token]?.[destinationChainId] != undefined;
+  l2TokenEnabledForL1Token(_l1Token: EvmAddress | string, destinationChainId: number): boolean {
```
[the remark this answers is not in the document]
bmzig: We do it here, unfortunately. https://github.com/across-protocol/sdk/blob/fa658be3981da4df540569bf47c228c236e1d15d/src/clients/HubPoolClient.ts#L217.

src/clients/BundleDataClient/utils/SuperstructUtils.ts
bmzig: Not yet, although this is pretty similar to what we do with BigNumber types. Do you anticipate running into deserialization issues here?

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: Good call. Done here https://github.com/across-protocol/sdk/pull/904/commits/16dbd7e71e3f07706b0b708ea5e94aba544c84f5

bmzig:
Closing since this was done in parts.

PR #905, opened 2025-02-27
feat(solana): cached solana provider

Reinis-FRP:
This implements similar cached provider logic as `src/providers/cachedProvider.ts`, but for Solana.

Also adds missing tests for the rate limited Solana provider

Fixes: https://linear.app/uma/issue/ACX-3741/solana-cached-provider


src/caching/Memory/MemoryCacheClient.ts
Reinis-FRP, line 1: currently this is used in tests only

src/providers/solana/rateLimitedRpcFactory.ts
```diff
@@ -13,7 +13,7 @@ export class RateLimitedSolanaRpcFactory extends SolanaClusterRpcFactory {
   private queue: QueueObject<SolanaRateLimitTask>;
 
   // Holds the underlying transport that the rate limiter wraps.
-  private readonly defaultTransport: RpcTransport;
+  protected defaultTransport: RpcTransport;
```
Reinis-FRP, line 16: had to change these for being able to use in mocks

src/utils/JSONUtils.ts
```diff
@@ -83,3 +83,37 @@ export function jsonReviverWithBigNumbers(_key: string, value: unknown): unknown
...
+
+/**
+ * A replacer for use in `JSON.stringify` that converts BigInt number objects containing numeric string value.
+ * @param _key Unused
+ * @param value The value to convert
+ * @returns The converted object
+ */
+export function jsonReplacerWithBigInts(_key: string, value: unknown): unknown {
+  if (typeof value === "bigint") {
+    return { type: "BigInt", value: value.toString() };
```
Reinis-FRP, line 95: Alternative would be to encode this as decimal strings followed by `n` character which would take bit less space.

src/providers/solana/cachedRpcFactory.ts
```diff
@@ -0,0 +1,126 @@
...
+      if (!this.isGetTransactionParams(params)) return this.rateLimitedTransport<TResponse>(...args);
+
+      // Check the confirmation status first to avoid caching non-finalized transactions.
+      const getSignatureStatusesResponse = await this.rateLimitedRpcClient
+        .getSignatureStatuses([params[0]], {
+          searchTransactionHistory: true,
+        })
+        .send();
+
+      const getTransactionResponse = await this.rateLimitedTransport<TResponse>(...args);
```
md0x, line 79: Can we run `getSignatureStatuses` and `this.rateLimitedTransport` in parallel using `Promise.all(...)` to reduce sequential waits?

src/providers/solana/cachedRpcFactory.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      // Cache the transaction only if it is finalized.
+      if (getSignatureStatusesResponse.value[0]?.confirmationStatus === "finalized") {
+        const redisKey = this.buildRedisKey(method, params);
+        await this.redisClient?.set(
```
md0x, line 84: I think there is no need to await here since it won't affect the returned value

src/providers/solana/cachedRpcFactory.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      return this.rateLimitedTransport<TResponse>(...args);
+    };
+  }
+
+  private async requestAndCacheFinalized<TResponse>(...args: Parameters<RpcTransport>): Promise<TResponse> {
+    const { method, params } = args[0].payload as { method: string; params?: unknown[] };
+
+    // Only handles getTransaction right now.
+    if (method === "getTransaction") {
```
md0x, line 68: Nit: You could use an early return `if (method !== "getTransaction") return this.rateLimitedTransport<TResponse>(...args);`
So we don't need the else at the end and improve readability

src/providers/solana/cachedRpcFactory.ts
```diff
@@ -0,0 +1,126 @@
...
+        // If cache has the result, parse the json and return it.
+        if (redisResult) {
+          return JSON.parse(redisResult, jsonReviverWithBigInts);
+        }
+
+        // Cache does not have the result. Query it directly and cache it if finalized.
+        return this.requestAndCacheFinalized<TResponse>(...args);
+      }
+
+      return this.rateLimitedTransport<TResponse>(...args);
```
pxrl, line 60: Minor refactor that Nick figured out with the EVM implementation:

```suggestion
      if (cacheType === CacheType.NONE) {
        return this.rateLimitedTransport<TResponse>(...args);
      }
       
      const redisKey = this.buildRedisKey(method, params);

      // Attempt to pull the result from the cache.
      const redisResult = await this.redisClient?.get<string>(redisKey);

      // If cache has the result, parse the json and return it.
      if (redisResult) {
        return JSON.parse(redisResult, jsonReviverWithBigInts);
      }

      // Cache does not have the result. Query it directly and cache it if finalized.
      return this.requestAndCacheFinalized<TResponse>(...args);
```

src/utils/JSONUtils.ts
pxrl: I favour storing them as strings :+1:. Ultimately I wonder if we'll just use BigInt everywhere, in which case we could unconditionally convert strings as numbers back to BigInt.

src/utils/JSONUtils.ts
```diff
@@ -83,3 +83,37 @@ export function jsonReviverWithBigNumbers(_key: string, value: unknown): unknown
...
+ */
+export function jsonReviverWithBigInts(_key: string, value: unknown): unknown {
+  if (
+    typeof value === "object" &&
+    value !== null &&
+    "type" in value &&
+    "value" in value &&
+    value.type === "BigInt" &&
+    typeof value.value === "string" &&
+    /^-?\d+$/.test(value.value) // Ensure it's a valid BigInt value
```
pxrl, line 114: ooc, could superstruct help here ?

pxrl (approved):
Only a couple of minor comments from me - this looks pretty good.

src/providers/solana/cachedRpcFactory.ts
nicholaspai, line 68: what does the result of `getTransaction` return? Does it include a block number or finalization status? If it returns a block number, then we might be able to skip the `getSignatureStatuses` call to check whether this transactions is finalized. Instead, we could defer the TTL setting until after we get the blocknumber from the `getTransaction` call. 

I introduced this recently to the ethereum `CachedProvider` [here](https://github.com/across-protocol/sdk/blob/2f07ff2f49be2f28f05bf40b68cd3adfec1e0cd9/src/providers/cachedProvider.ts#L62) by introducing a new `CacheType == DECIDE_TTL_POST_SEND`.

nicholaspai (commented):
Just one suggestion on deciding TTL when setting values in the cache

src/providers/solana/cachedRpcFactory.ts
Reinis-FRP: I'm afraid not. we want to guarantee that `getSignatureStatuses` gets resolved first. Otherwise we could cache transaction that is not finalized yet.

src/providers/solana/cachedRpcFactory.ts
Reinis-FRP: agree, but that would make error handling harder. And we also await in the EVM implementation, so maybe can keep it here for consistency?

src/utils/JSONUtils.ts
Reinis-FRP: ye, we might. but for now I'm going to follow your suggestion to use string representation (ending with `n`)

md0x (approved):

PR #906, opened 2025-02-28
chore: Drop legacy V3 support

pxrl:
The UMIP permits support for these events to be dropped as early as 7 days post-upgrade. This cleanup is a necessary prerequisite for adding non-EVM support.


src/utils/SpokeUtils.ts
```diff
@@ -14,6 +14,51 @@ import { toBytes32 } from "./AddressUtils";
 
 type BlockTag = providers.BlockTag;
 
+/**
+ * @param spokePool SpokePool Contract instance.
+ * @param deposit Deopsit instance.
+ * @param repaymentChainId Optional repaymentChainId (defaults to destinationChainId).
+ * @returns An Ethers UnsignedTransaction instance.
+ */
+export function populateRelay(
```
nicholaspai, line 23: Why'd you add this?

pxrl: Ah, I've had this kicking around for a while and Bennett fixed `populateV3Relay()` inbetween. I'll drop this.

pxrl: https://github.com/across-protocol/sdk/pull/906/commits/d1aaee98a31f08cf4da6e69cd6d4c280610f6732

src/clients/SpokePoolClient.ts
```diff
@@ -947,23 +935,13 @@ export class SpokePoolClient extends BaseAbstractClient {
...
-          this.spokePool.filters.V3FundsDeposited(null, null, null, null, null, depositId),
-          { fromBlock, toBlock, maxBlockLookBack }
-        ),
-        paginatedEventQuery(
-          this.spokePool,
-          this.spokePool.filters.FundsDeposited(null, null, null, null, null, depositId),
-          { fromBlock, toBlock, maxBlockLookBack }
-        ),
-      ])
-    ).flat();
```
pxrl, line 966: fwiw, might have to back this out. If a relayer fills an old V3FundsDeposited event then the deposit would not be found and the fill would not be refunded.

nicholaspai (approved):

src/clients/SpokePoolClient.ts
nicholaspai: That's true. We always have to keep this in even after we deprecate depositDeprecated()....

pxrl: Restored here: https://github.com/across-protocol/sdk/pull/906/commits/1a1e6c90920d493aeaf451f62690eac39b0a098b

src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -108,7 +107,7 @@ export class MockSpokePoolClient extends SpokePoolClient {
       });
 
     // Update latestDepositIdQueried.
-    const idx = eventsToQuery.indexOf("V3FundsDeposited");
```
bmzig, line 111: This looks like one of those instances where we used the fact that deposit IDs are incrementing to store the latest deposit ID. I think this means that there will now be just garbage in `this.latestDepositIdQueried`, which I _think_ is alright, but flagging since we may just want to remove that variable entirely since it will contain nothing of substance.

bmzig (approved):

src/clients/mocks/MockSpokePoolClient.ts
pxrl: I can mark this resolved now that we dropped that part of SpokePoolClient update entirely in this commit: https://github.com/across-protocol/sdk/commit/d804fbe2e207047309a567823a84e27f7d76c032.

james-a-morris (approved):
The indexer should be good to go - this may be worth as a breaking version bump

melisaguevara:
> @amateima @melisaguevara @james-a-morris This PR drops support for legacy V3 events, meaning they won't be able to be queried via the SpokePoolClient anymore. @amateima noted that you're still indexing older events, so I just wanted to check it off with you to confirm if/when this is OK to merge RE the indexer.

@james-a-morris @amateima I think we should wait until we finish the production backfilling to merge this. Do you agree?

melisaguevara (approved):

bmzig (approved):

PR #909, opened 2025-03-03
feat: expose a custom Address class

bmzig:
This is a part of https://github.com/across-protocol/sdk/pull/904. We want to expose this early and not integrate it into the SDK so that we may integrate the class into some aspects of the relayer, like the rebalancer or finalizer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.1.20",
+  "version": "4.1.21",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "files": [
@@ -112,6 +112,7 @@
     "async": "^3.2.5",
     "axios": "^0.27.2",
     "big-number": "^2.0.0",
+    "bs58": "^6.0.0",
     "decimal.js": "^10.3.1",
     "ethers": "^5.7.2",
     "lodash": "^4.17.21",
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -1,5 +1,7 @@
 import { providers, utils } from "ethers";
-import { BigNumber } from "./BigNumberUtils";
+import bs58 from "bs58";
+import { isAddress as isSvmAddress } from "@solana/web3.js";
+import { BigNumber, chainIsEvm } from "./";
 
 /**
  * Checks if a contract is deployed at the given address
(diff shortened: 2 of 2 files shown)


src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,170 @@ export function isValidEvmAddress(address: string): boolean {
     return false;
   }
 }
+
+// Creates the proper address type given the input chain ID corresponding to the address's origin network.
+// @todo: Change this to `toAddress` once we remove the other `toAddress` function.
```
dohaki, line 75: NIT: We should try to use TSDoc style comments for this type of function/method documentation.

dohaki (commented):
Looks really good 👍 One nit: A lot of the function/method comments are helpful but it would be nice to follow a standard way of formatting, e.g. TSDoc style.

Also can we add some unit tests? Or will they be part of a another PR?

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,170 @@ export function isValidEvmAddress(address: string): boolean {
...
+  // Checks if the address is valid on the given chain ID.
+  isValidOn(chainId: number): boolean {
+    if (chainIsEvm(chainId)) {
+      return this.isValidEvmAddress();
+    }
+    return this.isValidSvmAddress();
+  }
+
+  // Checks if the object is an address by looking at whether it has an Address constructor.
+  static isAddress(obj: Record<string, unknown>): boolean {
```
pxrl, line 147: Does `obj` need to be a Record type, or is `unknown` sufficient?

```suggestion
  static isAddress(obj: unknown): boolean {
```

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,170 @@ export function isValidEvmAddress(address: string): boolean {
...
+   */
+  async isContractDeployedToAddress(provider: providers.Provider): Promise<boolean> {
+    if (this.isValidEvmAddress()) {
+      const code = await provider.getCode(this.toAddress());
+      // If the code is not empty, then there is a contract at this address
+      return code !== "0x";
+    }
+    // @todo Make this work for SVM
+    return false;
+  }
```
pxrl, line 179: This feels like it could belong outside of the Address class - especially because it adds a dependency on the Provider class. wdyt about factoring it out to a standalone AddressUtils function?

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,170 @@ export function isValidEvmAddress(address: string): boolean {
...
+
+  // Compares Addresses by first converting them to BigNumbers.
+  compareAddresses(otherAddress: Address): 1 | -1 | 0 {
```
pxrl, line 182: Minor nit (and this is close to bikeshedding): I prefer shorter names because these expressions can get chained and create very long lines, leading to line wrapping.

```suggestion
  compare(otherAddress: Address): 1 | -1 | 0 {
```

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,170 @@ export function isValidEvmAddress(address: string): boolean {
...
+    // Convert address strings to BigNumbers and then sort numerical value of the BigNumber, which sorts the addresses
+    // effectively by their hex value.
+    const bnAddressA = BigNumber.from(this.toBytes32());
```
pxrl, line 185: I like that this is available as a utility function, but this repetitive example might consume quite some CPU if we're looping on it in a large array. I wonder if the Address class should lazily-evaluate this and then store it for subsequent accesses. A getter method could help with this.

bmzig: That's a good idea https://github.com/across-protocol/sdk/pull/909/commits/92539941d0dcc6c8a974822b9c84d5c5e524a465

src/utils/AddressUtils.ts
bmzig: I just removed it. I originally had it here since I wanted to move all the address utils functions to the class itself, but since we aren't migrating those other functions yet, I think it's fine to just remove here. https://github.com/across-protocol/sdk/pull/909/commits/92539941d0dcc6c8a974822b9c84d5c5e524a465

src/utils/AddressUtils.ts
bmzig: Unknown is sufficient. I just wanted to be more verbose here https://github.com/across-protocol/sdk/pull/909/commits/92539941d0dcc6c8a974822b9c84d5c5e524a465

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,170 @@ export function isValidEvmAddress(address: string): boolean {
...
+    // Further validation is done on the subclasses so that we can guarantee that, for example, an EvmAddress type will always contain a valid, 20-byte
+    // EVM address.
+    if (_rawAddress.length > 32) {
+      throw new Error(`Address ${utils.hexlify(_rawAddress)} cannot be longer than 32 bytes.`);
+    }
+    // Ensure all addresses in this class are internally stored as 32 bytes.
+    this.rawAddress = utils.zeroPad(_rawAddress, 32);
+  }
+
+  // Converts the address into a bytes32 string. Note that the output bytes will be lowercase  so that it matches ethers event data. This function will never
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/sdk/pull/909/commits/92539941d0dcc6c8a974822b9c84d5c5e524a465

src/utils/AddressUtils.ts
bmzig: Good call. https://github.com/across-protocol/sdk/pull/909/commits/92539941d0dcc6c8a974822b9c84d5c5e524a465.

I think it is generally good to have a single comment format. Should all functions, including class methods, be TSDoc style as well?

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,170 @@ export function isValidEvmAddress(address: string): boolean {
     return false;
   }
 }
+
+// Creates the proper address type given the input chain ID corresponding to the address's origin network.
+// @todo: Change this to `toAddress` once we remove the other `toAddress` function.
+export function toAddressType(address: string, chainId: number): EvmAddress | SvmAddress {
```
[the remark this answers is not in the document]
bmzig: I generally do not like using `undefined` if I can help it. Do we really want to do this? I'd think it would be preferable to not silently fail in these instances.

package.json
```diff
@@ -112,6 +112,7 @@
     "async": "^3.2.5",
     "axios": "^0.27.2",
     "big-number": "^2.0.0",
+    "bs58": "^6.0.0",
```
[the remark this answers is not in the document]
bmzig: Interesting. I also saw that ethers.js exposes a base58 encoding utility, but I wanted to use `bs58` since 1. We already have `bs58` in our `yarn.lock` since it is a dependency for a variety of our other packages, so we can just use it "for free," and 2. I wanted to move away from depending on ethers or other chain-specific SDKs since they tend to make assumptions about the data which is correct for that specific chain but might be somewhat annoying to deal with if you are using it for another unrelated purpose.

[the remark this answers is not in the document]
gsteenkamp89: Oh yeah that totally makes sense! we don't want any 3rd party assumptions ruining our weekend

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,175 @@ export function isValidEvmAddress(address: string): boolean {
...
+ * @param address Stringified address type to convert. Can be either hex encoded or base58 encoded.
+ * @param chainId Network ID corresponding to the input address, used to determine which address type to output.
+ * @returns a child `Address` type most fitting for the chain ID.
+ * @todo: Change this to `toAddress` once we remove the other `toAddress` function.
+ */
+export function toAddressType(address: string, chainId: number): EvmAddress | SvmAddress {
+  if (chainIsEvm(chainId)) {
+    return EvmAddress.from(address);
+  }
+  return SvmAddress.from(address);
```
pxrl, line 85: @bmzig Do we need another type to handle addresses that we consider invalid for the specified chain? For example, an EVM bytes32 address w/o the upper 12 bytes being padded w/ 0. Or alternatively, specifying an EVM address as the recipient for a deposit destined for SVM?

We're unable to drop these events, but we won't be able to handle them in EvmAddress or SvmAddress. I'm wondering if we also need an InvalidAddress type. This would have benefit in (for example) the BundleDataClient, where we currently have logic to detect this.

bmzig: Interesting. So I modified this function to be able to return an `Address` here https://github.com/across-protocol/sdk/pull/909/commits/f6c9966d6ea518a260c850abc9fb374593b2789d. I can see this potentially causing issues if we use `toAddress` to often in other parts of the code, so I also think it is necessary to override a `toAddress` modifier for the subclasses https://github.com/across-protocol/sdk/pull/909/commits/e323b526501ad2a930d185910c68ac1c782fa376. Do you think that's sufficient? 

As a side, when we are defining these addresses, we need to be aware that using `EvmAddress.from` or `SvmAddress.from` will return a validated address, while `toAddressType` may or may not return a validated address type. Not sure if that could cause issues or if there is a way around the edge, but I don't think we can avoid some way of instantiating unsafe addresses while also doing validation on the addresses.

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,170 @@ export function isValidEvmAddress(address: string): boolean {
...
+    }
+  }
+
+  // Override `toHexString` so `EvmAddress` types will be encoded as `address` types automatically by ethers.js
+  override toHexString(): string {
+    return this.toAddress();
+  }
+
+  // Constructs a new EvmAddress type.
+  static from(hexString: string): EvmAddress {
```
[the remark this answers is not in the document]
bmzig: I think the only downside to this is going back to a string type, since we will presumably need to continue calling functions to swap to/from address types, and I think it would just be cleaner if we could slot these functions in a class instead of static functions. I'm on the fence either way, but I do think that being able to lazily evaluate addresses in the class improves the case for keeping subclasses of Address.

src/utils/AddressUtils.ts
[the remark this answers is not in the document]
pxrl: I don't consider returning `undefined` as a silent failure, because tsc forces the caller to verify that the result was not `undefined` on return. If we instead throw, it requires the caller to remember to wrap the call in a try/catch clause, which clutters the code and is unlikely to be consistently used. So I think returning `undefined` is generally an OK approach.

(Returning a generic Address is probably a nice way around having to do this though)

src/utils/AddressUtils.ts
```diff
@@ -68,3 +70,187 @@ export function isValidEvmAddress(address: string): boolean {
...
+    return new this(utils.arrayify(hexString));
+  }
+}
+
+// Subclass of address which strictly deals SVM addresses. These addresses are guaranteed to be valid SVM addresses, so `toBase58` will always produce a valid Solana address.
+export class SvmAddress extends Address {
```
nicholaspai, line 237: How do you think we should handle translation from bytes20 to bytes32. This is important when we see a PoolRebalanceRoute/CrossChainContractsSet event from the HubPool which emits `address`'s that we'll have to map to Solana addresses. The [SolanaAdapter](https://github.com/across-protocol/contracts/blob/a6bd1bdc948b0eb0075cc0e753a26d2ff70acd6e/contracts/chain-adapters/Solana_Adapter.sol#L45) stores hardcoded btyes32 addresses of USDC (the only Solana token we expect to support for now) and the Solana_SpokePool, so I wonder if we need to hardcode those addresses in here as well.

amateima:
From my perspective this looks good. The indexer won't deal for now with creating Address instances, its usage being limited to calling `toAddress()` or `toHexString()` for getting the value that will end up being stored in the DB

src/utils/AddressUtils.ts
pxrl: In case of an invalid address, can we just revert to `toAddress()` printing in bytes32 base16 representation? That's the closest to a "raw" address and it should still fail to match on anything else.

src/utils/AddressUtils.ts
pxrl: Yeah; I think we need to read this translation onchain via config (i.e. for each PoolRebalanceRoute token, we query some other contract for the SVM representation of it). Ideally we can do this via a multicall, rather than having to index events.

pxrl (commented):
@bmzig I responded to a couple of open points above, but otherwise I think this is looking good and is worth getting in so we can iterate further in-tree.

src/utils/AddressUtils.ts
bmzig: Sounds good. I agree that this is probably OK to do, so it is done here https://github.com/across-protocol/sdk/blob/f4a99deea3a8401492aa5892c7f3ddf6c344490a/src/utils/AddressUtils.ts#L145.

pxrl (approved):

PR #910, opened 2025-03-06
refactor(SpokePoolClient): Drop historical depositId queries

pxrl:
These queries seem to be mostly superfluous; they can be dropped to ease non-EVM integrations.


src/clients/SpokePoolClient.ts
```diff
@@ -562,7 +544,7 @@ export class SpokePoolClient extends BaseAbstractClient {
     });
 
     const timerStart = Date.now();
-    const multicallFunctions = ["getCurrentTime", "numberOfDeposits"];
+    const multicallFunctions = ["getCurrentTime"];
```
pxrl, line 547: I left this intact multicall query in case there's anything else we want to query during update - it's easy to just append functions here. Is there likely to be anything else? If not we could alternatively just revert to a `SpokePool.getCurrentTime()` call.

src/utils/DepositUtils.ts
```diff
@@ -116,17 +99,24 @@ export async function queryHistoricalDepositForFill(
...
-    deposit = await spokePoolClient.findDeposit(fill.depositId, fill.destinationChainId);
+    const result = await spokePoolClient.findDeposit(fill.depositId);
+    if (!result.found) {
+      return result;
+    }
+
+    ({ deposit } = result);
     if (cache) {
-      await setDepositInCache(deposit, getCurrentTime(), cache, DEFAULT_CACHING_TTL);
+      await setDepositInCache(deposit!, getCurrentTime(), cache, DEFAULT_CACHING_TTL);
```
pxrl, line 109: Nasty - fix this.

src/clients/mocks/MockSpokePoolClient.ts
```diff
@@ -107,19 +106,10 @@ export class MockSpokePoolClient extends SpokePoolClient {
...
-      (depositId, event) => bnMax(depositId, event.args["depositId"] ?? bnZero),
-      this.latestDepositIdQueried
-    );
-
     return Promise.resolve({
       success: true,
       firstDepositId: bnZero,
-      latestDepositId,
       currentTime,
-      oldestTime: 0,
```
pxrl, line 122: Driveby fix - this hasn't been needed for a long time.

src/utils/DepositUtils.ts
pxrl: https://github.com/across-protocol/sdk/pull/910/commits/2adc1aab97c2d5d1e7dff761c8b7ce245576983a

nicholaspai (approved):

PR #911, opened 2025-03-06
fix(svm): handle errors in cached solana provider

Reinis-FRP:
This adds error handling in the cached Solana RPC transport. Specifically, now it does not cache and pass through json-rpc response to rpc client if:
- `getTransaction` underlying transport returns JSON-RPC error response
- `getSignatureStatuses` RPC client request errors


md0x (approved):
Looks perfect to me!

PR #912, opened 2025-03-11
feat(utils): Add findDepositBlock

pxrl:
This implements the bare minimum functionality needed to identify the block number for a deposit. It performs a similar function to getBlockRangeForDepositId() but has no dependency on a SpokePoolClient instance, and is less strict about the inputs. The underlying algorithm is simple: find the lowest blockNumber where numberOfDeposits is greater than the target deposit Id. This effectively mirrors the behaviour of findFillBlock().

The main motivator for this is to push chain-specific knowledge out of the SpokePoolClient in advance of non-EVM code being introduced.


src/utils/SpokeUtils.ts
```diff
@@ -333,6 +333,48 @@ export function getRelayHashFromEvent(e: RelayData & { destinationChainId: numbe
...
+  lowBlockNumber: number,
+  highBlockNumber?: number
+): Promise<number | undefined> {
+  // We can only perform this search when we have a safe deposit ID.
+  if (isUnsafeDepositId(depositId)) {
+    throw new Error(`Cannot binary search for depositId ${depositId}`);
+  }
+
+  // @todo this call can be optimised away by clever use of multicall
+  highBlockNumber ??= await spokePool.provider.getBlockNumber();
```
pxrl, line 348: This is fwiw how it can be done w/ multicall, where if `highBlockNumber` is not supplied, it is read back from `Multicall3.blockAndAggregate(..., { blockTag: "latest" })`. I like the idea but it seems like a lot of code without great readability in order to optimise a single RPC call away. Feels like some interface work is needed to abtract the complexity out.

```
 export async function findDepositBlock(
+  chainId: number,
   spokePool: Contract,
   depositId: BigNumber,
   lowBlockNumber: number,
@@ -344,18 +347,27 @@ export async function findDepositBlock(
     throw new Error(`Cannot binary search for depositId ${depositId}`);
   }
 
-  // @todo this call can be optimised away by clever use of multicall
-  highBlockNumber ??= await spokePool.provider.getBlockNumber();
-  assert(highBlockNumber > lowBlockNumber, `Block numbers out of range (${lowBlockNumber} >= ${highBlockNumber})`);
+  const multicall = getMulticall3(chainId, spokePool.provider);
+  assert(isDefined(multicall));
 
   // Make sure the deposit occurred within the block range supplied by the caller.
-  const [nDepositsLow, nDepositsHigh] = (
+  const [_nDepositsLow, multicallResult] = (
     await Promise.all([
       spokePool.numberOfDeposits({ blockTag: lowBlockNumber }),
-      spokePool.numberOfDeposits({ blockTag: highBlockNumber }),
+      multicall.callStatic.blockAndAggregate([{
+        target: spokePool.address,
+        callData: spokePool.interface.encodeFunctionData("numberOfDeposits")
+      }], { blockTag: highBlockNumber })
     ])
-  ).map((n) => toBN(n));
+  );
+
+  const nDepositsLow = toBN(_nDepositsLow);
+  const { blockNumber: _highBlockNumber, returnData } = multicallResult;
+  highBlockNumber = Number(_highBlockNumber);
+  assert(highBlockNumber > lowBlockNumber, `Block numbers out of range (${lowBlockNumber} >= ${highBlockNumber})`);
 
+  const [_nDepositsHigh] = spokePool.interface.decodeFunctionResult("numberOfDeposits", returnData.at(0)!.returnData);
+  const nDepositsHigh = toBN(_nDepositsHigh);
   if (nDepositsLow.gt(depositId) || nDepositsHigh.lte(depositId)) {
     return undefined; // Deposit did not occur within the specified block range.
   }
```

src/utils/SpokeUtils.ts
```diff
@@ -333,6 +333,49 @@ export function getRelayHashFromEvent(e: RelayData & { destinationChainId: numbe
...
+
+  // Find the lowest block number where numberOfDeposits is greater than the requested depositId.
+  do {
+    const midBlockNumber = Math.floor((highBlockNumber + lowBlockNumber) / 2);
+    const numberOfDeposits = toBN(await spokePool.numberOfDeposits({ blockTag: midBlockNumber }));
+
+    if (numberOfDeposits.gt(depositId)) {
+      highBlockNumber = midBlockNumber;
+    } else {
+      lowBlockNumber = midBlockNumber + 1;
```
bmzig, line 372: Should this be `midBlockNumber + 1` or just `midBlockNumber`? It seems like there is a fairly high chance that as we approach the block number of the deposit, `midBlockNumber == deposit.blockNumber`, which would cause the new bounds to be completely out of the deposit's block range range (i.e. [deposit.blockNumber + 1, highBlockNumber]).

src/utils/SpokeUtils.ts
pxrl: I factored out most of the complexity here: https://github.com/across-protocol/sdk/pull/915

src/utils/SpokeUtils.ts
pxrl: `lowBlockNumber = midBlockNumber` is prone to deadlock on the final iteration where `lowBlockNumber`and `highBlockNumber` are adjacent, because it reverts to `lowBlockNumber = lowBlockNumber` due to the rounding on the selection of `midBlockNumber`.

I did spend a while (probably too long) reasoning about this, just to be sure this was the correct approach. One thing to note is that this might all be a bit less intuitive because the loop is only indirectly checking for a specific `depositId` via the `numberOfDeposits` variable, which is always 1 ahead of the highest sequential `depositId`.

Playing out the scenario where the target block is actually `midBlockNumber`, this should revert to assigning `highBlockNumber = midBlockNumber` because `numberOfDeposits.gt(depositId)` implies that the deposit lies below or equal to `midBlockNumber`.

For the other case, `numberOfDeposits.lte(depositId) => true` implies that the deposit lies _only_ above `lowBlockNumber`, so it's safe to increment.

src/utils/SpokeUtils.ts
pxrl: Implemented: https://github.com/across-protocol/sdk/pull/912/commits/da4d11ff09ea79e96b285fb7cbf0b915828beea8

src/utils/SpokeUtils.ts
pxrl: Updated for clarity:

```
    if (nDeposits.gt(depositId)) {
      highBlock = midBlock; // depositId occurred at or earlier than midBlock.
    } else {
      lowBlock = midBlock + 1; // depositId occurred later than midBlock.
    }
```

src/clients/SpokePoolClient.ts
```diff
@@ -930,25 +930,18 @@ export class SpokePoolClient extends BaseAbstractClient {
...
+    const upperBound = this.latestBlockSearched || undefined; // Don't permit block 0 as the high block.
+    const fromBlock = await findDepositBlock(this.spokePool, depositId, this.deploymentBlock, upperBound);
+    if (!fromBlock) {
+      const chain = getNetworkName(this.chainId);
+      throw new Error(
+        `Unable to find ${chain} depositId ${depositId} within blocks [${this.deploymentBlock}, ${upperBound}]`
+      );
+    }
 
+    const toBlock = fromBlock;
```
pxrl, line 942: One nice side-effect of isolating the deposit to a single block is that we can also identify what type of event to expect. For this we'd need to load in the block number for each protocol upgrade, but that's actually easy enough to add.

bmzig (approved):

PR #913, opened 2025-03-11
improve(install): make SDK package-manager agnostic

james-a-morris:
The postinstall process currently relies on `yarn` to update call the `build-bigint-buffer`. This will cause issues if a repo attempting to install it is not using `yarn`. We see this error in the `indexer`.

These changes update the `postinstall` script process to use a OS/package-manager agnostic, dependency free JS script to imitate the existing `build-bigint-buffer` embedded script.


.eslintrc.js
james-a-morris, line 1: This was a fix for warnings received by prettier

pxrl (approved):

PR #920, opened 2025-03-17
improve: expose multicall3 test utilities

bmzig:
Context for this PR is here https://github.com/across-protocol/relayer/pull/2143#issuecomment-2730653767


src/utils/Multicall.ts
```diff
@@ -94,3 +96,32 @@ export async function blockAndAggregate(
...
+  return code !== "0x";
+}
+
+/**
+ * Deploy a Multicall3 instance.
+ * Guidance: https://github.com/mds1/multicall3?tab=readme-ov-file#new-deployments
+ * @param signer An Ethers Signer instance with at least 0.1 ETH.
+ */
+export async function deploy(signer: Signer) {
+  assert(signer.provider);
```
pxrl, line 113: @bmzig OK to restrict this when chainId is a testnet? The Multicall3 docs have some pretty strict wording about not messing up the deployment because you only get one shot at it, so it'd be nice for me to not be the reason that a chain can't have a nice deterministic address :sweat_smile:

src/utils/Multicall.ts
```diff
@@ -16,6 +17,7 @@ export type Call3 = {
 };
 
 const DETERMINISTIC_MULTICALL_ADDRESS = "0xcA11bde05977b3631167028862bE2a173976CA11";
+const MULTICALL_DEPLOYER = "0x05f32b3cc3888453ff71b01135b34ff8e41263f2";
```
pxrl, line 20: I missed making this a checksum address in the first pass.

```suggestion
const MULTICALL_DEPLOYER = "0x05f32B3cC3888453ff71B01135B34FF8e41263F2";
```

src/utils/Multicall.ts
```diff
@@ -94,3 +96,32 @@ export async function blockAndAggregate(
 
   return { blockNumber: blockNumber.toNumber(), returnData };
 }
+
+async function isDeployed(provider: providers.Provider): Promise<boolean> {
+  const chainId = (await provider.getNetwork()).chainId;
```
pxrl, line 101: Minor nit

```suggestion
  const { chainId } = await provider.getNetwork();
```

src/utils/Multicall.ts
bmzig: I think this does it, right? https://github.com/across-protocol/sdk/pull/920/commits/2c9fefee8d4ed67ebdc35fb00ed997b69c76ecb2 (also made some mistakes in that commit that I has to fix here https://github.com/across-protocol/sdk/pull/920/commits/f669d1e675c782ab3114a06c503d3e0f4a740aad)

pxrl (approved):
LGTM!

PR #930, opened 2025-03-21
refactor(SpokePoolClient): Squash redundant block-to-timestamp function

pxrl:
SpokePoolClient.getTimeAt() does the same.


gsteenkamp89 (approved):
Looks good.
ooc Is there any loss in performance by querying the block timestamp **via the contract**

PR #932, opened 2025-03-21
improve: implement a common interface in sdk providers

bmzig:
Ideally we can make custom implementations for our providers (e.g. `speedProvider`) and also make providers for other networks (e.g. a Solana provider) while still using a common utility type (e.g. BlockFinder). I think a good way to achieve this is to extend these providers from a common interface and start utilizing generic types more.

### DIFF
--- a/src/providers/solana/cachedRpcFactory.ts
+++ b/src/providers/solana/cachedRpcFactory.ts
@@ -6,8 +6,9 @@ import { SolanaClusterRpcFactory } from "./baseRpcFactories";
 import { RateLimitedSolanaRpcFactory } from "./rateLimitedRpcFactory";
 import { CacheType } from "../utils";
 import { jsonReplacerWithBigInts, jsonReviverWithBigInts } from "../../utils";
+import { CrosschainProvider } from "../";
 
-export class CachedSolanaRpcFactory extends SolanaClusterRpcFactory {
+export class CachedSolanaRpcFactory extends SolanaClusterRpcFactory implements CrosschainProvider {
   public readonly getTransactionCachePrefix: string;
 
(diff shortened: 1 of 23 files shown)


src/utils/BlockUtils.ts
```diff
@@ -69,10 +70,10 @@ export async function averageBlockTime(
   blockRange ??= defaultBlockRange;
 
   const earliestBlockNumber = highBlock - blockRange;
-  const [firstBlock, lastBlock] = await Promise.all([
+  const [firstBlock, lastBlock] = (await Promise.all([
     provider.getBlock(earliestBlockNumber),
     provider.getBlock(highBlock),
-  ]);
```
bmzig, line 75: Unsafe cast here. Marking as need to change.

src/providers/solana/cachedRpcFactory.ts
```diff
@@ -62,6 +63,26 @@ export class CachedSolanaRpcFactory extends SolanaClusterRpcFactory {
...
+    // TODO: Support other evm block tags.
+    let slot;
+    try {
+      slot = BigInt(blockNumber);
+    } catch {
+      // Assume block tag = "latest";
+      slot = BigInt(await this.getBlockNumber());
+    }
+    return this.rateLimitedRpcClient.getBlock(slot, {});
+  }
```
pxrl, line 80: This could be simplified by narrowing the string type.

```suggestion
  async getBlock(blockNumber: number | "latest"): Promise<unknown> {
    const slot = blockNumbe === "latest"
      ? BigInt(await this.getBlockNumber())
      : BigInt(blockNumber);

    return this.rateLimitedRpcClient.getBlock(slot, {});
  }
```

PR #936, opened 2025-03-26
refactor(addressAggregator): Consolidate functions into abstract class

pxrl:
This shrinks the adapters and makes them more configurable, such that an adapter can be instructed to throw or return an empty array on error.


dohaki (approved):
Nice refactor 👍

PR #937, opened 2025-03-26
chore: bump constants

bmzig:


bmzig:
This isn't necessary anymore due to https://github.com/across-protocol/sdk/pull/938

PR #941, opened 2025-03-26
fix(SpokePoolClient): Use chain time instead of SpokePool time

pxrl:
These are 1:1 in production but some quirks in test are causing failures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.1.30",
+  "version": "4.1.31",
   "license": "AGPL-3.0",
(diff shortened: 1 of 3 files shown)


src/clients/BundleDataClient/BundleDataClient.ts
pxrl, line 1: This could probably be backed out after having changed the implementation of `getTimeAt()`.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.1.29",
+  "version": "4.1.30",
```
bmzig, line 4:
```suggestion
  "version": "4.1.31",
```

bmzig (approved):

PR #943, opened 2025-03-27
fix(build): propagate build failures in parallel builds

james-a-morris:
Add --kill-others-on-fail flag to concurrently to ensure build failures are properly detected and propagated. This prevents scenarios where one build process fails silently while others succeed, masking the error.


package.json
```diff
@@ -16,7 +16,7 @@
     "build-bigint-buffer": "node scripts/build-bigint-buffer.js",
     "postinstall": "node scripts/build-bigint-buffer.js",
     "start": "yarn typechain && nodemon -e ts,tsx,json,js,jsx --watch ./src --ignore ./dist --exec 'yarn dev'",
-    "build": "yarn run clean && yarn typechain && yarn run build:cjs & yarn run build:esm & yarn run build:types; wait",
+    "build": "yarn run clean && yarn typechain && concurrently --kill-others-on-fail --names \"cjs,esm,types\" --prefix-colors \"blue,magenta,green\" \"yarn run build:cjs\" \"yarn run build:esm\" \"yarn run build:types\"",
     "dev": "yarn run build:cjs & yarn run build:esm & yarn run build:types; wait",
```
pxrl, line 20: How about something like this? btw, do we need the escaped quotes, or could we use `'` instead?

```suggestion
    "build": "yarn run clean && yarn typechain && yarn dev",
    "dev": "concurrently --kill-others-on-fail --names \"cjs,esm,types\" --prefix-colors \"blue,magenta,green\" \"yarn run build:cjs\" \"yarn run build:esm\" \"yarn run build:types\"",
```

james-a-morris: https://github.com/across-protocol/sdk/pull/943/commits/5861a3894d38ad39953a1b1468ab364355c06386

pxrl (approved):
Nice catch!

PR #953, opened 2025-03-31
improve: allow pool rebalance routes to be removed

bmzig:
This PR makes two improvements:
- Every time we observe a `L1TokenEnabledForLiquidityProvision`, we distinguish the token based on its symbol. This is unhelpful for testnet since there are multiple tokens we need to support with the same symbol (e.g. multiple definitions of different `WETH` tokens). We now discriminate these tokens based on their address.
-  Pool rebalance routes could not be set back to `address(0)` since we query `SetPoolRebalanceRoute` events and set `l1TokensToDestinationTokens` the most up-to-date event arguments. Now, if we do observe a `SetPoolRebalanceRoute` where `destinationToken` is `address(0)`, we interpret this as "removing" that pool rebalance route, so we delete the mapping. Note that since these events are sorted, we can still set it to zero and then later reset it to a proper value. This was needed since the Sepolia WETH address for Tatara actually had a defined pool rebalance route for Base (set to the zero address).

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -2,7 +2,7 @@ import assert from "assert";
 import { Contract, EventFilter } from "ethers";
 import _ from "lodash";
 import winston from "winston";
-import { DEFAULT_CACHING_SAFE_LAG, DEFAULT_CACHING_TTL } from "../constants";
+import { DEFAULT_CACHING_SAFE_LAG, DEFAULT_CACHING_TTL, ZERO_ADDRESS } from "../constants";
 import {
   CachingMechanismInterface,
   CancelledRootBundle,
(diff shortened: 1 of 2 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -918,21 +918,29 @@ export class HubPoolClient extends BaseAbstractClient {
...
+                blockNumber: args.blockNumber,
+                transactionIndex: args.transactionIndex,
+                logIndex: args.logIndex,
+                transactionHash: args.transactionHash,
+              },
+            ]
+          );
+        } else {
+          // Clear out the mapping for the L1 token entirely.
+          delete this.l1TokensToDestinationTokens[args.l1Token][args.destinationChainId];
```
nicholaspai, line 941: This might lead to some where artifacts where `this.l1TokensToDestinationTokens[args.l1Token]` is an empty dict because all of its destination chains are deleted--not sure if there are any problems downstream though

nicholaspai (approved):

mrice32 (approved):

src/clients/HubPoolClient.ts
bmzig: It probably doesn't hurt to delete these too. I'm sure we iterate over these values, https://github.com/across-protocol/sdk/pull/953/commits/c864ee6a0350a22f99f4a785f1e34913b01220aa

nicholaspai (approved):

PR #956, opened 2025-04-01
feat: enable support for SVM in hubpool client

james-a-morris:
Ongoing work to incorporate solana into the hubpool client

### DIFF
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -46,6 +46,9 @@ import {
   getL1TokenInfo,
   compareAddressesSimple,
   deleteFromJson,
+  chainIsSvm,
+  getDeployedAddress,
+  SvmAddress,
 } from "../utils";
 import { AcrossConfigStoreClient as ConfigStoreClient } from "./AcrossConfigStoreClient/AcrossConfigStoreClient";
 import { BaseAbstractClient, isUpdateFailureReason, UpdateFailureReason } from "./BaseAbstractClient";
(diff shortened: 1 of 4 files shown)


src/clients/HubPoolClient.ts
```diff
@@ -902,18 +905,33 @@ export class HubPoolClient extends BaseAbstractClient {
...
+          blockNumber: args.blockNumber,
+          transactionIndex: args.transactionIndex,
+          logIndex: args.logIndex,
+        };
+        // If the chain is SVM then our `args.spokePool` will be set to the `solanaSpokePool.toAddressUnchecked()` in the
+        // hubpool event because our hub deals with `address` types and not byte32. Therefore, we should confirm that the
+        // `args.spokePool` is the same as the `solanaSpokePool.toAddressUnchecked()`. We can derive the `solanaSpokePool`
+        // address by using the `getDeployedAddress` function.
+        if (chainIsSvm(args.l2ChainId)) {
+          const solanaSpokePool = getDeployedAddress("SvmSpoke", args.l2ChainId);
```
pxrl, line 919: Thought bubble - do we want to reference a Solana deployment as `SvmSpoke` ? It might be a mistake to assume a 1:1 link that SVM === Solana.

There are other chains using the SVM and these could feasibly be supported by Across (i.e. [Eclipse](https://www.eclipse.xyz)), in which case we'd end up with naming inconsistencies. Should we consider naming this deployment `SolanaSpoke` instead?

(nb. I realise this is mostly an upstream question - tagging @Reinis-FRP @md0x for their thoughts on this)

james-a-morris: Agreed here - I was referencing the contracts repo directly - should I propose that change?

src/clients/HubPoolClient.ts
```diff
@@ -902,35 +905,67 @@ export class HubPoolClient extends BaseAbstractClient {
...
+          const truncatedAddress = SvmAddress.from(solanaSpokePool).toEvmAddress();
+          // Verify the event address matches our expected truncated address
+          if (args.spokePool.toLowerCase() !== truncatedAddress.toLowerCase()) {
+            throw new Error(
+              `SVM spoke pool address mismatch for chain ${args.l2ChainId}. ` +
+                `Expected ${truncatedAddress}, got ${args.spokePool}`
+            );
+          }
+          // Store the full Solana address
+          dataToAdd.spokePool = solanaSpokePool;
```
pxrl, line 932: This address is formatted as a base58-encoded string, which makes a comparison with the truncated address difficult. Any considerations about whether we should reformat this to base 16? @bmzig's Address class might help with this.

src/utils/AddressUtils.ts
```diff
@@ -241,6 +241,10 @@ export class SvmAddress extends Address {
     return this.toBase58();
   }
 
+  override toEvmAddress(): string {
+    return `0x${this.toBytes32().slice(-40)}`;
+  }
```
pxrl, line 246: We probably want to convert this to checksum case after truncating it.

pxrl (approved):

PR #957, opened 2025-04-01
refactor: Relocate evm-specific SpokeUtils

pxrl:


src/arch/evm/SpokeUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+    inputToken: toBytes32(deposit.inputToken),
+    outputToken: toBytes32(deposit.outputToken),
+    inputAmount: deposit.inputAmount,
+    outputAmount: deposit.outputAmount,
+    originChainId: deposit.originChainId,
+    depositId: deposit.depositId,
+    fillDeadline: deposit.fillDeadline,
+    exclusivityDeadline: deposit.exclusivityDeadline,
+    message: deposit.message,
+  };
```
james-a-morris, line 49: Thoughts on breaking this out? I suspect that this data structure will be shared by the SVM counterpart

james-a-morris (commented):
drive by comment

src/arch/evm/SpokeUtils.ts
pxrl: Good call - added here: https://github.com/across-protocol/sdk/pull/957/commits/9afa6ef60a078d6e7b97d4f6154028c7d5648497

james-a-morris (approved):

PR #958, opened 2025-04-01
feat: Add CCTP V2 support on Linea

nicholaspai:


src/utils/NetworkUtils.ts
```diff
@@ -156,6 +156,7 @@ export function chainIsCCTPEnabled(chainId: number): boolean {
     CHAIN_IDs.OPTIMISM,
     CHAIN_IDs.POLYGON,
     CHAIN_IDs.UNICHAIN,
+    CHAIN_IDs.LINEA,
```
pxrl, line 159: wdyt about this? https://github.com/across-protocol/sdk/pull/959

It should be OK to use it without adding Linea to `cctpExceptions` because we've disabled Linea for USDC.

src/utils/NetworkUtils.ts
```diff
@@ -170,7 +171,7 @@ export function chainIsCCTPEnabled(chainId: number): boolean {
  * @returns True if chainId requires manual L1 -> L2 finalization, otherwise false.
  */
 export function chainRequiresL1ToL2Finalization(chainId: number): boolean {
-  return chainIsCCTPEnabled(chainId) || chainIsLinea(chainId);
```
pxrl, line 173: What about non-USDC deposits? We still need to finalise for those.

bmzig: I think I'm missing the point here, but doesn't this function only get used by the finalizer? The bridge adapters are configured per-bridge, and it looks like we only use this function to add some extra addresses to check when finalizing L1->L2 deposits https://github.com/across-protocol/relayer/blob/177420f58a66f11900206bb0eff916c53449b4d4/src/finalizer/index.ts#L240. 

To be clear, it seems like we would only then be running the cctpFinalizer on those chains (since that is the defined L1->L2 finalizer), so we wouldn't check other tokens.

nicholaspai: Yeah I agree this is only used by the finalizer

nicholaspai: Also we won't need this function anymore following the relayer PR accompanying this SDK PR: https://github.com/across-protocol/relayer/pull/2165/files#diff-d8c1bd620fbb6212fbcc45c4aed169001b34a3fb5dbe3731902fc9af923fab7d

So I'll remove this function

PR #965, opened 2025-04-04
chore: create skeleton for svm spoke client

james-a-morris:
This PR simply specs out the SVM spoke client - it's not meant to be merged to master and will live in the epic while @melisaguevara and I build out the function specifics


melisaguevara (approved):
looks good!

PR #966, opened 2025-04-04
improve(SpokePoolClient): Remove redundant relayFillStatus argument

pxrl:
The destinationChainId is implied by the SpokePool being queried, so
remove it as an argument.

While here, add `fillStatusArray` to the set of abstract methods
required by the SpokePoolClient. This is useful down in the relayer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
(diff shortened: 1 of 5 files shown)


pxrl:
Corresponding relayer PR: https://github.com/across-protocol/relayer/pull/2172

bmzig (approved):

PR #968, opened 2025-04-08
refactor(svm): Factor out individual event decoding

pxrl:
Isolate the dependency on Anchor to src/svm/utils/events.ts. Do this in
part by using ethers base58 and base64 implementations.

This was motivated by the need to add some utilities for fetching and 
decoding SVM transactions within the relayer repository, but without
wanting to depend directly on anchor.


gsteenkamp89 (approved):
nice refactor 👍

src/utils/common.ts
```diff
@@ -7,7 +7,7 @@ export type Decimalish = string | number | Decimal;
 export const AddressZero = ethers.constants.AddressZero;
 export const MAX_BIG_INT = BigNumber.from(Number.MAX_SAFE_INTEGER.toString());
 
-export const { keccak256 } = ethers.utils;
+export const { base58, base64, keccak256 } = ethers.utils;
```
pxrl, line 10: @bmzig I see we use `bs58` in `src/utils/AddressUtils.ts`. Is there a specific reason we use that?

(nb. the long term goal is to shuffle off ethers as much as possible, so if there's something better/smaller/whatever then I'm on board with using that. ethers is just convenient because it bundles base64 as well.

bmzig: Yeah you comment about shuffling off of ethers/having a way to work without a dependency on a network-specific library is why. `bs58` is a dependency for other libraries we have (presumably the solana and ethers ones), so we get to add it in our `package.json` without needing to modify `yarn.lock`.

pxrl: Migrated here: https://github.com/across-protocol/sdk/pull/968/commits/bc11c988b29c083a4a3d5031e3a6fd0bdac6aa08

PR #970, opened 2025-04-08
improve: Handle deposits referencing input/output/repayment tokens that do not map to PoolRebalanceRoutes

nicholaspai:
Prerequisite changes needed to prepare for SpokePool that do not enforce deposit route whitelist checking. In this upcoming paradigm, deposits can be sent where the `inputToken` does not correspond to a `PoolRebalanceRoute` mapping. Implements this [UMIP change](https://github.com/UMAprotocol/UMIPs/pull/619/files)

This change should be backwards compatible as all tokens that can be deposited are mapped to PoolRebalanceRoutes

The changes include:
- changing the bundle data client's choice of repayment chain for a deposit where the inputToken and originChain, or outputToken and destinationChain, do not map to a PoolRebalanceRoute. These cases should switch repayment to the originChain.
- building running balances based on refunds for a token that isn't mapped to a pool rebalance route
- SpokePoolClient should gracefully handle deposits where the `outputToken = 0x0` but there are not sufficient PoolRebalanceRoutes to resolve the outputToken


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -372,7 +373,8 @@ export class BundleDataClient {
         if (
           fill.blockNumber < blockRanges[chainIndex][0] ||
           fill.blockNumber > blockRanges[chainIndex][1] ||
-          isZeroValueFillOrSlowFillRequest(fill)
+          isZeroValueFillOrSlowFillRequest(fill) ||
+          invalidOutputToken(fill)
```
pxrl, line 377: What's the intention with this check? `invalidOutputToken()` is only checking against 0x0, but no fill should ever set that.

src/utils/DepositUtils.ts
```diff
@@ -208,6 +208,11 @@ export function isZeroValueDeposit(deposit: Pick<Deposit, "inputAmount" | "messa
   return deposit.inputAmount.eq(0) && isMessageEmpty(deposit.message);
 }
 
+export function invalidOutputToken(deposit: Pick<DepositWithBlock, "outputToken">): boolean {
+  // If the output token is zero address, then it is invalid.
+  return deposit.outputToken === ZERO_ADDRESS;
```
pxrl, line 213: Should this be `ZERO_BYTES` ? I think fwiw we already have a similar function (`isZeroAddress()`) used by the SpokePoolClient to determine whether it needs to sub in a valid address.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: Ah true good call

src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -1,128 +1,191 @@
...
-  // So, save the slow fill under the destination chain, and save the fast fill under its repayment chain.
-  let chainToSendRefundTo = isSlowFill(fill) ? fill.destinationChainId : fill.repaymentChainId;
-  // If the fill is for a deposit originating from the lite chain, the repayment chain is the origin chain
-  // regardless of whether it is a slow or fast fill (we ignore slow fills but this is for posterity).
-  // @note fill.repaymentChainId should already be set to originChainId but reset it to be safe.
-  if (fromLiteChain) {
-    chainToSendRefundTo = fill.originChainId;
+  assert(
+    !_repaymentAddressNeedsToBeOverwritten(relayData),
+    "getRefundInformationFromFill: fill repayment address must be overwritten"
```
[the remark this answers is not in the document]
nicholaspai: good comment. I added this initially as a programmer error where I wanted to prevent someone using this function without also verifying the repayment address is valid. This function only returns the repayment chain and repayment token.

However, we use this function in the SpokePoolClient to figure out whether a fill has been repaid, so I'm just going to remove it and keep the warnings in the Natspec.

src/utils/DepositUtils.ts
nicholaspai: https://github.com/across-protocol/sdk/pull/970/commits/cfabc9b1348c69b6faa5046898b4e8e15be211eb

bmzig (approved):
This looks good to me. I couldn't see anything wrong. A minor note: it seems like `getL1TokenForDeposit` and `getL2TokenForDeposit` will break with the contract upgrade, but the only time we use it is when calculating lp fees, which is addressed by the changes here.

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: Should this check even be in `getApproximateRefundsForBlockRange`? Would this not filter out valid deposits with 0x0 as the output token?

nicholaspai: @bmzig my thinking was that in `getApproximateRefundsForBlockRange` like in this `loadDataFromScratch` function, there is a possibility to match a fill with an `outputToken = 0x0` with a deposit. 

So that's why this check exists there now. WDYT?

nicholaspai:
Yeah I think this PR catches all instances where we use getL1TokenForDeposit and getL2TokenForDeposit but that's definitely a worry of mine. Thanks for reviewing

src/clients/BundleDataClient/BundleDataClient.ts
bmzig: Yep. That makes sense. I think we do want this here, then.

src/clients/BundleDataClient/utils/FillUtils.ts
```diff
@@ -1,128 +1,185 @@
...
-  // If the fill is for a deposit originating from the lite chain, the repayment chain is the origin chain
-  // regardless of whether it is a slow or fast fill (we ignore slow fills but this is for posterity).
-  // @note fill.repaymentChainId should already be set to originChainId but reset it to be safe.
-  if (fromLiteChain) {
-    chainToSendRefundTo = fill.originChainId;
+  const chainToSendRefundTo = _getRepaymentChainId(relayData, hubPoolClient);
+  if (chainToSendRefundTo === relayData.originChainId) {
+    return {
+      chainToSendRefundTo,
+      repaymentToken: relayData.inputToken,
```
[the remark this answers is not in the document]
nicholaspai: how would that work?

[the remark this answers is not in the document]
bmzig: I'm just worried here since there is a discrepancy between the repayment token we return in this case vs the repayment token we returned before, which is `hubPoolClient.getL2TokenForL1TokenAtBlock(l1TokenCounterpart, chainToSendRefundTo, endBlockForMainnet`. Obviously, in the case of USDC, we would be refunding USDC formerly and USDC.e with this change in the event that somebody does fill a deposit with inputToken USDC.e. I'm not sure if this affects running balance calculations, but if not, then  this is probably OK.

bmzig (approved):
I had one comment about USDC.e. Intuitively I think it's alright to do, but I just wanted to verify.

src/clients/BundleDataClient/utils/FillUtils.ts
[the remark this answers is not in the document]
nicholaspai: I think today no one can deposit with input token USDC.e because there are no deposit routes enabled for it. Moreover, USDC.e is not the official token returned by `hubPoolClient.getL2TokenForL1TokenAtBlock(l1TokenCounterpart, chainToSendRefundTo, endBlockForMainnet` on those chains where native USDC exists. This function returns the token mapped to by PoolRebalanceRoutes.

So I think in my PR, [this line will return false](https://github.com/across-protocol/sdk/blob/7d6c9fde7386f6a2276cb0382612a801f66e12a7/src/clients/BundleDataClient/utils/FillUtils.ts#L155) meaning that USDC.e is not mapped to a PoolRebalanceRoute, and therefore any deposit setting `inputToken = USDC.e` on a chain where native USDC exists will force the relayer to take repayment on the origin chain (in USDC.e)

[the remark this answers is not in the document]
bmzig: Ah of course. Thanks. You're right.

PR #974, opened 2025-04-08
improve(Address): Support new Address from nontraditional encoding

pxrl:
SVM events emit addresses with base58-encoding. At least two addresses emitted within an FundsDeposited or FilledRelay event will need to be converted to a base16 EVM address. Make it easy by permitting the caller of EvmAddress.from() to specify the encoding, defaulting to base16. Also do the same for the SVM side (i.e. permit base16-encoded 32-byte addresses).


src/utils/AddressUtils.ts
```diff
@@ -223,8 +223,20 @@ export class EvmAddress extends Address {
   }
 
   // Constructs a new EvmAddress type.
-  static from(hexString: string): EvmAddress {
-    return new this(utils.arrayify(hexString));
+  static from(address: string, encoding: "base16" | "base58" = "base16"): EvmAddress {
```
pxrl, line 226: nb. needs test; will follow up with that.

pxrl: https://github.com/across-protocol/sdk/pull/974/commits/b8aedadc4cd32767d11c2cb3a6870433cd4b3e45

pxrl, line 226: nb. We have the option of inferring the encoding with a regex. That'd definitely be a UX improvement...not sure if we want to go there though.

melisaguevara (approved):

PR #976, opened 2025-04-09
feat: create svm spoke _update

james-a-morris:
WIP (not working atm)

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -38,7 +38,7 @@ export async function getTimestampForSlot(provider: Provider, slotNumber: number
   const block = await provider.getBlock(BigInt(slotNumber)).send();
   let timestamp: number;
   if (!block?.blockTime) {
-    console.error(`Unable to resolve svm block ${slotNumber}`);
+    console.error(`Unable to resolve block for slot ${slotNumber}`);
     timestamp = 0; // @todo: How to handle this?
   } else {
     timestamp = Number(block.blockTime); // Unix timestamps fit within number.


src/clients/BaseAbstractClient.ts
```diff
@@ -58,15 +59,21 @@ export abstract class BaseAbstractClient {
...
     if (isDefined(toBlock)) {
       if (fromBlock > toBlock) {
         throw new Error(`Invalid event search config fromBlock (${fromBlock}) > toBlock (${toBlock})`);
       }
     } else {
-      toBlock = await provider.getBlockNumber();
+      if (provider instanceof providers.Provider) {
+        toBlock = await provider.getBlockNumber();
+      } else {
+        toBlock = Number(await provider.getBlockHeight({ commitment: "confirmed" }).send());
```
pxrl, line 75: @james-a-morris Did you check on the time delta for `confirmed` vs `processed`? I think the equivalent behaviour on EVM is `processed` since we typically don't want to impose any finality overheads on the data ingested here.

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -60,93 +69,86 @@ export class SvmSpokePoolClient extends SpokePoolClient {
...
+      ...eventsToQuery.map(async (eventName, idx) => {
+        const config = eventSearchConfigs[idx];
+        const events = await this.svmEventsClient.queryEvents(
+          eventName as SVMEventNames,
+          BigInt(config.fromBlock),
+          BigInt(config.toBlock)
+        );
+        return Promise.all(
+          events.map(async (event): Promise<SortableEvent> => {
+            const block = await this.rpc.getBlock(event.slot).send();
```
melisaguevara, line 117: I think we have to add `{maxSupportedTransactionVersion: 0}` as config for this call. This was failing locally for me with: 
```
SolanaError: Transaction version (0) is not supported by the requesting client.
Please try the request again with the following configuration parameter: "maxSupportedTransactionVersion": 0
```

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -60,93 +69,86 @@ export class SvmSpokePoolClient extends SpokePoolClient {
...
+            };
+          })
+        );
+      }),
+    ]);
+    this.log("debug", `Time to query new events from RPC for ${this.chainId}: ${Date.now() - timerStart} ms`);
+    if (!BigNumber.isBigNumber(currentTime) || currentTime.lt(this.currentTime)) {
+      const errMsg = BigNumber.isBigNumber(currentTime)
+        ? `currentTime: ${currentTime} < ${toBN(this.currentTime)}`
+        : `currentTime is not a BigNumber: ${JSON.stringify(currentTime)}`;
```
melisaguevara, line 139: I'm also getting an error here: `TypeError: Do not know how to serialize a BigInt`.

melisaguevara (commented):
Left two comments sharing some errors I had when testing locally.

src/arch/svm/utils.ts
```diff
@@ -64,3 +66,65 @@ export function getEventName(rawName: string): EventName {
...
+  // Handle null/undefined
+  if (data == null) {
+    return data;
+  }
+  // Handle Uint8Array and byte arrays
+  if (data instanceof Uint8Array || isUint8Array(data)) {
+    const bytes = data instanceof Uint8Array ? data : new Uint8Array(data as number[]);
+    const hex = "0x" + Buffer.from(bytes).toString("hex");
+    if (currentKey && uint8ArrayKeysAsBigInt.includes(currentKey)) {
+      return BigInt(hex);
```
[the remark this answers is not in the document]
melisaguevara: Also, should we parse bigint to bigNumber?

melisaguevara (approved):
Nice work! Looks good!

src/clients/SpokePoolClient/EVMSpokePoolClient.ts
```diff
@@ -123,17 +123,22 @@ export class EVMSpokePoolClient extends SpokePoolClient {
...
+      throw new Error(`EVMSpokePoolClient::update: ${errMsg}`);
     }
 
     // Sort all events to ensure they are stored in a consistent order.
     events.forEach((events) => sortEventsAscendingInPlace(events));
 
+    // Map events to SortableEvent
+    const eventsWithBlockNumber = events.map((eventList) =>
+      eventList.map((event) => spreadEventWithBlockNumber(event))
+    );
```
pxrl, line 135: wdyt about combining this with the `events.forEach()` above on line 130? Seems like we could avoid some re-iteration.

src/clients/SpokePoolClient/SpokePoolClient.ts
```diff
@@ -492,15 +488,15 @@ export abstract class SpokePoolClient extends BaseAbstractClient {
       // For each deposit, resolve its quoteTimestamp to a block number on the HubPool.
       // Don't bother filtering for uniqueness; the HubPoolClient handles this efficienctly.
       const quoteBlockNumbers = await this.getBlockNumbers(
-        depositEvents.map(({ args }) => Number(args["quoteTimestamp"]))
+        depositEvents.map((e) => (e as DepositWithBlock).quoteTimestamp)
       );
       for (const event of depositEvents) {
```
pxrl, line 493: Could we apply the type assertion on the loop to avoid having to do it multiple places within?

```suggestion
      for (const event of depositEvents as DepositWithBlock[]) {
```

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -60,93 +70,87 @@ export class SvmSpokePoolClient extends SpokePoolClient {
...
+
+    const [_currentTime, ...eventsQueried] = await Promise.all([
+      this.rpc.getBlockTime(BigInt(searchConfig.toBlock)).send(),
+      ...eventsToQuery.map(async (eventName, idx) => {
+        const config = eventSearchConfigs[idx];
+        const events = await this.svmEventsClient.queryEvents(
+          eventName as SVMEventNames,
+          BigInt(config.fromBlock),
+          BigInt(config.toBlock)
+        );
```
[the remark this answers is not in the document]
melisaguevara: I think that limit refers to the amount of signatures returned by [getSignaturesForAddress](https://solana.com/docs/rpc/http/getsignaturesforaddress) instead of the amount of slots being queried, so the 1k limit is imposed by Solana itself.

src/clients/BaseAbstractClient.ts
pxrl: This should be OK :+1:

src/arch/svm/eventsClient.ts
```diff
@@ -80,18 +79,18 @@ export class SvmCpiEventsClient {
...
-    toBlock?: bigint,
+    eventName: EventName,
+    fromSlot?: bigint,
+    toSlot?: bigint,
     options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
   ): Promise<EventWithData[]> {
-    const events = await this.queryAllEvents(fromBlock, toBlock, options);
+    const events = await this.queryAllEvents(fromSlot, toSlot, options);
     return events.filter((event) => event.name === eventName) as EventWithData[];
   }
```
melisaguevara, line 95: We also have too modify these in the new `queryDerivedAddressEvents` function

melisaguevara (commented):
Just one comment, otherwise I think this looks good

melisaguevara (approved):

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -68,93 +81,80 @@ export class SvmSpokePoolClient extends SpokePoolClient {
...
+      this.rpc.getBlockTime(BigInt(searchConfig.to)).send(),
+      ...eventsToQuery.map(async (eventName, idx) => {
+        const config = eventSearchConfigs[idx];
+        const events = await this.svmEventsClient.queryEvents(
+          eventName as SVMEventNames,
+          BigInt(config.from),
+          BigInt(config.to),
+          {
+            limit: config.maxLookBack,
+          }
```
melisaguevara, line 130: This limit doesn't work as it does in evm, if we want to use it this way then it can't be higher than 1000. Should we have an assertion for this?

melisaguevara: Or we could just avoid setting the limit here and use the 1k default.

amateima (approved):

pxrl (approved):

PR #978, opened 2025-04-10
feat(svmSpokeUtils): get fill deadline buffer implementation

melisaguevara:
This PR implements the utility function for getting svm fill deadline buffer. That's done by fetching the Spoke Pool's State PDA.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -1,4 +1,4 @@
-import { Rpc, SolanaRpcApi, Address, getProgramDerivedAddress, address, getU64Encoder } from "@solana/kit";
+import { Rpc, SolanaRpcApi, Address } from "@solana/kit";
 
 import { Deposit, FillStatus, FillWithBlock, RelayData } from "../../interfaces";
 import { BigNumber, isUnsafeDepositId } from "../../utils";
@@ -50,11 +50,10 @@ export async function getTimestampForBlock(provider: Provider, blockNumber: numb
 /**
  * Returns the current fill deadline buffer.
  * @param provider SVM Provider instance
- * @param programId Program ID from which State account is derived.
+ * @param statePda Spoke Pool's State PDA
  * @returns fill deadline buffer
  */
-export async function getFillDeadline(provider: Provider, programId: string): Promise<number> {
-  const statePda = await getStatePda(programId);
+export async function getFillDeadline(provider: Provider, statePda: Address): Promise<number> {
   const state = await fetchState(provider, statePda);
   return state.data.fillDeadlineBuffer;
 }
@@ -174,20 +173,4 @@ export function findFillEvent(
   _highBlockNumber?: number
 ): Promise<FillWithBlock | undefined> {
   throw new Error("fillStatusArray: not implemented");
-}
-
-/**
- * Returns the PDA for the State account.
- * @param programId The SpokePool program ID.
- * @param extraSeed An optional extra seed. Defaults to 0.
- * @returns The PDA for the State account.
- */
-export async function getStatePda(programId: string, extraSeed = 0): Promise<Address> {
-  const seedEncoder = getU64Encoder();
-  const encodedExtraSeed = seedEncoder.encode(extraSeed);
-  const [statePda] = await getProgramDerivedAddress({
-    programAddress: address(programId),
-    seeds: ["state", encodedExtraSeed],
-  });
-  return statePda;
 }
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -1,20 +1,8 @@
-import {
-  Rpc,
-  SolanaRpcApi,
-  Address,
-  getProgramDerivedAddress,
-  address,
-  getStructDecoder,
-  getBooleanCodec,
-  getU32Codec,
-  getU64Codec,
-  getAddressCodec,
-  getAddressDecoder,
-} from "@solana/kit";
+import { Rpc, SolanaRpcApi, Address, getProgramDerivedAddress, address, getU64Encoder } from "@solana/kit";
 
 import { Deposit, FillStatus, FillWithBlock, RelayData } from "../../interfaces";
 import { BigNumber, isUnsafeDepositId } from "../../utils";
-import { snakeToCamel } from "../../svm/utils/events";
+import { fetchState } from "@across-protocol/contracts/dist/src/svm/clients/SvmSpoke";
 
 type Provider = Rpc<SolanaRpcApi>;
 
@@ -60,22 +48,15 @@ export async function getTimestampForBlock(provider: Provider, blockNumber: numb
 }
 
 /**
- * Return maximum of fill deadline buffer at start and end of block range.
- * @param spokePool SpokePool contract instance
- * @param startBlock start block
- * @param endBlock end block
- * @returns maximum of fill deadline buffer at start and end block
+ * Returns the current fill deadline buffer.
+ * @param provider SVM Provider instance
+ * @param programId Program ID from which State account is derived.
+ * @returns fill deadline buffer
  */
-export async function getMaxFillDeadlineInRange(
-  _spokePool: { provider: Provider },
-  _startBlock: number,
-  _endBlock: number
-): Promise<any> {
-  const programId = "JAZWcGrpSWNPTBj8QtJ9UyQqhJCDhG9GJkDeMf5NQBiq"; // TODO: read this from the spokePool
+export async function getFillDeadline(provider: Provider, programId: string): Promise<number> {
   const statePda = await getStatePda(programId);
-  const state = await queryStateFromStatePda(statePda, _spokePool.provider);
-
-  return state.fillDeadlineBuffer;
+  const state = await fetchState(provider, statePda);
+  return state.data.fillDeadlineBuffer;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -176,3 +194,38 @@ export function findFillEvent(
...
+    ["remote_domain", getU32Codec()],
+    ["cross_domain_admin", getAddressCodec()],
+    ["root_bundle_id", getU32Codec()],
+    ["deposit_quote_time_buffer", getU32Codec()],
+    ["fill_deadline_buffer", getU32Codec()],
+  ]);
+  const stateData = pdaAccountInfo.value.data[0];
+  const stateDataBytes = Buffer.from(stateData, "base64");
+  const discriminatorBytesSize = 8;
+  const decodedState = stateDecoder.decode(Uint8Array.from(stateDataBytes), discriminatorBytesSize);
```
melisaguevara, line 220: This is the only way I was able to decode the state pda data so far. I guess there should be an easier way leveraging the IDL, I'm still looking into this.

test/SpokeUtils.ts
```diff
@@ -26,6 +28,21 @@ describe("SpokeUtils", function () {
...
+      0,
+      undefined,
+      "https://api.mainnet-beta.solana.com",
+      34268394551451
+    );
+    const provider = rpcFactory.createRpcClient();
+    const fillDeadline = await getMaxFillDeadlineInRange({ provider }, 0, 100);
+    console.log(fillDeadline);
+  });
+
```
melisaguevara, line 45: This is temporary, is intended only to test the function.

src/arch/svm/SpokeUtils.ts
```diff
@@ -176,3 +194,38 @@ export function findFillEvent(
...
+    ["seed", getU64Codec()],
+    ["number_of_deposits", getU32Codec()],
+    ["chain_id", getU64Codec()],
+    ["current_time", getU32Codec()],
+    ["remote_domain", getU32Codec()],
+    ["cross_domain_admin", getAddressCodec()],
+    ["root_bundle_id", getU32Codec()],
+    ["deposit_quote_time_buffer", getU32Codec()],
+    ["fill_deadline_buffer", getU32Codec()],
+  ]);
```
pxrl, line 216: @melisaguevara Can we use `getStateDecoder()` from the contracts repo here? It's available in `src/svm/clients/SvmSpoke/accounts/state.ts`. I believe it should do most of the heavy lifting for state decoding.

melisaguevara: Absolutely, ty! I missed those methods.
I changed it to use `fetchState` instead here [9cf2918](https://github.com/across-protocol/sdk/pull/978/commits/9cf29183cb164d3343934925f8d8a63ccca0306f).

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -19,6 +20,7 @@ export class SvmSpokePoolClient extends SpokePoolClient {
     chainId: number,
     deploymentSlot: bigint, // Using slot instead of block number for SVM
     eventSearchConfig: MakeOptional<EventSearchConfig, "toBlock">,
+    protected programId: string,
```
melisaguevara, line 23: I've added `programId` here because I think it's useful to have direct access to it. That said, I'm also not sure this is the best approach since `svmSpokeEventsClient` already resolves `programId` internally, and that value might differ from the one passed in here. To address that I think we can either (1) use the resolved value from `svmEventsClient` or (2) let the callers provide `programId` and pass that to `svmSpokeEventsClient`.

For option 1, we only need to make `svmSpokeAddress` a public property ([here](https://github.com/across-protocol/sdk/blob/ad52cf26d9d3987c80eac24d9d740d309e01ea11/src/svm/eventsClient.ts#L31)).

For option 2 we'll have to remove the [resolution logic](https://github.com/across-protocol/sdk/blob/ad52cf26d9d3987c80eac24d9d740d309e01ea11/src/svm/eventsClient.ts#L53-L55) from `svmSpokeEventsClient` entirely and turn it into a constructor param instead. 

I'm curious to hear other's thoughts about this. Personally, I'm leaning toward option 2. I think it's closer to what we have for EVM and it also improves flexibility. As an example, I had to override the program signature to test some Solana features while working in the Indexer and there was other way to do so than changing the `svmSpokeEventsClient`.

src/arch/svm/SpokeUtils.ts
```diff
@@ -176,3 +175,19 @@ export function findFillEvent(
...
+ * @param extraSeed An optional extra seed. Defaults to 0.
+ * @returns The PDA for the State account.
+ */
+export async function getStatePda(programId: string, extraSeed = 0): Promise<Address> {
+  const seedEncoder = getU64Encoder();
+  const encodedExtraSeed = seedEncoder.encode(extraSeed);
+  const [statePda] = await getProgramDerivedAddress({
+    programAddress: address(programId),
+    seeds: ["state", encodedExtraSeed],
+  });
```
pxrl, line 191: @melisaguevara wdyt about factoring this out so that we can derive the PDA in the SpokePoolClient, and then supply it as an input to `getStatePda()`. That would permit us to query it once and then retain it in the SpokePoolClient for subsequent reuse.

melisaguevara: Good point. I think it makes sense since we might fetch its data to get other properties. I've changed that here [fe2b225](https://github.com/across-protocol/sdk/pull/978/commits/fe2b2259fca16a0c8e244e4d6e1bab3439bc28e7), wdyt?

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
pxrl: @melisaguevara Is it necessary for the eventsClient to be instantiated outside of the SVMSpokePoolClient? Could we instead instantiate it entirely within the `.create()` static method?

src/arch/svm/SpokeUtils.ts
```diff
@@ -46,18 +48,14 @@ export async function getTimestampForBlock(provider: Provider, blockNumber: numb
...
+ * @param statePda Spoke Pool's State PDA
+ * @returns fill deadline buffer
  */
-export function getMaxFillDeadlineInRange(
-  _spokePool: unknown,
-  _startBlock: number,
-  _endBlock: number
-): Promise<number> {
-  throw new Error("getMaxFillDeadlineInRange: not implemented");
+export async function getFillDeadline(provider: Provider, statePda: Address): Promise<number> {
```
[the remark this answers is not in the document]
melisaguevara: Well, since we might need to fetch other variables from the `statePda` I think it makes sense to have it as a client property and as a param in here. In the end, if you are using this function through the SvmSpokePoolClient you don't have to resolve it yourself. Wdyt?

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
melisaguevara: > Is it necessary for the eventsClient to be instantiated outside of the SVMSpokePoolClient? Could we instead instantiate it entirely within the .create() static method?

@pxrl I don't think so, I think we'll always use it through the clients. This could be a follow up if we all agree to do so.

melisaguevara: > Fwiw I think it makes sense to also expose the svmSpokeAddress - it's not required to be private, and could theoretically be encapsulated in a getter fn

@james-a-morris Yes I saw you did that in your PR. I think that's also an option, though as an end user you would have to do `svmSpokeClient.svmEventsClient.getSvmSpokeAddress()`. I think it would be nice to have that directly as property of SvmSpokeClient. For now I'll follow your approach, we can revisit this later.

melisaguevara: This is how it looks for now: [d7e3dd6](https://github.com/across-protocol/sdk/pull/978/commits/d7e3dd696d67a778dd5866525be71f0fdf406fdc)

src/arch/svm/utils.ts
```diff
@@ -64,3 +65,19 @@ export function getEventName(rawName: string): EventName {
...
   throw new Error(`Unknown event name: ${rawName}`);
 }
+
+/**
+ * Returns the PDA for the State account.
+ * @param programId The SpokePool program ID.
+ * @param extraSeed An optional extra seed. Defaults to 0.
+ * @returns The PDA for the State account.
+ */
+export async function getStatePda(programId: string, extraSeed = 0): Promise<Address> {
```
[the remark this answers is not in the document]
melisaguevara: [d7e3dd6](https://github.com/across-protocol/sdk/pull/978/commits/d7e3dd696d67a778dd5866525be71f0fdf406fdc)

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
pxrl: > @pxrl I don't think so, I think we'll always use it through the clients. This could be a follow up if we all agree to do so.

Yeah; I think it's probably cleaner to incorporate it directly into the SpokePoolClient, rather than requiring it to be supplied as an external dependency. This should make it easier to instantiate.

melisaguevara: I think we can modify this after this PR goes in: https://github.com/across-protocol/sdk/pull/982

src/arch/svm/utils.ts
```diff
@@ -64,3 +65,19 @@ export function getEventName(rawName: string): EventName {
...
   throw new Error(`Unknown event name: ${rawName}`);
 }
+
+/**
+ * Returns the PDA for the State account.
+ * @param programId The SpokePool program ID.
+ * @param extraSeed An optional extra seed. Defaults to 0.
+ * @returns The PDA for the State account.
+ */
+export async function getStatePda(programId: Address, extraSeed = 0): Promise<Address> {
```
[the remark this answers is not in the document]
melisaguevara: I think for state pda it'll always be zero. I removed it here: [d37346b](https://github.com/across-protocol/sdk/pull/978/commits/d37346b126d901ce19ff89ac47bfcb49d98e8766)

pxrl (approved):
LGTM!

PR #979, opened 2025-04-10
improve(SpokeUtils): Use viem for RelayData hash computation

pxrl:
The Viem ABI encoder permits supplying the ABI as an object, which is faster than ethers, which apparently requires that the ABI be supplied to AbiCoder.encode() as a string. This implies re-parsing the same ABI on each call.

Brief testing showed that the new implementation is 50% faster than the existing one. Producing 500 RelayData hashes locally with takes ~0.5 seconds with viem and ~1 second with ethers. This is relevant in the relayer because performs bulk fillStatus queries of up to 250 RelayData hashes per chain on update, and it has a loop time in the order of low-digit seconds.

### DIFF
--- a/src/utils/SpokeUtils.ts
+++ b/src/utils/SpokeUtils.ts
@@ -1,8 +1,7 @@
-import { utils as ethersUtils } from "ethers";
+import { encodeAbiParameters, keccak256 } from "viem";
 import { MAX_SAFE_DEPOSIT_ID, ZERO_ADDRESS, ZERO_BYTES } from "../constants";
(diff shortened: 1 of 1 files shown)


pxrl:
One comment on the safety of this change - `test/SpokePoolClient.ValidateFill.ts` has tests that make fills and then query the FillStatus in the SpokePool, using `getRelayDataHash()` to compute the hash. If the hashes are not computed identically to the SpokePool's own RelayData hashes then these tests will fail.

melisaguevara (approved):

PR #980, opened 2025-04-10
feat: Solana relayFeeCalculator and gasPriceOracle

bmzig:
This should implement the relayFeeCalculator (and by extension, the gasPriceOracle) for svm. It will contain some other utilities which are used by those structures.


bmzig:
I can confirm that the gas price oracle is giving reasonable estimates for priority fees/base fees (assuming that the fill transaction only requires a single signature). 

The relay fee calculator simulation is still failing somewhere in the fill instruction, and I haven't figured out if this is an issue with how the instruction is being created or if I am just testing with relay data values that would cause the SvmSpoke to revert, but even if it is incorrect, this should at most require a small change in `fillRelayInstruction`.

src/arch/svm/SpokeUtils.ts
```diff
@@ -150,6 +173,22 @@ export function fillStatusArray(
   throw new Error("fillStatusArray: not implemented");
 }
 
+/**
+ * Returns the PDA for the State account.
+ * @param programId The SpokePool program ID.
+ * @param extraSeed An optional extra seed. Defaults to 0.
+ * @returns The PDA for the State account.
+ */
+export async function getStatePda(programId: string, extraSeed = 0): Promise<Address> {
```
bmzig, line 182: I generally just copied the format for these functions from https://github.com/across-protocol/sdk/pull/978

bmzig:
I'm now getting both reasonable gas price estimations and reasonable compute unit estimations.

src/arch/svm/SpokeUtils.ts
```diff
@@ -176,3 +215,166 @@ export function findFillEvent(
...
+  if (!isDefined(mintDecimals)) {
+    throw new Error(`No mint decimals found for token ${mint.toBase58()}`);
+  }
+
+  return getApproveCheckedInstruction({
+    source: relayerTokenAccount,
+    mint: mint.toV2Address(),
+    delegate: statePda,
+    owner: relayer.toV2Address(),
+    amount: amount.toBigInt(),
```
pxrl, line 343: Neat - TIL!

src/gasPriceOracle/adapters/solana.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * @notice Returns result of getFeeForMessage and getRecentPrioritizationFees RPC calls.
+ * @returns GasPriceEstimate
+ */
+export async function messageFee(provider: Provider, opts: GasPriceEstimateOptions): Promise<GasPriceEstimate> {
+  const { unsignedTx: _unsignedTx } = opts;
+
+  // Cast the opaque unsignedTx type to a solana-kit CompilableTransactionMessage.
+  const unsignedTx = _unsignedTx as CompilableTransactionMessage;
```
pxrl, line 15: It'd be nice to verify this via superstruct - then we'd inherit type inference as well.

src/gasPriceOracle/adapters/solana.ts
```diff
@@ -0,0 +1,34 @@
...
+    "base64"
+  ) as TransactionMessageBytesBase64;
+  const baseFeeResponse = await provider.getFeeForMessage(encodedTransactionMessage).send();
+
+  // Get the priority fee by calling `getRecentPrioritzationFees` on all the addresses in the transaction's instruction array.
+  const instructionAddresses = dedupArray(unsignedTx.instructions.map((instruction) => instruction.programAddress));
+  const recentPriorityFees = await provider.getRecentPrioritizationFees(instructionAddresses).send();
+
+  const nonzeroPrioritizationFees = recentPriorityFees.map((value) => value.prioritizationFee).filter((fee) => fee > 0);
+  const totalPrioritizationFees = nonzeroPrioritizationFees.reduce((acc, fee) => acc + fee, BigInt(0));
```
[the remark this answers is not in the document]
pxrl: The [JSON-RPC docs](https://solana.com/docs/rpc/http/getrecentprioritizationfees) for Solana say they retain fee data for 150 slots - do you only get the last 100 on query?

In any case, would we consider just taking the last n slots - i.e. 25? That'd give 10 seconds of coverage, which is pretty long in relative terms. Is there additionally some minimum priority fee that we might want to set, just so that we don't submit at 0?

src/gasPriceOracle/oracle.ts
```diff
@@ -53,6 +55,21 @@ export async function getGasPriceEstimate(
     `Require 1.0 < priority fee multiplier (${priorityFeeMultiplier}) <= 5.0 for a total gas multiplier within [+1.0, +5.0]`
   );
 
+  const isEthersProvider = provider instanceof providers.Provider;
+  // Exit here if we need to estimate on Solana.
+  if (!isEthersProvider) {
```
pxrl, line 60: I wonder if we should make `chainId` a mandatory field in `GasPriceEstimateOptions`, because we can't resolve the Solana "chain Id" via an RPC call, so this won't scale beyond one non-evm/svm chain.

src/gasPriceOracle/oracle.ts
```diff
@@ -53,6 +55,21 @@ export async function getGasPriceEstimate(
...
+    const optsWithDefaults: GasPriceEstimateOptions = {
+      ...GAS_PRICE_ESTIMATE_DEFAULTS,
+      baseFeeMultiplier,
+      priorityFeeMultiplier,
+      ...opts,
+      chainId: CHAIN_IDs.SOLANA,
+    };
+    return solana.messageFee(provider as SolanaProvider, optsWithDefaults);
```
pxrl, line 68: Is type inference not working on `provider` ? It might work if line 60 is changed `if (!provider instanceof providers.Provider)`. We might alternatively want a superstruct discriminator to handle it for us.

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -213,7 +213,9 @@ export class QueryBase implements QueryInterface {
         ? Promise.resolve({ maxFeePerGas: _gasPrice })
         : getGasPriceEstimate(provider, { chainId, baseFeeMultiplier, priorityFeeMultiplier, transport, unsignedTx }),
     ] as const;
-    const [nativeGasCost, { maxFeePerGas: gasPrice }] = await Promise.all(queries);
+    const [nativeGasCost, _gasPriceEstimate] = await Promise.all(queries);
+    // It should be safe to cast to an EvmGasPriceEstimate here since QueryBase is only used for EVM chains.
```
pxrl, line 217: It's probably a good use case for an `assert(EvmGasPriceEstimate.is(gasPriceEstimate))`. We'd need superstruct defs for this.

src/utils/SpokeUtils.ts
```diff
@@ -89,3 +93,40 @@ export function isZeroAddress(address: string): boolean {
...
+    bufferFromInt(relayData.outputAmount, 8),
+    bufferFromInt(toBN(relayData.originChainId), 8),
+    bufferFromInt(relayData.depositId, 32, false),
+    bufferFromInt(toBN(relayData.fillDeadline), 4),
+    bufferFromInt(toBN(relayData.exclusivityDeadline), 4),
+    bufferFromHexString(getMessageHash(relayData.message)),
+    bufferFromInt(toBN(destinationChainId), 8),
+  ]);
+  return keccak256(contentToHash);
+}
```
pxrl, line 132: Damn; this one caught me by surprise. Can we stick it in `src/arch/svm/SpokeUtils.ts` ? We might also want to relocate the EVM part as well, so `getRelayDataHash()` should be pretty thin as a result.

src/arch/svm/SpokeUtils.ts
```diff
@@ -176,3 +215,166 @@ export function findFillEvent(
...
+}
+
+export async function getAssociatedTokenAddress(
+  owner: SvmAddress,
+  mint: SvmAddress,
+  tokenProgramId: Address<string> = TOKEN_PROGRAM_ADDRESS
+): Promise<Address<string>> {
+  const [associatedToken] = await getProgramDerivedAddress({
+    programAddress: ASSOCIATED_TOKEN_PROGRAM_ADDRESS,
+    seeds: [owner.toBuffer(), SvmAddress.from(tokenProgramId).toBuffer(), mint.toBuffer()],
```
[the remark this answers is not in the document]
bmzig: I'm actually not too sure how the string encoding is supposed to work here, but each character is treated as a byte, so passing in a string here will have `getProgramDerivedAddress` interpret the input as 66 byte strings, not 32 bytes.

src/utils/SpokeUtils.ts
bmzig: I agree. This was moved https://github.com/across-protocol/sdk/pull/980/commits/6186021dc7a654baf1f85ca1268914ed9ced8d73

src/relayFeeCalculator/chain-queries/baseQuery.ts
bmzig: I agree. Probably another place for superstruct.

src/gasPriceOracle/oracle.ts
bmzig: I also agree. It's annoying that the default value is {} since I expect that we pass in an empty set of options quite often. I at least made a change to use the opts to get the chain Id if applicable before assuming `CHAIN_IDs.SOLANA` https://github.com/across-protocol/sdk/pull/980/commits/61b885440f23df0eb26b7df511315a4eabe9b1e7

src/arch/svm/SpokeUtils.ts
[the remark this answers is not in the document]
pxrl: Ah yeah, `.toAddress()` would be completely wrong. The kit readme has an example on how to supply the seeds correctly: https://github.com/anza-xyz/kit/blob/0bfe90b1c1e1c1a2dd90c46408c7cab995692f7b/packages/addresses/README.md#getprogramderivedaddress

src/gasPriceOracle/oracle.ts
pxrl: hm, odd - I tried it locally and this builds for me. What error do you get?

```
diff --git a/src/gasPriceOracle/oracle.ts b/src/gasPriceOracle/oracle.ts
index 30985f95..bcec2e54 100644
--- a/src/gasPriceOracle/oracle.ts
+++ b/src/gasPriceOracle/oracle.ts
@@ -55,9 +55,8 @@ export async function getGasPriceEstimate(
     `Require 1.0 < priority fee multiplier (${priorityFeeMultiplier}) <= 5.0 for a total gas multiplier within [+1.0, +5.0]`
   );
 
-  const isEthersProvider = provider instanceof providers.Provider;
   // Exit here if we need to estimate on Solana.
-  if (!isEthersProvider) {
+  if (!(provider instanceof providers.Provider)) {
     const optsWithDefaults: GasPriceEstimateOptions = {
       ...GAS_PRICE_ESTIMATE_DEFAULTS,
       baseFeeMultiplier,
@@ -65,7 +64,7 @@ export async function getGasPriceEstimate(
       ...opts,
       chainId: opts.chainId ?? CHAIN_IDs.SOLANA,
     };
-    return solana.messageFee(provider as SolanaProvider, optsWithDefaults);
+    return solana.messageFee(provider, optsWithDefaults);
   }
 
   // Cast the provider to an ethers provider, which should be given to the oracle when querying an EVM network.
```

pxrl (approved):

src/gasPriceOracle/oracle.ts
bmzig: I see. I misinterpreted the comment, then. I changed the `const isEthersProvider` to a bunch of different values and not the actual boolean expression in the `if()` block. Changed here https://github.com/across-protocol/sdk/pull/980/commits/af90716d0106ee4a8d06b7bb04de7a62678fd2ae

PR #982, opened 2025-04-18
improve: allow SVM events client to take in non-spoke IDLs

bmzig:
The events client basically serves as an SVM `paginatedEventQuery`, which is sometimes used in the bots (e.g. the rebalancer/finalizer). I would like to generalize this class a bit so that it can be used for contracts beyond just the SvmSpoke.

### DIFF
--- a/src/arch/svm/eventsClient.ts
+++ b/src/arch/svm/eventsClient.ts
@@ -82,14 +82,14 @@ export class SvmEventsClient {
    * @param options - Options for fetching signatures.
    * @returns A promise that resolves to an array of events matching the eventName.
    */
-  public async queryEvents<T extends EventData>(
+  public async queryEvents(
     eventName: EventName,
     fromBlock?: bigint,
     toBlock?: bigint,
     options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
-  ): Promise<EventWithData<T>[]> {
+  ): Promise<EventWithData[]> {
     const events = await this.queryAllEvents(fromBlock, toBlock, options);
-    return events.filter((event) => event.name === eventName) as EventWithData<T>[];
+    return events.filter((event) => event.name === eventName) as EventWithData[];
   }
 
   /**
@@ -104,7 +104,7 @@ export class SvmEventsClient {
     fromBlock?: bigint,
     toBlock?: bigint,
     options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
-  ): Promise<EventWithData<EventData>[]> {
+  ): Promise<EventWithData[]> {
     const allSignatures: GetSignaturesForAddressTransaction[] = [];
     let hasMoreSignatures = true;
     let currentOptions = options;
@@ -189,7 +189,7 @@ export class SvmEventsClient {
    */
   private processEventFromTx(
     txResult: GetTransactionReturnType
-  ): { program: Address; data: EventData; name: EventName }[] {
+  ): { program: Address; data: unknown; name: EventName }[] {
     if (!txResult) return [];
     const events: { program: Address; data: EventData; name: EventName }[] = [];
 
--- a/src/arch/svm/eventsClient.ts
+++ b/src/arch/svm/eventsClient.ts
@@ -26,10 +26,11 @@ type GetSignaturesForAddressConfig = Parameters<GetSignaturesForAddressApi["getS
 type GetSignaturesForAddressTransaction = ReturnType<GetSignaturesForAddressApi["getSignaturesForAddress"]>[number];
 type GetSignaturesForAddressApiResponse = readonly GetSignaturesForAddressTransaction[];
 
-export class SvmSpokeEventsClient {
+export class SvmEventsClient {
   private rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>;
-  private svmSpokeAddress: Address;
-  private svmSpokeEventAuthority: Address;
+  private svmAddress: Address;
+  private svmEventAuthority: Address;
+  private customIdl;
 
   /**
    * Private constructor. Use the async create() method to instantiate.
@@ -37,19 +38,19 @@ export class SvmSpokeEventsClient {
   private constructor(
     rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>,
     svmSpokeAddress: Address,
-    eventAuthority: Address
+    eventAuthority: Address,
+    customIdl = SvmSpokeIdl
   ) {
     this.rpc = rpc;
-    this.svmSpokeAddress = svmSpokeAddress;
-    this.svmSpokeEventAuthority = eventAuthority;
+    this.svmAddress = svmSpokeAddress;
+    this.svmEventAuthority = eventAuthority;
+    this.customIdl = customIdl;
   }
 
   /**
    * Factory method to asynchronously create an instance of SvmSpokeEventsClient.
    */
-  public static async create(
-    rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>
-  ): Promise<SvmSpokeEventsClient> {
+  public static async create(rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>): Promise<SvmEventsClient> {
     const isTestnet = await isDevnet(rpc);
     const programId = getDeployedAddress("SvmSpoke", getSolanaChainId(isTestnet ? "devnet" : "mainnet").toString());
     if (!programId) throw new Error("Program not found");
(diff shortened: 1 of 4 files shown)


src/arch/svm/eventsClient.ts
```diff
@@ -26,39 +29,48 @@ type GetSignaturesForAddressConfig = Parameters<GetSignaturesForAddressApi["getS
...
+  public static async createFor(
+    rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>,
+    programId: string,
+    idl: Idl
+  ): Promise<SvmEventsClient> {
+    const svmAddress = web3.address(programId);
+    const [svmEventAuthority] = await web3.getProgramDerivedAddress({
+      programAddress: svmAddress,
       seeds: ["__event_authority"],
     });
```
melisaguevara, line 72: ooc, do you know if this would work for any address?

src/arch/svm/eventsClient.ts
```diff
@@ -26,39 +29,48 @@ type GetSignaturesForAddressConfig = Parameters<GetSignaturesForAddressApi["getS
...
-  ): Promise<SvmSpokeEventsClient> {
+  public static async create(rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>): Promise<SvmEventsClient> {
     const isTestnet = await isDevnet(rpc);
     const programId = getDeployedAddress("SvmSpoke", getSolanaChainId(isTestnet ? "devnet" : "mainnet").toString());
     if (!programId) throw new Error("Program not found");
-    const svmSpokeAddress = web3.address(programId);
-    const [svmSpokeEventAuthority] = await web3.getProgramDerivedAddress({
-      programAddress: svmSpokeAddress,
+    return this.createFor(rpc, programId, SvmSpokeIdl);
+  }
```
melisaguevara, line 61: I think we can keep both create functions for now but wdyt about deleting this one in the future? I think it is ok to have the users of the class provide the programId.

src/arch/svm/eventsClient.ts
```diff
@@ -26,39 +29,48 @@ type GetSignaturesForAddressConfig = Parameters<GetSignaturesForAddressApi["getS
...
 type GetSignaturesForAddressApiResponse = readonly GetSignaturesForAddressTransaction[];
 
-export class SvmSpokeEventsClient {
+export class SvmEventsClient {
   private rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>;
-  private svmSpokeAddress: Address;
-  private svmSpokeEventAuthority: Address;
+  private svmAddress: Address;
+  private svmEventAuthority: Address;
+  private customIdl;
```
melisaguevara, line 36: nit: could we name this just 'idl'?

melisaguevara, line 36: nit: could we name this just 'idl'?

melisaguevara (approved):
Just a few questions but it looks good! Thanks for working on this, I'm also needing it for fetching pda events.

src/arch/svm/types.ts
```diff
@@ -36,12 +36,12 @@ export enum SVMEventNames {
...
 
-export type EventWithData<T extends EventData> = {
+export type EventWithData = {
   confirmationStatus: string | null;
   blockTime: UnixTimestamp | null;
   signature: Signature;
   slot: bigint;
   name: EventName;
-  data: T;
+  data: unknown;
```
bmzig, line 45: I don't love getting rid of this type bound, but I don't think we can have this if we want to use the eventsClient to query events for other contracts.

src/arch/svm/eventsClient.ts
bmzig: Definitely https://github.com/across-protocol/sdk/pull/982/commits/b5712a57cd41115abcceb9338c5f35c730cdb8c4

src/arch/svm/eventsClient.ts
bmzig: Yeah I think you're right. I've kept the original function in so I wouldn't interfere with the svm spoke pool client in case there were some pretty large changes being made there, but we don't need to have this function anymore.

src/arch/svm/eventsClient.ts
bmzig: I asked @md0x about this. I think that not all events will have an event authority, but the ones that do follow this structure. So basically this events client can only query events which have an event authority field.

src/arch/svm/eventsClient.ts
melisaguevara: Cool 👍🏼  Yes I agree this change would've created a huge diff so we can def work on this after we finish merging the outstanding PRs.

src/arch/svm/eventsClient.ts
```diff
@@ -70,14 +82,14 @@ export class SvmSpokeEventsClient {
...
+  public async queryEvents(
+    eventName: string,
     fromBlock?: bigint,
     toBlock?: bigint,
     options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
-  ): Promise<EventWithData<T>[]> {
+  ): Promise<EventWithData[]> {
     const events = await this.queryAllEvents(fromBlock, toBlock, options);
-    return events.filter((event) => event.name === eventName) as EventWithData<T>[];
+    return events.filter((event) => event.name.toLowerCase() === eventName.toLowerCase()) as EventWithData[];
```
pxrl, line 92: Do we want to permit events to be specified in the wrong casing on the `queryEvents` call? On the EVM side we require that they're exact, but here we normalise via `.toLowerCase()`. I'd think we could skip the normalisation. wdyt?

src/arch/svm/eventsClient.ts
```diff
@@ -26,39 +29,48 @@ type GetSignaturesForAddressConfig = Parameters<GetSignaturesForAddressApi["getS
 type GetSignaturesForAddressTransaction = ReturnType<GetSignaturesForAddressApi["getSignaturesForAddress"]>[number];
 type GetSignaturesForAddressApiResponse = readonly GetSignaturesForAddressTransaction[];
 
-export class SvmSpokeEventsClient {
+export class SvmEventsClient {
   private rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>;
-  private svmSpokeAddress: Address;
-  private svmSpokeEventAuthority: Address;
+  private svmAddress: Address;
+  private svmEventAuthority: Address;
```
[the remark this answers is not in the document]
md0x: Agree, maybe we can call them `programAddress` and `programEventAuthority`

src/arch/svm/eventsClient.ts
```diff
@@ -9,10 +9,13 @@ import web3, {
   Signature,
 } from "@solana/kit";
 import { bs58 } from "../../utils";
-import { EventData, EventName, EventWithData } from "./types";
+import { EventWithData } from "./types";
 import { decodeEvent, isDevnet } from "./utils";
 import { getSlotForBlock } from "./SpokeUtils";
 
+// Blanket type definition so that we don't need to import the `Idl` type from anchor directly.
+type Idl = typeof SvmSpokeIdl;
```
pxrl, line 17: Is this `Idl` generic, or specific to the `SvmSpokeIdl` in any way?

fwiw we already rely on anchor for the `Idl` type in `utils.ts`, so I think it's OK to re-export `Idl` from there.

bmzig: Good point https://github.com/across-protocol/sdk/pull/982/commits/5f3d674305951620c0567d116fe63096b6dba173

src/arch/svm/eventsClient.ts
```diff
@@ -26,39 +29,48 @@ type GetSignaturesForAddressConfig = Parameters<GetSignaturesForAddressApi["getS
 type GetSignaturesForAddressTransaction = ReturnType<GetSignaturesForAddressApi["getSignaturesForAddress"]>[number];
 type GetSignaturesForAddressApiResponse = readonly GetSignaturesForAddressTransaction[];
 
-export class SvmSpokeEventsClient {
+export class SvmEventsClient {
```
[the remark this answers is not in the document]
bmzig: I agree. https://github.com/across-protocol/sdk/pull/982/commits/466f977e29ce3a6412595be11dbdce06945e41b7

src/arch/svm/eventsClient.ts
bmzig: Yeah you're probably right. https://github.com/across-protocol/sdk/pull/982/commits/466f977e29ce3a6412595be11dbdce06945e41b7

src/arch/svm/eventsClient.ts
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/sdk/pull/982/commits/466f977e29ce3a6412595be11dbdce06945e41b7

PR #983, opened 2025-04-21
feat(sdk): implement Solana-optimized deposit block finder

james-a-morris:
Implements findDepositBlock for Solana with a 2-day lookback window strategy instead of full chain traversal. This approach accounts for Solana PDA state limitations while maintaining practical deposit discovery capabilities. Key changes:
- Uses 2-day block range for deposit event queries
- Handles Solana slot/block height mapping
- Documents architectural decisions and limitations

### DIFF
--- a/src/clients/SpokePoolClient/SVMSpokePoolClient.ts
+++ b/src/clients/SpokePoolClient/SVMSpokePoolClient.ts
@@ -17,6 +17,7 @@ import {
   DepositSearchResult,
   EventSearchConfig,
   InvalidFill,
+  isZeroAddress,
   MakeOptional,
   sortEventsAscendingInPlace,
 } from "../../utils";
@@ -220,9 +221,20 @@ export class SvmSpokePoolClient extends SpokePoolClient {
         reason: `Deposit with ID ${depositId} not found`,
       };
     }
+    // Because we have additional context about this deposit, we can enrich it
+    // with additional information.
     return {
       found: true,
-      deposit,
+      deposit: {
+        ...deposit,
+        quoteBlockNumber: await this.getBlockNumber(Number(deposit.quoteTimestamp)),
+        originChainId: this.chainId,
+        fromLiteChain: this.isOriginLiteChain(deposit),
+        toLiteChain: this.isDestinationLiteChain(deposit),
+        outputToken: isZeroAddress(deposit.outputToken)
+          ? this.getDestinationTokenForDeposit(deposit)
+          : deposit.outputToken,
+      },
     };
   }
 


src/arch/svm/SpokeUtils.ts
```diff
@@ -113,17 +116,73 @@ export async function getSlotForBlock(
...
+  const block = await provider.getBlock(depositEvent.slot).send();
+  if (!block) {
+    return undefined;
+  }
+
+  return {
+    block: BigNumber.from(block.blockHeight),
+    slot: BigNumber.from(depositEvent.slot),
+    deposit: {
+      transactionHash: depositEvent.signature.toLowerCase(),
```
melisaguevara, line 179: since signatures are case sensitive, should we keep them as they are?

melisaguevara (approved):
I think this makes sense but since an undefined return value here would result in marking a fill as invalid in the BundleDataClient I would like to hear other's thoughts.
Other than that, just one question.

src/arch/svm/SpokeUtils.ts
```diff
@@ -113,17 +116,73 @@ export async function getSlotForBlock(
...
+ * typically not relevant to current operations.
+ *
+ * @param provider - Solana RPC provider
+ * @param depositId - The deposit ID to search for
+ * @returns The deposit, block number, and slot where the deposit occurred, or undefined if not found
+ */
+export async function findDeposit(
+  provider: Provider,
+  depositId: BigNumber
+): Promise<{ block: BigNumber; slot: BigNumber; deposit: DepositWithBlock } | undefined> {
```
pxrl, line 146: We're already returning `DepositWithBlock`; I assume that'll now contain a slot, but is there any reason to return `block` or `slot` separately? There's at least some duplication there.

src/arch/svm/SpokeUtils.ts
```diff
@@ -113,17 +116,73 @@ export async function getSlotForBlock(
...
+  const currentBlock = await provider.getBlockHeight({ commitment: "confirmed" }).send();
+  const startBlock = currentBlock - elapsedBlocks;
+  // Query for the deposit events with this limited block range. Filter by deposit id.
+  const depositEvent = (
+    await eventClient.queryEvents<SvmSpokeClient.FundsDeposited>("FundsDeposited", startBlock, currentBlock)
+  )?.find((event) => depositId.eq(event.data.depositId));
+
+  if (!depositEvent) {
+    return undefined;
+  }
```
pxrl, line 169: For `depositId < uint32_max - 1`, we can actually infer whether the search bounds were sufficient based on whether the earliest sequential `depositId` is above or below the requested `depositId`. We could use that as an indication for whether we should shuffle the search bounds (or not).

pxrl:
> I think this makes sense but since an undefined return value here would result in marking a fill as invalid in the BundleDataClient I would like to hear other's thoughts.

There might be some considerations around the validation of historical bundles. I _think_ it should be OK if we ensure to set the SpokePoolClient update bounds correctly for the relevant time period of whatever we're validating; in that case we should ideally avoid having to use `findDeposit()`.

If we passed in more than `depositId` to this function then we make a much better guess about the correct slot range based on `deposit.fillDeadline` and `deposit.quoteTimestamp`. That implies an interface change; I suspect that's probably OK though.

src/arch/svm/SpokeUtils.ts
melisaguevara: +1

src/arch/svm/SpokeUtils.ts
```diff
@@ -67,17 +69,72 @@ export function getDepositIdAtBlock(_contract: unknown, _blockTag: number): Prom
...
+  const slotDurationInMs = 300;
+
+  // We know we're likely to not be searching for a deposit id that is
+  // greater than two days older than the provided timestamp (or current time if no timestamp is provided)
+  // therefore we can use the current block and timestamp to calculate the slot at the timestamp and then
+  // calculate two days ago from that slot.
+  const provider = eventClient.getRpc();
+  const currentSlot = await provider.getSlot({ commitment: "confirmed" }).send();
+  const endSlot = currentSlot - BigInt(Math.round((currentTimeInMs - timestampInMs) / slotDurationInMs));
+  const startSlot = endSlot - BigInt(Math.round((2 * 24 * 60 * 60 * 1000) / slotDurationInMs));
```
md0x, line 120: Quick question: are we sure the 2-day value won’t ever change? If not it might be worth making it a function param with a default of 2.

src/arch/svm/SpokeUtils.ts
```diff
@@ -67,17 +69,72 @@ export function getDepositIdAtBlock(_contract: unknown, _blockTag: number): Prom
...
+ *                   for deposits between (timestamp - 2 days) and timestamp. Time must be in seconds.
+ * @returns The deposit if found within the time window, undefined otherwise
+ */
+export async function findDeposit(
+  eventClient: SvmCpiEventsClient,
   depositId: BigNumber,
-  _lowBlock: number,
-  _highBlock?: number
-): Promise<number | undefined> {
+  timestamp: number = getCurrentTime()
```
md0x, line 104: Should we validate that timestamp is not in the future?

src/arch/svm/SpokeUtils.ts
melisaguevara: @james-a-morris Do you think is worth adding some logic to shuffle bounds when possible?

melisaguevara (approved):
Just one question but this LGTM

src/arch/svm/SpokeUtils.ts
pxrl, line 104: ooc, why do we set the bounds based on timestamp rather than ~block~ slot number?

src/arch/svm/SpokeUtils.ts
```diff
@@ -39,6 +38,11 @@ import {
 import { getStatePda, SvmCpiEventsClient, getFillStatusPda, unwrapEventData, getEventAuthority } from "./";
 import { SVMEventNames, SVMProvider } from "./types";
 
+/**
+ * @note: Average Solana slot duration is about 400-500ms, using 500ms to be conservative
+ */
+export const SLOT_DURATION_MS = 500;
```
melisaguevara, line 44: I think for this case it'd be better to use 400 since that would mean we include more slots in our search later. wdyt?

src/arch/svm/SpokeUtils.ts
```diff
@@ -103,45 +107,49 @@ export function getDepositIdAtBlock(_contract: unknown, _blockTag: number): Prom
...
 
+  // If no deposit event is found, return undefined
   if (!depositEvent) {
     return undefined;
   }
 
+  // Return the deposit event with block info
   return {
     txnRef: depositEvent.signature.toString(),
     blockNumber: Number(depositEvent.slot),
```
melisaguevara, line 155: I see the EVM equivalent also adds the keys `originChainId`, `quoteBlockNumber`, `fromLiteChain` and `toLiteChain`. Should we include them for compatibility?
It also resolves the output token in case it is the zero address, is that something we have to do here? I think there's no need but flagging it just in case.

md0x (approved):

PR #989, opened 2025-04-24
improve(svmEventsClient): enable pda events querying

melisaguevara:
This PR adds support for querying events associated with PDAs in the SvmCpiEventsClient. 

When querying events from a PDA, there are two considerations:
1. The event authority is still derived from the program address, not from the PDA.
2. While we query transaction history by the PDA address, we still need to filter the events that were executed by the program address.

To support this, this PR:
- Adds an optional `pda` parameter to `createFor` method
- Introduces a `queryDerivedAddressEvents` method to specifically query PDA events
- Modifies event filtering logic to handle both program and PDA-specific queries

### DIFF
--- a/src/arch/svm/eventsClient.ts
+++ b/src/arch/svm/eventsClient.ts
@@ -32,6 +32,7 @@ export class SvmCpiEventsClient {
   private programAddress: Address;
   private programEventAuthority: Address;
   private idl: Idl;
+  private derivedAddress?: Address;
 
   /**
    * Private constructor. Use the async create() method to instantiate.
@@ -40,12 +41,14 @@ export class SvmCpiEventsClient {
     rpc: web3.Rpc<web3.SolanaRpcApiFromTransport<RpcTransport>>,
     address: Address,
     eventAuthority: Address,
-    idl: Idl
+    idl: Idl,
+    derivedAddress?: Address
   ) {
     this.rpc = rpc;
     this.programAddress = address;
     this.programEventAuthority = eventAuthority;
     this.idl = idl;
+    this.derivedAddress = derivedAddress;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/arch/svm/eventsClient.ts
```diff
@@ -61,14 +64,16 @@ export class SvmCpiEventsClient {
...
-    idl: Idl
+    idl: Idl,
+    pda?: string
   ): Promise<SvmCpiEventsClient> {
     const address = web3.address(programId);
+    const derivedAddress = pda ? web3.address(pda) : undefined;
     const [eventAuthority] = await web3.getProgramDerivedAddress({
       programAddress: address,
       seeds: ["__event_authority"],
     });
```
melisaguevara, line 75: Just to clarify: I couldn't instantiate the client using the derived address directly since the event authority should be calculated based on the program address.

src/arch/svm/eventsClient.ts
```diff
@@ -122,7 +153,7 @@ export class SvmCpiEventsClient {
 
     while (hasMoreSignatures) {
       const signatures: GetSignaturesForAddressApiResponse = await this.rpc
-        .getSignaturesForAddress(this.programAddress, currentOptions)
+        .getSignaturesForAddress(addressToQuery!, currentOptions)
```
melisaguevara, line 156: We can use the derived address in here but when filtering by executing address ([here](https://github.com/across-protocol/sdk/pull/989/files#diff-1a3493f875cbfb86bf3cfe72994cbce480bc3467eb6a83050ca8815200bcba54R237)) we still have to use the program address.

bmzig (approved):
I think this is fine. What if we made `eventAuthority` an optional field in `createFor` and only called `getProgramDerivedAddress` if it is undefined. i.e.
```
public static async createFor(rpc: ..., programId: string, idl: Idl, eventAuthority?: string) {
  eventAuthority ??= (await web3.getProgramDerivedAddress({ ... })[0];
}
```
Would that work? And then you wouldn't have to make another function for a derived address? Not sure if it fits your use case.

bmzig:
Oh never mind. I see that you still need to filter on the spoke pool address later on.

melisaguevara:
> Oh never mind. I see that you still need to filter on the program ID address later on.

@bmzig yes exactly. I tried what you suggested but I still needed both the program and its derived address.

src/arch/svm/eventsClient.ts
```diff
@@ -90,18 +95,39 @@ export class SvmCpiEventsClient {
...
+   */
+  public async queryDerivedAddressEvents(
+    eventName: string,
+    fromBlock?: bigint,
+    toBlock?: bigint,
+    options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
+  ): Promise<EventWithData[]> {
+    const events = await this.queryAllEvents(fromBlock, toBlock, options, true);
+    return events.filter((event) => event.name === eventName) as EventWithData[];
+  }
```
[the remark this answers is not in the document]
melisaguevara: Just for the record: we synced on this and since `queryAllEvents` is a private function we agreed on keeping the boolean flag for specifying which events to look for.

PR #990, opened 2025-04-24
feat(SvmSpokePoolClient): relayFillStatus and fillStatusArray implementation

melisaguevara:
This PR introduces the SVM implementation of `relayFillStatus`, which resolves a relay's fill status using PDA data or, if necessary, PDA events.

The resolution process is:

1. Retrieve the fill status PDA for the relay.
2. If querying the current slot:
   - If the PDA exists, fetch the status directly.
   - If the PDA doesn't exist but the fill deadline hasn’t passed, solve the status as Unfilled. This relies on the fact that PDAs can't be closed before the fill deadline.
3. If the status can't be determined using the PDA, or if a specific slot is requested, fetch and sort relevant PDA events (fills and slow fill requests) to resolve it.

This PR also introduces `fillStatusArray`, a batched version of the same logic for resolving multiple deposits. Note that resolving status via events for large arrays is slow. If it becomes a bottleneck, we could introduce a new `Unknown` fill status to skip event reconstruction entirely. This should be acceptable, as event-based resolution is only needed for deposits whose fill deadline has already passed or for queries at a specific slot. 

Some execution times observed during local testing:
- Fetching data from 100 accounts takes ~250ms. Multiple requests can run concurrently.
- Resolving statuses via events takes ~9000ms for each batch of 100 deposits.
Note: These timings were measured without caching which could reduce them further.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -87,11 +87,7 @@ export function findDepositBlock(
 /**
  * Resolves the fill status of a deposit at a specific slot or at the current confirmed one.
  *
- * If no slot is provided, the function first tries to fetch the status from the fill status PDA.
- * - If the PDA exists, it reads the status directly.
- * - If the PDA is missing and the deposit's fill deadline has not passed, the deposit is considered unfilled.
- *
- * If a specific slot is requested, or if the PDA is unavailable, the status is reconstructed from PDA events.
+ * If no slot is provided, attempts to solve the fill status using the PDA. Otherwise, it is reconstructed from PDA events.
  *
  * @param programId - The spoke pool program ID.
  * @param relayData - Deposit information used to locate the fill status.
@@ -137,9 +133,8 @@ export async function relayFillStatus(
 }
 
 /**
- * Resolves the fill status of an array of deposits at the requested slot or at the current confirmed one.
- * For current slot queries, first tries to fetch statuses using the PDAs.
- * When PDAs are unavailable, or for specific slot queries, it reconstructs status using PDA events.
+ * Resolves fill statuses for multiple deposits at a specific slot or current confirmed slot.
+ * Uses PDAs for current slot queries, falls back to event reconstruction when needed.
  *
  * @param programId The spoke pool program ID.
  * @param relayData An array of relay data to resolve fill statuses for.
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -1,4 +1,5 @@
 import assert from "assert";
+import { Logger } from "winston";
 import { Rpc, SolanaRpcApi, Address, fetchEncodedAccounts, fetchEncodedAccount } from "@solana/kit";
 import { fetchState, decodeFillStatusAccount } from "@across-protocol/contracts/dist/src/svm/clients/SvmSpoke";
 
@@ -153,7 +154,8 @@ export async function fillStatusArray(
   destinationChainId: number,
   provider: Provider,
   svmEventsClient: SvmCpiEventsClient,
-  atHeight?: number
+  atHeight?: number,
+  logger?: Logger
 ): Promise<(FillStatus | undefined)[]> {
   assert(chainIsSvm(destinationChainId), "Destination chain must be an SVM chain");
 
@@ -169,11 +171,12 @@ export async function fillStatusArray(
     )
   ).flat();
 
-  if (atHeight !== undefined) {
-    console.warn(
-      "fillStatusArray: Querying specific slots for large arrays is slow. " +
-        "For current status, omit 'atHeight' param to use latest confirmed slot instead."
-    );
+  if (atHeight !== undefined && logger) {
+    logger.warn({
+      at: "SvmSpokeUtils#fillStatusArray",
+      message:
+        "Querying specific slots for large arrays is slow. For current status, omit 'atHeight' param to use latest confirmed slot instead.",
+    });
   }
 
   // If no specific slot is requested, try fetching current statuses from PDAs
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -95,10 +95,10 @@ export function findDepositBlock(
 export async function relayFillStatus(
   programId: Address,
   relayData: RelayData,
-  blockTag: number | "confirmed" = "confirmed",
   destinationChainId: number,
   provider: Provider,
-  svmEventsClient: SvmCpiEventsClient
+  svmEventsClient: SvmCpiEventsClient,
+  blockTag?: number
 ): Promise<FillStatus> {
   assert(chainIsSvm(destinationChainId), "Destination chain must be an SVM chain");
 
@@ -107,7 +107,7 @@ export async function relayFillStatus(
 
   // Set search range
   let toSlot: bigint;
-  if (blockTag === "confirmed") {
+  if (!blockTag) {
     toSlot = await provider.getSlot({ commitment: "confirmed" }).send();
   } else {
     toSlot = BigInt(blockTag);
@@ -148,10 +148,10 @@ export async function relayFillStatus(
 export async function fillStatusArray(
   programId: Address,
   relayData: RelayData[],
-  blockTag: number | "confirmed" = "confirmed",
   destinationChainId: number,
   provider: Provider,
-  svmEventsClient: SvmCpiEventsClient
+  svmEventsClient: SvmCpiEventsClient,
+  blockTag?: number
 ): Promise<(FillStatus | undefined)[]> {
   assert(chainIsSvm(destinationChainId), "Destination chain must be an SVM chain");
   const chunkSize = 2;
@@ -160,7 +160,7 @@ export async function fillStatusArray(
   for (const chunk of chunkedRelayData) {
     const chunkResults = await Promise.all(
       chunk.map((relayData) =>
-        relayFillStatus(programId, relayData, blockTag, destinationChainId, provider, svmEventsClient)
+        relayFillStatus(programId, relayData, destinationChainId, provider, svmEventsClient, blockTag)
       )
     );
     results.push(...chunkResults);
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -1,8 +1,13 @@
+import assert from "assert";
 import { Rpc, SolanaRpcApi, Address } from "@solana/kit";
-
-import { Deposit, FillStatus, FillWithBlock, RelayData } from "../../interfaces";
-import { BigNumber, isUnsafeDepositId } from "../../utils";
+import { SvmSpokeIdl } from "@across-protocol/contracts";
 import { fetchState } from "@across-protocol/contracts/dist/src/svm/clients/SvmSpoke";
+
+import { SvmCpiEventsClient } from "./eventsClient";
+import { Deposit, FillStatus, FillWithBlock, RelayData } from "../../interfaces";
+import { BigNumber, chainIsSvm, chunk, isUnsafeDepositId } from "../../utils";
+import { getFillStatusPda } from "./utils";
+import { SVMEventNames } from "./types";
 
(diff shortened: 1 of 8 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -125,27 +130,86 @@ export function findDepositBlock(
...
+
+  // Get fill status PDA using relayData
+  const fillStatusPda = await getFillStatusPda(programId, relayData, destinationChainId);
+
+  // Set search range
+  let toSlot: bigint;
+  if (blockTag === "confirmed") {
+    toSlot = await provider.getSlot({ commitment: "confirmed" }).send();
+  } else {
+    toSlot = BigInt(blockTag);
```
pxrl, line 158: Is it worth validating `toSlot >= fromSlot` here?

src/arch/svm/SpokeUtils.ts
```diff
@@ -125,27 +130,86 @@ export function findDepositBlock(
...
+  programId: Address,
+  relayData: RelayData[],
+  fromSlot: number,
+  blockTag: number | "confirmed" = "confirmed",
+  destinationChainId: number,
+  provider: Provider
 ): Promise<(FillStatus | undefined)[]> {
-  throw new Error("fillStatusArray: not implemented");
+  assert(chainIsSvm(destinationChainId), "Destination chain must be an SVM chain");
+  const chunkSize = 2;
```
pxrl, line 203: Is this intended to be 2 or 200? If I read the commit message properly then I _think_ it hints at 200. 2 seems quite low, though we seem to lack any multicall-like read capabilities here, so going for high-parallelism adds the risk of rate-limiting as well :(

As background, the relayer can hit this function with an array of hundreds of relayData objects. If we're querying 2 at a time it's going to take quite a while to resolve that. I also wonder if we might see some nasty overheads given that `relayFillStatus()` instantiates a new events client for each individual query.

src/arch/svm/SpokeUtils.ts
```diff
@@ -125,27 +130,86 @@ export function findDepositBlock(
...
-export function relayFillStatus(
-  _spokePool: unknown,
-  _relayData: RelayData,
-  _blockTag?: number | "latest",
-  _destinationChainId?: number
+export async function relayFillStatus(
+  programId: Address,
+  relayData: RelayData,
+  fromSlot: number,
+  blockTag: number | "confirmed" = "confirmed",
```
pxrl, line 144: If the delta between `processed` and `confirmed` is reliably 0 - 2 slots then I wonder if we can just revert to a raw slot number for the upper bound (i.e. `toSlot?: number`), where `toSlot` would be a caller-specified slot, or otherwise resolved dynamically to the latest `confirmed` slot. wdyt?

src/clients/SpokePoolClient/SpokePoolClient.ts
```diff
@@ -813,7 +813,10 @@ export abstract class SpokePoolClient extends BaseAbstractClient {
    * @param blockTag The block at which to query the fill status.
    * @returns The fill status for the given relay data.
    */
-  public abstract relayFillStatus(relayData: RelayData, blockTag?: number | "latest"): Promise<FillStatus>;
+  public abstract relayFillStatus(
+    relayData: RelayData,
+    blockTag?: number | "latest" | "confirmed"
```
pxrl, line 818: Exposing `confirmed` at this interface means that the caller needs to be aware of which chain type they're on, and we then need to validate in each chain-specific implementation that the correct tag was passed in. If we modify the type to be just an optional number then we could infer "chain-specific latest" if the input is undefined. That seems more platform-agnostic, though it means that we internally must decide on `processed` or `confirmed` for SVM. Based on @james-a-morris feedback about the delta being only 1 - 2 slots, it seems like we could probably live with `confirmed` for now. wdyt?

```suggestion
    blockTag?: number
```

pxrl (commented):
@melisaguevara This looks great!

src/arch/svm/SpokeUtils.ts
melisaguevara: [0015fb4](https://github.com/across-protocol/sdk/pull/990/commits/0015fb4aca82a68ecf0b12b995abec627ffb52b2) This would use a single events client instance.

I'm doing some testing regarding the chunk size to have an idea of how long will it take. Will post some numbers in a bit.

src/arch/svm/SpokeUtils.ts
```diff
@@ -125,27 +130,86 @@ export function findDepositBlock(
...
+    toSlot = BigInt(blockTag);
+  }
+
+  // Get fill and requested slow fill events from fillStatusPda
+  const pdaEventsClient = await SvmCpiEventsClient.createFor(provider, programId, SvmSpokeIdl, fillStatusPda);
+  const eventsToQuery = [SVMEventNames.FilledRelay, SVMEventNames.RequestedSlowFill];
+  const relevantEvents = (
+    await Promise.all(
+      eventsToQuery.map((eventName) =>
+        pdaEventsClient.queryDerivedAddressEvents(eventName, BigInt(fromSlot), toSlot, { limit: 50 })
```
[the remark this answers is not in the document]
melisaguevara: From the events client's perspective, if we don't set `fromSlot`, it will fetch all events for the address. I think that should be fine for PDAs, since we expect them to have only a few events. The only case where they'd have many is if they were targeted with spam. So it seems safe to remove it, do you agree?

I've optimistically removed it here: [988c565](https://github.com/across-protocol/sdk/pull/990/commits/988c565f62ea8906f8d7239058de223241c8b258). For now I'm only setting it as undefined when calling `svmEventsClient.queryDerivedAddressEvents`, but I think we can improve readability if we pass the search range as an object rather than individual params to the events client functions. I can follow up with that change after we merge the update function to avoid conflicts.

src/clients/SpokePoolClient/SpokePoolClient.ts
melisaguevara: I like this idea! Fwiw, we aren't relying too much on this option, the only place where I see we are using "latest" explicitly is in [this line](https://github.com/across-protocol/relayer/blob/master/src/dataworker/Dataworker.ts#L1234) of the Dataworker. So I think this makes sense, wdyt @james-a-morris?

melisaguevara: This is how it would look like [d3cfa15](https://github.com/across-protocol/sdk/pull/990/commits/d3cfa155f4dfe77ace924d00ca560123ee59cfdc)

I wonder if we should change the name of this param too, blockTag might be confusing when used from the SVM Spoke Pool. Wdyt? It could be "atHeight" or something similar.

src/arch/svm/SpokeUtils.ts
melisaguevara: I think we can get rid of `fromSlot`.

src/arch/svm/SpokeUtils.ts
melisaguevara: I agree, I covered this below.

src/arch/svm/SpokeUtils.ts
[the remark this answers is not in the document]
melisaguevara: @pxrl The performance of this is really poor. It takes up to 20 seconds to fetch 100 deposits. So I’m looking into using the fill status PDA data when possible. There’s a method that allows querying 100 accounts at once. When checking the latest slot (which is what the relayer does), we can use it to filter out deposits that have either been filled or have a slow fill requested and still have an open PDA.

For deposits that are unfilled or filled but have a closed PDA, I'm afraid we’ll still have to reconstruct from events 😞. I'm still thinking if there's any faster approach for this case.

src/arch/svm/SpokeUtils.ts
melisaguevara: @pxrl The performance of this is really poor. It takes up to 20 seconds to fetch 100 deposits. So I’m looking into using the fill status PDA data when possible. There’s a method that allows querying 100 accounts at once. When checking the latest slot (which is what the relayer does), we can use it to filter out deposits that have either been filled or have a slow fill requested and still have an open PDA.

For deposits that are unfilled or filled but have a closed PDA, I'm afraid we’ll still have to reconstruct from events 😞. I'm still thinking if there's any faster approach for this case.

melisaguevara: [182cde9](https://github.com/across-protocol/sdk/pull/990/commits/182cde9aa01c75484c6eaee42a2b430a135470b2) Based on our discussion, now we are also relying on the fill deadline to filter unfilled deposits easily and reduce the amount of statuses we have to resolve with events.

md0x (approved):
Amazing work!

src/arch/svm/SpokeUtils.ts
```diff
@@ -81,27 +85,136 @@ export function findDepositBlock(
...
+    const fillStatusAccount = await fetchEncodedAccount(provider, fillStatusPda, { commitment: "confirmed" });
+
+    // If the PDA exists, return the stored fill status
+    if (fillStatusAccount.exists) {
+      const decodedAccountData = decodeFillStatusAccount(fillStatusAccount);
+      return decodedAccountData.data.status;
+    }
+    // If the PDA doesn't exist and the deadline hasn't passed yet, the deposit must be unfilled,
+    // since PDAs can't be closed before the fill deadline.
+    else if (getCurrentTime() < relayData.fillDeadline) {
```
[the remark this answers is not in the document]
melisaguevara: No, it's on seconds which is the same unit that getCurrentTime returns.

src/arch/svm/SpokeUtils.ts
```diff
@@ -81,27 +85,136 @@ export function findDepositBlock(
...
+        Promise.all(relayDataChunk.map((relayData) => getFillStatusPda(programId, relayData, destinationChainId)))
+      )
+    )
+  ).flat();
+
+  if (atHeight !== undefined) {
+    console.warn(
+      "fillStatusArray: Querying specific slots for large arrays is slow. " +
+        "For current status, omit 'atHeight' param to use latest confirmed slot instead."
+    );
```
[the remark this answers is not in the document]
melisaguevara: No, it was intentional, since reconstructing status from events is really slow I thought it would be useful to have this just in case.

src/arch/svm/SpokeUtils.ts
Reinis-FRP, line 124: At the current implementation `getCurrentTime()` returns cached value from the latest `update()` call. This might be lower than the latest block time at the confirmed commitment when fill status account was fetched above, so there is some chance that this was filled and fill status account was closed just between the `update()` and `fetchEncodedAccount` calls. It would be safer to use the latest block time with the same confirmed commitment updating it immediately after the `fetchEncodedAccount` call.

src/arch/svm/SpokeUtils.ts
```diff
@@ -130,3 +243,86 @@ export function findFillEvent(
...
+
+  const fillStatuses = pdaAccounts.map((account, index) => {
+    // If the PDA exists, we can fetch the status directly.
+    if (account.exists) {
+      const decodedAccount = decodeFillStatusAccount(account);
+      return decodedAccount.data.status;
+    }
+    // If the PDA doesn't exist and the deadline hasn't passed yet, the deposit must be unfilled,
+    // since PDAs can't be closed before the fill deadline.
+    else if (getCurrentTime() < relayDataArray[index].fillDeadline) {
```
Reinis-FRP, line 320: same comment as above when fetching single account to use updated block time for this comparison

melisaguevara: Also covered here [86f6008](https://github.com/across-protocol/sdk/pull/990/commits/86f6008c8689be91f054c38f372b4024e1f6a4dc)

Reinis-FRP (approved):

md0x (approved):

src/arch/svm/SpokeUtils.ts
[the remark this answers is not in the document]
melisaguevara: Right, good call! [a5b03ce](https://github.com/across-protocol/sdk/pull/990/commits/a5b03cea8a37a0499fe763ef15aa1f81bb65e5e6)

PR #1001, opened 2025-04-30
improve: getL1TokenInfo should work for testnet-only tokens

bmzig:


nicholaspai (approved):

nicholaspai:
I think this is redundant with https://github.com/across-protocol/sdk/commit/3ef0608b2fd3edeb38f561d04236a506bc416aed

PR #1002, opened 2025-04-30
fix(TokenUtils): Handle testnet chainId being passed into getL1TokenInfo

nicholaspai:


src/utils/TokenUtils.ts
```diff
@@ -138,10 +138,10 @@ export function getUsdcSymbol(l2Token: string, chainId: number): string | undefi
 }
 
 export function getL1TokenInfo(l2TokenAddress: string, chainId: number): L1Token {
-  if (chainId === CHAIN_IDs.MAINNET)
+  if ((chainIsProd(chainId) && chainId === CHAIN_IDs.MAINNET) || chainId === CHAIN_IDs.SEPOLIA)
```
bmzig, line 141: I think you just need to do `if(chainIsL1(chainId))` here.

PR #1006, opened 2025-04-30
chore: Bump constants and add USDT-BNB to stablecoin mapping

nicholaspai:


src/constants.ts
```diff
@@ -73,6 +73,7 @@ export const STABLE_COIN_SYMBOLS = [
   TOKEN_SYMBOLS_MAP.USDB.symbol,
   TOKEN_SYMBOLS_MAP.USDC.symbol,
   TOKEN_SYMBOLS_MAP.USDT.symbol,
+  TOKEN_SYMBOLS_MAP["USDT-BNB"].symbol,
```
pxrl, line 76: @nicholaspai Do we also need USDC-BNB in here?

package.json
```diff
@@ -104,7 +104,7 @@
   },
   "dependencies": {
     "@across-protocol/across-token": "^1.0.0",
-    "@across-protocol/constants": "^3.1.59",
+    "@across-protocol/constants": "^3.1.60",
```
bmzig, line 107: Can we also include the newest version of constants here?
```suggestion
    "@across-protocol/constants": "^3.1.61",
```

src/constants.ts
bmzig: I added it in a previous commit here https://github.com/across-protocol/sdk/blob/7457901f003e714bfaeae0614386cea85900a27e/src/constants.ts#L68, but I forgot to add USDT.

PR #1010, opened 2025-05-01
feat: Mock SVMSpokePoolClient

melisaguevara:
This PR introduces a MockSvmSpokePoolClient with util functions to mock onchain events using the MockSvmEventsClient.

It also re-arranges our mock files into /src so that they are in consistent folders instead of separated into /src and /test.


test/mocks/MockSvmSpokePoolClient.ts
pxrl, line 1: Just as a general comment to the location, we have existing "client" mocks (including `SpokePoolClient`) up in `src/clients/mocks`. It's debatable whether they belong up there, so we could consider relocating them to `test/mocks`, or vice versa. I think it makes sense to keep them together though.

test/mocks/MockSolanaEventClient.ts
```diff
@@ -46,4 +60,149 @@ export class MockSolanaEventClient extends SvmCpiEventsClient {
...
+    let { depositId, inputAmount, outputAmount, fillDeadline } = fill;
+    depositId ??= Uint8Array.from([random(1, 100_000, false)]);
+    inputAmount ??= BigInt(random(1, 1000, false));
+    outputAmount ??= (inputAmount * 95n) / 100n;
+    fillDeadline ??= getCurrentTime() + 60;
+
+    const depositor = fill.depositor ?? getRandomSvmAddress();
+    const recipient = fill.recipient ?? depositor;
+    const inputToken = fill.inputToken ?? getRandomSvmAddress();
+    const outputToken = fill.outputToken ?? inputToken;
```
pxrl, line 115: Should these be reversed for an SVM fill, such that we revert to SVM addreses on recipient and outputToken ? Also, what happens to `recipient` and `outputToken` if `fill.depositor` or `fill.inputToken`  are undefined?

test/mocks/MockSvmSpokePoolClient.ts
```diff
@@ -0,0 +1,119 @@
...
+  getDestinationTokenForDeposit(deposit: DepositWithBlock): string {
+    return this.destinationTokenForChainOverride[deposit.originChainId] ?? super.getDestinationTokenForDeposit(deposit);
+  }
+
+  setLatestBlockNumber(blockNumber: number): void {
+    this.latestHeightSearched = blockNumber;
+  }
+
+  async _update(eventsToQuery: string[]): Promise<SpokePoolUpdate> {
+    const from = this.eventSearchConfig.from ? BigInt(this.eventSearchConfig.from) : undefined;
```
pxrl, line 58: ooc, should the `from` slot revert to `this.deploymentBlock` in case of not being specified by the caller?

```suggestion
    const from = this.eventSearchConfig.from ? BigInt(this.eventSearchConfig.from) : undefined;
```

melisaguevara: Yes that's a good idea. Done [here](https://github.com/across-protocol/sdk/pull/1010/files#diff-687f55ed02679a4c62667e363806b6797a5c7830d76f7ce98138c9816f87e87cR58).

test/mocks/MockSolanaEventClient.ts
melisaguevara: I reversed those 👍🏼 
And also to have the mocked events be more consistent to what we have onchain, now I'm setting either depositor or recipient (depending on the event) to the base58 representation of an EVM address. I'm also doing that for inputToken and outputToken. See for example [mock deposit fn](https://github.com/across-protocol/sdk/pull/1010/files#diff-b23e46fa5f9ce6278bca54f0fd312bf34c10af70e5d70294eaf1d0f72ef8d9e9R72-R75).

test/mocks/MockSvmSpokePoolClient.ts
melisaguevara: I agree yes. I decided to keep them under /src because the relayer is importing some of those classes but we are not exporting the /test folder. To be consistent I also moved all the mock providers to /src/providers/mocks. [483a879](https://github.com/across-protocol/sdk/pull/1010/commits/483a879a591168ecc0ca419dc5bce269a0baab1d)
lmk what you think!

src/clients/mocks/MockSolanaEventClient.ts
melisaguevara, line 1: This is the same file we already had. I only relocated it.

test/mocks/MockSolanaEventClient.ts
melisaguevara, line 1: This was relocated to /src/clients/mocks

src/clients/mocks/MockSolanaEventClient.ts
```diff
@@ -0,0 +1,222 @@
...
+import { FillType } from "../../interfaces";
+import { getRandomSvmAddress } from "../../arch/svm/utils";
+
+export class MockSolanaEventClient extends SvmCpiEventsClient {
+  private events: Record<EventName, EventWithData[]> = {} as Record<EventName, EventWithData[]>;
+  private slotHeight: bigint = BigInt(0);
+  public chainId: number;
+  public minBlockRange = 10;
+  public numberOfDeposits = bnZero;
+  public SVM_ZERO_ADDRESS = bs58.encode(new Uint8Array(32));
```
md0x, line 23: We now have a constante exported with the  `SVM_DEFAULT_ADDRESS` here:
https://github.com/across-protocol/sdk/blob/9b3bb77407182843eb71e87f3e3412bcdbca4351/src/arch/svm/constants.ts#L1

melisaguevara: oh right, let me update this.

src/clients/mocks/MockSolanaEventClient.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    return this.generateEvent({
+      event: SVMEventNames.FundsDeposited,
+      address: this.getProgramAddress(),
+      args,
+      slot,
+    });
+  }
+
+  public fillRelay(fill: SvmSpokeClient.FilledRelay & Partial<EventWithData>): EventWithData {
```
md0x, line 105: Why is fillRelay function in the Mock events client? Shouldn't it be in the MockSVM SpokePoolClient?

src/clients/mocks/MockSolanaEventClient.ts
```diff
@@ -0,0 +1,222 @@
...
+import { CHAIN_IDs } from "@across-protocol/constants";
+
+import { MockSolanaRpcFactory } from "../../providers/mocks";
+import { SvmCpiEventsClient } from "../../arch/svm/eventsClient";
+import { EventName, EventWithData, SVMEventNames, SVMProvider } from "../../arch/svm";
+import { bnZero, bnOne, bs58, getCurrentTime, randomAddress, EvmAddress } from "../../utils";
+import { FillType } from "../../interfaces";
+import { getRandomSvmAddress } from "../../arch/svm/utils";
+
+export class MockSolanaEventClient extends SvmCpiEventsClient {
```
md0x, line 17: Maybe name it `MockSvmCpiEventsClient` for consistency

src/clients/mocks/MockSolanaEventClient.ts
melisaguevara: Here [0b04d83](https://github.com/across-protocol/sdk/pull/1010/commits/0b04d83e527d8c5208bcb06cfd2ac1aedf70eabe)

src/clients/mocks/MockSolanaEventClient.ts
melisaguevara: I wrote such functions in this mock client because here we also have the `generateEvent` and `setEvents` functions which uses the slotHeight property of this mock client.

src/clients/mocks/MockSolanaEventClient.ts
melisaguevara: Yes good call [d6c2bca](https://github.com/across-protocol/sdk/pull/1010/commits/d6c2bca4ef301a303ce2d0303b70aa43c60ee813)

src/clients/mocks/MockSolanaEventClient.ts
md0x, line 1: Can also rename the file name `MockSvmCpiEventsClient`

melisaguevara: [c970ee7](https://github.com/across-protocol/sdk/pull/1010/commits/c970ee730afc484d8a087fd83971b754a8c96773)

md0x (approved):
Nice job!

PR #1015, opened 2025-05-02
feat(svm): integration tests with test validator

md0x:
Changes proposed in this PR:
- Added SVM integration test scaffolding using `solana-test-validator` and integrated it into the CI pipeline.
- Implemented setup utilities to initialize SVM Spoke state, create a test mint, and enable a test route.
- Added SvmSpokeEventsClient integration tests.


test/utils/svm/validator.setup.ts
```diff
@@ -0,0 +1,37 @@
...
+import { SvmSpokeClient } from "@across-protocol/contracts";
+
+const LEDGER_DIR = path.resolve(__dirname, "..", ".ledger");
+
+export default async () => {
+  // Always start with a clean ledger
+  await fs.rm(LEDGER_DIR, { recursive: true, force: true });
+
+  const args = [
+    "--clone-upgradeable-program",
```
md0x, line 13: Currently, we’re cloning the program from Solana mainnet, which requires querying an RPC and introduces associated latency (~10-15 seconds). This will be replaced by the SVM Spoke binaries once they’re exported from the contracts repo.

test/utils/svm/utils.ts
```diff
@@ -0,0 +1,359 @@
...
+  const depositInput: DepositInput = {
+    ...depositData,
+    ...depositAccounts,
+    eventAuthority,
+    signer,
+  };
+
+  const depositIx = await SvmSpokeClient.getDepositInstructionAsync(depositInput);
+
+  return pipe(
```
md0x, line 353: Note: These are test-only helper functions designed to simplify interactions with the program. They do not estimate optimal compute units or configure priority fees for better block inclusion. However, they can serve as a reference for implementing the program interaction interface in the SDK.

test/integration/spoke.events.client.integration.test.ts
md0x, line 1: These tests will be executed using Mocha as part of the `"test"` script integrated through hardhat.

.github/workflows/test.yml
```diff
@@ -8,6 +8,41 @@ jobs:
...
+          WORKDIR="/tmp/contracts"
+          mkdir -p "$WORKDIR"
+
+          curl -sSfL "https://raw.githubusercontent.com/${REPO}/${REF}/Cargo.lock" \
+               -o "${WORKDIR}/Cargo.lock" || true
+
+          cd "$WORKDIR"
+
+          if [[ -n "${OVERRIDE_SOLANA_VERSION:-}" ]]; then
+            SOLANA_VERSION="${OVERRIDE_SOLANA_VERSION}"
```
Reinis-FRP, line 30: where would `OVERRIDE_SOLANA_VERSION` env come from?

md0x: `OVERRIDE_SOLANA_VERSION` isn’t set anywhere currently. I just left the door open for it in case we want to override the extracted version in the future. If needed, it can be set via repo/org variables or job env

test/Solana.SvmCpiEventsClient.Integration.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  let state: Address;
+  let decimals: number;
+
+  const destinationChainId = 1n;
+  const tokenAmount = 100000000n;
+
+  // helper to create a deposit
+  const createDeposit = async (payerAta: Address, inputAmount: bigint, outputAmount: bigint) => {
+    const lastBlock = await solanaClient.rpc.getBlockHeight({ commitment: "confirmed" }).send();
+    const currentTime = await solanaClient.rpc.getBlockTime(lastBlock).send();
```
melisaguevara, line 35: `getBlockTime` receives a slot number actually (which is annoying 😕). So we should get the latest slot instead.

test/Solana.SvmCpiEventsClient.Integration.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    const tx1 = await solanaClient.rpc
+      .getTransaction(firstSig, {
+        commitment: "confirmed",
+        maxSupportedTransactionVersion: 0,
+      })
+      .send();
+    if (!tx1) throw new Error("first deposit tx not found");
+
+    /* Second deposit (should be returned) */
+    await mintTokens(signer, solanaClient, mint.address, tokenAmount + 1n);
```
melisaguevara, line 129: ooc, is this call intended for the signer to have more tokens available for the second deposit?

test/utils/svm/utils.ts
```diff
@@ -0,0 +1,385 @@
...
+const STATE_SEED = 0n;
+
+// Derives the global State PDA for SVM Spoke.
+export const getStatePda = async () => {
+  const [state] = await getProgramDerivedAddress({
+    programAddress: address(SvmSpokeClient.SVM_SPOKE_PROGRAM_ADDRESS),
+    seeds: [Buffer.from("state"), toLEBuffer(STATE_SEED, 8)],
+  });
+  return state;
+};
```
melisaguevara, line 227: nit: We also have this function [here](https://github.com/across-protocol/sdk/blob/epic/svm-client/src/arch/svm/utils.ts#L142). Might be worth reusing it?

test/utils/svm/utils.ts
```diff
@@ -0,0 +1,385 @@
...
+    inputToken: Address;
+    outputToken: Address;
+    inputAmount: bigint;
+    outputAmount: bigint;
+    destinationChainId: number;
+    exclusiveRelayer: Address;
+    quoteTimestamp: number;
+    fillDeadline: number;
+    exclusivityParameter: number;
+    message: ReadonlyUint8Array;
```
melisaguevara, line 344: Is it possible to use `SvmSpokeClient.DepositInstructionDataArgs` here?

melisaguevara (commented):
Amazing work! Just a few questions

test/Solana.SvmCpiEventsClient.Integration.test.ts
md0x: You are absolutely right, good call!

Reinis-FRP (approved):
looks good, though better to reuse getStatePda as suggested by @melisaguevara

test/Solana.SvmCpiEventsClient.Integration.test.ts
md0x: Actually I could merge them in a single commit. Initially I wanted to make sure both deposits landed in different slots that's why I added the mint in between them but it's not necessary :)

test/utils/svm/utils.ts
md0x: Good call! I will use `SvmSpokeClient.DepositInput` that includes both params and accounts. Same for the rest of interactions

package.json
```diff
@@ -111,6 +111,8 @@
     "@ethersproject/bignumber": "^5.7.0",
     "@pinata/sdk": "^2.1.0",
     "@solana/kit": "^2.1.0",
+    "@solana-program/system": "^0.7.0",
+    "@solana-program/token-2022": "^0.4.0",
```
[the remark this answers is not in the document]
md0x: This will need to be in the sdk package in order to interact with the programs so I think they need to be installed as regular dependencies.

PR #1026, opened 2025-05-08
improve: SVM getTimestampForSlot

melisaguevara:
This PR changes the implementation of `getTimestampForSlot` to use [getBlockTime](https://solana.com/docs/rpc/http/getblocktime). This RPC method does exactly what we need. It takes a slot (despite its misleading name 😅) and returns its timestamp. By using it we no longer need to handle the case when there was no block mined in the requested slot.

This PR also implements `getTimeAt` which also uses `getTimestampForSlot`.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -25,29 +25,12 @@ export function populateV3Relay(
 }
 
 /**
- * Retrieves the time from the SpokePool contract at a particular block.
- * @returns The time at the specified block tag.
- */
-export function getTimeAt(_spokePool: unknown, _blockNumber: number): Promise<number> {
-  throw new Error("getTimeAt: not implemented");
-}
-
-/**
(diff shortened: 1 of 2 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -24,30 +24,13 @@ export function populateV3Relay(
...
 }
 
-/**
- * Retrieves the time from the SpokePool contract at a particular block.
- * @returns The time at the specified block tag.
- */
-export function getTimeAt(_spokePool: unknown, _blockNumber: number): Promise<number> {
-  throw new Error("getTimeAt: not implemented");
-}
-
```
melisaguevara, line 34: From what I saw, the relayer doesn't use this function directly. It always calls it via the SpokePoolClient. So I think we are ok to delete it.

pxrl (approved):

PR #1028, opened 2025-05-09
feat: add request slow fill and fill functions and event tests

md0x:
Changes proposed in this PR:
- Add `RequestedSlowFill`  and `FilledRelay` events integration tests.
- Add `requestSlowFill` and `createFill` functions


test/Solana.SvmCpiEventsClient.Integration.test.ts
```diff
@@ -12,10 +22,13 @@ import {
...
   initializeSvmSpoke,
   mintTokens,
+  requestSlowFill,
 } from "./utils/svm/utils";
 import { validatorSetup, validatorTeardown } from "./utils/svm/validator.setup";
+import { SvmAddress, getRelayDataHash } from "../src/utils";
+import { RelayData } from "../src/interfaces";
 
-describe("SvmCpiEventsClient (integration)", () => {
+describe.only("SvmCpiEventsClient (integration)", () => {
```
pxrl, line 31: Watch out for this one :eyes: 

```suggestion
describe("SvmCpiEventsClient (integration)", () => {
```

test/Solana.SvmCpiEventsClient.Integration.test.ts
```diff
@@ -62,6 +75,83 @@ describe("SvmCpiEventsClient (integration)", () => {
...
+    if (arr.length === 0) {
+      return BigNumber.from(0); // or handle this case appropriately
+    }
+
+    const hex = Array.from(arr)
+      .map((byte) => byte.toString(16).padStart(2, "0"))
+      .join("");
+
+    return BigNumber.from("0x" + hex);
+  };
```
pxrl, line 88: This should fwiw be unnecessary - BigNumber seems to handle Uint8Arrays natively.

```
> arr = new Uint8Array([255])
Uint8Array(1) [ 255 ]
> ethers.BigNumber.from(arr).toNumber()
255

```

test/Solana.SvmCpiEventsClient.Integration.test.ts
```diff
@@ -62,6 +75,83 @@ describe("SvmCpiEventsClient (integration)", () => {
...
+      depositor: signer.address,
+      recipient: signer.address,
+      exclusiveRelayer: signer.address,
+      inputToken: mint.address,
+      outputToken: getRandomSvmAddress(),
+      inputAmount: tokenAmount,
+      outputAmount: tokenAmount,
+      originChainId: 1,
+      depositId: new Uint8Array(),
+      fillDeadline: Number(currentTime) + 60 * 30, // 30‑minute deadline
```
pxrl, line 106: There might be a misunderstanding here - this should be copied 1:1 from the deposit arguments, otherwise the RelayData hash won't match and the slow fill request will be invalid.

test/Solana.SvmCpiEventsClient.Integration.test.ts
```diff
@@ -147,4 +237,16 @@ describe("SvmCpiEventsClient (integration)", () => {
     expect(events).to.have.lengthOf(1);
     expect(events[0].data.inputAmount).to.equal(secondDeposit.inputAmount);
   });
+
+  it.only("creates and reads a single request slow fill event", async () => {
```
pxrl, line 241: One more

```suggestion
  it("creates and reads a single request slow fill event", async () => {
```

test/Solana.SvmCpiEventsClient.Integration.test.ts
```diff
@@ -147,4 +237,16 @@ describe("SvmCpiEventsClient (integration)", () => {
...
+    await sendRequestSlowFill();
+
+    const [requestSlowFillEvent] = await client.queryEvents("RequestedSlowFill");
+
+    const { data } = requestSlowFillEvent as { data: SvmSpokeClient.RequestedSlowFill };
+
+    expect(data.depositor).to.equal(signer.address.toString());
+    expect(data.recipient).to.equal(signer.address.toString());
+    expect(data.inputToken).to.equal(mint.address.toString());
```
pxrl, line 250: Is it work constructing `sendRequestSlowFill()` here such that it accepts a RelayData input? Then we can compare the RelayData hashes on the output side to verify that the request was submitted correctly. This would also be convenient as a utility function down in the relayer, because we currently don't have any slow fill request support for SVM.

md0x: I added the generic `deposit`, `requestSlowFill`, and `createFill` functions here:
https://github.com/across-protocol/sdk/blob/02128c254326f137826c10654a31d60bde5548a4/test/utils/svm/utils.ts#L264-L339

They take SVM-typed inputs to keep things flexible.

For the event tests, I made simpler wrappers that also return the relayData object, so it’s easy to check that the events are logging correctly:
https://github.com/across-protocol/sdk/blob/02128c254326f137826c10654a31d60bde5548a4/test/Solana.SvmCpiEventsClient.Integration.test.ts#L70-L205

src/arch/svm/SpokeUtils.ts
```diff
@@ -412,29 +415,24 @@ export async function getAssociatedTokenAddress(
...
+    Uint8Array.from(addressEncoder.encode(SvmAddress.from(relayData.exclusiveRelayer, "base16").toV2Address())),
+    Uint8Array.from(addressEncoder.encode(SvmAddress.from(relayData.inputToken, "base16").toV2Address())),
+    Uint8Array.from(addressEncoder.encode(SvmAddress.from(relayData.outputToken, "base16").toV2Address())),
+    Uint8Array.from(uint64Encoder.encode(BigInt(relayData.inputAmount.toString()))),
+    Uint8Array.from(uint64Encoder.encode(BigInt(relayData.outputAmount.toString()))),
+    Uint8Array.from(uint64Encoder.encode(BigInt(relayData.originChainId.toString()))),
+    arrayify(hexZeroPad(hexlify(relayData.depositId), 32)),
+    Uint8Array.from(uint32Encoder.encode(relayData.fillDeadline)),
+    Uint8Array.from(uint32Encoder.encode(relayData.exclusivityDeadline)),
+    hashNonEmptyMessage(Buffer.from(relayData.message)),
```
md0x, line 434: empty messages need to use [hashNonEmptyMessage](https://github.com/across-protocol/contracts/blob/31356a285cc345267a75b5ff466eaf53c1b03bc9/src/svm/web3-v1/relayHashUtils.ts#L70-L77)

melisaguevara (commented):
left a comment regarding getRelayDataHash

src/arch/svm/SpokeUtils.ts
md0x: Good call, @melisaguevara!

Since relay messages on Solana are passed as binary data to the program, and relayHash expects a string for EVM compatibility, we now hex-encode the Solana relay message before passing it in. We can then decode it prior to hashing.

I’ve added those changes in my last commit.

Also, note that I updated [formatRelayData](https://github.com/across-protocol/sdk/blob/f2aac6169d9c2a1902214c23cec17babebec6c60/test/Solana.SvmCpiEventsClient.Integration.test.ts#L33-L48) to hexlify the original binary message accordingly.

melisaguevara: Awesome, @md0x! It is working now! 😃

melisaguevara (approved):
LGTM!

src/arch/svm/SpokeUtils.ts
gsteenkamp89: So @md0x @melisaguevara  does this mean that empty messages on SVM chains will be a non-empty buffer? ("0x")

melisaguevara: @gsteenkamp89 onchain they will be an empty buffer (see [this deposit](https://solscan.io/tx/3XkpJMmETwcTW6MX3i95psi9ogTyePPdXUHnhvbZNW1xNiETtvuQvGD1Cqi5gkmKmNg1x8vCabnhqAow7nN8KDjw) for example). But the SvmSpokePoolClient and any other function using the svm util [unwrapEventData](https://github.com/across-protocol/sdk/blob/epic/svm-client/src/arch/svm/utils.ts#L97) will parse those empty buffers and set the message to '0x' (see [this line](https://github.com/across-protocol/sdk/blob/epic/svm-client/src/arch/svm/utils.ts#L144-L147)). Do you think that's ok?

src/arch/svm/constants.ts
```diff
@@ -1 +1,4 @@
+import { address } from "@solana/kit";
+
 export const SVM_SPOKE_SEED = BigInt(0);
+export const SVM_ZERO_ADDRESS = address("11111111111111111111111111111111");
```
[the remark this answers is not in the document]
md0x: Good point.
I’ve updated it. Technically it’s not the 0 address, but the default address, which happens to be the same as `SYSTEM_PROGRAM_ADDRESS`.
I went ahead and exported it as `SVM_DEFAULT_ADDRESS` so it’s clearer what it’s for.

PR #1031, opened 2025-05-12
improve(gasPriceOracle): Add basic gas price type discriminators

pxrl:
Prompted by linking against the relayer and seeing a few type incompatibility errors.


gsteenkamp89 (approved):
nice this will be useful

PR #1033, opened 2025-05-12
feat(svm-eventsClient): add methods to find deposit and fills from signature

gsteenkamp89:
Adds 2 methods to cpi events client:
- find deposit event from signature
- find fill event from signature


src/arch/svm/SpokeUtils.ts
```diff
@@ -276,6 +277,62 @@ export async function findFillEvent(
...
+
+  return events.map((event) => {
+    const unwrappedEventArgs = unwrapEventData(event as Record<"data", Record<string, unknown>>) as Record<
+      string,
+      unknown
+    >;
+    return {
+      originChainId,
+      fromLiteChain: false, // to be updated
+      toLiteChain: false, // to be updated
```
gsteenkamp89, line 325: is it worthwhile to exclude these 2 fields here so the consumer is forced to calculate this?

src/arch/svm/SpokeUtils.ts
```diff
@@ -276,6 +277,63 @@ export async function findFillEvent(
...
+  if (!txDetails) {
+    return;
+  }
+
+  const slot = BigInt(txDetails.slot);
+
+  // find FundsDeposited events for that slot
+  const depositEvents = await eventsClient.queryEvents(SVMEventNames.FundsDeposited, slot, slot, {
+    commitment: "confirmed",
+  });
```
melisaguevara, line 311: This works, but given how Solana works, this function will start querying events from the most recent slot even if you provide a `toSlot`.
Since we already have the transaction signature, what do you think about parsing its events by turning the function [readEventsFromSignature](https://github.com/across-protocol/sdk/blob/pablo/slow-fill/src/arch/svm/eventsClient.ts#L166-L173) from the events client into a public one? Then we'll have to filter by event name but the result would be the same.

src/arch/svm/SpokeUtils.ts
melisaguevara: I'd say is better to exclude them to avoid having anyone trusting the false values.

gsteenkamp89: Yeah I tend to agree.

src/arch/svm/SpokeUtils.ts
gsteenkamp89: It did feel a bit slow in testing, this explains it. Great suggestion, thanks!

src/arch/svm/SpokeUtils.ts
```diff
@@ -276,6 +277,63 @@ export async function findFillEvent(
...
 
+/**
+ * Finds all FundsDeposited events for a given transaction signature.
+ *
+ * @param originChainId - The chain ID where the deposit originated.
+ * @param txSignature - The transaction signature to search for events.
+ * @param eventsClient - SvmCpiEventsClient instance for querying events.
+ * @returns A promise that resolves to an array of deposit events for the transaction, or undefined if none found.
+ */
+export async function getDepositEventsFromSignature(
```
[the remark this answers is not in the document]
gsteenkamp89: yeah good idea.

gsteenkamp89:
> We may want to add unit tests now that we're including this function in the events client

I can dd them as a fast-follow once this [PR](https://github.com/across-protocol/sdk/pull/1028) is merged with the required `createFill` test utility

melisaguevara (approved):

PR #1037, opened 2025-05-14
feat(svm): add svm test node hook at root level

md0x:


test/multicall.test.ts
```diff
@@ -26,8 +26,12 @@ describe("getMulticallAddress", () => {
   });
 });
 
-describe("getMulticall3", async () => {
-  const provider = (await ethers.getSigners()).at(0).provider;
+describe("getMulticall3", function () {
+  let provider;
```
md0x, line 30: Unrelated, but having an async test prevented me from using the only option to run specific tests. Moving the async initialization to a before hook resolved the issue and follows best practices.

PR #1038, opened 2025-05-14
improve: generalize block finder interface for EVM

bmzig:
This is the first part of a two-part PR which will adapter the block finder to work for both EVM and SVM. The strategy I am going for is to have a base interface and then passing in generic types to resolve EVM/SVM block types for each block finder.

### DIFF
--- a/src/arch/evm/BlockUtils.ts
+++ b/src/arch/evm/BlockUtils.ts
@@ -1,17 +1,16 @@
 import assert from "assert";
 import { Provider, Block as EthersBlock } from "@ethersproject/abstract-provider";
 import { clamp, sortedIndexBy } from "lodash";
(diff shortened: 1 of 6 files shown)


src/arch/evm/BlockUtils.ts
bmzig, line 1: This is basically a copy/paste of all the EVM-specific stuff in `src/utils/BlockUtils`. It implements an `EVMBlockFinder` which is really just a `BlockFinder<EVMBlock>` that gets those blocks via an ethers provider.

bmzig:
> Is the expectation that we will be following the same rollout pattern as SVM?

Yeah exactly. I'm working on the SVM one now.

melisaguevara (approved):

PR #1039, opened 2025-05-15
feat: expose `getNativeGasCosts` in `svmQuery`

dohaki:
We need to expose `getNativeGasCosts` in the SVM implementation of the relay fee calculator.


bmzig (approved):
Good catch

PR #1044, opened 2025-05-15
feat: SVM Block finder

bmzig:
The SVM Block finder is slightly different from the EVM Block finder in the sense that input values are slots. Obviously, a slot may not have a block, but most functions assume that a block is to be returned. I addressed this by taking a range of values starting from that slot onward and returning the earliest block within that range. This means we get the closest block to that slot such that the timestamp/slot is still less than or equal to the returned block.


bmzig:
There are a lot of merge conflicts now so it's going to be easier to just re-open on master.

PR #1048, opened 2025-05-16
feat(svm): svm spoke binaries integration tests

md0x:
Changes proposed in this PR:
- Update the solana test validator to use the svm verified test binaries
- Add close pda test

### DIFF
--- a/test/SVMSpokePoolClient.fills.ts
+++ b/test/SVMSpokePoolClient.fills.ts
@@ -1,4 +1,4 @@
-import { KeyPairSigner, address } from "@solana/kit";
+import { KeyPairSigner, address, fetchEncodedAccount } from "@solana/kit";
 import { CHAIN_IDs } from "@across-protocol/constants";
 import { SvmSpokeClient } from "@across-protocol/contracts";
 import { intToU8Array32 } from "@across-protocol/contracts/dist/src/svm/web3-v1";
@@ -12,6 +12,8 @@ import {
   formatRelayData,
   mintTokens,
   sendRequestSlowFill,
+  closeFillPda,
+  setCurrentTime,
 } from "./utils/svm/utils";
 import { SVM_DEFAULT_ADDRESS, findFillEvent, getRandomSvmAddress } from "../src/arch/svm";
 import { SvmSpokePoolClient } from "../src/clients";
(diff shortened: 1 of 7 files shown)


test/utils/svm/utils.ts
```diff
@@ -364,6 +365,12 @@ export const setCurrentTime = async (signer: KeyPairSigner, solanaClient: RpcCli
   );
 };
 
+export const getCurrentTime = async (solanaClient: RpcClient) => {
+  const statePda = await getStatePda(SvmSpokeClient.SVM_SPOKE_PROGRAM_ADDRESS);
```
md0x, line 369: The SVM Spoke now uses a test feature that retrieves the time from the state account, rather than the actual on-chain time. See [get_current_time](https://github.com/across-protocol/contracts/blob/da166638625b01851cd998920dd3acebc81dfb2d/programs/svm-spoke/src/utils/testable_utils.rs#L37-L47).

test/SVMSpokePoolClient.fills.ts
```diff
@@ -201,4 +203,36 @@ describe("SVMSpokePoolClient: Fills", function () {
...
+
+    await setCurrentTime(signer, solanaClient, fillRelayData.fillDeadline + 1);
+
+    await closeFillPda(signer, solanaClient, fillInput.fillStatus);
+
+    const fillStatusAccount = await fetchEncodedAccount(solanaClient.rpc, fillInput.fillStatus, {
+      commitment: "confirmed",
+    });
+
+    expect(fillStatusAccount.exists).to.be.false;
```
melisaguevara, line 236: Could we add another check here calling relayFillStatus at the end to be sure that it continues to return Filled even if the fillStatusPDA is closed?
Something like:
```suggestion
    expect(fillStatusAccount.exists).to.be.false;
    
    const fillStatusWithPdaClosed = await spokePoolClient.relayFillStatus(formattedRelayData);
    expect(fillStatusWithPdaClosed).to.equal(FillStatus.Filled);
```

melisaguevara (approved):
looks good!
Only one suggestion

test/utils/svm/validator.setup.ts
```diff
@@ -1,19 +1,58 @@
...
 import fs from "node:fs/promises";
-import { SvmSpokeClient } from "@across-protocol/contracts";
+import path from "node:path";
+import tar from "tar";
 
+// Directory and file constants
 const LEDGER_DIR = path.resolve(__dirname, "..", ".ledger");
+const TARGET_DIR = path.resolve(__dirname, "..", "..", "..");
+const SVM_SPOKE_SO_PATH = path.resolve(TARGET_DIR, "target", "deploy", "svm_spoke.so");
+const BINARY_RELEASE_TAG = process.env.SVM_BINARY_RELEASE_TAG || "v4.0.11";
```
Reinis-FRP, line 12: might be better to try to resolve the installed contracts version and set the GitHub binary release tag accordingly

md0x:
> any chance we can rebase so that we don't have the tonne of commits

@james-a-morris There are a bunch of conflicts and file renames when rebasing, though. All the commits will get squashed into one when we merge anyway.

test/utils/svm/validator.setup.ts
md0x: @Reinis-FRP , I’ll follow your suggestion in a follow-up PR since that’ll need some logic updates now that we don’t have the enableRoutes function and deposit and fill both require a new delegate PDA.

Reinis-FRP (approved):

PR #1050, opened 2025-05-16
feat: SVM BlockFinder 

bmzig:
Basically a clone of the EVM Block finder in implementation, but the notable thing about this is **this block finder returns blocks, not slots**. This is because the information of a block, such as its number, timestamp, and hash, are not all available when we just return a slot. It should be assumed that everything in this block finder relates to blocks and not slots.

### DIFF
--- a/src/arch/svm/BlockUtils.ts
+++ b/src/arch/svm/BlockUtils.ts
@@ -14,63 +14,21 @@ import { SVMProvider } from "./";
 
 interface SVMBlock extends Block {}
 
-const defaultHighBlockOffset = 10;
-
-const defaultBlockRange = 120;
-const cacheTTL = 60 * 15;
-const blockTimes: { [chainId: number]: BlockTimeAverage } = {};
+const now = getCurrentTime();
+const averageBlockTimes: { [chainId: number]: BlockTimeAverage } = {
+  [CHAIN_IDs.SOLANA]: { average: 0.4, timestamp: now, blockRange: 1 },
+};
 
 /**
  * @description Compute the average block time over a block range.
  * @returns Average number of seconds per block.
  */
 export async function averageBlockTime(
-  provider: SVMProvider,
-  { highBlock, highBlockOffset, blockRange }: Opts = {}
+  _provider: SVMProvider,
+  _opts: Opts = {}
 ): Promise<Pick<BlockTimeAverage, "average" | "blockRange">> {
   // @todo This may need to be expanded to work without assuming that chainId = CHAIN_IDs.SOLANA.
-  const chainId = CHAIN_IDs.SOLANA;
-
-  const cache = blockTimes[chainId];
-
-  const now = getCurrentTime();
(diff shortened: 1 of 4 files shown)


src/arch/svm/BlockUtils.ts
```diff
@@ -0,0 +1,173 @@
...
+    private readonly blocks: SVMBlock[] = []
+  ) {
+    super();
+  }
+
+  /**
+   * @notice Gets the latest block whose timestamp is <= the provided timestamp.
+   * @param number Timestamp timestamp to search.
+   * @param hints Optional low and high block to bound the search space.
+   */
```
melisaguevara, line 52: should these docs say "slot" instead of "block" despite the function name?
Or maybe we can add a clarifying comment as you did [here](https://github.com/across-protocol/sdk/pull/1050/files#diff-287b987d66c4d702eef88efed15abcdcdb0bb1c23eee195d6e1a9e1cc58cbb37R115)

src/arch/svm/BlockUtils.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    // Rerun this check to avoid duplicate insertion.
+    if (this.blocks[index]?.number === number) return this.blocks[index];
+    this.blocks.splice(index, 0, block); // A simple insert at index.
+    return block;
+  }
+
+  // Return the latest block, between startBlock and endBlock, whose timestamp is <= timestamp.
+  // Effectively, this is an interpolation search algorithm to minimize block requests.
+  // Note: startBlock and endBlock _must_ be different blocks.
```
melisaguevara, line 140: Same comment here.

melisaguevara (commented):
Looks good!
Just one suggestion, since the terminology around "blocks" and "slots" can sometimes be confusing, it might be helpful to ensure that the comments clearly explain whenever they are dealing with slot numbers.

src/arch/svm/BlockUtils.ts
bmzig: Good point. I've changed all comments to `slot` since they are slots. https://github.com/across-protocol/sdk/pull/1050/commits/52fc51767700a2edcbde87b9a48a732989c39f9c

melisaguevara (approved):

src/arch/svm/BlockUtils.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (this.blocks.length === 0 || this.blocks[this.blocks.length - 1].timestamp < timestamp) {
+      const block = await this.getLatestBlock();
+      if (timestamp >= block.timestamp) return block;
+    }
+
+    // Prime the BlockFinder cache with any supplied hints.
+    // If the hint is accurate, then this will bypass the subsequent estimation.
+    await Promise.all(
+      Object.values(hints)
+        .filter((blockNumber) => isDefined(blockNumber))
```
pxrl, line 67:
```suggestion
        .filter(isDefined)
```

pxrl (approved):
LGTM!

PR #1051, opened 2025-05-19
improve(gasPriceOracle): Add basic gas price type discriminators (#1031)

pxrl:
Prompted by linking against the relayer and seeing a few type incompatibility errors.

### DIFF
--- a/src/gasPriceOracle/types.ts
+++ b/src/gasPriceOracle/types.ts
@@ -1,5 +1,5 @@
 import { type Chain, type Transport, PublicClient, FeeValuesEIP1559 } from "viem";
-import { BigNumber } from "../utils";
+import { BigNumber, isDefined } from "../utils";
 
 export type InternalGasPriceEstimate = FeeValuesEIP1559;
 export type GasPriceEstimate = EvmGasPriceEstimate | SvmGasPriceEstimate;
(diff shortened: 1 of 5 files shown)


src/gasPriceOracle/types.ts
```diff
@@ -17,3 +17,13 @@ export type SvmGasPriceEstimate = {
...
+
+export function isEVMGasPrice(gasPrice: GasPriceEstimate): gasPrice is EvmGasPriceEstimate {
+  const { maxFeePerGas, maxPriorityFeePerGas } = gasPrice as EvmGasPriceEstimate;
+  return isDefined(maxFeePerGas) && isDefined(maxPriorityFeePerGas);
+}
+
+export function isSVMGasPrice(gasPrice: GasPriceEstimate): gasPrice is SvmGasPriceEstimate {
+  const { baseFee, microLamportsPerComputeUnit } = gasPrice as SvmGasPriceEstimate;
+  return isDefined(baseFee) && isDefined(microLamportsPerComputeUnit);
+}
```
gsteenkamp89, line 29: nit: I wonder if in practice it might be more convenient to accept args of type `unknown` in these functions?

gsteenkamp89 (approved):
Looks good!
actually have one question

src/gasPriceOracle/types.ts
pxrl: Updated out of band - @gsteenkamp89 nice work on this; I really like this approach. It'll be interesting to see if we have other areas we can apply it.

PR #1052, opened 2025-05-19
improve: factor out addresses in spoke pool client

bmzig:
We make a lot of assumptions about the structure of the spoke pool client in the relayer. The SVMSpokePoolClient and EVMSpokePoolClient both extend SpokePoolClient but largely do not share a similar API (mainly due to the latter's extensive use of the `spokePool` contract). We can refactor some of this by simply storing the address as a state variable in the SpokePoolClient's class.

### DIFF
--- a/src/clients/SpokePoolClient/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient/SpokePoolClient.ts
@@ -87,7 +87,7 @@ export abstract class SpokePoolClient extends BaseAbstractClient {
   protected relayerRefundExecutions: RelayerRefundExecutionWithBlock[] = [];
   protected configStoreClient: AcrossConfigStoreClient | undefined;
   protected invalidFills: Set<string> = new Set();
-  protected spokePoolAddress: Address | undefined;
+  public spokePoolAddress: Address | undefined;
   public fills: { [OriginChainId: number]: FillWithBlock[] } = {};
 
   /**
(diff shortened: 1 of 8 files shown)


src/clients/SpokePoolClient/SpokePoolClient.ts
```diff
@@ -86,6 +87,7 @@ export abstract class SpokePoolClient extends BaseAbstractClient {
   protected relayerRefundExecutions: RelayerRefundExecutionWithBlock[] = [];
   protected configStoreClient: AcrossConfigStoreClient | undefined;
   protected invalidFills: Set<string> = new Set();
+  protected spokePoolAddress: Address | undefined;
```
pxrl, line 90: Do we want this to be protected? We do have some use cases where we grab the address via `SpokePoolClient.spokePool.address`, but we can't rely on that working for SVM. If  `address` is public then we could read it directly.

Also, is it possible to be `undefined` because we only set it in the extended (i.e. EVM/SVM) constructor? That could be annoying for any users; maybe we could have an abstract method `getAddress()` that the `SpokePoolClient` calls, but which is implemented in the EVM/SVM implementations.

```suggestion
  readonly address: Address | undefined;
```

bmzig: Good catch. That's what this PR was meant to address https://github.com/across-protocol/sdk/pull/1052/commits/71d591f2bf8c678bdee0cd83e07610f339430b33

pxrl (approved):

PR #1053, opened 2025-05-20
improve: expose more spoke pool client objects

bmzig:
We will likely need to expose more spoke pool client functions/variables to support the relayer.

### DIFF
--- a/src/arch/svm/eventsClient.ts
+++ b/src/arch/svm/eventsClient.ts
@@ -38,7 +38,7 @@ export type DepositEventFromSignature = Omit<DepositWithTime, "fromLiteChain" | 
 export type FillEventFromSignature = FillWithTime;
 
 export class SvmCpiEventsClient {
-  private rpc: SVMProvider;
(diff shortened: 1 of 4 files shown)


melisaguevara (approved):

src/arch/svm/eventsClient.ts
```diff
@@ -38,7 +38,7 @@ export type DepositEventFromSignature = Omit<DepositWithTime, "fromLiteChain" |
 export type FillEventFromSignature = FillWithTime;
 
 export class SvmCpiEventsClient {
-  private rpc: SVMProvider;
+  public rpc: SVMProvider;
```
melisaguevara, line 41: there's also a getRpc() function in this class in case that's useful.

bmzig: Nice I didn't realize that. I reverted this https://github.com/across-protocol/sdk/pull/1053/commits/3d5dfa2cc6720638eee4bf2635bbee7d8d555284

PR #1054, opened 2025-05-20
fix: parse chain id fields of SVM events as Number instead of BigNumber

melisaguevara:
This PR aligns SVM Spoke Pool event parsing with EVM by changing chain id fields from BigNumber to Number.


src/arch/svm/utils.ts
```diff
@@ -107,6 +107,10 @@ export function unwrapEventData(
   }
   // Handle BigInt
   if (typeof data === "bigint") {
+    const bigIntKeysAsNumber = ["originChainId", "destinationChainId", "repaymentChainId", "chainId"];
+    if (currentKey && bigIntKeysAsNumber.includes(currentKey)) {
+      return Number(data);
+    }
```
melisaguevara, line 113: This is similar to how this handle this on EVM [here](https://github.com/across-protocol/sdk/blob/master/src/utils/EventUtils.ts#L53-L77).

bmzig (approved):

amateima (approved):

gsteenkamp89 (approved):

PR #1064, opened 2025-05-30
fix(svm): svm gas queries

md0x:
Changes proposed in this PR:
- Fix Svm gas estimation functions
- Add unit test for `getNativeGasCost` and `getGasCosts`

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -26,19 +26,9 @@ import assert from "assert";
 import { arrayify, hexZeroPad, hexlify } from "ethers/lib/utils";
 import { Logger } from "winston";
 
-import { CHAIN_IDs } from "../../constants";
+import { SYSTEM_PROGRAM_ADDRESS } from "@solana-program/system";
 import { Deposit, DepositWithBlock, FillStatus, FillWithBlock, RelayData } from "../../interfaces";
-import {
-  BigNumber,
-  SvmAddress,
-  chainIsSvm,
-  chunk,
-  getTokenInfo,
-  isDefined,
-  isUnsafeDepositId,
-  keccak256,
-  toAddressType,
-} from "../../utils";
+import { BigNumber, SvmAddress, chainIsSvm, chunk, isUnsafeDepositId, keccak256, toAddressType } from "../../utils";
 import {
   SvmCpiEventsClient,
   createDefaultTransaction,
@@ -48,7 +38,6 @@ import {
   unwrapEventData,
 } from "./";
 import { SVMEventNames, SVMProvider } from "./types";
-import { SYSTEM_PROGRAM_ADDRESS } from "@solana-program/system";
 
 /**
  * @note: Average Solana slot duration is about 400-500ms. We can be conservative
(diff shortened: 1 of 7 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -426,41 +427,146 @@ export function createTokenAccountsInstruction(
...
-  amount: BigNumber,
-  relayer: SvmAddress,
-  spokePool: SvmAddress,
-  mintDecimals?: number
-) {
-  const [relayerTokenAccount, statePda] = await Promise.all([
-    getAssociatedTokenAddress(relayer, mint, TOKEN_PROGRAM_ADDRESS),
-    getStatePda(spokePool.toV2Address()),
-  ]);
+export const createFillInstruction = async (
```
md0x, line 438: I moved these helpers to SpokeUtils. They were originally in test/utils, but I think they make more sense here. We can use them through wrapper functions like [getFillRelayTx](https://github.com/across-protocol/sdk/blob/9543adb024421815fc1bdb182f343a1ecbcad93b/src/relayFeeCalculator/chain-queries/svmQuery.ts#L131).

src/relayFeeCalculator/chain-queries/svmQuery.ts
```diff
@@ -137,65 +130,72 @@ export class SvmQuery implements QueryInterface {
...
+      fillStatus,
+      tokenProgram: mintInfo.programAddress,
+      associatedTokenProgram: ASSOCIATED_TOKEN_PROGRAM_ADDRESS,
+      systemProgram: SYSTEM_PROGRAM_ADDRESS,
+      eventAuthority: eventAuthority,
+      program: this.spokePoolAddress.toV2Address(),
+      relayHash: relayDataHash,
+      relayData,
+      repaymentChainId: BigInt(repaymentChainId),
+      repaymentAddress: toSvmAddress(deposit.recipient),
```
Reinis-FRP, line 194: why deposit recipient is used as relayer repayment address?

md0x: Nice catch, I've added an option repaymentAddress param that defaults to the simulatedRelayer

src/relayFeeCalculator/chain-queries/svmQuery.ts
```diff
@@ -137,65 +130,72 @@ export class SvmQuery implements QueryInterface {
...
+    };
+    // Pass createRecipientAtaIfNeeded =true to the createFillInstruction function to create the recipient token account
+    // if it doesn't exist.
+    return createFillInstruction(simulatedSigner, this.provider, fillInput, mintInfo.data.decimals, true);
```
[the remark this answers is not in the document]
md0x: Is this what you meant? [Here](https://github.com/across-protocol/sdk/blob/c1dc6eef65001db51886479228e8aa2729d73e7f/src/arch/svm/SpokeUtils.ts#L460-L468) we’re creating the recipient ATA.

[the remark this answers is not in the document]
Reinis-FRP: I meant expressing the deposited lamports in terms of bridged token and adding that to bridging fee calculation. But I'm not sure if it belongs here, or in the API that frontend uses to suggest the amount of destination tokens received less bridging fee

Reinis-FRP (approved):

gsteenkamp89:
@md0x @Reinis-FRP Can we bump the version here for the next release?

PR #1067, opened 2025-05-30
scale gas to output token decimals

gsteenkamp89:


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -302,7 +302,7 @@ export class RelayFeeCalculator {
           throw error;
         }),
     ]);
-    const gasFeesInToken = nativeToToken(tokenGasCost, tokenPrice, inputTokenInfo.decimals, this.nativeTokenDecimals);
+    const gasFeesInToken = nativeToToken(tokenGasCost, tokenPrice, outputTokenInfo.decimals, this.nativeTokenDecimals);
```
pxrl, line 305: I initially thought that it was obviously wrong to use `inputToken.decimals`, but now I'm not so sure. We do typically want to communicate gas fee information in terms of the input token because that's what the depositor is paying in. It'd be good to confirm whether switching to `outputToken.decimals` is just masking the issue because we're quoting from 18 -> 6 decimals; it'd be good to double-check how it looks in the opposite direction (i.e. 6 -> 18 decimals).

PR #1069, opened 2025-06-01
fix: return relayerFeeDetails decimals in inputToken decimals 

mrice32:
This relates to the open question in https://github.com/across-protocol/frontend/pull/1612.

We should quote all fees in terms of input tokens, not output tokens.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -302,7 +302,7 @@ export class RelayFeeCalculator {
           throw error;
         }),
     ]);
-    const gasFeesInToken = nativeToToken(tokenGasCost, tokenPrice, inputTokenInfo.decimals, this.nativeTokenDecimals);
+    const gasFeesInToken = nativeToToken(tokenGasCost, tokenPrice, outputTokenInfo.decimals, this.nativeTokenDecimals);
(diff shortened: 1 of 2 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -508,14 +510,15 @@ export class RelayFeeCalculator {
...
+    const gasFeeTotal = gasFeePercent.mul(outToInDecimals(amountToRelay.toString())).div(fixedPointAdjustment);
     const capitalFeePercent = this.capitalFeePercent(
       amountToRelay,
-      token.symbol,
+      inputTokenInfo.symbol,
       deposit.originChainId.toString(),
       deposit.destinationChainId.toString()
     );
-    const capitalFeeTotal = capitalFeePercent.mul(amountToRelay).div(fixedPointAdjustment);
+    const capitalFeeTotal = capitalFeePercent.mul(outToInDecimals(amountToRelay.toString())).div(fixedPointAdjustment);
```
pxrl, line 521: wdyt about renaming `amountToRelay` to `outputAmount` ? That makes it absolutely clear that it's set in `outputTokenInfo.decimals`.

mrice32: Good call. That might be clearer.

PR #1074, opened 2025-06-04
chore: Make contract check EIP-7702 aware

pxrl:
isContractDeployedToAddress() currently mistakes EIP-7702 delegations for contracts. For most/all purposes in Across, 7702 delegation is preferred to be treated as an EOA. This change permits the caller to specify whether EIP-7702 delegations should be considered as contract code. EIP-7702 delegation is considered as a contract by default, so this is backwards compatible with the existing implementation and must requires opt-in.


src/utils/AddressUtils.ts
```diff
@@ -1,23 +1,47 @@
...
 import { BigNumber, chainIsEvm } from "./";
 
+/**
+ * Verify whether an address' bytecode resembles an EIP-7702 delegation.
+ * @param code Bytecode for a given address.
+ * @returns True if the bytecode resembles an EIP-7702 delegation, otherwise false.
+ */
+export function is7702Delegate(code: string): boolean {
+  // Sample 7702 delegation bytecode: 0xef010063c0c19a282a1b52b07dd5a65b58948a07dae32b
+  return code.length === 48 && code.startsWith("0xef0100") && isAddress(`0x${code.slice(8)}`);
```
pxrl, line 14: nb. an EOA revokes delegation by setting the delegated address to address(0x0). From what I've understood, this removes the code at the EOA and resets the code hash to keccak256(""). I haven't verified this with any test yet, though.

nicholaspai (commented):
One q on dependencies

src/utils/AddressUtils.ts
nicholaspai: According to [this](https://eips.ethereum.org/EIPS/eip-7702) it does seem like setting delegated address to 0x0 resets code hash:

> If address is 0x0000000000000000000000000000000000000000, do not write the delegation indicator. Clear the account’s code by resetting the account’s code hash to the empty code hash

nicholaspai (approved):

PR #1077, opened 2025-06-06
chore: remove protected constructor of SvmCpiEventsClient

melisaguevara:
Being forced to use the async create method of the SvmCpiEventsClient is causing some overload in the indexer. It would be easier if we could instantiate it directly resolving the required params ourselves.


pxrl (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.2.10",
+  "version": "4.2.11",
```
pxrl, line 4: @melisaguevara We published 4.2.11 yesterday so you probably want to bump this.

PR #1078, opened 2025-06-07
Convert Buffer to Uint8Array in SVM event parsing

fusmanii:
Issue: https://linear.app/uma/issue/ACX-4111/sdk-fix-depositid-and-message-formatting

`message` is being received as `Buffer`: https://solscan.io/tx/4LtMQn6UAV2M13b1JDz9m9dT7wYYm9UQ5GC8pT9pqPdXmKM73ZQBuw2EUg3CZ7xd75Ws5yi62mjT3fwwu3mSQTAq

Converting to `Uint8Array` so that it properly gets converted to hex string in `unwrapEventData`

~Also removing the double `0x` that can be encoded in the message~

### DIFF
--- a/src/arch/svm/utils.ts
+++ b/src/arch/svm/utils.ts
@@ -120,12 +120,7 @@ export function unwrapEventData(
   // Handle Uint8Array and byte arrays
   if (data instanceof Uint8Array || isUint8Array(data)) {
     const bytes = data instanceof Uint8Array ? data : new Uint8Array(data as number[]);
-    // 3078 is the prefix for 0x in hex
-    let hex = Buffer.from(bytes).toString("hex");
-    if (hex.startsWith("3078")) {
-      hex = hex.slice(2);
-    }
-    hex = "0x" + hex;
+    const hex = ethers.utils.hexlify(bytes);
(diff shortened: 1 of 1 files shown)


src/arch/svm/utils.ts
```diff
@@ -116,7 +120,12 @@ export function unwrapEventData(
   // Handle Uint8Array and byte arrays
   if (data instanceof Uint8Array || isUint8Array(data)) {
     const bytes = data instanceof Uint8Array ? data : new Uint8Array(data as number[]);
-    const hex = "0x" + Buffer.from(bytes).toString("hex");
+    // 3078 is the prefix for 0x in hex
+    let hex = Buffer.from(bytes).toString("hex");
+    if (hex.startsWith("3078")) {
+      hex = hex.slice(2);
+    }
+    hex = "0x" + hex;
```
grasphoper, line 128: @fusmanii Could we use `const hex = ethers.utils.hexlify(bytes);` here?

grasphoper: 2 other points:
- if hex starts with 3078, we'd have to trim 4 chars right?
- However! What if a valid hex starts from 3078?

bmzig (approved):

PR #1083, opened 2025-06-11
improve(SpokeUtils): Be more strict on SVM fill instructions

pxrl:
Throwing up a suggestion here after reviewing https://github.com/across-protocol/sdk/pull/1068

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -374,7 +374,11 @@ export async function findFillEvent(
  */
 export async function fillRelayInstruction(
   spokePool: SvmAddress,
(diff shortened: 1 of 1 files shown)


src/arch/evm/SpokeUtils.ts
```diff
@@ -33,6 +33,7 @@ export function populateV3Relay(
   spokePool: Contract,
   deposit: Omit<Deposit, "messageHash" | "fromLiteChain" | "toLiteChain"> & {
     recipient: EvmAddress;
+    outputToken: EvmAddress;
```
pxrl, line 36: After preparing the SVM side it occurred to me we missed narrowing on `outputToken` in the EVM implementation.

PR #1084, opened 2025-06-11
chore(SpokePool): Add ClaimedRelayerRefund and BridgedToHubPool events

melisaguevara:
We need the SpokePoolClient to support `ClaimedRelayerRefund` and `BridgedToHubPool` events for the indexer.

### DIFF
--- a/src/clients/SpokePoolClient/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient/SpokePoolClient.ts
@@ -24,6 +24,8 @@ import {
 import { duplicateEvent, sortEventsAscendingInPlace } from "../../utils/EventUtils";
 import { ZERO_ADDRESS } from "../../constants";
 import {
+  BridgedToHubPoolWithBlock,
+  ClaimedRelayerRefundWithBlock,
   Deposit,
   DepositWithBlock,
   EnabledDepositRouteWithBlock,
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient/SpokePoolClient.ts
```diff
@@ -60,14 +62,11 @@ export const knownEventNames = [
   "TokensBridged",
   "RelayedRootBundle",
   "ExecutedRelayerRefundRoot",
-  "V3FundsDeposited",
   "FundsDeposited",
-  "RequestedSpeedUpV3Deposit",
   "RequestedSpeedUpDeposit",
-  "RequestedV3SlowFill",
   "RequestedSlowFill",
-  "FilledV3Relay",
```
melisaguevara, line 69: Are we ok to delete these now that we've dropped V3 support?

src/interfaces/SpokePool.ts
```diff
@@ -133,6 +133,18 @@ export interface TokensBridged extends SortableEvent {
   l2TokenAddress: string;
 }
 
+export interface ClaimedRelayerRefundWithBlock extends SortableEvent {
+  l2TokenAddress: string;
+  refundAddress: string;
+  amount: BigNumber;
+  caller: string;
```
Reinis-FRP, line 140: SVM event does not have a caller - would that create a problem?

melisaguevara: Good catch, thought I saw it there.
I made it an optional field here: [1d7b5a9](https://github.com/across-protocol/sdk/pull/1084/commits/1d7b5a92b3049993265d28518ed5e361687a557e)

Reinis-FRP (approved):

amateima (approved):

PR #1087, opened 2025-06-19
improve: adjust methods on `Address` class

grasphoper:


src/utils/AddressUtils.ts
```diff
@@ -270,17 +228,34 @@ export class Address {
...
+    const parseRawAddress = () => {
+      const hexString = utils.hexlify(this.rawAddress);
+      const rawAddress = utils.hexZeroPad(utils.hexStripZeros(hexString), 20);
+      return utils.getAddress(rawAddress);
+    };
+    return (this.evmAddress ??= parseRawAddress());
+  }
+
+  // Override `formatAsNativeAddress` to return the 20-byte representation address.
+  override formatAsNativeAddress(): string {
```
pxrl, line 250: I'd probably also err towards having a short name for this - this function is heavily used inline so when it's long it will cause lots of line wrapping. Something as short as `EvmAddress.toNative()` would suffice IMO; naming stuff is hard :\

src/utils/AddressUtils.ts
```diff
@@ -270,17 +228,34 @@ export class Address {
...
+    if (!isValidEvmAddress(hexString)) {
       throw new Error(`${hexString} is not a valid EVM address`);
     }
   }
 
-  // Override `toAddress` to return the 20-byte representation address.
-  override toAddress(): string {
-    return this.toEvmAddress();
+  // Formats the address as a valid EVM address. This function should never throw as constructor-check will have ensured we have a valid Evm address
+  formatAsChecksummedEvmAddress(): string {
```
pxrl, line 240: idk if it's necessary to have this as a separate function - for me it's OK to have it under the "to native address" function.

src/utils/AddressUtils.ts
```diff
@@ -328,3 +327,30 @@ export class SvmAddress extends Address {
...
+    return false;
+  }
+
+  override __unsafeStaticCastToSvmAddress(): SvmAddress {
+    throw new Error("Cannot cast `InvalidAddress` to `SvmAddress`.");
+  }
+
+  override __unsafeStaticCastToEvmAddress(): EvmAddress {
+    throw new Error("Cannot cast `InvalidAddress` to `EvmAddress`.");
+  }
```
pxrl, line 355: I think we will need implementations for these because they will ultimately need to be used for inclusion in relayer refund leaves for depositor refunds, for example. So I guess the Svm component should coerce into base58 encoding, and the evm one should be to to base16 bytes32.

src/utils/TokenUtils.ts
```diff
@@ -14,7 +14,7 @@ type SignerOrProvider = providers.Provider | Signer;
 export async function fetchTokenInfo(address: string, signerOrProvider: SignerOrProvider): Promise<TokenInfo> {
   const token = new Contract(address, ERC20__factory.abi, signerOrProvider);
   const [symbol, decimals] = await Promise.all([token.symbol(), token.decimals()]);
-  return { address: toAddressType(address).forceEvmAddress(), symbol, decimals };
+  return { address: toAddressType(address).__unsafeStaticCastToSvmAddress(), symbol, decimals };
```
pxrl, line 17: This is EVM-specific code; was it intended to specify `...CastToSvmAddress()` ?

src/utils/TokenUtils.ts
```diff
@@ -114,7 +114,7 @@ export function isStablecoin(tokenSymbol: string): boolean {
 export function getTokenInfo(l2TokenAddress: string, chainId: number, tokenMapping = TOKEN_SYMBOLS_MAP): TokenInfo {
   const parsedAddress = chainIsSvm(chainId)
     ? toAddressType(l2TokenAddress).toBase58()
-    : toAddressType(l2TokenAddress).toEvmAddress();
+    : toAddressType(l2TokenAddress).__unsafeStaticCastToEvmAddress().formatAsChecksummedEvmAddress();
```
pxrl, line 117: I think this is a good use case for `toAddressType(l2TokenAddress, chainId).toNative()` (...or whatever the "to native representation" method is named. Then we can drop the local chainId evaluation.

(nb. this is however a bit of a weird place to use this, since `l2TokenAddress` is already a string, and we're just converting it back to a string. I feel like this function should probably accept an Address type instead of a string.)

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -126,7 +126,18 @@ export class QueryBase implements QueryInterface {
     deposit: Omit<Deposit, "messageHash">,
     relayer = toAddressType(getDefaultSimulatedRelayerAddress(deposit.destinationChainId), deposit.destinationChainId)
   ): Promise<PopulatedTransaction> {
-    return populateV3Relay(this.spokePool, deposit, relayer);
+    // TODO? We should confirm that these deposit fields were verified on ingestion, otherwise we will throw on casts here
+    // TODO? If anything, this `getUnsignedTxFromDeposit` function should already accept deposit & { ...fields: EvmAddress }-type argument
```
nicholaspai, line 130: have you tried changing the `deposit` type to `deposit & { ...fields: EvmAddress }`?

src/utils/AddressUtils.ts
nicholaspai: definitely agreed

nicholaspai: you probably don't need an Address in this name since its part of an XAddress class

src/utils/AddressUtils.ts
```diff
@@ -78,42 +78,40 @@ export function isValidEvmAddress(address: string): boolean {
...
-      return addressType.forceEvmAddress();
-    } catch {
-      return addressType;
+  }
+
+  try {
+    if (utils.isHexString(address)) {
+      return EvmAddress.from(address);
     }
+    return SvmAddress.from(address);
```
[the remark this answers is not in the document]
grasphoper: chainId is undefined for this part

src/utils/AddressUtils.ts
```diff
@@ -299,21 +274,45 @@ export class EvmAddress extends Address {
...
     return new this(evmAddress);
   }
+
+  override __unsafeStaticCastToSvmAddress(): SvmAddress {
+    throw new Error("Cannot cast `EvmAddress` to `SvmAddress`.");
+  }
+
+  override __unsafeStaticCastToEvmAddress(): EvmAddress {
+    return this;
+  }
```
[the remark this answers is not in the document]
grasphoper: I preferred the longer name because how do we separate the functions that return strings from the functions that cast and return a different type? For me, the answer was using `cast...` and `format...` prefixes. If we do `__unsafeToXXXAddress()` and `toNative()` , it's really hard to know from the name what the function is doing. `.toNative()` might be giving us a narrowed down type, or might be giving us a string. Same with `__unsafeToXXXAddress`

src/utils/AddressUtils.ts
grasphoper: These `__unsafe...` functions are type casts rather than formatting tools. If we want to represent a `RawAddress` as `EVM-style-20-byte-truncated-address`, we can have something like `formatForOnchainEVM` or something. Or a more general `formatForChain(chainId)`. Wdythink? That formatting function would solve a lot of the confusion between type casting versus formatting.

I don't think we should ever cast between child classes of `Address` after they're created

PR #1090, opened 2025-06-19
feat(svm): update to latest contracts and refactor SVM functions for u256 amounts

md0x:
Summary

This PR updates the SDK to use the latest SVM contract version and migrates relevant SVM logic to support u256 (as Uint8Array) for inputAmount and outputAmount.

Changes
- Bumped @across-protocol/contracts version
- Replaced u64 amount handling with u256 (Uint8Array)
- Updated utils, mock clients, and tests accordingly

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -18,6 +18,7 @@ import {
   getU32Encoder,
   getU64Encoder,
   pipe,
+  ReadonlyUint8Array,
   some,
   type TransactionSigner,
 } from "@solana/kit";
@@ -39,6 +40,7 @@ import {
 } from "../../utils";
 import {
   SvmCpiEventsClient,
+  bigToU8a32,
   createDefaultTransaction,
   getEventAuthority,
   getFillStatusPda,
@@ -447,7 +449,7 @@ export async function fillRelayInstruction(
       exclusiveRelayer,
       inputToken,
       outputToken,
-      inputAmount: relayData.inputAmount.toBigInt(),
+      inputAmount: bigToU8a32(relayData.inputAmount.toBigInt()),
       outputAmount: relayData.outputAmount.toBigInt(),
       originChainId: BigInt(relayData.originChainId),
       fillDeadline: relayData.fillDeadline,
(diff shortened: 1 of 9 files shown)


src/arch/svm/utils.ts
```diff
@@ -304,3 +305,21 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
   );
 };
+
+/**
+ * Converts a BigNumber or bigint to a Uint8Array of 32 bytes.
+ * @param bn - The BigNumber or bigint to convert.
+ * @returns The Uint8Array of 32 bytes.
+ */
+export const bigToU8a32 = (bn: BigNumber | bigint) => {
+  return new Uint8Array(intToU8Array32(Number(bn)));
```
Reinis-FRP, line 315: `Number(bn)` would loose precision for larger amounts in JS.

src/arch/svm/utils.ts
```diff
@@ -19,6 +19,7 @@ import { SvmSpokeClient } from "@across-protocol/contracts";
 import { FillType, RelayData } from "../../interfaces";
 import { BigNumber, SvmAddress, getRelayDataHash, isUint8Array } from "../../utils";
 import { EventName, SVMEventNames, SVMProvider } from "./types";
+import { intToU8Array32 } from "@across-protocol/contracts/dist/src/svm/web3-v1";
```
Reinis-FRP, line 22: Not sure we should use helper methods from contracts if they were not exported from the package root. And we need to handle `bigint` here anyway so maybe its ok to rewrite it here.

test/SVMSpokePoolClient.fills.ts
```diff
@@ -56,7 +57,7 @@ describe("SVMSpokePoolClient: Fills", function () {
       exclusiveRelayer: SVM_DEFAULT_ADDRESS,
       inputToken: address(inputToken.toBase58()),
       outputToken: mint.address,
-      inputAmount: 10,
+      inputAmount: numberToU8a32(10),
```
Reinis-FRP, line 60: we might want to use larger amounts above `u64` here in tests

src/arch/svm/SpokeUtils.ts
```diff
@@ -596,7 +598,7 @@ export function getRelayDataHash(relayData: RelayData, destinationChainId: numbe
     Uint8Array.from(addressEncoder.encode(SvmAddress.from(relayData.exclusiveRelayer, "base16").toV2Address())),
     Uint8Array.from(addressEncoder.encode(SvmAddress.from(relayData.inputToken, "base16").toV2Address())),
     Uint8Array.from(addressEncoder.encode(SvmAddress.from(relayData.outputToken, "base16").toV2Address())),
-    Uint8Array.from(uint64Encoder.encode(BigInt(relayData.inputAmount.toString()))),
+    arrayify(hexZeroPad(hexlify(relayData.inputAmount), 32)),
```
nicholaspai, line 601: Shouldn't this be the same as the `outputAmount` line on 602?

Reinis-FRP (approved):

src/arch/svm/SpokeUtils.ts
md0x: `inputAmount` is now a u8 array in the SVM Spoke
https://github.com/across-protocol/contracts/blob/8da0a9013db1f01b30cea125245d3b1ce1f7e3f3/programs/svm-spoke/src/common/relay_data.rs#L10

PR #1092, opened 2025-06-19
chore: Frontload SVM Address class changes

pxrl:
...in advance of making Address types mandatory across the codebase.

### DIFF
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -1,7 +1,8 @@
 import { providers, utils } from "ethers";
+import assert from "assert";
 import bs58 from "bs58";
 import { Address as V2Address } from "@solana/kit";
-import { BigNumber, chainIsEvm } from "./";
+import { BigNumber, chainIsEvm, isDefined } from "./";
 
 /**
  * Checks if a contract is deployed at the given address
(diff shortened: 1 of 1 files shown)


src/arch/svm/utils.ts
```diff
@@ -42,6 +42,13 @@ export async function isDevnet(rpc: SVMProvider): Promise<boolean> {
   return genesisHash === "EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG";
 }
 
+/**
+ * Small utility to convert an Address to a Solana Kit branded type.
+ */
+export function toAddress(address: EvmAddress | SvmAddress): Address<string> {
```
pxrl, line 48: Not a huge fan of this naming, but not sure what else to call it :shrug:

It's only used in a couple of places, so I figure we can easily change it if we think of something better.

src/relayFeeCalculator/chain-queries/svmQuery.ts
```diff
@@ -134,38 +137,51 @@ export class SvmQuery implements QueryInterface {
...
+    const [recipientAta, relayerAta, fillStatus, eventAuthority] = await Promise.all([
+      getAssociatedTokenAddress(
+        toAddressType(deposit.recipient, destinationChainId).forceSvmAddress(),
+        _mint,
+        mintInfo.programAddress
+      ),
+      getAssociatedTokenAddress(SvmAddress.from(relayer.toBase58()), _mint, mintInfo.programAddress),
+      getFillStatusPda(program, deposit, destinationChainId),
+      getEventAuthority(),
+    ]);
```
pxrl, line 177: Opportunistic refactor to parallelise some independent async calls.

src/arch/svm/utils.ts
grasphoper: Yeah I don't mind it either as the scope of the function is limited. I think any more descriptive name will be long :D `toSolanaSDKAddress` could be one

src/relayFeeCalculator/chain-queries/svmQuery.ts
```diff
@@ -134,38 +137,51 @@ export class SvmQuery implements QueryInterface {
     repaymentChainId = deposit.destinationChainId,
     repaymentAddress = getDefaultSimulatedRelayerAddress(deposit.destinationChainId)
   ) {
-    const toSvmAddress = (address: string) => toAddressType(address).forceSvmAddress().toV2Address();
-    const relayer = _relayer ? toAddressType(_relayer).forceSvmAddress() : this.simulatedRelayerAddress;
-    const state = await getStatePda(this.spokePoolAddress.toV2Address());
-    const _relayDataHash = getRelayDataHash(deposit, deposit.destinationChainId);
+    const toSvmAddress = (address: string, chainId: number) =>
+      toAddress(toAddressType(address, chainId).forceSvmAddress());
```
grasphoper, line 141: I'm not a huge fan of calling `.forceSvmAddress()` like this because it will always succeed even if address string is e.g. 20 bytes or 5 bytes or 25 bytes. I'd rather use `instanceof` after `toAddressType` to see that the conversion suceeded. Wdythink?

src/utils/AddressUtils.ts
```diff
@@ -220,6 +225,14 @@ export class Address {
...
     }
   }
+
+  static isEvmAddress(address: Address): boolean {
+    return address instanceof EvmAddress;
+  }
+
+  static isSvmAddress(address: Address): boolean {
+    return address instanceof SvmAddress;
+  }
```
grasphoper, line 235: I found these to not be useful, as Typescript cannot infer the type after the call. I found myself using `instanceof` directly instead

src/arch/svm/SpokeUtils.ts
```diff
@@ -341,33 +350,32 @@ export async function findFillEvent(
...
-    deposit.depositor,
-    deposit.recipient,
-    deposit.exclusiveRelayer,
-    deposit.inputToken,
-    deposit.outputToken,
-  ].map((addr) => toAddressType(addr).forceSvmAddress());
+  assert(
+    repaymentAddress.isValidOn(repaymentChainId),
+    `Invalid repayment address for chain ${repaymentChainId}: ${repaymentAddress.toAddress()}.`
+  );
```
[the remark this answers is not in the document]
pxrl: Yeah; returning could be an option but there are layers of calling code that don't really handle that. It's also a pretty critical error to attempt this whilst mis-specifying the repaymentAddress, so I'm probably OK with failing hard on this.

src/utils/AddressUtils.ts
pxrl: We should be able to fix that by modifying the return type:

```
static isSvmAddress(address: Address): address is SvmAddress {
```

PR #1095, opened 2025-06-20
improve: address a todo and remove some functions from Address class

grasphoper:


src/utils/AddressUtils.ts
```diff
@@ -153,34 +151,19 @@ export class Address {
...
-    } catch {
-      return false;
-    }
-  }
-
   // Checks if the address is valid on the given chain ID.
   isValidOn(chainId: number): boolean {
     if (chainIsEvm(chainId)) {
-      return this.isValidEvmAddress();
+      return isValidEvmAddress(this.toHexString());
```
pxrl, line 166: How about resolving the chainId to a class type X (EvmAddress, SvmAddress) and then calling `X.validate(this.rawAddress)`. 

In general I'm not a fan of `isValidEvmAddress()` because it involves a lot of conversions to infer whether or not the address is valid. I think we should be able to do that based on address lengths alone, since address casing is not a consideration here.

grasphoper: I think this makes a lot of sense.

Smth like 

```
  // Checks if the address is valid on the given chain ID.
  isValidOn(chainId: number): boolean {
    return chainIsEvm(chainId) ? EvmAddress.validate(this.rawAddress) : SvmAddress.validate(this.rawAddress);
  }
```

grasphoper: @pxrl changed here [aef36b6](https://github.com/across-protocol/sdk/pull/1095/commits/aef36b6e41d21b0d7f0a65283ef29185edacf6a1)

PR #1099, opened 2025-06-24
feat(svm): handle receive message pause/unpause deposits remotely from remote admin

md0x:
Changes proposed in this PR:
Add support for receiving CCTP messages on Solana
- Implements `createReceiveMessageInstruction` and related utilities
- Adds integration test to pause/unpause deposits using a mocked CCTP message
- Sets up local test validator with cloned CCTP programs and account overrides


src/arch/svm/utils.ts
```diff
@@ -311,3 +348,280 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+ * @param attestedMessages The CCTP messages to Solana.
+ * @param signer A base signer to be converted into a Solana signer.
+ * @param simulate Whether to simulate the transaction.
+ * @param hubChainId The chain ID of the hub.
+ * @param solanaClient The Solana client.
+ * @returns A list of executed transaction signatures.
+ */
+
+export async function finalizeSvmMessages(
+  attestedMessages: AttestedCCTPMessage[],
```
grasphoper, line 555: I think here we want to minimize the amount of information per message that the caller has to pass in. `AttestedCCTPMessage` is an internal relayer repo type and can change a lot. E.g. it changed a lot when we were adding the possibility to finalize *tokenless* messages.

If we minimized the amount of info caller has to supply, working with this function is easier (e.g. we don't need to change SDK code next time we want to, say, remove a field from `AttestedCCTPMessage` as we would then be unable to construct the full struct required by this function)

I've identified these vars that this function uses for each message: `{nonce, sourceDomain, messageBytes, attestation, type: "transfer" | "message"}`, it'd be ideal if we restricted each message to that info.

I'd also suggest renaming it to something like `finalizeCCTPV1Messages`

src/arch/svm/utils.ts
```diff
@@ -311,3 +348,280 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+
+  const tokenAccount = await getAssociatedTokenAddress(SvmAddress.from(svmSigner.keyPair.publicKey.toString()), l2Usdc);
+
+  // Define accounts dependent on deposit information.
+  const [tokenPairPda] = await getProgramDerivedAddress({
+    programAddress: tokenMessengerMinter,
+    seeds: [
+      Buffer.from("token_pair"),
+      Buffer.from(String(message.sourceDomain)),
+      Buffer.from(l1Usdc.toBytes32().slice(2), "hex"),
```
grasphoper, line 520: Here as well as in other places in the repo, TS gives me errors.

They are language server errors and otherwise the code seems to build fine via `yarn build`, but it's generally nice if we fix those.

I think most are a matter of just converting from `Buffer` to `Uint8Array`
image

grasphoper (commented):
Nice! Left just 2 comments, I think we should polish these details

src/arch/svm/utils.ts
```diff
@@ -311,3 +348,280 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+
+/**
+ * Checks if a CCTP message has been processed.
+ * @param solanaClient The Solana client.
+ * @param signer The signer of the transaction.
+ * @param nonce The nonce to check.
+ * @param sourceDomain The source domain.
+ * @returns True if the message has been processed, false otherwise.
+ */
+export const hasCCTPMessageBeenProcessed = async (
```
grasphoper, line 426: Can also add `V1` to the fn name here if you end up renaming the finalizeMessages function for consistency.

Only `V1` messages have `nonce` as a number, `V2`s have `nonceHash` instead so this would work for them

src/arch/svm/utils.ts
```diff
@@ -311,3 +348,280 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+    programAddress: tokenMessengerMinter,
+    seeds: [Buffer.from("__event_authority")],
+  });
+
+  const [custodyTokenAccountPda] = await getProgramDerivedAddress({
+    programAddress: tokenMessengerMinter,
+    seeds: [Buffer.from("custody"), bs58.decode(l2Usdc.toBase58())],
+  });
+
+  const tokenAccount = await getAssociatedTokenAddress(SvmAddress.from(svmSigner.keyPair.publicKey.toString()), l2Usdc);
```
Reinis-FRP, line 512: shouldn't this be the vault owned by state PDA when receiving tokens on SVM spoke?

src/arch/svm/utils.ts
Reinis-FRP, line 555: its bit unclear to me where do we perform the filtering of `attestedMessages` - is that the responsibility of the caller? cause this function expects this to be either tokenless message targeted to `svm_spoke` or receiving tokens to the `vault`

src/arch/svm/utils.ts
```diff
@@ -311,3 +341,283 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+  nonce: number,
+  sourceDomain: number
+): Promise<boolean> => {
+  const noncePda = await getCCTPNoncePda(solanaClient, signer, nonce, sourceDomain);
+  const isNonceUsedIx = await MessageTransmitterClient.getIsNonceUsedInstruction({
+    nonce: nonce,
+    usedNonces: noncePda,
+  });
+  const parserFunction = (buf: Buffer): boolean => {
+    return buf.length == 1 && Boolean(buf[0]);
```
Reinis-FRP, line 431: nit: I meant throwing if `buf.length !== 1`

src/arch/svm/utils.ts
```diff
@@ -311,3 +350,286 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+ * @param solanaClient The Solana client.
+ * @returns A list of executed transaction signatures.
+ */
+
+export async function finalizeCCTPV1Messages(
+  attestedMessages: AttestedCCTPMessage[],
+  signer: KeyPairSigner,
+  simulate = false,
+  hubChainId = 1,
+  solanaClient: RpcClient
```
grasphoper, line 568: Related to the question in Slack about Solana Tx execution. This function accepts `RpcClient`, whereas we use `SVMProvider` everywhere instead. Is there an easy way to get one from the other?

md0x: @grasphoper I’ve switched to the SVMProvider and started sending raw transactions without tracking their inclusion, like we agreed with @pxrl, at least for the short term.

src/arch/svm/utils.ts
```diff
@@ -312,8 +350,298 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+
+    if (simulate) {
+      await solanaClient
+        .simulateTransaction(
+          getBase64EncodedWireTransaction(await signTransactionMessageWithSigners(receiveMessageIx)),
+          {
+            encoding: "base64",
+          }
+        )
+        .send();
```
grasphoper, line 626: Does the simulateTransaction call throw on simulation errors? 

I saw that it has something like `result.value.error` field, so I thought it just returned it?

On a related note, while implementing Solana-side fill execution in the relayer repo, I used a pattern that I think works really nicely.

SDK has a function to prepare a transaction completely ([getFillRelayTx](https://github.com/across-protocol/sdk/blob/e009e0aa3ef39043c420a2ec223531e5318e9d26/src/arch/svm/SpokeUtils.ts#L491)) and then the execution / simulation steps are done on the relayer side.

I know this is a departure from what I previously suggested(sorry!!), but I think that pattern works really well by giving relayer repo control over actually sending the tx, while wrapping the complexity of preparing it fully. Could we perhaps follow the same pattern here?

We could have `getCCTPV1ReceiveMessageTx` function that would handle a single `AttestedCCTPMessage` and return the full transaction (`receiveMessageIx` in the current impl)

I think this option is a sweet spot between having just the lower level `createReceiveMessageInstruction` and the complete wrapping of tx creating and execution (current impl)

src/arch/svm/types.ts
```diff
@@ -61,3 +61,14 @@ export type RpcClient = {
...
+export type AttestedCCTPMessage = {
+  nonce: number;
+  sourceDomain: number;
+  messageBytes: string;
+  attestation: string;
+  type: "transfer" | "message";
+  amount?: string;
+  mintRecipient?: string;
+  burnToken?: string;
+};
```
grasphoper, line 74: Following the approach of minimal required info, I think we should remove the `amount`, `mintRecipient`, `burnToken` fields and let the `type` be the deciding factor in deciding depositforburn / tokenless message difference

src/arch/svm/utils.ts
```diff
@@ -312,8 +350,298 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+  });
+  const getNonceIx = await MessageTransmitterClient.getGetNoncePdaInstruction({
+    messageTransmitter: messageTransmitterPda,
+    nonce,
+    sourceDomain: sourceDomain,
+  });
+
+  const parserFunction = (buf: Buffer): Address => {
+    if (buf.length === 32) {
+      return address(bs58.encode(buf));
```
grasphoper, line 408: `Buffer` / `Uint8Array` mismatch here

src/arch/svm/utils.ts
```diff
@@ -312,8 +350,298 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+    seeds: [
+      new Uint8Array(Buffer.from("token_pair")),
+      new Uint8Array(Buffer.from(String(message.sourceDomain))),
+      new Uint8Array(Buffer.from(l1Usdc.toBytes32().slice(2), "hex")),
+    ],
+  });
+
+  const [remoteTokenMessengerPda] = await getProgramDerivedAddress({
+    programAddress: tokenMessengerMinter,
+    seeds: ["remote_token_messenger", Buffer.from(String(message.sourceDomain))],
```
grasphoper, line 531: `Buffer` / `Seed` mismatch here

src/arch/svm/utils.ts
```diff
@@ -312,8 +350,298 @@ export const createDefaultTransaction = async (rpcClient: SVMProvider, signer: T
...
+      payer: signer,
+      caller: signer,
+      authorityPda,
+      messageTransmitter: messageTransmitterPda,
+      eventAuthority: eventAuthorityPda,
+      usedNonces,
+      receiver: SvmSpokeClient.SVM_SPOKE_PROGRAM_ADDRESS,
+      systemProgram: SYSTEM_PROGRAM_ADDRESS,
+      message: messageBytes,
+      attestation: Buffer.from(message.attestation.slice(2), "hex"),
```
grasphoper, line 613: `Buffer` / `ReadonlyUint8Array` mismatch in the above 2 lines

src/arch/svm/utils.ts
grasphoper: @md0x hmmmm, I didn't realize that this was just my-IDE-type-thing. I do get this on both `Cursor` and `VSCode`, but I don't get the same errors, in, say, `Zed`. Weird ...

grasphoper: @Reinis-FRP do you get the same type errors in your IDE?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.3.4",
+  "version": "4.3.5",
```
grasphoper, line 4: Should update this version after a sync up with master

grasphoper:
@md0x @Reinis-FRP I found the culprit of these red squigglies. Apparently, VSCode or Cursor will use a version of Typescript shipped with their distro by default (uses a newer TS version than we have as dev deps in `package.json`).

The simplest solution is to tell them to use a version from `node_modules` instead(then errors disappear). Done via this config in `.vscode/settings.json`:
```
{
  "typescript.tsdk": "node_modules/typescript/lib"
}
```

For this we need to add `.vscode` to `.gitignore`.

The other option is updating our dev deps, such as typescript and adjacent things + fixing a couple of Buffer <> Uint8Array mismatches. I tried it [here](https://github.com/across-protocol/sdk/compare/pablo/acx-4185-handle-receive-message-pauseunpause-deposits-remotely-from...if/upg-typescript?expand=1), seems to work well too. Having fresher dev tools is always nice

PR #1101, opened 2025-06-24
improve: Address class updates

pxrl:
A collection of updates to Address based on review & discussion:
- Convert Address.isX() to be a proper type predicate.
- Be more strict on the input types for platform-specific functions.
- Bypass Address in low-level SVM event decoding.
- Update constraints for instantiating EVM & SVM addresses.
- Prevent instantiating SVM addresses with the upper 12 bytes zeroed.
- Update parsing/rehydration of Address type from Arweave data.

### DIFF
--- a/src/clients/BundleDataClient/utils/SuperstructUtils.ts
+++ b/src/clients/BundleDataClient/utils/SuperstructUtils.ts
@@ -36,10 +36,9 @@ const BigNumberType = coerce(instance(BigNumber), union([string(), number()]), (
 const AddressInstanceSS = union([instance(EvmAddress), instance(SvmAddress), instance(RawAddress)]);
 
 const AddressType = coerce(AddressInstanceSS, string(), (value) => {
-  // The chainId for the event data is not directly available, so infer it based on the shape of the address.
-  // Addresses are posted to arweave in their native format (base16 for EVM, base58 for SVM). Infer the relevant
-  // chain family based on the address format. RawAddress will be instantiated if the address format does not
-  // match the expected family.
+  // Addresses are posted to arweave in their native format (base16 for EVM, base58 for SVM). The chainId for
+  // for the event data is not directly available, so infer it based on the shape of the address. RawAddress
+  // will be instantiated if the address format does not match the expected family.
   const chainId = value.startsWith("0x") ? CHAIN_IDs.MAINNET : CHAIN_IDs.SOLANA;
   return toAddressType(value, chainId);
 });
--- a/src/clients/BundleDataClient/utils/SuperstructUtils.ts
+++ b/src/clients/BundleDataClient/utils/SuperstructUtils.ts
@@ -15,7 +15,7 @@ import {
   type,
 } from "superstruct";
 import { UNDEFINED_MESSAGE_HASH } from "../../../constants";
-import { BigNumber, toAddressType, Address } from "../../../utils";
+import { BigNumber, toAddressType, EvmAddress, SvmAddress, RawAddress } from "../../../utils";
 
 const PositiveIntegerStringSS = pattern(string(), /\d+/);
 const Web3AddressSS = pattern(string(), /^0x[a-fA-F0-9]{40}$/);
@@ -31,7 +31,11 @@ const BigNumberType = coerce(instance(BigNumber), union([string(), number()]), (
   }
 });
 
-const AddressType = coerce(instance(Address), string(), (value) => {
+// Accept any concrete implementation of `Address` (Evm, Svm, or Raw) but avoid using the
+// abstract `Address` class directly to keep TypeScript happy.
+const AddressInstanceSS = union([instance(EvmAddress), instance(SvmAddress), instance(RawAddress)]);
+
+const AddressType = coerce(AddressInstanceSS, string(), (value) => {
   return toAddressType(value, 1); // @todo Don't merge without fixing this!
 });
 
--- a/src/clients/BundleDataClient/utils/SuperstructUtils.ts
+++ b/src/clients/BundleDataClient/utils/SuperstructUtils.ts
@@ -14,8 +14,8 @@ import {
   union,
   type,
 } from "superstruct";
-import { UNDEFINED_MESSAGE_HASH } from "../../../constants";
-import { BigNumber, toAddressType, EvmAddress, SvmAddress, RawAddress } from "../../../utils";
+import { CHAIN_IDs, UNDEFINED_MESSAGE_HASH } from "../../../constants";
+import { BigNumber, EvmAddress, RawAddress, SvmAddress, toAddressType } from "../../../utils";
 
 const PositiveIntegerStringSS = pattern(string(), /\d+/);
 const Web3AddressSS = pattern(string(), /^0x[a-fA-F0-9]{40}$/);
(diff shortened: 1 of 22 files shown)


src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -93,7 +94,12 @@ export class QueryBase implements QueryInterface {
       transport,
     } = options;
 
-    const tx = await this.getUnsignedTxFromDeposit(deposit, relayer);
+    const { recipient, outputToken, exclusiveRelayer } = deposit;
+    assert(recipient.isEVM());
+    assert(outputToken.isEVM());
+    assert(exclusiveRelayer.isEVM());
+
+    const tx = await this.getUnsignedTxFromDeposit({ ...deposit, recipient, outputToken, exclusiveRelayer }, relayer);
```
pxrl, line 102: Ideally this wouldn't be needed, but I think we can live with it for now.

A better way would be to update the `RelayFeeCalculator` interface to permit a union of types for `getGasCosts()` - i.e. such that implementations can choose to implement _either_ the EVM variant or SVM variant of the fill. There's precedent for doing that in the `GasPriceOracle` RE EVM/SVM gas price estimates.

nicholaspai: I think this is fine if we want to get this code running and be backwards compatible and then fast follow by removing that seems like a solid plan

src/utils/AddressUtils.ts
```diff
@@ -239,34 +232,24 @@ export class EvmAddress extends Address {
     );
   }
 
+  override isEVM(): this is EvmAddress {
```
mrice32, line 235: +1 -- this seems incredibly useful.

src/utils/AddressUtils.ts
```diff
@@ -207,20 +203,18 @@ export class Address {
     }
   }
 
-  static isEvmAddress(address: Address): boolean {
-    return address instanceof EvmAddress;
+  isEVM(): this is EvmAddress {
```
mrice32, line 206: +1

Should we also make isAddress a type predicate?

isAddress is a bit tricker to avoid `instanceof`, but a potential workaround would be:
```ts
export abstract class Address {
  __address_type_brand = true;
...
  static isAddress(obj: unknown): obj is Address {
    return "__address_type_brand" in obj;
  }
...
}
```

Even though it's a bit nastier, I think it's safer.

mrice32 (commented):
+1 great addition!

src/utils/AddressUtils.ts
```diff
@@ -81,15 +81,15 @@ export function isValidEvmAddress(address: string): boolean {
...
   const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
 
-  const isEvm = chainIsEvm(chainId);
-  if (isEvm && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
-  if (!isEvm && SvmAddress.validate(rawAddress)) return new SvmAddress(rawAddress);
-  return new Address(rawAddress);
+  if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
+  else if (chainIsSvm(chainId) && SvmAddress.validate(rawAddress)) return new SvmAddress(rawAddress);
+
+  return new RawAddress(rawAddress);
```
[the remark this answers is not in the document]
grasphoper: I think `new SvmAddress` / `new EvmAddress` , same as `SvmAddress.from` / `EvmAddressFrom` are the strict constructors you're looking for?

Do you have a fn like `toAddressTypeStrict` in mind instead?

src/utils/AddressUtils.ts
```diff
@@ -239,34 +232,24 @@ export class EvmAddress extends Address {
...
+  static from(address: string, encoding: "base58" | "base16" = "base16"): EvmAddress {
+    return encoding === "base16" ? new this(utils.arrayify(address)) : new this(bs58.decode(address));
   }
 }
 
 // Subclass of address which strictly deals SVM addresses. These addresses are guaranteed to be valid SVM addresses, so `toBase58` will always produce a valid Solana address.
 export class SvmAddress extends Address {
-  // @dev This property is required for Typescript typechecker to know to distinguish between `Address`, `SvmAddress` and `EvmAddress`.
-  // Otherwise it lets any of these to use in place where other is expected.
-  private readonly _brandSvmAddress!: void;
```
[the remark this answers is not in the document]
pxrl: Yeah; here's my understanding: since `_type` is declared locally in each class, and is assigned to a specific string, tsc narrows the type of the `_type` variable to be the exact string it has been assigned (`evm`, `svm` or `raw`), rather than `string`. So this should work OK as long as we ensure no 2 `Address` implementations have the same underlying `_type` string.

src/utils/AddressUtils.ts
pxrl: ooc, do you have a specific use case in mind for this? If I've understood correctly, it'd be most useful for narrowing from `unknown` or `any`. For `Address` types we typically pass them around as `Address` and have to subsequently narrow them down to the `EvmAddress` or `SvmAddress` variants, so I'm not sure where we'd actually need to use an `isAddress()` predicate.

src/clients/BundleDataClient/utils/SuperstructUtils.ts
```diff
@@ -31,8 +31,16 @@ const BigNumberType = coerce(instance(BigNumber), union([string(), number()]), (
...
+// abstract `Address` class directly to keep TypeScript happy.
+const AddressInstanceSS = union([instance(EvmAddress), instance(SvmAddress), instance(RawAddress)]);
+
+const AddressType = coerce(AddressInstanceSS, string(), (value) => {
+  // Addresses are posted to arweave in their native format (base16 for EVM, base58 for SVM). The chainId for
+  // for the event data is not directly available, so infer it based on the shape of the address. RawAddress
+  // will be instantiated if the address format does not match the expected family.
+  const chainId = value.startsWith("0x") ? CHAIN_IDs.MAINNET : CHAIN_IDs.SOLANA;
+  return toAddressType(value, chainId);
 });
```
pxrl, line 44: Just calling attention to this update, for review.

src/utils/AddressUtils.ts
```diff
@@ -275,12 +258,22 @@ export class SvmAddress extends Address {
...
-    this._brandSvmAddress;
+    this._type; // tsc noUnusedLocals appeasement.
   }
 
   static validate(rawAddress: Uint8Array): boolean {
-    return rawAddress.length === 32;
+    // Deliberately invalidate SVM addresses w/ the upper 12 bytes zeroed. These addresses are technically valid
+    // but highly improbable and are much more likely to be a mistaken interpretation of an EVM address. Err on
+    // the side of caution for the time being. Exception: Permit the zero address (i.e. for exclusiverRelayer).
+    // @todo Add test for this.
```
pxrl, line 268: To be fixed prior to merging.

pxrl: https://github.com/across-protocol/sdk/pull/1101/commits/fc55febb240c67c6982caee50d57c5c665b8d27d

src/utils/AddressUtils.ts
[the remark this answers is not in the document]
mrice32: > I think new SvmAddress / new EvmAddress , same as SvmAddress.from / EvmAddress.from are the strict constructors you're looking for? They throw on incorrect format

Good point, as long as these are ergonomic to use when reading in these types of addresses. The only issue I see is that we use chainId for the fallback one, but we require the caller to resolve the chainId themselves for the strict one.

> @mrice32 This blows up a lot in test where random chain IDs are used. For example, supplying a valid EVM address with an unrecognised chain ID forced it to be validated as an SVM address. This caused a lot of failures with the new logic to be strict on the upper 12 bytes being non-zero for SVM address.
> 
> This is however also representative of what we can reasonably expect to see in production. Reverting to RawAddress for these cases seems to work really well.

Yeah, it's a bit disappointing that the tests block us from making the asserts we want, but I get it.

src/utils/TokenUtils.ts
```diff
@@ -127,7 +129,7 @@ export function getTokenInfo(l2TokenAddress: string, chainId: number, tokenMappi
     tokenObject = tokenMapping[l1TokenSymbol as keyof typeof tokenMapping];
   }
   return {
-    address: toAddressType(l2TokenAddress, chainId),
+    address: EvmAddress.from(l2TokenAddress),
```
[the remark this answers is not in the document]
pxrl: My bad; that's totally right. I will revert.

bmzig (approved):

src/utils/AddressUtils.ts
pxrl: Dropped here: https://github.com/across-protocol/sdk/pull/1101/commits/5b6fd8cf393870057dca47751cd7b4ce054e477a

PR #1102, opened 2025-06-25
feat(SpokePoolClients): Create SpokePoolClientManager

dijanin-brat:
Create and leverage SpokePoolClientManager so we can mitigate threats from invalid chainid.

### DIFF
--- a/src/clients/BundleDataClient/BundleDataClient.ts
+++ b/src/clients/BundleDataClient/BundleDataClient.ts
@@ -59,7 +59,7 @@ import {
   verifyFillRepayment,
 } from "./utils";
 import { isEVMSpokePoolClient, isSVMSpokePoolClient } from "../SpokePoolClient";
-import { SpokePoolClientManager } from "../SpokePoolClient/SpokePoolClientManager";
+import { SpokePoolManager } from "../SpokePoolClient/SpokePoolClientManager";
 
 // max(uint256) - 1
 export const INFINITE_FILL_DEADLINE = bnUint32Max;
@@ -164,7 +164,7 @@ export class BundleDataClient {
   private arweaveDataCache: Record<string, Promise<LoadDataReturnValue | undefined>> = {};
 
   private bundleTimestampCache: Record<string, { [chainId: number]: number[] }> = {};
-  readonly spokePoolClientManager: SpokePoolClientManager;
+  readonly spokePoolClientManager: SpokePoolManager;
 
   // eslint-disable-next-line no-useless-constructor
   constructor(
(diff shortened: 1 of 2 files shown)


src/clients/SpokePoolClient/SpokePoolClientManager.ts
```diff
@@ -0,0 +1,36 @@
+import winston from "winston";
+import { SpokePoolClient } from "./SpokePoolClient";
+
+/**
+ * SpokePoolClientManager is a wrapper around spokePoolClients. We want to use wrapper almost always
+ * instead of direct access to spokePoolClients because chainId can be invalid and we want to return undefined.
+ */
+export class SpokePoolClientManager {
```
pxrl, line 8: I wasn't sure about the naming, but `SpokePoolClientManager` turns out to be a mouthful. wdyt about `SpokePoolManager` instead?

```suggestion
export class SpokePoolManager {
```

src/clients/SpokePoolClient/SpokePoolClientManager.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.spokePoolClients = spokePoolClients;
+  }
+
+  /**
+   * Retrieves a SpokePoolClient for a given chainId.
+   * @param chainId - The chainId of the spokePoolClient to retrieve.
+   * @returns SpokePoolClient | undefined
+   * @note This method returns SpokePoolClient for given chainId. If its not found, it returns undefined.
+   */
+  getSpokePoolClientByChainId(chainId: number): SpokePoolClient | undefined {
```
pxrl, line 24: As above, since this is used fairly often in multi-level nested contexts, it might be nice to keep it on the shorter side. wdyt?

```suggestion
  getClient(chainId: number): SpokePoolClient | undefined {
```

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -445,20 +451,27 @@ export class BundleDataClient {
...
+        const matchingDeposit = originSpokePoolClient.getDeposit(fill.depositId);
         const hasMatchingDeposit =
           matchingDeposit !== undefined && getRelayEventKey(fill) === getRelayEventKey(matchingDeposit);
         return hasMatchingDeposit;
       });
       await forEachAsync(fillsToCount, async (_fill) => {
-        const matchingDeposit = this.spokePoolClients[_fill.originChainId].getDeposit(_fill.depositId);
+        const originSpokePoolClient = this.spokePoolClientManager.getSpokePoolClientByChainId(_fill.originChainId);
+        const matchingDeposit = originSpokePoolClient?.getDeposit(_fill.depositId);
         assert(isDefined(matchingDeposit), "Deposit not found for fill.");
```
pxrl, line 467: This one gets a bit tricky to debug because the assert() will fire if `originSpokePoolClient` does not exist, OR if it does not contain the deposit that we were looking for. Additionally, the assert doesn't identify which origin chain or fill we were looking for.

(This was existing implementation, but wdyt about having two assert statements, and identifying the relevant chains & deposit IDs?)

```suggestion
        const originChain = getNetworkName(_fill.originChainId);
        const originSpokePoolClient = this.spokePoolClientManager.getSpokePoolClientByChainId(_fill.originChainId);
        assert(isDefined(originSpokePoolClient), `No SpokePoolClient for chain ${originChain}`);
        const matchingDeposit = originSpokePoolClient.getDeposit(_fill.depositId);
        assert(isDefined(matchingDeposit), `No ${originChain} deposit found for ${getNetworkName(_fill.destinationChainId)} fill ${_fill.depositId}`);
```

PR #1105, opened 2025-06-26
feat: Replace getValidUnfilledAmountForDeposit in SpokePoolClient

dijanin-brat:
Replace getValidUnfilledAmountForDeposit in SpokePoolClient so we don't check for amount anymore

### DIFF
--- a/src/clients/SpokePoolClient/SpokePoolClient.ts
+++ b/src/clients/SpokePoolClient/SpokePoolClient.ts
@@ -1,4 +1,5 @@
 import winston from "winston";
+import { utils as ethersUtils } from "ethers";
 import {
   AnyObject,
   BigNumber,
@@ -6,6 +7,7 @@ import {
   DefaultLogLevels,
   DepositSearchResult,
   EventSearchConfig,
+  EvmAddress,
   MAX_BIG_INT,
   MakeOptional,
   assign,
@@ -13,16 +15,13 @@ import {
   isDefined,
   getMessageHash,
   isSlowFill,
-  isValidEvmAddress,
-  isZeroAddress,
-  toEvmAddress,
   validateFillForDeposit,
-  chainIsEvm,
   chainIsProd,
   Address,
+  toAddressType,
 } from "../../utils";
 import { duplicateEvent, sortEventsAscendingInPlace } from "../../utils/EventUtils";
-import { ZERO_ADDRESS } from "../../constants";
+import { CHAIN_IDs, ZERO_ADDRESS } from "../../constants";
 import {
   BridgedToHubPoolWithBlock,
   ClaimedRelayerRefundWithBlock,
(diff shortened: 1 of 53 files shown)


src/clients/SpokePoolClient/SpokePoolClient.ts
```diff
@@ -363,6 +364,56 @@ export abstract class SpokePoolClient extends BaseAbstractClient {
     return this.depositHashesToFills[this.getDepositHash(deposit)];
   }
 
+  public getInvalidFillsForDeposit(deposit: Deposit): UnfilledDeposit {
```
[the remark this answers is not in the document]
dijanin-brat: i am okey with that as well. My idea was to have list of all fills for given deposit so we can log whatever we want, but i that we just want to know if the deposit was filled or not.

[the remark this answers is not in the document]
dijanin-brat: [d19dd7c](https://github.com/across-protocol/sdk/pull/1105/commits/d19dd7c90963fa213883ce69a7cc6af29b46ba74)

src/clients/SpokePoolClient/SpokePoolClient.ts
```diff
@@ -363,6 +363,23 @@ export abstract class SpokePoolClient extends BaseAbstractClient {
...
 
+  public getFillStatusForDeposit(deposit: Deposit): FillStatus {
+    const depositHash = this.getDepositHash(deposit);
+    const fills = this.depositHashesToFills[depositHash] ?? [];
+
+    for (const fill of fills) {
+      const validation = validateFillForDeposit(fill, deposit);
+
+      if (validation.valid) {
+        return isSlowFill(fill) ? FillStatus.RequestedSlowFill : FillStatus.Filled;
```
pxrl, line 374: @dijanin-brat Actually, one thing here - if the fill is valid then `FillStatus` is unconditionally `Filled`.

dijanin-brat: But in that case we will never return `RequestedSlowFill`. I mean, should we return it at all?

src/clients/SpokePoolClient/SpokePoolClient.ts
```diff
@@ -363,6 +363,14 @@ export abstract class SpokePoolClient extends BaseAbstractClient {
     return this.depositHashesToFills[this.getDepositHash(deposit)];
   }
 
+  public isDepositFilled(deposit: Deposit): boolean {
+    const depositHash = this.getDepositHash(deposit);
+    const fills = this.depositHashesToFills[depositHash] ?? [];
+
+    return fills.some((fill) => validateFillForDeposit(fill, deposit).valid) ? true : false;
```
pxrl, line 370:
```suggestion
    return fills.some((fill) => validateFillForDeposit(fill, deposit).valid);
```

PR #1106, opened 2025-06-26
improve: Cleanup chain-specific address type handling

pxrl:
- Don't require exclusiveRelayer to be valid on the destination chain,
  there's no requirement for this to be true.
- Refocus on RelayData as the core type for "proto" fill types, and 
  add supplemental fields as required. This is cleaner.
- Slight re-org in the SpokePool interfaces to permit reuse of some
  SpeedUp fields. There's more work to be done to clean this up, but 
  hat's deferred for now because it's unrelated to SVM work.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -114,7 +114,7 @@ export const DEFAULT_LOGGER: Logger = {
   error: (...args) => console.error(args),
 };
 
-export function getDefaultSimulatedRelayerAddress(chainId?: number) {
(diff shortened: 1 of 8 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -114,7 +114,7 @@ export const DEFAULT_LOGGER: Logger = {
   error: (...args) => console.error(args),
 };
 
-export function getDefaultSimulatedRelayerAddress(chainId?: number) {
```
pxrl, line 117: Renamed because this function had a very long name and often caused line wrapping (especially because it's often used to set default vales in function interfaces).

PR #1108, opened 2025-06-27
chore: Post-Address cleanup

pxrl:
An assortment of minor cleanups. These are submitted separately to avoid delaying merge of the Address PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.3.0",
+  "version": "4.3.1",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "files": [


src/relayFeeCalculator/chain-queries/svmQuery.ts
```diff
@@ -144,9 +136,9 @@ export class SvmQuery implements QueryInterface {
       recipient: SvmAddress;
       outputToken: SvmAddress;
     },
-    relayer = toAddressType(getDefaultRelayer(relayData.destinationChainId), relayData.destinationChainId),
+    relayer = getDefaultRelayer(relayData.destinationChainId),
     repaymentChainId = relayData.destinationChainId,
-    repaymentAddress = toAddressType(getDefaultRelayer(relayData.destinationChainId), relayData.destinationChainId)
+    repaymentAddress = getDefaultRelayer(relayData.destinationChainId)
```
pxrl, line 141: nb. this is actually pretty dangerous because it permits _any_ caller to generate an SVM fill txn without specifying their repayment address.

We could mark `getFillRelayTx()` `protected` to ensure it's only used internally for gas cost estimation. Alternatively, we could make it mandatory to specify `repaymentAddress`.

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -224,9 +224,6 @@ export class SVMSpokePoolClient extends SpokePoolClient {
         originChainId: this.chainId,
         fromLiteChain: this.isOriginLiteChain(deposit),
         toLiteChain: this.isDestinationLiteChain(deposit),
-        outputToken: isZeroAddress(deposit.outputToken)
```
bmzig, line 227: Why can we get rid of this?

src/utils/SpokeUtils.ts
```diff
@@ -72,11 +71,6 @@ export function isUnsafeDepositId(depositId: BigNumber): boolean {
   return maxSafeDepositId.lt(depositId);
 }
 
-// Determines if the input address (either a bytes32 or bytes20) is the zero address.
-export function isZeroAddress(address: Address): boolean {
-  return address.toBytes32() === ZERO_BYTES;
-}
```
pxrl, line 78: This function is redundant because it accepts `Address`, which already has an `isZeroAddress()` method.

src/relayFeeCalculator/chain-queries/svmQuery.ts
pxrl: Just wanted to flag it for comments before fixing it. I updated it here: https://github.com/across-protocol/sdk/pull/1108/commits/886b894c57c1f0d81f0d63cb2f012358098349ec

PR #1111, opened 2025-06-30
improve: parse arweave refund addresses as bytes32

bmzig:


src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -327,7 +328,9 @@ export class BundleDataClient {
               Object.entries(tokenData).map(([token, data]) => [
                 token,
                 {
-                  refunds: data.refunds,
+                  refunds: Object.fromEntries(
+                    Object.entries(data.refunds).map(([refundAddress, refund]) => [toBytes32(refundAddress), refund])
```
nicholaspai, line 332: so follow up PR will be required to add SVM address support?

PR #1115, opened 2025-07-02
feat(svm): Support resolving latest finalized slot with block

pxrl:
Required by the proposer when selecting a bundle end block range.

### DIFF
--- a/src/arch/svm/utils.ts
+++ b/src/arch/svm/utils.ts
@@ -66,7 +66,7 @@ export async function getLatestFinalizedSlotWithBlock(
   let slot = BigInt(endSlot);
   do {
     const block = await provider.getBlock(slot, opts).send();
-    if (isDefined(block)) {
+    if (isDefined(block) && [block.blockHeight, block.blockTime].every(isDefined)) {
       break;
     }
   } while (--maxLookback > 0 && --slot > 0);
--- a/src/arch/svm/utils.ts
+++ b/src/arch/svm/utils.ts
@@ -54,13 +54,18 @@ export function toAddress(address: SdkAddress): Address<string> {
  * In most cases the first-resolved slot should also have a block. Avoid making arbitrary decisions about
  * how many slots to rotate through.
  */
-export async function getLatestFinalizedSlotWithBlock(provider: SVMProvider, maxLookback = 1000): Promise<number> {
-  let slot = await provider.getSlot({ commitment: "finalized" }).send();
-  const endSlot = slot;
+export async function getLatestFinalizedSlotWithBlock(
+  provider: SVMProvider,
+  maxSlot: bigint,
+  maxLookback = 1000
+): Promise<number> {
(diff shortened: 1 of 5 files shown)


src/arch/svm/utils.ts
```diff
@@ -49,6 +49,30 @@ export function toAddress(address: SdkAddress): Address<string> {
...
+ * how many slots to rotate through.
+ */
+export async function getLatestFinalizedSlotWithBlock(provider: SVMProvider, maxLookback = 1000): Promise<number> {
+  let slot = await provider.getSlot({ commitment: "finalized" }).send();
+  const endSlot = slot;
+  const opts = { maxSupportedTransactionVersion: 0, transactionDetails: "none", rewards: false } as const;
+
+  do {
+    const block = (await provider.getBlock(slot, opts).send());
+    if (isDefined(block)) {
```
bmzig, line 64: It seems like the block can be defined but the timestamp is undefined (because of the return value [here](https://solana.com/ja/docs/rpc/http/getblock)) which is not what we want. Do you think this is sufficient, or should we specifically ensure that `block.blockTime` is defined, too?

pxrl: Great catch... I _assume_ this is because validators aren't required to submit timestamps for each slot, so there might be cases where none submit their timestamp.

Updated: https://github.com/across-protocol/sdk/pull/1115/commits/6616695371c9e212299cf34be4dfb631f22c2be6

src/clients/BundleDataClient/utils/PoolRebalanceUtils.ts
```diff
@@ -18,25 +19,47 @@ export type PoolRebalanceRoot = {
...
   enabledChains: number[]
-): number[][] {
+): Promise<number[][]> {
   // We impose a buffer on the head of the chain to increase the probability that the received blocks are final.
   // Reducing the latest block that we query also gives partially filled deposits slightly more buffer for relayers
   // to fully fill the deposit and reduces the chance that the data worker includes a slow fill payment that gets
   // filled during the challenge period.
-  const latestPossibleBundleEndBlockNumbers = chainIdListForBundleEvaluationBlockNumbers.map(
-    (chainId: number, index) =>
-      spokeClients[chainId] && Math.max(spokeClients[chainId].latestHeightSearched - endBlockBuffers[index], 0)
```
pxrl, line 35: nb. The initial precondition `spokePoolClients[chainId]` is initially dropped here - I don't see how this code can work if it ever evaluates `unknown`.

bmzig (approved):

src/clients/BundleDataClient/utils/PoolRebalanceUtils.ts
nicholaspai: Actually this is important to keep perhaps because the spoke pool client for frozen chains like 288/Boba is undefined. So we don't want to query a timestamp on such chains

nicholaspai (commented):
Just one point about the refactor.

PR #1117, opened 2025-07-03
feat: add getSlowFillRequestTx

grasphoper:
`getSlowFillRequestTx` is analogous to [`getFillRelayTx`](https://github.com/across-protocol/sdk/blob/e009e0aa3ef39043c420a2ec223531e5318e9d26/src/arch/svm/SpokeUtils.ts#L491). It prepares the full transaction for requesting a slow fill, while allowing the caller control over whether to send / simulate the prepared transaction.

Note to self: to bump package V before merging

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.3.6",
+  "version": "4.3.7",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "files": [


src/arch/svm/SpokeUtils.ts
```diff
@@ -876,7 +951,7 @@ export async function getDepositDelegatePda(
 
   const [pda] = await getProgramDerivedAddress({
     programAddress: programId,
-    seeds: [Buffer.from("delegate"), seedHash],
```
bmzig, line 879: Why make these changes?

grasphoper: Let me just revert these for now I think, and this can be an isolated `slowfillrequest` change

grasphoper: @bmzig removed here [3e98a0b](https://github.com/across-protocol/sdk/pull/1117/commits/3e98a0b4d6d09ecab41cb330c11ab4b98da35e92)

bmzig: Ah OK. I wasn't sure if it was just an extra memory allocation or not.

bmzig (approved):

PR #1127, opened 2025-07-12
improve(BlockUtils): Handle svm undefined block at slot

pxrl:


gsteenkamp89 (approved):
Thank you!

src/arch/svm/BlockUtils.ts
```diff
@@ -92,15 +93,32 @@ export class SVMBlockFinder extends BlockFinder<SVMBlock> {
...
-    const estimatedSlotTime = await this.provider.getBlockTime(latestSlot).send();
+    const { slot, timestamp } = await this.getBlockTime();
 
     // Cast the return type to an SVMBlock.
     const block: SVMBlock = {
-      timestamp: Number(estimatedSlotTime),
-      number: Number(latestSlot),
+      timestamp: Number(slot),
+      number: timestamp,
     };
```
grasphoper, line 122: It seems like we mixed up the meaning of these 2 fields.

grasphoper (approved):

PR #1128, opened 2025-07-14
feat: get next bundle start blocks optimistically

bmzig:
We only have the option to get the next bundle's start block based on the last fully executed root bundle. This means it is not possible to optimistically enqueue the next root bundle proposal during a pending root bundle's liveness period. This adds the option to the hub pool client to get the "optimistic" bundle start block numbers for a chain ID. This means we must also change `getWidestPossibleExpectedBlockRange` to support this new function, so `getWidestPossibleExpectedBlockRange` now takes a parameter `optimistic` which defaults to **false** (since it is widely used by the disputer) which, when set to true, calls this new function to get the block range.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.3.13-alpha.0",
+  "version": "4.3.13",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "files": [


nicholaspai (commented):
one q

src/clients/HubPoolClient.ts
```diff
@@ -789,6 +789,29 @@ export class HubPoolClient extends BaseAbstractClient {
...
+    // If there is no pending root bundle, then return block ranges based on the latest fully executed root bundle.
+    if (!this.hasPendingProposal()) {
+      return this.getNextBundleStartBlockNumber(chainIdList, latestMainnetBlock, chainId);
+    }
+    // We cannot normally index `this.proposedRootBundles` since a bundle there may have been previously disputed, so only index `this.proposedRootBundles`
+    // if we have a pending proposal, since this must mean that the pending root bundle is the most recent proposed root bundle.
+    const latestProposedBundle = this.proposedRootBundles[this.proposedRootBundles.length - 1];
+
+    // If there is no previous root bundle, then return 0.
+    if (!isDefined(latestValidBundle)) {
```
grasphoper, line 803: `Cannot find name 'latestValidBundle'`

src/clients/HubPoolClient.ts
```diff
@@ -789,6 +789,29 @@ export class HubPoolClient extends BaseAbstractClient {
...
 
+  // @dev Returns the start block of the next bundle assuming that if there is a currently outstanding root bundle proposal, it will pass liveness.
+  getOptimisticBundleStartBlockNumber(chainIdList: number[], latestMainnetBlock: number, chainId: number): number {
+    // If there is no pending root bundle, then return block ranges based on the latest fully executed root bundle.
+    if (!this.hasPendingProposal()) {
+      return this.getNextBundleStartBlockNumber(chainIdList, latestMainnetBlock, chainId);
+    }
+    // We cannot normally index `this.proposedRootBundles` since a bundle there may have been previously disputed, so only index `this.proposedRootBundles`
+    // if we have a pending proposal, since this must mean that the pending root bundle is the most recent proposed root bundle.
+    const latestProposedBundle = this.proposedRootBundles[this.proposedRootBundles.length - 1];
```
grasphoper, line 800: Is it at all possible to use `this.pendingRootBundle` here?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.3.12",
+  "version": "4.3.13-alpha.0",
```
nicholaspai, line 4: are you going to change this before merge?

bmzig: Just updated here. I wanted to make sure that this worked as expected in the relayer. https://github.com/across-protocol/sdk/pull/1128/commits/7ab863ee76019ec7672e2038b4ed1ddd831ab6a0

src/clients/HubPoolClient.ts
bmzig: I don't think so since it's a different type entirely from a `ProposedRootBundle` type. From [this block](https://github.com/across-protocol/sdk/blob/433b1ba1c6916d9546eb30a45e4fd27f1dc79885/src/clients/HubPoolClient.ts#L1039), I think they should be equivalent if we do have a pending proposal, however. (Obviously, that assumption breaks down if there is not a pending proposal).

nicholaspai (approved):

PR #1137, opened 2025-07-21
improve(svm): Factor out "find timestamp for slot" logic

pxrl:
For subsequent reuse in the relayer, to address ongoing runtime issues.

### DIFF
--- a/src/arch/svm/BlockUtils.ts
+++ b/src/arch/svm/BlockUtils.ts
@@ -5,7 +5,7 @@ import { isDefined } from "../../utils/TypeGuards";
 import { getCurrentTime } from "../../utils/TimeUtils";
 import { CHAIN_IDs } from "../../constants";
 import { SVMProvider } from "./";
-import { getTimestampForSlot } from "./SpokeUtils";
+import { findNearestTime } from "./utils";
 
 interface SVMBlock extends Block {}
 
@@ -95,17 +95,8 @@ export class SVMBlockFinder extends BlockFinder<SVMBlock> {
    * immediate preceding block timestamp will be used. Note that this may return an eventually-incorrect timestamp for
    * future slots.
    */
-  private async getBlockTime(_slot?: bigint): Promise<{ slot: bigint; timestamp: number }> {
-    let timestamp: number | undefined;
-    let slot = _slot ?? (await this.provider.getSlot({ commitment: "finalized" }).send());
-
-    do {
-      timestamp = await getTimestampForSlot(this.provider, slot);
-    } while (!isDefined(timestamp) && --slot);
-    assert(isDefined(timestamp), `Unable to resolve block time for SVM slot ${_slot ?? "latest"}`);
-    assert(BigInt(Number(timestamp) === timestamp), `Unexpected SVM block timestamp: ${timestamp}`);
-
-    return { slot, timestamp: Number(timestamp) };
+  private getBlockTime(_slot?: bigint): Promise<{ slot: bigint; timestamp: number }> {
+    return findNearestTime(this.provider, _slot);
   }
 
   // Grabs the most recent slot and caches it.
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -65,6 +65,7 @@ import {
 import { SvmCpiEventsClient } from "./eventsClient";
 import { SVM_NO_BLOCK_AT_SLOT, isSolanaError } from "./provider";
 import { AttestedCCTPMessage, SVMEventNames, SVMProvider } from "./types";
+import { findNearestTime } from "./utils";
 
 /**
  * @note: Average Solana slot duration is about 400-500ms. We can be conservative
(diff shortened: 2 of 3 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -816,7 +819,7 @@ export const createReceiveMessageInstruction = async (
   input: MessageTransmitterClient.ReceiveMessageInput,
   remainingAccounts: IAccountMeta<string>[]
 ) => {
-  const receiveMessageIx = await MessageTransmitterClient.getReceiveMessageInstruction(input);
+  const receiveMessageIx = MessageTransmitterClient.getReceiveMessageInstruction(input);
```
pxrl, line 822: Driveby fix - my editor was complaining.

src/arch/svm/utils.ts
```diff
@@ -56,6 +58,28 @@ export function toAddress(address: SdkAddress): Address<string> {
...
+ * @param provider SVM Provider instance.
+ * @param _slot Optional starting slot number, or the latest slot if undefined.
+ * @returns An object containing the slot number and the relevant timestamp for the block.
+ */
+export async function findNearestTime(
+  provider: SVMProvider,
+  _slot?: bigint
+): Promise<{ slot: bigint; timestamp: number }> {
+  let timestamp: number | undefined;
+  let slot = _slot ?? (await provider.getSlot({ commitment: "confirmed" }).send());
```
pxrl, line 72: This previously used `commitment: "finalized` but now uses `confirmed`. The reason for the change is that every other RPC query is tagged to the latest confirmed slot, so it's inconsistent to require finalized for timestamps.

In general we might need to centralise the confirmation definition to protect against accidentally working against different slots.

test/SVMSpokePoolClient.fills.ts
```diff
@@ -60,11 +64,11 @@ describe("SVMSpokePoolClient: Fills", function () {
...
-      depositor,
-      recipient,
-      exclusiveRelayer: toAddressType(SVM_DEFAULT_ADDRESS, CHAIN_IDs.SOLANA),
-      inputToken,
-      outputToken: toAddressType(mint.address, CHAIN_IDs.SOLANA),
+      depositor: toAddress(depositor),
+      recipient: toAddress(recipient),
+      exclusiveRelayer: toAddress(SvmAddress.from(SVM_DEFAULT_ADDRESS)),
+      inputToken: toAddress(inputToken),
+      outputToken: toAddress(SvmAddress.from(mint.address)),
```
pxrl, line 71: This change (and most others in this file) are just to address type complaints. My editor lit up like a christmas tree in this file.

test/SVMSpokePoolClient.fills.ts
```diff
@@ -221,48 +225,44 @@ describe("SVMSpokePoolClient: Fills", function () {
...
 
-    try {
-      const closePdaInstruction = await createCloseFillPdaInstruction(signer, solanaClient.rpc, fillInput.fillStatus);
-      await signAndSendTransaction(solanaClient, closePdaInstruction);
-    } catch (error) {
-      expect(error.context.logs.some((log) => log.includes("The fill deadline has not passed!"))).to.be.true;
-    }
+    // Verify that it's not possible to close the PDA at present.
+    let closePdaInstruction = await createCloseFillPdaInstruction(signer, solanaClient.rpc, fillInput.fillStatus);
+    await assertPromiseError(signAndSendTransaction(solanaClient, closePdaInstruction));
```
pxrl, line 250: Refactored this because it's not guaranteed that `signAndSendTransaction()` actually throws, in which case the subsequent `expect()` statement won't actually fire.

test/utils/svm/utils.ts
pxrl, line 1: My editor was pretty loud about type issues in this file, so I've re-worked them until it stopped complaining.

pxrl (commented):
Dropping a few introspective breadcrumbs.

src/arch/svm/utils.ts
grasphoper: It makes a bit more sense to me if we have `findNearestTime` always accept `slot: bigint` (no optionality) and let the caller decide the commitment required. Otherwise it's unclear "nearest to what" we're getting the time to.

I'm looking at the uses of it, e.g.:

```
    const [fillStatusAccount, { slot: currentSlot, timestamp }] = await Promise.all([
      fetchEncodedAccount(provider, fillStatusPda, { commitment: "confirmed" }),
      findNearestTime(provider),
    ]);
```

It'd make sense if the caller had commitment control here.

Alternatively, we could rename the function into smth like `fetchSlotTime` similar to `fetchEncodedAccount` above and have it take config that could have commitment, or could have slot number. What do you think?

src/arch/svm/BlockUtils.ts
```diff
@@ -95,17 +95,8 @@ export class SVMBlockFinder extends BlockFinder<SVMBlock> {
...
-    do {
-      timestamp = await getTimestampForSlot(this.provider, slot);
-    } while (!isDefined(timestamp) && --slot);
-    assert(isDefined(timestamp), `Unable to resolve block time for SVM slot ${_slot ?? "latest"}`);
-    assert(BigInt(Number(timestamp) === timestamp), `Unexpected SVM block timestamp: ${timestamp}`);
-
-    return { slot, timestamp: Number(timestamp) };
+  private getBlockTime(slot?: bigint): Promise<{ slot: bigint; timestamp: number }> {
+    return findNearestTime(this.provider, slot);
   }
```
grasphoper, line 100: Should we rename this to `getNearestSlotTime`? 
And if we do end up changing `findNearestTime` function, we can pass in either commitment into that, or can fetch a slot from commitment here if `slot?: bigint` is undefined

grasphoper (commented):
The change looks good, I have 2 stylistic comments

grasphoper (approved):

PR #1139, opened 2025-07-21
fix: Added space between explorer link and short url

fusmanii:
When trying to follow the explorer link the short url get concatenated resulting in a invalid url


pxrl:
> When trying to follow the explorer link the short url get concatenated resulting in a invalid url

ooc, how are you triggering this? Is this via double-click selecting past the pipe delimiter in the shell?

PR #1142, opened 2025-07-21
feat: running balance calculations for optimistic proposals

bmzig:
We need to know the pending root bundle's running balance totals, which is only retrievable by recomputing the pending root bundle's pool rebalance root.

### DIFF
--- a/src/clients/BundleDataClient/utils/DataworkerUtils.ts
+++ b/src/clients/BundleDataClient/utils/DataworkerUtils.ts
@@ -434,7 +434,8 @@ export function _buildPoolRebalanceRoot(
     });
   });
 
-  // Add to the running balance value from the last valid root bundle proposal. A root bundle proposal which is still in liveness and has not yet been executed counts as a valid proposal.
+  // Add to the running balance value from the last valid root bundle proposal for {chainId, l1Token}
+  // combination if found.
   addLastRunningBalance(latestMainnetBlock, runningBalances, clients.hubPoolClient);
 
   const leaves: PoolRebalanceLeaf[] = constructPoolRebalanceLeaves(
--- a/src/clients/BundleDataClient/utils/DataworkerUtils.ts
+++ b/src/clients/BundleDataClient/utils/DataworkerUtils.ts
@@ -1,5 +1,6 @@
 // Create a combined `refunds` object containing refunds for V2 + V3 fills
 import assert from "assert";
+import _ from "lodash";
 import {
   BundleDepositsV3,
   BundleExcessSlowFills,
@@ -21,6 +22,9 @@ import {
   toAddressType,
   getImpliedBundleBlockRanges,
   EvmAddress,
+  isDefined,
+  deleteFromJson,
+  BigNumber,
 } from "../../../utils";
 import {
   addLastRunningBalance,
@@ -118,7 +122,7 @@ export function getEndBlockBuffers(
   return chainIdListForBundleEvaluationBlockNumbers.map((chainId: number) => blockRangeEndBlockBuffer[chainId] ?? 0);
 }
 
-export async function _buildPoolRebalanceRoot(
+export async function buildPoolRebalanceRoot(
   latestMainnetBlock: number,
   mainnetBundleEndBlock: number,
   bundleV3Deposits: BundleDepositsV3,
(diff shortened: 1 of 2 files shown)


src/clients/BundleDataClient/utils/DataworkerUtils.ts
```diff
@@ -297,9 +306,62 @@ export function _buildPoolRebalanceRoot(
     });
   });
 
-  // Add to the running balance value from the last valid root bundle proposal for {chainId, l1Token}
-  // combination if found.
-  addLastRunningBalance(latestMainnetBlock, runningBalances, clients.hubPoolClient);
+  // Add to the running balance value from the last valid root bundle proposal. A root bundle proposal which is still in liveness and has not yet been executed counts as a valid proposal.
+  // If the mainnetBundleEndBlock for this pool rebalance root corresponds to the pending root bundle or some root bundle before the pending root bundle, then we can fetch running balances directly from `ExecutedRootBundle` events.
+  if (
+    clients.hubPoolClient.getPendingRootBundle()?.bundleEvaluationBlockNumbers[0] === mainnetBundleEndBlock ||
```
bmzig, line 312: The idea here is: if the proposal for which we are trying to build a pool rebalance root is either the pending or a historical proposal, then do what we normally do. Otherwise, this proposal must be "optimistic" so re-calculate the running balance by getting the pool rebalance root of the current pending proposal.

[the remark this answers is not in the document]
bmzig: Yes, you're right. That is better https://github.com/across-protocol/sdk/pull/1142/commits/bea0bc6da12f2479d9da50fc3bce79fc83603be8

bmzig: I think your other comment makes the check constant-time, so IMO keeping the same function arguments as before seems pretty nice to me.

nicholaspai (approved):

src/clients/BundleDataClient/utils/DataworkerUtils.ts
```diff
@@ -297,9 +306,59 @@ export function _buildPoolRebalanceRoot(
...
+      clients,
+      maxL1TokenCountOverride
+    );
+    // Update the running balances by adding the pending root bundle's running balance amounts to this bundle's running balances.
+    const { runningBalances: prevRunningBalances } = prevPoolRebalanceRoot;
+    Object.keys(runningBalances).forEach((repaymentChainId) => {
+      Object.keys(runningBalances[Number(repaymentChainId)]).forEach((l1TokenAddress) => {
+        const updateRunningBalanceAmount = prevRunningBalances[Number(repaymentChainId)]?.[l1TokenAddress];
+        // If the pending root bundle didn't have a running balance, fall back to the last running balance for that token and chain pair.
+        const { runningBalance: fallbackRunningBalance } = clients.hubPoolClient.getRunningBalanceBeforeBlockForChain(
```
nicholaspai, line 348: Instead of precomputing all running balances for `prevPoolRebalanceRoot` by calling `await _buildPoolRebalanceRoot`, could we instead remove that call and just use the result of `hubPoolClient.getRunningBalanceBeforeBlockForChain` in order to set the `runningBalances` value?

I guess i'm saying that I don't think we need the `prevRunningBalances` value as you have written it here, since this value will be equal to `hubPoolClient.getRunningBalanceBeforeBlockForChain`if you pass in `latestMainnetBlock`.

Maybe I'm wrong here, just a thought.

nicholaspai: yeah It hink its helpful to have a function that spits out running balances just from all the `bundle` events, and then you can choose how to add the `lastRunningBalance`

src/clients/BundleDataClient/utils/DataworkerUtils.ts
```diff
@@ -114,17 +117,96 @@ export function getEndBlockBuffers(
...
+  const { runningBalances, realizedLpFees, chainWithRefundsOnly } = _getMarginalRunningBalances(
+    mainnetBundleEndBlock,
+    bundleV3Deposits,
+    bundleFillsV3,
+    bundleSlowFillsV3,
+    unexecutableSlowFills,
+    expiredDepositsToRefundV3,
+    clients
+  );
+  addLastRunningBalance(latestMainnetBlock, runningBalances, clients.hubPoolClient);
```
nicholaspai, line 145: Do you always add the last executed running balances? I thought you only add it if the `clients.hubPoolClient.getPendingRootBundle()!.bundleEvaluationBlockNumbers[1] >= latestMainnetBundleEndBlock`? Otherwise you do all the stuff below

src/clients/BundleDataClient/utils/DataworkerUtils.ts
bmzig: Yeah this is much simpler. I did that here https://github.com/across-protocol/sdk/pull/1142/commits/f4d7d7284e6aff9ab83c61a2ff318ff025f02fd6

nicholaspai (approved):
Ok I think this should work... best way to test is to wait until a new root bundle is proposed, then run the proposer locally using the same bundle block ranges.

Then, delete the `ExecutedRootBundle` events in the HubPoolClient locally so the just-recently executed bundle doesn't look like a "validated root bundle" and I think the running balances should match.

WDYT of this testing plan? I could be wrong

src/clients/BundleDataClient/utils/DataworkerUtils.ts
```diff
@@ -114,17 +119,244 @@ export function getEndBlockBuffers(
...
+  // Get the pool rebalance root for the pending bundle so that we may account for its calculated running balances.
+  // @dev It is safe to index the hub pool client's proposed root bundles here since there is guaranteed to be a pending proposal in this code block.
+  const mostRecentProposedRootBundle = clients.hubPoolClient.getLatestProposedRootBundle();
+  const blockRangesForChains = getImpliedBundleBlockRanges(
+    clients.hubPoolClient,
+    clients.configStoreClient,
+    mostRecentProposedRootBundle
+  );
+  // We are loading data from a pending root bundle which should be well into liveness, so we want to use arweave if possible.
+  const pendingRootBundleData = await clients.bundleDataClient.loadData(
```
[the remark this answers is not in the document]
bmzig: Good idea https://github.com/across-protocol/sdk/pull/1142/commits/18af35715d9558be6f51bc88b11b69b8af4c13a0

src/clients/BundleDataClient/utils/DataworkerUtils.ts
```diff
@@ -114,17 +119,244 @@ export function getEndBlockBuffers(
...
+  Object.keys(runningBalances).forEach((_repaymentChainId) => {
+    Object.keys(runningBalances[Number(_repaymentChainId)]).forEach((_l1TokenAddress) => {
+      const repaymentChainId = Number(_repaymentChainId);
+      const l1TokenAddress = EvmAddress.from(_l1TokenAddress);
+      const pendingPoolRebalanceLeaf = leaves.find(
+        (leaf) => leaf.chainId === repaymentChainId && leaf.l1Tokens.some((l1Token) => l1Token.eq(l1TokenAddress))
+      );
+      // If the pending pool rebalance root has running balances defined, then add it to `runningBalances`.
+      if (isDefined(pendingPoolRebalanceLeaf)) {
+        const pendingLeafTokenIdx = pendingPoolRebalanceLeaf.l1Tokens.findIndex((l1Token) =>
```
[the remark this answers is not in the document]
bmzig: Yeah it shouldn't hurt to do. https://github.com/across-protocol/sdk/pull/1142/commits/18af35715d9558be6f51bc88b11b69b8af4c13a0

nicholaspai (approved):

PR #1148, opened 2025-07-24
feat(ArweaveClient): Add retry logic to Arweave client

nicholaspai:


src/caching/Arweave/ArweaveClient.ts
```diff
@@ -225,21 +233,45 @@ export class ArweaveClient {
...
+    try {
+      return request();
+    } catch (e) {
+      if (retryCount < this.retries) {
+        this.logger.debug({
+          at: "ArweaveClient:getBalance",
+          message: `Arweave request failed, retrying after waiting ${this.retryDelaySeconds} seconds: ${e}`,
+          retryCount,
+        });
+        await delay(this.retryDelaySeconds);
```
pxrl, line 246: wdyt about an exponential backoff? We have an example here: https://github.com/across-protocol/relayer/blob/fc1c174fc2cb6cf5ee25ed0d57d03e3a42897047/src/utils/ProviderUtils.ts#L175

src/caching/Arweave/ArweaveClient.ts
```diff
@@ -225,21 +233,45 @@ export class ArweaveClient {
     return this.client.wallets.jwkToAddress(this.arweaveJWT);
   }
 
+  private async _retryRequest(request: () => Promise<unknown>, retryCount: number): Promise<unknown> {
```
pxrl, line 236: ooc, can this work as a generic so we don't need the type assertion to `BigNumber` down on line 275?

src/caching/Arweave/ArweaveClient.ts
pxrl: It requires tracking how many retries, and increasing the backoff for each subsequent one via a multiplier.

In general, if we're being rate-limited more than once, a backoff may be required to get unstuck.

nicholaspai: it makes sense, i was referring to the delay function in the example you linked but that one takes ms and this takes s

here's a go: 0627ba2b7649609e872553b9e356e2bd4f123993

src/caching/Arweave/ArweaveClient.ts
nicholaspai: 690d87977daee0aa83f602ac1421e349da8850df 


good idea

pxrl (approved):
Nice!

PR #1151, opened 2025-07-25
improve: Add utility function for fetching the token programId

dijanin-brat:
Add utility function for fetching the token programId


pxrl (approved):
LGTM!

PR #1154, opened 2025-07-25
refactor(svm): Simplify getTimestampForSlot() error handling

pxrl:


src/arch/svm/SpokeUtils.ts
```diff
@@ -97,11 +101,19 @@ export async function getTimestampForSlot(provider: SVMProvider, slotNumber: big
...
+        return undefined;
+
+      case SVM_BLOCK_NOT_AVAILABLE:
+        if (retries <= 0) {
+          throw new Error(`Timeout on SVM getBlockTime() for slot ${slotNumber.toString()}`);
+        }
+        return getTimestampForSlot(provider, slotNumber, --retries);
+
+      default:
+        throw new Error(`Unhandled error on SVM getBlockTime() for slot ${slotNumber.toString()}: ${code}`);
```
nicholaspai, line 115: We should still include the full error object here right?

src/arch/svm/SpokeUtils.ts
```diff
@@ -97,11 +101,19 @@ export async function getTimestampForSlot(provider: SVMProvider, slotNumber: big
...
 
     const { __code: code } = err.context;
-    if ([SVM_NO_BLOCK_AT_SLOT].includes(code)) {
-      return undefined;
+    switch (err.context.__code) {
+      case SVM_NO_BLOCK_AT_SLOT:
+        return undefined;
+
+      case SVM_BLOCK_NOT_AVAILABLE:
+        if (retries <= 0) {
```
nicholaspai, line 109: What if the underlying call to getBlockTime also has retries, then we're nesting retries right? I kind of prefer not having any retry logic here and returning undefined. 

Also in the specific use case of trying to grab the latest block from the block finder it will just try the next lowest slot, right? So in a sense it naturally retries

pxrl: This function already returns `undefined` for the definitive "no block was _ever_ produced for this slot". Callers use this to determine whether a block exists at the given slot. This new -32004 "temporarily no block at slot (...due to timeout)" is different because it implies that there may be a block there if we retry. I think it's important for this distinction to be handled because we might otherwise do bad things like invalidate fills because we temporarily couldn't query the block where a fill occurred.

nicholaspai: wouldn't the retry provider retry as well?

src/arch/svm/SpokeUtils.ts
pxrl: Good call - the `cause` field is designed for this: https://github.com/across-protocol/sdk/pull/1154/commits/4ca82ca03539ae961888c75da0841d69e947c601

src/arch/svm/SpokeUtils.ts
nicholaspai: As discussed offline, i'm not sure we want to retry here since the RetryProvider would have already retried so adding another retry here is akin to just bumping the retry configuration param for this type of error. Perhaps we can discuss allowing the retry provider to let the caller specify a per-method+error_code retry param, or we should bump the retry param in the bots.

Regardless, I'm wary of future tech debt being added by adding retry logic here where it exists elsewhere

PR #1157, opened 2025-07-28
[WIP] improve(SpokePoolClient): Validate raw event types

pxrl:
Initial WIP to demonstrate a possible way forward.

### DIFF
--- a/src/utils/ValidatorUtils.ts
+++ b/src/utils/ValidatorUtils.ts
@@ -5,7 +5,7 @@ import { BigNumber } from "../utils";
 
 const AddressValidator = define<string>("AddressValidator", (v) => ethersUtils.isAddress(String(v)));
 const HexValidator = define<string>("HexValidator", (v) => ethersUtils.isHexString(String(v)));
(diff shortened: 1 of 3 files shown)


src/utils/ValidatorUtils.ts
```diff
@@ -5,7 +5,7 @@ import { BigNumber } from "../utils";
 
 const AddressValidator = define<string>("AddressValidator", (v) => ethersUtils.isAddress(String(v)));
 const HexValidator = define<string>("HexValidator", (v) => ethersUtils.isHexString(String(v)));
-const BigNumberValidator = define<BigNumber>("BigNumberValidator", (v) => BigNumber.isBigNumber(v));
+export const BigNumberValidator = define<BigNumber>("BigNumberValidator", (v) => BigNumber.isBigNumber(v));
 
 const V3DepositSchema = object({
```
pxrl, line 10: todo: Refactor to share some definitions with `V3DepositSchema`. It's unfortunately not composable because it defines all possible fields, but it has the advantage of being more granular in its validation. I would propose to refactor such that it is defined as a product of smaller, composable types.

PR #1159, opened 2025-07-28
feat: pass in recipient to svm cctp finalizations

bmzig:
As of right now, we can only finalize messages destined to the svm spoke. This lets us pass in a `recipient` field to cctp deposit functions (which we already have when finalizing) so that we can also finalize EOA transfers.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -1183,14 +1183,14 @@ export async function getAccountMetasForTokenlessMessage(
  * @param message The CCTP message.
  * @param hubChainId The chain ID of the hub.
  * @param tokenMessengerMinter The token messenger minter address.
- * @param recipient The ATA of the recipient address.
+ * @param recipientAta The ATA of the recipient address.
  * @returns The account metas for a deposit message.
  */
 async function getAccountMetasForDepositMessage(
   message: AttestedCCTPMessage,
   hubChainId: number,
   tokenMessengerMinter: Address,
-  recipient: SvmAddress
+  recipientAta: SvmAddress
 ): Promise<IAccountMeta<string>[]> {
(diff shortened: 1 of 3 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -1338,11 +1336,12 @@ export function finalizeCCTPV1Messages(
   solanaClient: SVMProvider,
   attestedMessages: AttestedCCTPMessage[],
   signer: KeyPairSigner,
+  recipient: SvmAddress,
   simulate = false,
```
dijanin-brat, line 1340: I think we are using this function in `test/Solana.SvmCctpMessages.ts`

We should add recepient there as well

bmzig: https://github.com/across-protocol/sdk/pull/1159/commits/2d4c5275602be2854153c308afc0f3bd0144ff4c

src/arch/svm/SpokeUtils.ts
```diff
@@ -1183,12 +1183,14 @@ export async function getAccountMetasForTokenlessMessage(
...
  * @param tokenMessengerMinter The token messenger minter address.
+ * @param recipient The ATA of the recipient address.
  * @returns The account metas for a deposit message.
  */
 async function getAccountMetasForDepositMessage(
   message: AttestedCCTPMessage,
   hubChainId: number,
-  tokenMessengerMinter: Address
+  tokenMessengerMinter: Address,
+  recipient: SvmAddress
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/sdk/pull/1159/commits/2d4c5275602be2854153c308afc0f3bd0144ff4c

dijanin-brat (approved):

PR #1168, opened 2025-07-30
improve: Create SVM relayDataHash with messageHash instead of message

dijanin-brat:
Create SVM relayDataHash with messageHash instead of message.

### DIFF
--- a/src/arch/svm/utils.ts
+++ b/src/arch/svm/utils.ts
@@ -125,7 +125,8 @@ export function parseEventData(eventData: any): any {
   }
 
   if (typeof eventData === "object") {
-    if (eventData.constructor.name === "PublicKey") {
+    const stringTag = Object.prototype.toString.call(eventData);
+    if (stringTag.includes("PublicKey")) {
       return address(eventData.toString());
     }
     if (BN.isBN(eventData)) {


pxrl (commented):
@dijanin-brat Just one general comment here - we already have `messageHash` on both `FundsDeposited` and `FilledRelay` events - does it make sense to just update the type of `RelayData` on each of these function prototypes to include `messageHash` ?

 i.e.:
```suggestion
  relayData: RelayData: & { messageHash?: string },
```

src/arch/svm/SpokeUtils.ts
```diff
@@ -257,11 +264,12 @@ export async function findDeposit(
  * @param provider - SVM provider instance.
  * @param svmEventsClient - SVM events client for querying events.
  * @param atHeight - (Optional) Specific slot number to query. Defaults to the latest confirmed slot.
+ * @param messageHash - (Optional) Hash of the message that will be used to create relayDataHash from fill.
```
pxrl, line 267: Should be OK to drop this one now :+1: 

```suggestion
```

dijanin-brat: [93ef97a](https://github.com/across-protocol/sdk/pull/1168/commits/93ef97a38007b14afa52ccf0f81806991a476941)

PR #1170, opened 2025-07-31
feat: support message relays on Solana

bmzig:
In order to support message relays on Solana, we need to deserialize the message locally and append those accounts within the message to the `remainingAccounts` of the solana-side transaction. 

Another utility is the ability to get the instructions needed to write relayData to an instruction params PDA. As it stands, it is up to the caller to verify that there is no existing instruction params PDA when the instructions returned by `getFillRelayViaInstructionParamsInstructions` are executed.


src/arch/svm/SpokeUtils.ts
```diff
@@ -613,9 +619,20 @@ export async function getFillRelayTx(
     depositId: new Uint8Array(intToU8Array32(relayData.depositId.toNumber())),
     fillDeadline: relayData.fillDeadline,
     exclusivityDeadline: relayData.exclusivityDeadline,
-    message: new Uint8Array(Buffer.from(relayData.message, "hex")),
```
bmzig, line 616: I don't see how this wasn't a bug before, which is bad, since we could fill relays before this.

src/arch/svm/utils.ts
```diff
@@ -507,6 +507,29 @@ export const getEmergencyDeleteRootBundleRootBundleId = (body: Buffer): number =
   return result.rootBundleId.toNumber();
 };
 
+/**
+ * Converts a common `RelayData` type to an SvmSpokeClient.RelayData` type. This is useful for when we need
+ * to interface directly with the SvmSpoke.
+ * @param relayData The common RelayData TS type.
+ * @returns RelayData which conforms to the typing of the SvmSpoke.
+ */
+export function toSvmRelayData(relayData: RelayData): SvmSpokeClient.RelayData {
```
bmzig, line 516: Moved this from the relayer https://github.com/across-protocol/relayer/blob/2a57f93cba62d64a82130ab2506aaabf0f29592b/src/utils/TypeUtils.ts#L18

src/arch/svm/SpokeUtils.ts
```diff
@@ -635,7 +657,105 @@ export async function getFillRelayTx(
...
+ * Creates a fill instruction with an instruction params PDA as the relayData input.
+ * @param spokePoolAddr Address of the spoke pool we're trying to fill through
+ * @param solanaClient RPC client to interact with Solana chain
+ * @param relayData RelayData instance, supplemented with destinationChainId
+ * @param signer signer associated with the relayer creating a Fill. Can be VoidSigner for gas estimation
+ * @param repaymentChainId Chain id where relayer repayment is desired
+ * @param repaymentAddress Address to which repayment will go to on repaymentChainId
+ * @returns FillRelay transaction
+ */
+export async function getIPFillRelayTx(
```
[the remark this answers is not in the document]
bmzig: `getFillRelayTx` requires `relayData` to be defined. When filling with instruction params, at least some of the relay data needs to be undefined, so this creates a fill relay tx with all undefined relay data (which is best since we would just be wasting tx size by defining e.g. the repayment chain but no relay data, see https://github.com/across-protocol/contracts/blob/ad84307d106524a3ef1e84191f04714151dc01df/programs/svm-spoke/src/instructions/fill.rs#L180).

PR #1171, opened 2025-08-01
feat(SVMProvider): Add getSlot wrapper function

nicholaspai:
We are seeing errors that look to be coming from direct calls to `provider.getSlot` so this wrapper function will better log errors that go through that function. Eventually we can add better error handling once we better understand this RPC endpoint behavior.

Secondarily, this PR adds a `logger` param to the `arch.svm.SpokeUtils` functions to log about error codes more

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -36,7 +36,7 @@ import {
 } from "@solana/kit";
 import assert from "assert";
 import { arrayify, hexZeroPad, hexlify } from "ethers/lib/utils";
-import winston, { Logger } from "winston";
+import winston from "winston";
 import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "../../constants";
 import { DepositWithBlock, FillStatus, FillWithBlock, RelayData, RelayExecutionEventInfo } from "../../interfaces";
 import {
@@ -257,6 +257,7 @@ export function getDepositIdAtBlock(_contract: unknown, _blockTag: number): Prom
 export async function findDeposit(
   eventClient: SvmCpiEventsClient,
   depositId: BigNumber,
+  logger: winston.Logger,
   slot?: bigint,
   secondsLookback = 2 * 24 * 60 * 60 // 2 days
 ): Promise<DepositWithBlock | undefined> {
@@ -266,7 +267,7 @@ export async function findDeposit(
   }
 
   const provider = eventClient.getRpc();
-  const { slot: currentSlot } = await getNearestSlotTime(provider);
+  const { slot: currentSlot } = await getNearestSlotTime(provider, logger);
 
   // If no slot is provided, use the current slot
   // If a slot is provided, ensure it's not in the future
@@ -324,6 +325,7 @@ export async function relayFillStatus(
   relayData: RelayData,
   destinationChainId: number,
   svmEventsClient: SvmCpiEventsClient,
+  logger: winston.Logger,
   atHeight?: number
 ): Promise<FillStatus> {
   assert(chainIsSvm(destinationChainId), "Destination chain must be an SVM chain");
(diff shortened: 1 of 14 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -89,22 +90,63 @@ type ProtoFill = Omit<RelayData, "recipient" | "outputToken"> & {
...
+    const { __code: code } = err.context;
+
+    logger.debug({
+      at: "getSlot",
+      message: "Caught error from getSlot()",
+      errorCode: code,
+      commitment,
+      retryAttempt,
+      maxRetries,
+    });
```
pxrl, line 125: Minor simplification - we can reduce the number of places where we `throw` by changing the if() statement.

```suggestion
    if (isSolanaError(err)) {
      const { __code: code } = err.context;
      logger.debug({
        at: "_getSlotWithRetry",
        message: "Caught error from getSlot()",
        code,
        commitment,
        retryAttempt,
        maxRetries,
      });
    }
```

src/arch/svm/SpokeUtils.ts
```diff
@@ -118,6 +160,16 @@ async function _callGetTimestampForSlotWithRetry(
...
     const slot = slotNumber.toString();
+
+    logger.debug({
+      at: "getTimestampForSlot",
+      message: "Caught error from getBlockTime()",
+      errorCode: code,
+      slot,
+      retryAttempt,
+      maxRetries,
+    });
```
pxrl, line 171: I think this will be quite noisy for errors where we already have good handling, and it'll make it harder to find the actual unhandled cases. Should we skip logging here and instead log under the `default` case?

src/clients/BaseAbstractClient.ts
```diff
@@ -72,7 +74,7 @@ export abstract class BaseAbstractClient {
       if (provider instanceof providers.Provider) {
         to = await provider.getBlockNumber();
       } else {
-        const { slot } = await getNearestSlotTime(provider);
+        const { slot } = await getNearestSlotTime(provider, this.logger);
```
pxrl, line 77: In general I think we should be doing EVM- and SVM-specific implementations of `updateSearchConfig()` downstream of this class, but that's for follow-up refactoring.

src/arch/svm/SpokeUtils.ts
nicholaspai: that's a good call sure

src/clients/BaseAbstractClient.ts
nicholaspai: Actually, I had to go with this implementation of two different `updateSearchConfig` functions in the `BaseAbstractClient` because you cannot have multiple inheritance in TS and I couldn't get the `Evm/SvmSpokePoolClient` to work because the `SpokePoolClient` needs to extend `BaseAbstractClient`.:

https://github.com/across-protocol/sdk/pull/1171/commits/ae2ff12b62c0d00f428e2205c8885a65b82b6cb7

bmzig (approved):

src/arch/svm/SpokeUtils.ts
```diff
@@ -89,22 +90,60 @@ type ProtoFill = Omit<RelayData, "recipient" | "outputToken"> & {
...
+  maxRetries = 2
+): Promise<bigint> {
+  return _callGetSlotWithRetry(provider, commitment, 0, maxRetries, logger);
+}
+
+async function _callGetSlotWithRetry(
+  provider: SVMProvider,
+  commitment: Commitment,
+  retryAttempt: number,
+  maxRetries: number,
```
pxrl, line 106: Should we drop `maxRetries` for now, since they don't actually do anything? Otherwise it's a bit misleading.

src/clients/BaseAbstractClient.ts
nicholaspai: I looked into mixins but they have tradeoffs and the resultant hybrid class isn't a normal class, I'd rather not use mixins

PR #1176, opened 2025-08-04
refactor: consolidate RelayData to Solana RelayData conversion

bmzig:
We encode to/from `RelayData` and `SvmSpokeClient.RelayDataInput` quite a bit, so we should standardize the encoding, which consequently shrinks the amount of code.


src/arch/svm/utils.ts
pxrl, line 1: @bmzig Should we add a short test for this to protect against breaking it in the future?

PR #1177, opened 2025-08-04
fix(provider): Update types & validation on QuickNode

pxrl:


src/providers/quicknode.ts
```diff
@@ -2,12 +2,13 @@ import { CHAIN_IDs, MAINNET_CHAIN_IDs as _MAINNET_CHAIN_IDs, PUBLIC_NETWORKS } f
 import { RPCTransport } from "./types";
 
 const SNOWFLAKES = {
+  [CHAIN_IDs.ARBITRUM]: "arbitrum-mainnet",
```
pxrl, line 5: Needed because we reference Arbitrum as "Arbitrum One", so it gets named "arbitrum-one-mainnet" unless an override is supplied here.

nicholaspai (approved):

PR #1181, opened 2025-08-05
feat: Add Fallback and Quorum Solana RPC Factory

nicholaspai:


src/providers/solana/fallbackRpcFactory.ts
pxrl, line 1: Is there a tldr on the hierarchy for the new SVM provider structure? Just trying to map them to the EVM side, where we have:
- RetryProvider instantiates multiple CacheProviders
- CacheProvider extends RateLimitedProvider
- RateLimitedProvider extends StaticJsonRpcProvider

pxrl (approved):
Looks good - a couple of questions and an update.

src/providers/solana/fallbackRpcFactory.ts
nicholaspai: yeah good question, and i'm curious where this documentation should live:
- FallbackFactory stores multiple CachedFactories and:
    - Checks quorum amongst them and/or
    - Provides fallback logic between them
- CachedFactory extends RetryFactory by caching the latter's results
- RetryFactory extends RateLimitFactory
- RateLimitFactory implements `@solana/kit#RpcTransport`

This hierarchy actually is parallel to the EVM provider structure, but the EVM provider RetryProvider is kind of inappropriately named because it also implements fallback and quorum logic. Let's track an example RPC request:
- The EVM RetryProvider sends multiple rpc requests out to its CachedProviders, identical to the SVM FallbackFactory
- The RetryProvider also implements fallback and retry logic for these rpc requests, identical to the FallbackFactory
- The RetryProvider tries to reach quorum amongst the resultant rpc responses
- The EVM CacheProvider caches each individual RetryProvider responses, identical to the SVM CachedFactory

One action item here is perhaps to rename the SVM `FallbackFactory` as the `QuorumFallbackFactory`

PR #1185, opened 2025-08-11
fix: account for ATA creation in Solana fee queries

bmzig:
Quite a bit of new branches here, unfortunately:

- If the recipient ATA exists, nothing should change. ELSE
    - If the output token is a regular (non-2022) token, then the cost for rent should be the minimum rent for 165 bytes. ELSE
        - Add 165 with the size of all the extensions in the token account and get the minimum rent with this new size.
        
I will test this locally before merging.

### DIFF
--- a/src/relayFeeCalculator/chain-queries/svmQuery.ts
+++ b/src/relayFeeCalculator/chain-queries/svmQuery.ts
@@ -14,7 +14,7 @@ import { Address, BigNumber, BigNumberish, SvmAddress, TransactionCostEstimate, 
 import { Logger, QueryInterface, getDefaultRelayer } from "../relayFeeCalculator";
 import { SymbolMappingType } from "./";
 import { TOKEN_PROGRAM_ADDRESS } from "@solana-program/token";
-import { TOKEN_2022_PROGRAM_ADDRESS, getTokenSize, fetchMint } from "@solana-program/token-2022";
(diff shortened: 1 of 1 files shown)


src/relayFeeCalculator/chain-queries/svmQuery.ts
```diff
@@ -76,24 +83,49 @@ export class SvmQuery implements QueryInterface {
...
+    let tokenGasCost = gasPrice;
+
+    // If the ATA does not exist, we need to factor the rent amount into the token gas cost.
+    if (!encodedAta.exists) {
+      // If the ATA is a non-2022 token, then it will always have a fixed size of 165.
+      let extensions = undefined;
+      if (tokenOwner === TOKEN_2022_PROGRAM_ADDRESS) {
+        const mint = await fetchMint(this.provider, toAddress(outputToken));
+        extensions = isSome(mint.data.extensions) ? mint.data.extensions.value : undefined;
+      }
```
pxrl, line 120: Are we able to narrow the type of `extensions` ? It's resolving to `any` for me.

bmzig: I believe this does it? https://github.com/across-protocol/sdk/pull/1185/commits/880776eb699a441158ee74773a63e7b4c8bec4f2

pxrl (approved):

PR #1186, opened 2025-08-12
improve: getFillStatusPda based on messageHash

dijanin-brat:
Create relayDataHash based on messageHash instead of message

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -862,10 +862,7 @@ export async function getAssociatedTokenAddress(
   return associatedToken;
 }
 
-export function getRelayDataHash(
-  relayData: RelayData & { messageHash: string },
-  destinationChainId: number
-): string {
+export function getRelayDataHash(relayData: RelayData & { messageHash: string }, destinationChainId: number): string {
   assert(relayData.messageHash.startsWith("0x"), "Message hash must be a hex string");
 
   const uint64Encoder = getU64Encoder();


src/arch/svm/SpokeUtils.ts
```diff
@@ -848,21 +855,24 @@ export async function getAssociatedTokenAddress(
...
 
-export function getRelayDataHash(relayData: RelayData, destinationChainId: number): string {
+export function getRelayDataHash(relayData: RelayDataWithMessageHash, destinationChainId: number): string {
   assert(relayData.message.startsWith("0x"), "Message must be a hex string");
   const uint64Encoder = getU64Encoder();
 
   const svmRelayData = toSvmRelayData(relayData);
   const relayDataEncoder = SvmSpokeClient.getRelayDataEncoder();
   const encodedRelayData = relayDataEncoder.encode(svmRelayData);
   const encodedMessage = Buffer.from(relayData.message.slice(2), "hex");
```
bmzig, line 869: I thought the point of using `RelayDataWithMessageHash` is to no longer require the object being passed in to have `message` defined? I'd actually think it would be better to re-define `RelayDataWithMessageHash` as a `Omit<RelayData, "message"> & { messageHash: string}`. Right now, it seems like if I wanted to call this function, I still need a `message` field in relay data, so there isn't really a point to add a message hash ever.

src/clients/SpokePoolClient/SpokePoolClient.ts
bmzig, line 1: Do you need to change the function interfaces from `RelayData` to `RelayDataWithMessageHash`? I don't love that we are using this in EVM, and I'd also just prefer to not have more spoke pool relay data definitions if possible (or at least, greatly limit their usage).

src/arch/evm/SpokeUtils.ts
bmzig, line 1: It's not clear to me why you need to change the evm function interfaces to take `RelayDataWithMessageHash`. Can you explain why?

dijanin-brat: [a58f2ad](https://github.com/across-protocol/sdk/pull/1186/commits/a58f2ad5ac9d30df4c59a9b5a7accbf8e857fdf7)

src/arch/svm/SpokeUtils.ts
```diff
@@ -848,21 +859,25 @@ export async function getAssociatedTokenAddress(
...
+export function getRelayDataHash(
+  relayData: Omit<RelayData, "message"> & { messageHash: string },
+  destinationChainId: number
+): string {
+  assert(relayData.messageHash.startsWith("0x"), "Message hash must be a hex string");
+
   const uint64Encoder = getU64Encoder();
 
-  const svmRelayData = toSvmRelayData(relayData);
+  const svmRelayData = toSvmRelayData({ ...relayData, message: "0x" });
```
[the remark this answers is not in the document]
dijanin-brat: [a58f2ad](https://github.com/across-protocol/sdk/pull/1186/commits/a58f2ad5ac9d30df4c59a9b5a7accbf8e857fdf7)

src/clients/SpokePoolClient/SpokePoolClient.ts
dijanin-brat: Yup, you are right. Reverted:
[a58f2ad](https://github.com/across-protocol/sdk/pull/1186/commits/a58f2ad5ac9d30df4c59a9b5a7accbf8e857fdf7)

src/arch/svm/SpokeUtils.ts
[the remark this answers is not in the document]
pxrl: ooc, why do we have to drop `message` from `relayData` and impose the hardcoded `0x`. That seems to be a fairly niche use case, but `getRelayDataHash()` should otherwise be a very general utility function.

pxrl (approved):

PR #1189, opened 2025-08-12
feat(SVMProvider): Cache getBlockTime

nicholaspai:
Currently we only cache getTransaction, we should also cache the result of getBlockTime. This PR adds internal methods to keep track of the latest `confirmed` and `finalized` blocks to determine how long to keep data in the cache for.

### DIFF
--- a/src/providers/solana/cachedRpcFactory.ts
+++ b/src/providers/solana/cachedRpcFactory.ts
@@ -1,22 +1,43 @@
-import { RpcTransport, GetTransactionApi, RpcFromTransport, SolanaRpcApiFromTransport } from "@solana/kit";
+import {
+  RpcTransport,
+  GetTransactionApi,
+  RpcFromTransport,
+  SolanaRpcApiFromTransport,
+  GetBlockTimeApi,
+} from "@solana/kit";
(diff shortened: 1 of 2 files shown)


src/providers/solana/cachedRpcFactory.ts
```diff
@@ -60,67 +95,125 @@ export class CachedSolanaRpcFactory extends SolanaClusterRpcFactory {
...
+    method: string,
+    params: unknown[] = [],
+    latestFinalizedSlot = 0,
+    latestConfirmedSlot = 0
+  ): CacheType {
+    if (method === "getBlockTime") {
+      const targetSlot = (params as Parameters<GetBlockTimeApi["getBlockTime"]>)[0];
+      if (targetSlot <= latestFinalizedSlot) {
+        return CacheType.NO_TTL;
+      } else if (targetSlot > latestFinalizedSlot && targetSlot <= latestConfirmedSlot) {
```
nicholaspai, line 211: Loading both `finalized` and `confirmed` might be overkill and we can reduce the RPC usage by only checking one of these and making a CacheType decision off that.

For example, only fetching the `finalized` block and storing all blocks < that block with NO_TTL and all blocks above that with NONE

PR #1194, opened 2025-08-18
fix: Construct relayer address with repaymentChainId

dijanin-brat:


test/SpokePoolClient.fills.ts
```diff
@@ -254,4 +264,53 @@ describe("SpokePoolClient: Fills", function () {
...
+    await fillRelay(spokePool, targetDeposit2, relayer2, {
+      repaymentChainId: destinationChainId,
+      repaymentAddress: EvmAddress.from(relayer1.address),
+    });
+    const svmAddress = SvmAddress.from(Keypair.generate().publicKey.toBase58());
+    await fillRelay(spokePool, { ...depositTemplate, depositId: depositTemplate.depositId.add(3) }, relayer2, {
+      repaymentChainId: CHAIN_IDs.SOLANA,
+      repaymentAddress: svmAddress,
+    });
+    await hre.network.provider.send("evm_mine");
```
pxrl, line 287: I _think_ this should be unnecessary since a block gets produced each time we submit a fill.

test/SpokePoolClient.fills.ts
```diff
@@ -254,4 +264,53 @@ describe("SpokePoolClient: Fills", function () {
...
+
+    let fill1 = await findFillEvent(spokePool, targetDeposit1, startBlock);
+    expect(fill1).to.not.be.undefined;
```
pxrl, line 290: V. minor, but `to.exist` should do the same thing and is simpler.

```suggestion
    expect(fill1).to.exist;
```

test/SpokePoolClient.fills.ts
```diff
@@ -254,4 +264,53 @@ describe("SpokePoolClient: Fills", function () {
...
+    fill1 = fill1!;
+    expect(fill1.repaymentChainId).to.equal(originChainId2);
+    expect(fill1.relayer.isEVM()).to.equal(true);
```
pxrl, line 293: Also v. minor

```suggestion
    expect(fill1.relayer.isEVM()).to.be.true;
```

pxrl (approved):

test/SpokePoolClient.fills.ts
dijanin-brat: [e422ff5](https://github.com/across-protocol/sdk/pull/1194/commits/e422ff558c15725fbdae1397285394bc8763ed21)

PR #1208, opened 2025-08-26
chore: Drop Aleph Zero definitions & handling

pxrl:


pxrl:
This should be OK to go in now that Aleph Zero has been disabled for more than a week.

PR #1212, opened 2025-08-28
chore: Import minimal/stripped uma/sdk functions for poolClient

pxrl:
This permits dropping @uma/sdk as a dependency. This functionality is not actively maintained by the UMA team, so it makes sense to host it locally. Anything not required directly or indirectly by the poolClient was stripped out of the uma/sdk package on import.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "@solana/kit": "^2.1.0",
     "@solana/web3.js": "^1.31.0",
     "@types/mocha": "^10.0.1",
-    "@uma/sdk": "^0.34.10",
+    "@uma/contracts-node": "^0.4.26",
     "arweave": "^1.14.4",
     "async": "^3.2.5",
     "axios": "^0.27.2",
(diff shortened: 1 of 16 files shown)


yarn.lock
pxrl, line 1: Benefit: we can drop a bunch of dependencies here.

package.json
```diff
@@ -117,7 +117,7 @@
     "@solana/kit": "^2.1.0",
     "@solana/web3.js": "^1.31.0",
     "@types/mocha": "^10.0.1",
-    "@uma/sdk": "^0.34.10",
+    "@uma/contracts-node": "^0.4.26",
```
pxrl, line 120: nb. We can ultimately propose to relocate the `poolClient` implementation to the fe, which is afaik the only place it's actually used. At that point we'd then be able to drop this dependency as well.

dohaki (commented):
I merged a version bump here https://github.com/across-protocol/sdk/pull/1213. So we need to bump the alpha version. Otherwise looking good!

dohaki (approved):

PR #1232, opened 2025-09-18
feat: add `getAuxiliaryNativeTokenCost` to `QueryInterface` 

grasphoper:
To account for native token cost, like [`value_amount` in SVM Spoke](https://github.com/across-protocol/contracts/blob/f722857a9cd5edefb8fb33c70b2904a8822725e1/programs/svm-spoke/src/utils/message_utils.rs#L15)

We don't rely on TX simulation, but rather rely on contract enforcing the fixed relayer native token spend when passing in the `value_amount` as a part of `AcrossPlusMessage`

### DIFF
--- a/src/relayFeeCalculator/chain-queries/svmQuery.ts
+++ b/src/relayFeeCalculator/chain-queries/svmQuery.ts
@@ -176,6 +176,7 @@ export class SvmQuery implements QueryInterface {
    * @notice Return the native token cost of filling a deposit beyond gas cost. If `value_amount` is specified in a message,
    * `value_amount` of SOL gets forwarded to the first Account. We account for that in Fill cost estimation
    * @param deposit RelayData associated with Deposit we're estimating for
+   * @throws If deposit.message is malformed (unable to be deserialized into `AcrossPlusMessage`)
    * @returns Native token cost
    */
   getAuxiliaryNativeTokenCost(deposit: RelayData): BigNumber {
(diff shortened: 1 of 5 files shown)


src/relayFeeCalculator/chain-queries/svmQuery.ts
```diff
@@ -156,6 +172,19 @@ export class SvmQuery implements QueryInterface {
...
+   * @notice Return the native token cost of filling a deposit beyond gas cost. If `value_amount` is specified in a message,
+   * `value_amount` of SOL gets forwarded to the first Account. We account for that in Fill cost estimation
+   * @param deposit RelayData associated with Deposit we're estimating for
+   * @returns Native token cost
+   */
+  getAuxiliaryNativeTokenCost(deposit: RelayData): BigNumber {
+    // Notice. We return `message.value_amount` here instead of simulating the Transaction. The reason is, we choose to
+    // rely hard on Solana program to protect us from not taking more than `value_amount` rather than relying on
+    // simulation. Chain state may change between simulation and execution, so simulation alone is unreliable
+    return isMessageEmpty(deposit.message) ? bnZero : BigNumber.from(deserializeMessage(deposit.message).value_amount);
```
fusmanii, line 185: should you catch malformed message here first before attempting to deserialize?

grasphoper: `deserializeMessage(deposit.message)` will throw on malformed message
This is the idea: throw and let the caller catch it. I believe this is what the neighbouring `getNativeGasCost` , `getGasCosts` do as well

Maybe I could add a `@throws` comment

grasphoper: b1aaeb4

PR #1234, opened 2025-09-19
feat: Added Plasma const

fusmanii:


pxrl:
Still pending:
- Bump contracts

PR #1243, opened 2025-09-24
improve: harden `deserializeMessage`

grasphoper:
When deserializing `AcrossPlusMessage`, add a 2nd way of deserializing the `value_amount` so that have higher confidence in `value_amount` being correct as we potentially will send that amount to the Depositor.

Supposed to protect from a scenario:
- receive messageBytes, deserialization shows low value_amount
- estimate low fee
- send messageBytes to SvmSpoke
- SvmSpoke's impl of deserialization is a bit different, so deserializes into a different value, which pulls money from relayer

Although this scenario is unlikely, this 2nd check is not costly so I think it's worth it.


nicholaspai (approved):
+1 for adding more safety

PR #1253, opened 2025-10-07
improve: move cctp try/catches downstream

bmzig:
I think we need to be stricter on the return values of `hasCCTPV1MessageBeenProcessed`. It looks like in practice, this function returns a lot of false negatives due to RPC issues like rate-limiting and other `fetch failed` errors and not because of Solana program errors. I argue that handling for the try/catch here should be done downstream.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.3.70",
+  "version": "4.3.71",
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

pxrl:
@bmzig wdyt about adding a comment here to explain the issues that we've seen? Otherwise we'll probably stumble on it again and conclude that we're missing error handling here.

PR #1409, opened 2026-04-20
fix(arweave): harden write failover and logging

nicholaspai:
## Problem
`ArweaveClient.set()` sits on the write hot path for Arweave persistence. Before this change, only the final `post()` call participated in gateway failover. `createTransaction()` and `sign()` still ran against the first gateway client, so failures like price lookup, anchor fetch, signing errors, or upstream gateway failures could fail the write before failover started. The method also logged intermediate gateway failures at `error`, which made transient failover events look like terminal write failures.

## Solution
- wrap the full write lifecycle in sequential gateway failover: `createTransaction -> sign -> post`
- retry write attempts within each gateway before failing over to the next gateway
- preserve aggregated terminal errors after all gateways are exhausted
- emit `error` only for terminal write failure, while keeping successful failover writes visible in debug logs
- add focused write-path coverage for create-time failover, post-time failover, and terminal logging

## Scope
This PR is intentionally limited to write-path behavior in `ArweaveClient.set()` and its direct tests.

### DIFF
--- a/src/caching/Arweave/ArweaveClient.ts
+++ b/src/caching/Arweave/ArweaveClient.ts
@@ -29,6 +29,7 @@ interface Gateway {
 }
 
 type WritePhase = "createTransaction" | "sign" | "post";
+type ArweavePostResponse = Awaited<ReturnType<Arweave["transactions"]["post"]>>;
 
 class ArweaveWriteError extends Error {
   constructor(
@@ -183,7 +184,7 @@ export class ArweaveClient {
           signedTransaction = createdTransaction;
         }
 
-        let result;
+        let result: ArweavePostResponse;
         try {
           result = await client.transactions.post(signedTransaction);
         } catch (error) {
--- a/src/caching/Arweave/ArweaveClient.ts
+++ b/src/caching/Arweave/ArweaveClient.ts
@@ -4,7 +4,15 @@ import { JWKInterface } from "arweave/node/lib/wallet";
 import { Struct, create } from "superstruct";
 import winston from "winston";
 import { ARWEAVE_TAG_APP_NAME, ARWEAVE_TAG_APP_VERSION, DEFAULT_ARWEAVE_STORAGE_ADDRESS } from "../../constants";
-import { BigNumber, delay, fetchWithTimeout, isDefined, jsonReplacerWithBigNumbers, toBN } from "../../utils";
+import {
+  BigNumber,
+  delay,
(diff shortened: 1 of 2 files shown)


src/caching/Arweave/ArweaveClient.ts
```diff
@@ -116,49 +153,61 @@ export class ArweaveClient {
...
-    transaction.addTag("App-Version", ARWEAVE_TAG_APP_VERSION.toString());
-    if (isDefined(topicTag)) {
-      transaction.addTag("Topic", topicTag);
-    }
+    try {
+      return await this._failoverGateways("set", async ({ client, url }, attempt) => {
+        let transaction;
+        try {
+          transaction = await client.createTransaction({ data: payload }, this.arweaveJWT);
+        } catch (error) {
```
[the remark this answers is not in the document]
nicholaspai: Fixed. `set()` now keeps a single signed transaction once create/sign succeeds and reuses that same transaction across retry and gateway failover on the `post()` path. That means a transient `post()`-side transport/proxy failure no longer mints a fresh transaction ID on the next attempt. I also tightened the failover test to assert that `createTransaction()` and `sign()` are only called once when the second gateway succeeds.

nicholaspai:
@codex

nicholaspai:
@codex

src/caching/Arweave/ArweaveClient.ts
```diff
@@ -116,49 +154,66 @@ export class ArweaveClient {
...
+          }
 
-      // Ensure that the result is successful
-      if (result.status !== 200) {
-        const message = result?.data?.error?.msg ?? "Unknown error";
-        this.logger.error({
+          signedTransaction = createdTransaction;
+        }
+
+        let result;
```
pxrl: Can we type this?

pxrl (approved):

src/caching/Arweave/ArweaveClient.ts
nicholaspai: Typed it inline. `result` now uses `Awaited<ReturnType<Arweave["transactions"]["post"]>>` directly at the declaration site, so the `post()` response keeps its SDK-derived shape without introducing a separate alias.

I also tried to run `tsc`, but this checkout is currently blocked by unrelated missing/generated `typechain` artifacts, so I kept the change minimal and isolated to this thread.

PR #1410, opened 2026-04-20
fix(arweave): harden topic reads and fallback

nicholaspai:
## Problem
The Arweave read hot path was fragile in two places.

First, `ArweaveClient.getByTopic()` relied on the SDK GraphQL transport, which is where the `maxContentLength` / transport-style failures were occurring. It also logged expected per-entry `404` misses as warnings, which inflated noise during gateway propagation delays.

Second, `BundleDataClient.loadData()` treated Arweave-assisted reads too much like a dependency. If the Arweave path threw, the bundle load could fail instead of degrading cleanly to on-chain reconstruction.

## Solution
- replace topic GraphQL lookups with a repo-owned HTTP transport
- classify per-entry `404` topic misses as debug-level noise instead of warn-level failures
- summarize unexpected per-entry topic fetch failures instead of emitting one warning per miss
- clear rejected in-memory Arweave promises so a single failed lookup does not poison later attempts in the same process
- fall back to `loadDataFromScratch()` when the Arweave-assisted path throws
- add focused coverage for 404 classification and bundle-load fallback behavior

## Scope
This PR is limited to read-path hardening for `ArweaveClient.getByTopic()` and `BundleDataClient.loadData()`.

### DIFF
--- a/src/caching/Arweave/ArweaveClient.ts
+++ b/src/caching/Arweave/ArweaveClient.ts
@@ -4,7 +4,16 @@ import { JWKInterface } from "arweave/node/lib/wallet";
 import { Struct, create } from "superstruct";
 import winston from "winston";
 import { ARWEAVE_TAG_APP_NAME, ARWEAVE_TAG_APP_VERSION, DEFAULT_ARWEAVE_STORAGE_ADDRESS } from "../../constants";
-import { BigNumber, delay, fetchWithTimeout, isDefined, jsonReplacerWithBigNumbers, toBN } from "../../utils";
+import {
+  BigNumber,
+  delay,
+  fetchWithTimeout,
+  isDefined,
+  isHttpError,
+  jsonReplacerWithBigNumbers,
+  postWithTimeout,
+  toBN,
+} from "../../utils";
 
 export interface ArweaveGatewayConfig {
   host: string;
@@ -17,6 +26,14 @@ export const DEFAULT_ARWEAVE_GATEWAYS: ArweaveGatewayConfig[] = [{ host: "arweav
 interface Gateway {
   client: Arweave;
   url: string;
+}
+
+interface GraphQLTransactionsResponse {
+  data?: {
+    transactions?: {
+      edges?: { node: { id: string } }[];
+    };
+  };
 }
 
 export class ArweaveClient {
@@ -105,6 +122,14 @@ export class ArweaveClient {
         throw e;
       }
     }
+  }
+
+  private _isNotFoundError(error: unknown): boolean {
+    if (isHttpError(error)) {
+      return error.status === 404;
+    }
+    const message = error instanceof Error ? error.message : String(error);
+    return message.includes("HTTP 404") || message.includes("404: Not Found");
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/caching/Arweave/ArweaveClient.ts
```diff
@@ -107,6 +124,14 @@ export class ArweaveClient {
     }
   }
 
+  private _isNotFoundError(error: unknown): boolean {
+    if (isHttpError(error)) {
+      return error.status === 404;
+    }
+    const message = error instanceof Error ? error.message : String(error);
+    return message.includes("HTTP 404") || message.includes("404: Not Found");
```
pxrl: Can we actually end up here? i.e. if we got a 404, in which case is it not thrown as a fetch error?

Separately, if we can end up here, thoughts on just looking for 404? It's case-insensitive, and avoids any server-side quirks.

src/caching/Arweave/ArweaveClient.ts
```diff
@@ -218,17 +243,11 @@ export class ArweaveClient {
...
-      if (!response.ok) {
-        throw new Error(`Arweave GraphQL request failed with status ${response.status}`);
-      }
-      return response;
+    const response = await this._raceGateways("getByTopic", async ({ url }) => {
+      return await postWithTimeout<GraphQLTransactionsResponse>(`${url}/graphql`, { query }, {}, {}, 20_000);
     });
 
-    const entries = response?.data?.data?.transactions?.edges ?? [];
+    const entries = response?.data?.transactions?.edges ?? [];
```
pxrl, line 297: Was this a bug?

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -384,7 +387,17 @@ export class BundleDataClient {
...
+        try {
+          arweaveData = await this.loadArweaveData(blockRangesForChains);
+        } catch (error) {
+          this.logger.warn({
+            at: "BundleDataClient#loadData",
+            message: "Failed to load bundle data from Arweave, falling back to on-chain reconstruction",
+            blockRanges: JSON.stringify(blockRangesForChains),
+            error: String(error),
+          });
+          arweaveData = undefined;
```
pxrl: Minor, but `arweaveData` should be undefined by default here. Is tsc complaining that it's never assigned? Might be better to initialise `let arweaveData: <x> | undefined` on line 388. That'd also permit dropping the `else` clause on lines 401 - 403.

src/caching/Arweave/ArweaveClient.ts
nicholaspai: Yes. The fallback string check is for the wrapped `_raceGateways()` error shape, where `this.get()` can surface something like `All Arweave gateways failed for get: ... HTTP 404 ...` instead of a raw `HttpError`. I simplified it to `/404/i` so we just key off the presence of `404` in that aggregated message.

src/caching/Arweave/ArweaveClient.ts
nicholaspai: It’s adapting to the new transport rather than fixing old behavior. `client.api.post()` returned an axios-style envelope, so the old path was `response.data.data.transactions`. `postWithTimeout()` returns the parsed GraphQL JSON body directly, so the extra `.data` layer goes away here.

src/clients/BundleDataClient/BundleDataClient.ts
nicholaspai: Adjusted this to `let arweaveData: LoadDataReturnValue | undefined` and dropped the redundant explicit `undefined` assignments.

nicholaspai:
@codex

src/clients/BundleDataClient/BundleDataClient.ts
```diff
@@ -441,11 +444,18 @@ export class BundleDataClient {
...
     // eslint-disable-next-line @typescript-eslint/no-misused-promises
     if (!this.loadDataCache[key]) {
-      let arweaveData;
+      let arweaveData: LoadDataReturnValue | undefined;
       if (attemptArweaveLoad) {
-        arweaveData = await this.loadArweaveData(blockRangesForChains);
-      } else {
-        arweaveData = undefined;
+        try {
+          arweaveData = await this.loadArweaveData(blockRangesForChains);
```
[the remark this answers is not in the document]
nicholaspai: this is fine since we won't have multiple functions calling this function from within one instance

bmzig (approved):

PR #1412, opened 2026-04-20
fix(arweave): fetch transaction metadata directly

nicholaspai:
## Problem
`ArweaveClient.getMetadata()` currently goes through `client.transactions.get(transactionID)`. In the Arweave SDK, that path is heavier than it looks: for format-2 transactions with small enough payloads it can eagerly fall through into transaction data loading, even though this method only needs the transaction tags. That makes metadata reads more fragile than necessary and keeps them coupled to SDK transaction-loading behavior.

## Solution
- replace the SDK transaction fetch in `getMetadata()` with a repo-owned `fetchWithTimeout()` call to `GET /tx/{id}`
- decode the returned base64url tag names and values directly
- keep the public return shape unchanged: `contentType`, `appName`, and `topic`

## Verification
- `eslint src/caching/Arweave/ArweaveClient.ts`
- `prettier --check src/caching/Arweave/ArweaveClient.ts`
- attempted `hardhat test test/arweaveClient.ts`, but this environment still hits the unrelated local `viem/@noble/curves` runtime failure after compilation

### DIFF
--- a/src/caching/Arweave/ArweaveClient.ts
+++ b/src/caching/Arweave/ArweaveClient.ts
@@ -29,6 +29,10 @@ interface ArweaveTransactionResponse {
 }
 
 function decodeBase64UrlUtf8(value: string): string {
+  // `/tx/{id}` returns tag names and values in RFC 4648 base64url form.
+  // Convert back to standard base64 alphabet, restore omitted `=` padding,
+  // then decode the bytes as UTF-8 so metadata parsing matches the SDK's
+  // `tag.get(..., { decode: true, string: true })` behavior.
   const normalized = value.replace(/-/g, "+").replace(/_/g, "/");
   const padding = normalized.length % 4 === 0 ? "" : "=".repeat(4 - (normalized.length % 4));
   return Buffer.from(`${normalized}${padding}`, "base64").toString("utf-8");
--- a/src/caching/Arweave/ArweaveClient.ts
+++ b/src/caching/Arweave/ArweaveClient.ts
@@ -19,6 +19,21 @@ interface Gateway {
   url: string;
 }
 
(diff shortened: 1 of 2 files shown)


src/caching/Arweave/ArweaveClient.ts
```diff
@@ -19,6 +19,25 @@ interface Gateway {
...
+
+function decodeBase64UrlUtf8(value: string): string {
+  // `/tx/{id}` returns tag names and values in RFC 4648 base64url form.
+  // Convert back to standard base64 alphabet, restore omitted `=` padding,
+  // then decode the bytes as UTF-8 so metadata parsing matches the SDK's
+  // `tag.get(..., { decode: true, string: true })` behavior.
+  const normalized = value.replace(/-/g, "+").replace(/_/g, "/");
+  const padding = normalized.length % 4 === 0 ? "" : "=".repeat(4 - (normalized.length % 4));
+  return Buffer.from(`${normalized}${padding}`, "base64").toString("utf-8");
+}
```
pxrl, line 46: Are there any standard utils we can use for this?

nicholaspai: Good call. I switched this over to Node's built-in base64url decoder in `1fed9c51`, so the helper is now just `Buffer.from(value, "base64url").toString("utf-8")`.

There isn't a repo-local utility for this already, and once we stopped using the SDK's `tag.get(..., { decode: true, string: true })` path we still needed to preserve the same decode semantics somewhere. Using the standard Node primitive is cleaner than keeping the manual alphabet/padding transform.

nicholaspai:
@codex

pxrl (approved):

PR #1413, opened 2026-04-20
refactor(retry): Caller-supplied retry gating

pxrl:
There's a downstream need to handle retries intelligently - both for regular HTTPS queries, but also for arbitrary promises that hide the underlying requests. Upgrade retry() to allow the caller to supply a handler that determines whether a failed request should be retried or not.

This can be used intelligently to evaluate (for example) HTTP(S) failures - i.e. to discriminate on the resulting status.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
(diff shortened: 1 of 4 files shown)


nicholaspai (approved):
I like this, having a universal retry util will go a long way

PR #1418, opened 2026-04-24
feat: Add Avalanche

dijanin-brat:


dijanin-brat:
Avalanche is using EIP1559, so we shouldn't need to change gasPriceOracle.

pxrl (approved):

PR #1424, opened 2026-05-02
fix(coingecko): URL-encode params and surface a typed price-not-found error

mrice32:
## Summary

Two bugs in the Coingecko client surfaced as opaque 500s in `quote-api`'s `/api/coingecko`. Datadog traces:

```
Error: Request path contains unescaped characters     // symbol contained CJK chars
AssertionError [ERR_ASSERTION]: false == true         // assert(tokenPrice !== undefined) for unknown symbol
```

### Fixes

1. **URL encoding.** `encodeURIComponent` user-controlled URL components in all six fetch paths (`getCurrentPriceBySymbol`/`ById`/`ByContract`, `getContractPrices`, `getContractDetails`, `getHistoricContractPrices`, `getContractHistoricDayPrice`). No-op for every real identifier (alphanumeric + `.` + `-`); only encodes pathological inputs (CJK, `/`, `&`, `+`, …).

2. **`CoingeckoPriceNotFoundError`** carrying `{ identifier, currency, lookupType: "symbol" | "address" | "id" }`. Thrown from all three lookup variants, replacing the generic `throw new Error(...)` *and* the downstream `assert(tokenPrice !== undefined)`. Exported from the `coingecko` namespace so callers can `instanceof`-detect it and map to a 404 instead of a 500. **`HttpError` (5xx/429), network failures, and parse errors are deliberately not caught** — callers can distinguish "token has no price" from "Coingecko is unhealthy".

### Tests

`test/Coingecko.test.ts` — 34 tests, fully offline via `sinon.stub(globalThis, "fetch")`:

- **Encoding no-op** (parameterized): every common token (`USDC`/`USDT`/`DAI`/`ETH`/`WETH`/`WBTC`/`USDE`/`MATIC`/`POL`/`SOL`/`ARB`/`OP`/`USDC.e`/`USDH-SPOT`/`PATHUSD`/`CENT`) and currency (`usd`/`eth`/`sol`/`btc`) reaches fetch byte-identical. `getContractPrices` URLs also unchanged for hex addresses and hyphenated platform ids (`polygon-pos`, `arbitrum-one`, …).
- **Encoding active**: `熊猫头` → `%E7%86%8A%E7%8C%AB%E5%A4%B4`, `BTC/USD` → `BTC%2FUSD`, both reach fetch.
- **Error discrimination**: empty CG body and wrong-currency response throw `CoingeckoPriceNotFoundError`; CG 5xx/429 throw `HttpError`; `fetch` rejection propagates as-is; non-JSON throws generic `Error`. None of the upstream-error cases are `instanceof CoingeckoPriceNotFoundError`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.3.153",
+  "version": "4.3.154",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "repository": {
--- a/src/coingecko/Coingecko.ts
+++ b/src/coingecko/Coingecko.ts
@@ -1,6 +1,7 @@
 import assert from "assert";
 import { fetchWithTimeout, getCoingeckoTokenIdByAddress, retry } from "../utils";
 import { Logger } from "../relayFeeCalculator";
+import { CoingeckoPriceNotFoundError } from "./CoingeckoErrors";
 
 export function msToS(ms: number) {
   return Math.floor(ms / 1000);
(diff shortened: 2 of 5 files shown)


mrice32:
@codex review

bmzig (approved):

pxrl (approved):
LGTM. In general I wonder if we should be doing more up-front validation on some of these inputs - we generally know what shape they should be, so it might be a way of narrowing the failure modes.

PR #1425, opened 2026-05-03
fix(gasPriceOracle/solana): handle null getFeeForMessage with confirmed-blockhash retry

mrice32:
## Summary

`gasPriceOracle/adapters/solana.ts` dereferenced the result of Solana's `getFeeForMessage` with a non-null assertion (`baseFeeResponse!.value!`). When the RPC returns `{ value: null }` — common on load-balanced pools where the request lands on a node that hasn't yet seen the blockhash from the earlier `getLatestBlockhash` — `toBN(null)` crashes with `TypeError: Cannot read properties of null (reading 'toString')`.

Caller-visible impact in `quote-api`'s `/api/limits` for SVM destinations: ~1.4% of requests (≈1,800/day, ≈30,000 over the past two weeks) become opaque 500s.

## Fix

Two layers, both in `messageFee`:

1. **Retry once with a `confirmed` blockhash.** When `value` is null, refresh the message's lifetime via `getLatestBlockhash({ commitment: "confirmed" })` and call `getFeeForMessage` again. A confirmed blockhash is by definition propagated to all healthy nodes, so the retry is immune to the race. Fee estimation is never sent on-chain — blockhash freshness is irrelevant.
2. **Typed error on residual null.** If the retry also returns null, throw `SvmGasPriceUnavailableError` (exported from the `gasPriceOracle` namespace). Callers can `instanceof`-detect it and map to a 502/503 instead of crashing with `toBN(null)`.

## Tests

`test/SvmGasPriceOracle.test.ts` (5 tests, fully offline via sinon stubs on the provider):
- Happy path: no retry, returns the base fee
- Null on first call → retries once and asserts the refresh requests `commitment: "confirmed"`
- Null on both calls → throws `SvmGasPriceUnavailableError` (and is `instanceof Error`)
- Exactly one retry — never a third call
- Priority-fee averaging unaffected by the change

### DIFF
--- a/src/gasPriceOracle/adapters/solana.ts
+++ b/src/gasPriceOracle/adapters/solana.ts
@@ -1,5 +1,5 @@
 import { SVMProvider } from "../../arch/svm";
-import { toBN, dedupArray } from "../../utils";
+import { BN, toBN, dedupArray } from "../../utils";
 import { SvmGasPriceEstimate } from "../types";
 import { GasPriceEstimateOptions } from "../oracle";
 import { SvmGasPriceUnavailableError } from "../errors";
@@ -11,6 +11,8 @@ import {
   setTransactionMessageLifetimeUsingBlockhash,
 } from "@solana/kit";
 
+const MAX_BASE_FEE_ATTEMPTS = 2;
+
 /**
  * @notice Returns result of getFeeForMessage and getRecentPrioritizationFees RPC calls.
  * @returns GasPriceEstimate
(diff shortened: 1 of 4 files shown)


src/gasPriceOracle/adapters/solana.ts
```diff
@@ -40,8 +50,27 @@ export async function messageFee(provider: SVMProvider, opts: GasPriceEstimateOp
...
-    baseFee: toBN(baseFeeResponse!.value!),
+    baseFee: toBN(baseFeeResponse.value),
     microLamportsPerComputeUnit,
   };
 }
+
+function getFeeForCompiledMessage(
+  provider: SVMProvider,
+  tx: TransactionMessage & TransactionMessageWithFeePayer
+): Promise<Awaited<ReturnType<ReturnType<SVMProvider["getFeeForMessage"]>["send"]>>> {
```
bmzig: You can probably just make this a `Promise<bigint | undefined>`. Or maybe even just move the retry logic into this function and then either throw the error here or return a `Promise<BigNumber>`.

bmzig (approved):

PR #1431, opened 2026-05-14
chore: bump contracts

bmzig:
This is needed to bump contracts in the relayer.


bmzig:
The contracts release is still in alpha, actually, so will revisit this later.

PR #1440, opened 2026-05-23
fix(providers): preserve Error prototype and cause across quorum failures

droplet-rl:
## Summary

`createSendErrorWithMessage` returned `{ ...sendError, ...new Error(message) }`. Spread loses the Error prototype, and Error's own `message`/`stack` are non-enumerable in V8, so the wrapper collapsed to a plain object — usually `{}` when the rejection reason was itself a generic `Error`. Downstream `instanceof Error` failed and error loggers surfaced `"could not extract error from 'Object' instance {}"` instead of the underlying RPC failure.

Switch to `new Error(message, { cause: sendError })`. Wrapper is a real Error; original reason reachable via `.cause`.

Companion: the SVM short-circuit at `quorumFallbackRpcFactory.ts:71` now also passes `{ cause: error }`, so the underlying `SolanaError` (preflight `__code` / `__serverMessage`) survives the climb.

## Scope

Both EVM (`retryProvider.ts:116`) and SVM (`quorumFallbackRpcFactory.ts:101`) quorum/fallback paths run through the helper. One-line fix each.

## Caller-visible behavior change

Callers can no longer rely on enumerable props of the rejection reason leaking onto the wrapper (e.g. `(error as any).context` for `SolanaError`). Walk `.cause` instead. No such callers in this repo.

## Test plan

- [x] `test/providers/utils.test.ts`: `instanceof Error`, message/stack preservation, `.cause` for Error and non-Error rejection reasons.
- [ ] CI lint + type-check (`Error(message, { cause })` requires Node ≥16.9 / TS lib `es2022`; already in use).

## Related

- Investigation: PD-325142 — https://risk-labs.pagerduty.com/incidents/Q2WQ66XKTQ8YIU
- Relayer companion PR (conditional PDA reuse, the proximal fix for the dataworker page): https://github.com/across-protocol/relayer/pull/3406
- Commit that started triggering this code path more often: e8ab986

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.3.159",
+  "version": "4.3.160",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "repository": {
--- a/src/providers/solana/quorumFallbackRpcFactory.ts
+++ b/src/providers/solana/quorumFallbackRpcFactory.ts
@@ -66,9 +66,12 @@ export class QuorumFallbackSolanaRpcFactory extends SolanaBaseRpcFactory {
               throw error;
             }
 
-            // If one RPC provider reverted, others likely will too. Skip them.
+            // If one RPC provider reverted, others likely will too. Skip them. Preserve the
+            // underlying RPC error (e.g. the SolanaError carrying the preflight `__code` and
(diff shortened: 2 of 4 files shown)


pxrl (approved):

droplet-rl:
No changes needed — pxrl's review on #1440 was a plain approval (empty body, zero inline comments). The PR was merged at 2026-05-23T10:49:35Z, ~1 minute after the approval, so there's no outstanding feedback to address.

PR #1441, opened 2026-05-23
fix(providers): summarize per-provider errors instead of dumping err.stack

droplet-rl:
## Summary

`RetryProvider.send` (and the Solana `QuorumFallbackSolanaRpcFactory` counterpart) accumulate per-provider failures via `(err as any)?.stack || err?.toString()`. For ethers transport errors (`SERVER_ERROR`, `NETWORK_ERROR`, etc.), both `err.stack` and the stringified form embed the failed RPC URL with its API key in cleartext, alongside a multi-line ethers stack trace. That accumulator then gets concatenated into the aggregate `Not enough providers succeeded. Errors: …` error message **and** into the `logQuorumMismatchOrFailureDetails` warn log — both reach Slack/Datadog. So in addition to being unreadable, this is leaking provider API keys to log destinations.

Real example surfaced from the Across relayer warning logs (URL/key redacted here):

```
reason: "Not enough providers succeeded. Errors:\nProvider https://arb-mainnet.g.alchemy.com failed with error: Error: processing response error (body=\"{...message:\\\"execution reverted: ERC20: burn amount exceeds balance\\\"...}\", url=\"https://arb-mainnet.g.alchemy.com/v2/<API_KEY>\", code=SERVER_ERROR, version=web/5.7.1)\n    at Logger.makeError (...)\n    at Logger.throwError (...)\n    ... <truncated, ~2KB total>"
```

After this change, the same failure produces:

```
Not enough providers succeeded. Errors:
Provider https://arb-mainnet.g.alchemy.com failed with error: execution reverted: ERC20: burn amount exceeds balance
```

## Approach

New `summarizeProviderError(err)` helper in `src/providers/utils.ts` that produces a single-line, log-safe summary. Preference order:

1. Parsed JSON-RPC error message via the existing `parseJsonRpcError` (drilling into `err.error` if needed) — this is the actual revert reason that callers want.
2. `err.reason` — ethers' short, URL-free reason string (`processing response error`, `execution reverted`, etc.).
3. `err.message`, **only** for non-ethers errors (no `err.code` present) — safe because there's no transport-layer URL to leak.
4. `err.code` (`SERVER_ERROR`, `NETWORK_ERROR`, …) as a last resort.

`err.message` is deliberately skipped for ethers-shaped errors because that's where ethers' `Logger.makeError` splices the failed URL into the message. The original error remains reachable via `error.cause` on the aggregate thrown by `createSendErrorWithMessage`, so any forensic detail isn't lost — it just stops being on the user-visible log path.

Wired into both `errors.push([provider, …])` sites in `retryProvider.ts` and both in `solana/quorumFallbackRpcFactory.ts`. No other behavior change: `failImmediate` still calls `parseJsonRpcError` on the raw error; the original `err` is still thrown/cascaded as before; classification downstream (e.g. relayer's `MultiCallerClient#canIgnoreRevertReason`) is `.includes()` based and continues to match the cleaner string.

## Test plan

- [x] 8 new unit tests for `summarizeProviderError` in `test/providers/utils.test.ts` covering: null/undefined, string inputs, JSON-RPC body extraction (top-level + nested `err.error`), `.reason` fallback, `.code` fallback, non-ethers `.message`, and an explicit URL-leak guard.
- [x] Existing `createSendErrorWithMessage` + `providers.test.ts` suites still pass.
- [x] `prettier --check` and `eslint` clean on touched files.
- [x] `tsc -p tsconfig.build.json --noEmit` clean.

## Context

Found while investigating noisy `MultiCallerClient#LogSimulationFailures` warnings in the Across relayer. A relayer-side regex sanitizer was considered first ([across-protocol/relayer#3411](https://github.com/across-protocol/relayer/pull/3411), now closed) but fixing it once at the source — here — is cleaner and benefits every consumer (relayer, dataworker, finalizer, monitor) with a single version bump.

### DIFF
--- a/src/providers/retryProvider.ts
+++ b/src/providers/retryProvider.ts
@@ -4,7 +4,13 @@ import { CHAIN_IDs } from "../constants";
 import { delay, isDefined, isPromiseFulfilled, isPromiseRejected } from "../utils";
 import { getOriginFromURL } from "../utils/NetworkUtils";
 import { CacheProvider } from "./cachedProvider";
-import { compareRpcResults, createSendErrorWithMessage, formatProviderError, parseJsonRpcError } from "./utils";
+import {
+  compareRpcResults,
+  createSendErrorWithMessage,
+  formatProviderError,
+  parseJsonRpcError,
+  summarizeProviderError,
+} from "./utils";
 import { PROVIDER_CACHE_TTL } from "./constants";
 import { Logger } from "winston";
 
@@ -82,9 +88,11 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
       return this._trySend(provider, method, params)
         .then((result): [ethers.providers.StaticJsonRpcProvider, unknown] => [provider, result])
         .catch((err: unknown) => {
-          // Append the provider and error to the error array.
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          errors.push([provider, (err as any)?.stack || err?.toString()]);
+          // Record a short, log-safe summary of the failure. We intentionally do not retain
+          // `err.stack` / `err.toString()` here: for ethers transport errors those embed the
+          // failed RPC URL (including its API key) and a multi-line stack trace, which then
+          // get concatenated into the aggregate error message and surfaced to logs.
+          errors.push([provider, summarizeProviderError(err)]);
 
           // If there are no new fallback providers to use, terminate the recursion by throwing an error.
           // Otherwise, we can try to call another provider.
@@ -187,7 +195,7 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
         this._trySend(provider, method, params)
           .then((result): [ethers.providers.StaticJsonRpcProvider, unknown] => [provider, result])
           .catch((err) => {
-            errors.push([provider, err?.stack || err?.toString()]);
+            errors.push([provider, summarizeProviderError(err)]);
             throw new Error("No fallbacks during quorum search");
           })
       )
--- a/src/providers/retryProvider.ts
+++ b/src/providers/retryProvider.ts
@@ -128,7 +128,10 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
       return values[0][1];
     }
 
-    const getMismatchedProviders = (values: [ethers.providers.StaticJsonRpcProvider, unknown][]) => {
+    const getMismatchedProviders = (
+      quorumResult: unknown,
+      values: [ethers.providers.StaticJsonRpcProvider, unknown][]
+    ) => {
       return values
         .filter(([, result]) => !compareRpcResults(method, result, quorumResult))
         .map(([provider]) => getOriginFromURL(provider.connection.url));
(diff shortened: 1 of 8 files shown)


droplet-rl:
Rebased onto #1483 and retargeted the base to its branch per Paul, so the two stack: merge #1483 first, then this one (GitHub retargets this PR to `master` automatically when #1483 merges).

Conflict resolution note: `master` meanwhile introduced `formatRpcError()` on the Solana path, which preserves the SolanaError JSON-RPC code in the wrap message. Kept `formatRpcError()` at the Solana call sites and replaced only its non-Solana fallback (`err.stack || err.toString()`) with `summarizeProviderError()`, so both intents compose: code preservation for SolanaErrors, no URL/API-key leakage for everything else. The `formatRpcError` fallback test was updated to match (`undefined` → "unknown error", no stack in the summary).

No separate version bump here — this ships with the 4.4.12 bump from #1483 in the same release.

src/providers/solana/quorumFallbackRpcFactory.ts
```diff
@@ -57,7 +57,8 @@ export class QuorumFallbackSolanaRpcFactory extends SolanaBaseRpcFactory {
           .transport<TResponse>(...args)
           .then((result): [SolanaClusterRpcFactory, RpcResponse<TResponse>] => [factory.rpcFactory, result])
           .catch((error) => {
-            // Preserve the underlying JSON-RPC error code in the wrap message.
+            // Preserve the underlying JSON-RPC error code in the wrap message. For non-Solana
+            // errors this renders a short, log-safe summary rather than `error.stack`.
```
pxrl: Revert this change; this is the Solana code - not relevant what other architectures do.

src/providers/retryProvider.ts
```diff
@@ -82,9 +88,11 @@ export class RetryProvider extends ethers.providers.StaticJsonRpcProvider {
       return this._trySend(provider, method, params)
         .then((result): [ethers.providers.StaticJsonRpcProvider, unknown] => [provider, result])
         .catch((err: unknown) => {
-          // Append the provider and error to the error array.
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          errors.push([provider, (err as any)?.stack || err?.toString()]);
+          // Record a short, log-safe summary of the failure. We intentionally do not retain
+          // `err.stack` / `err.toString()` here: for ethers transport errors those embed the
+          // failed RPC URL (including its API key) and a multi-line stack trace, which then
+          // get concatenated into the aggregate error message and surfaced to logs.
```
pxrl, line 94: Comment far too verbose. Keep it to 1 line.

src/providers/solana/quorumFallbackRpcFactory.ts
droplet-rl: Reverted to the original comment in 739b395.

droplet-rl:
Addressed the review: reverted the comment in `src/providers/solana/quorumFallbackRpcFactory.ts` back to its original wording ("Preserve the underlying JSON-RPC error code in the wrap message.") in 739b395, and replied on the inline thread.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=a82d91e395ea1e691f3fc845ffe1fcea&trace_start_ts=1784131137&trace_end_ts=1784217837)

pxrl (approved):

droplet-rl:
Thanks for the review and approval! No outstanding feedback to address — this is ready to merge whenever you are.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=58c05cc78e223337c7b19222f95eef61&trace_start_ts=1784131740&trace_end_ts=1784218440)

PR #1445, opened 2026-05-27
refactor(svm): delete getLatestFinalizedSlotWithBlock, use getNearestSlotTime directly

droplet-rl:
## Summary

- `getLatestFinalizedSlotWithBlock` re-implemented the walk-back already provided by `getNearestSlotTime`, but over `getBlock` and without catching `SVM_SLOT_SKIPPED` / `SVM_LONG_TERM_STORAGE_SLOT_SKIPPED`. A single skipped slot — or a post-snapshot ledger gap on the RPC — crashed the dataworker run (seen in `zion-across-disputer` 2026-05-27, slot 422438019).
- This was the missing half of the refactor in #1137: that PR factored out `getNearestSlotTime` and migrated five other callers, but left the second `do/while` in `getLatestFinalizedSlotWithBlock` untouched, so the function never picked up the slot-skipped error handling it was supposed to.
- The function added no value over `getNearestSlotTime` once that abstraction existed, so it's deleted entirely. The four lines it would collapse to are inlined into the single caller (`getWidestPossibleExpectedBlockRange` in `PoolRebalanceUtils`).
- No external callers in `across-protocol/relayer` either.

## Test plan
- [x] `tsc --noEmit` (clean)
- [x] `yarn lint-check` (clean)
- [ ] Watch a dataworker run on `zion-across-disputer` once SDK bump lands and confirm the error stops recurring on snapshot/skip events.

### DIFF
--- a/src/clients/BundleDataClient/utils/PoolRebalanceUtils.ts
+++ b/src/clients/BundleDataClient/utils/PoolRebalanceUtils.ts
@@ -2,7 +2,7 @@ import { MerkleTree } from "@across-protocol/contracts/dist/utils/MerkleTree";
 import { RunningBalances, PoolRebalanceLeaf, Clients, SpokePoolTargetBalance } from "../../../interfaces";
 import { isSVMSpokePoolClient, SpokePoolClient } from "../../SpokePoolClient";
 import { BigNumber, bnZero, chainIsEvm, chainIsSvm, compareAddresses, EvmAddress, isDefined } from "../../../utils";
-import { getLatestFinalizedSlotWithBlock } from "../../../arch/svm";
+import { getNearestSlotTime } from "../../../arch/svm";
 import { HubPoolClient } from "../../HubPoolClient";
 import { V3DepositWithBlock } from "./shims";
 import { AcrossConfigStoreClient } from "../../AcrossConfigStoreClient";
@@ -36,16 +36,17 @@ export async function getWidestPossibleExpectedBlockRange(
     Math.max(spokeClients[chainId].latestHeightSearched - endBlockBuffers[idx], 0);
 
   // Across bundles are bounded by slots on Solana. The UMIP requires that the bundle end slot is backed by a block.
-  const resolveSVMEndBlock = (chainId: number, idx: number): Promise<number> => {
+  const resolveSVMEndBlock = async (chainId: number, idx: number): Promise<number> => {
     const spokePoolClient = spokeClients[chainId];
     assert(isSVMSpokePoolClient(spokePoolClient));
 
-    const maxSlot = resolveEndBlock(chainId, idx); // Respect any configured buffer for Solana.
-    return getLatestFinalizedSlotWithBlock(
-      spokePoolClient.svmEventsClient.getRpc(),
-      spokePoolClient.logger,
-      BigInt(maxSlot)
-    );
+    const provider = spokePoolClient.svmEventsClient.getRpc();
+    const { logger } = spokePoolClient;
+    const maxSlot = BigInt(resolveEndBlock(chainId, idx)); // Respect any configured buffer for Solana.
+    const { slot: finalizedSlot } = await getNearestSlotTime(provider, { commitment: "finalized" }, logger);
+    if (finalizedSlot <= maxSlot) return Number(finalizedSlot);
+    const { slot } = await getNearestSlotTime(provider, { slot: maxSlot }, logger);
+    return Number(slot);
   };
 
   const latestPossibleBundleEndBlockNumbers = await Promise.all(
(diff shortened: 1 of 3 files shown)


droplet-rl:
Both Codex comments target the deleted wrapper at `src/arch/svm/utils.ts:99-101`. Commit `0dc0703` removed the wrapper entirely and inlined the two `getNearestSlotTime` calls into `getWidestPossibleExpectedBlockRange` (`PoolRebalanceUtils.ts:46-48`), so the line numbers in the review no longer exist — but the underlying behaviour the bot is flagging is still present at the inlined site. Pushing back on both, no code changes:

**1. "Keep the capped lookback for pruned Solana slots"** — the 1000-slot cap lived only in the deleted wrapper. `getNearestSlotTime` itself has no cap, and 8 other callers across `SpokeUtils.ts`, `BlockUtils.ts`, `BaseAbstractClient.ts`, etc. already rely on it that way. Re-introducing a cap *only* at this call site would be inconsistent and re-create the band-aid the refactor is trying to remove. If unbounded walk-back in pruned ranges is genuinely a concern, the right fix is on `getNearestSlotTime` (or on `getTimestampForSlot` distinguishing `SVM_SLOT_SKIPPED` from `SVM_LONG_TERM_STORAGE_SLOT_SKIPPED` and throwing on the latter), which is out of scope here. In practice `maxSlot` is `latestHeightSearched - endBlockBuffer`, which is recent for any non-bootstrap run.

**2. "Restore block validation before returning Solana end slots"** — same shape: if `getBlockTime` returns `null` per RPC spec, `getTimestampForSlot` does indeed `Number(null) → 0` and the loop accepts it. But that's a pre-existing `getTimestampForSlot` issue affecting every caller of `getNearestSlotTime`, not something introduced by this PR. The motivating bug the PR fixes (dataworker crashing on skipped slots in `getBlock`) is real and observed; the null-from-`getBlockTime` case is theoretical against current Solana RPC behaviour (skip/prune cases throw `SVM_*_SKIPPED`). If we want defence-in-depth there, the fix belongs in `getTimestampForSlot`.

Leaving the PR as-is.

src/clients/BundleDataClient/utils/PoolRebalanceUtils.ts
```diff
@@ -36,16 +36,17 @@ export async function getWidestPossibleExpectedBlockRange(
...
-      spokePoolClient.logger,
-      BigInt(maxSlot)
-    );
+    const provider = spokePoolClient.svmEventsClient.getRpc();
+    const { logger } = spokePoolClient;
+    const maxSlot = BigInt(resolveEndBlock(chainId, idx)); // Respect any configured buffer for Solana.
+    const { slot: finalizedSlot } = await getNearestSlotTime(provider, { commitment: "finalized" }, logger);
+    if (finalizedSlot <= maxSlot) return Number(finalizedSlot);
+    const { slot } = await getNearestSlotTime(provider, { slot: maxSlot }, logger);
+    return Number(slot);
```
pxrl, line 53: @droplet-rl Re-work this into a `do { ... } while (...)` loop, so there's only one call to `getNearestSlotTime()`.

droplet-rl:
Done in 60f48eb — single `getNearestSlotTime` call inside a `do { ... } while (slot > maxSlot)`:

```ts
let slot: bigint;
let opts: { slot: bigint } | { commitment: "finalized" } = { commitment: "finalized" };
do {
  ({ slot } = await getNearestSlotTime(provider, opts, logger));
  opts = { slot: maxSlot };
} while (slot > maxSlot);
return Number(slot);
```

First iteration walks back from the finalized head; if that lands above `maxSlot`, the second iteration walks back from `maxSlot` (which by construction returns ≤ `maxSlot`), so the loop exits in at most two iterations. `tsc` and `eslint` clean.

pxrl:
@codex review

pxrl:
@droplet-rl Remember to bump the patch version on this PR.

droplet-rl:
This is the same concern raised in the prior Codex round (review 4370323619, comment 3309188101) — pushing back on it again for the same reasons:

- The 1000-slot cap lived only inside the deleted wrapper. `getNearestSlotTime` itself has no cap and is the canonical primitive used at ~8 other call sites in this codebase.
- Re-introducing a cap *only* at this site would be inconsistent and would re-create the kind of bandaid the refactor is intentionally removing.
- If unbounded walkback against pruned ledgers is genuinely a concern, the fix belongs in `getNearestSlotTime` / `getTimestampForSlot` (e.g. distinguishing `SVM_LONG_TERM_STORAGE_SLOT_SKIPPED` from `SVM_SLOT_SKIPPED` and throwing on the former). That's out of scope for this PR.
- The human reviewer (@pxrl) reviewed the diff after my previous response on this thread and only requested the do/while restructure — they're aware of the tradeoff and are fine with the current behaviour.

Leaving the PR as-is.

pxrl (approved):

PR #1447, opened 2026-05-27
fix(svm): resolve slot timestamps via getBlocks instead of per-slot walk

droplet-rl:
## Summary

- `SVMSpokePoolClient.getTimestampForBlock()` and `getNearestSlotTime()` recover from skipped slots by decrementing `_slot` one at a time and re-calling `getBlockTime`. When a target slot lands inside a real skipped-slot run on Solana mainnet, the cost compounds — every `BundleDataClient.getBundleBlockTimestamps` lookup × NODE_QUORUM providers × propose/validate/execute pass.
- In a production dataworker run on 2026-05-27, the bundle boundary fell inside slots 422461540–422461547 (a real ~8-slot skip run on mainnet). The SVM executor primary kept calling `getBlockTime` for 57+ minutes against the same 8 slots before the serverless-orchestration hub's `rejectSpokeDelay` fired a retry that then raced the primary.
- Replace the per-slot walk with `findNearestProducedSlot(provider, targetSlot)`: a single `getBlocks(lower, upper)` call asks the RPC which slots in a 32-slot window produced a block, then iterates the window backwards in fixed steps if the initial range is empty.
- `findNearestProducedSlot` is exported from `arch/svm/SpokeUtils.ts` and used by both `SVMSpokePoolClient.getTimestampForBlock` and `getNearestSlotTime`.
- Also memoise per-block timestamp lookups inside `SVMSpokePoolClient`: bundle data loading hits the same boundary slots repeatedly across propose/validate/execute passes; storing the Promise (not the resolved value) also dedupes concurrent lookups.

## Why this is safe

- `getBlocks` is a standard Solana JSON-RPC method (`GetBlocksApi` in `@solana/kit`; in the spec since v1.7 / mid-2021 when `getConfirmedBlocks` was renamed). Verified live against `api.mainnet-beta.solana.com` and shape-checked against the kit type surface.
- Benchmarked latency on mainnet-beta is flat ~75 ms up to ~1024 slot windows, +5–10 ms per doubling past that. A 32-slot window is one round-trip ~75 ms, vs the current path that issues 8+ `getBlockTime` RPCs × 2 providers per single resolution.
- Iteration is bounded: `maxIterations × window` (default 32 × 32 = 1024 slots backward). If we still find nothing, the function returns `undefined` and the caller throws the same `Unable to resolve time…` error the old code did.
- Function signatures of `getTimestampForBlock` and `getNearestSlotTime` are unchanged; this is a pure implementation swap.

## Test plan

- [x] `yarn lint-check` clean on touched files.
- [x] `yarn build` (`tsc`) clean.
- [x] New unit test `test/Solana.findNearestProducedSlot.ts` covers: produced slot in initial window, iterate backwards when initial window is empty, return undefined when no produced slot inside the iteration bound, stop descending past slot 0.
- [ ] CI to run the full suite under `yarn test`.

## Follow-ups (separate PRs)

- The relayer-side companion fixes for the SVM duplicate-execution race that this stall triggered are tracked separately in `across-protocol/relayer`.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -203,29 +203,14 @@ async function _callGetTimestampForSlotWithRetry(
   return timestamp;
 }
 
-/**
- * Defaults for {@link findNearestProducedSlot}. The 32-slot initial window covers ~13s of Solana
- * time at 400ms/slot, which is more than enough for typical skipped-slot runs around bundle
- * boundaries. The iteration bound caps total backward coverage at 32 windows × 32 slots = 1024
- * slots; beyond that something is wrong upstream and we want to surface that rather than walk
- * for hours.
- */
+// 32 slots ≈ 13s at 400ms/slot; 32 iterations caps total coverage at 1024 slots.
 const DEFAULT_GET_BLOCKS_WINDOW = 32;
 const DEFAULT_GET_BLOCKS_ITERATIONS = 32;
 
 /**
- * Find the largest slot ≤ targetSlot that produced a block. Issues one getBlocks() call per
- * window, iterating backwards in fixed-size windows when the initial range contains no produced
- * slots (e.g. when targetSlot lands inside a skipped-slot run).
- *
- * Prefer this over scanning slot-by-slot with getBlockTime(): a single getBlocks() round-trip
- * resolves the entire window at once, and the cost is bounded even for long skipped-slot runs.
- *
- * @param provider SVM Provider instance.
- * @param targetSlot The slot at or below which to search.
- * @param opts Optional commitment level, per-iteration window size, and max iteration count.
- * @returns The largest slot ≤ targetSlot with a produced block, or undefined if none is found
- *          within the bounded search range.
+ * Find the largest slot ≤ targetSlot that produced a block, iterating backwards in fixed-size
+ * windows via getBlocks(). Returns undefined if no produced slot is found within the bounded
+ * search range.
  */
 export async function findNearestProducedSlot(
   provider: SVMProvider,
@@ -243,13 +228,15 @@ export async function findNearestProducedSlot(
   const commitmentArg = opts.commitment ? { commitment: opts.commitment } : undefined;
 
   let upper = targetSlot;
-  for (let i = 0; i < maxIterations; i++) {
+  for (let i = 0; i < maxIterations; ++i) {
     const lower = upper >= window ? upper - window + 1n : 0n;
     const blocks = await provider.getBlocks(lower, upper, commitmentArg).send();
     if (blocks.length > 0) {
       return BigInt(blocks[blocks.length - 1]);
     }
-    if (lower === 0n) return undefined;
+    if (lower === 0n) {
+      break;
+    }
     upper = lower - 1n;
   }
   return undefined;
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -204,6 +204,58 @@ async function _callGetTimestampForSlotWithRetry(
 }
 
 /**
+ * Defaults for {@link findNearestProducedSlot}. The 32-slot initial window covers ~13s of Solana
+ * time at 400ms/slot, which is more than enough for typical skipped-slot runs around bundle
+ * boundaries. The iteration bound caps total backward coverage at 32 windows × 32 slots = 1024
+ * slots; beyond that something is wrong upstream and we want to surface that rather than walk
+ * for hours.
+ */
+const DEFAULT_GET_BLOCKS_WINDOW = 32;
+const DEFAULT_GET_BLOCKS_ITERATIONS = 32;
+
+/**
+ * Find the largest slot ≤ targetSlot that produced a block. Issues one getBlocks() call per
+ * window, iterating backwards in fixed-size windows when the initial range contains no produced
+ * slots (e.g. when targetSlot lands inside a skipped-slot run).
+ *
+ * Prefer this over scanning slot-by-slot with getBlockTime(): a single getBlocks() round-trip
+ * resolves the entire window at once, and the cost is bounded even for long skipped-slot runs.
+ *
+ * @param provider SVM Provider instance.
+ * @param targetSlot The slot at or below which to search.
+ * @param opts Optional commitment level, per-iteration window size, and max iteration count.
+ * @returns The largest slot ≤ targetSlot with a produced block, or undefined if none is found
+ *          within the bounded search range.
+ */
+export async function findNearestProducedSlot(
+  provider: SVMProvider,
+  targetSlot: bigint,
+  opts: {
+    commitment?: Exclude<Commitment, "processed">;
+    window?: number;
+    maxIterations?: number;
+  } = {}
+): Promise<bigint | undefined> {
+  const window = BigInt(opts.window ?? DEFAULT_GET_BLOCKS_WINDOW);
+  const maxIterations = opts.maxIterations ?? DEFAULT_GET_BLOCKS_ITERATIONS;
+  assert(window > 0n, `findNearestProducedSlot: window must be > 0 (got ${window})`);
+  assert(maxIterations > 0, `findNearestProducedSlot: maxIterations must be > 0 (got ${maxIterations})`);
+  const commitmentArg = opts.commitment ? { commitment: opts.commitment } : undefined;
+
+  let upper = targetSlot;
+  for (let i = 0; i < maxIterations; i++) {
+    const lower = upper >= window ? upper - window + 1n : 0n;
+    const blocks = await provider.getBlocks(lower, upper, commitmentArg).send();
+    if (blocks.length > 0) {
+      return BigInt(blocks[blocks.length - 1]);
+    }
+    if (lower === 0n) return undefined;
+    upper = lower - 1n;
+  }
+  return undefined;
+}
+
+/**
  * Returns the current fill deadline buffer.
  * @param provider SVM Provider instance
  * @param statePda Spoke Pool's State PDA
--- a/src/clients/SpokePoolClient/SVMSpokePoolClient.ts
+++ b/src/clients/SpokePoolClient/SVMSpokePoolClient.ts
@@ -203,6 +203,9 @@ export class SVMSpokePoolClient extends SpokePoolClient {
     if (!isDefined(cached)) {
       cached = this._resolveTimestampForBlock(slot);
       this.timestampForBlockCache.set(slot, cached);
+      // Evict failed lookups so a transient RPC error doesn't poison the cache for the lifetime of
+      // the client. The caller's own await still observes the rejection.
+      cached.catch(() => this.timestampForBlockCache.delete(slot));
     }
     return cached;
   }
@@ -215,9 +218,6 @@ export class SVMSpokePoolClient extends SpokePoolClient {
     }
     const timestamp = await getTimestampForSlot(rpc, producedSlot, undefined, this.logger);
     if (!isDefined(timestamp)) {
-      // findNearestProducedSlot reported a produced block at this slot, but getBlockTime came back
-      // empty. Treat this as a transient RPC inconsistency rather than caching a bad answer.
-      this.timestampForBlockCache.delete(slot);
       throw new Error(
         `Missing block time for produced ${getNetworkName(this.chainId)} slot ${producedSlot} (queried for ${slot})`
       );
--- a/src/clients/SpokePoolClient/SVMSpokePoolClient.ts
+++ b/src/clients/SpokePoolClient/SVMSpokePoolClient.ts
@@ -5,6 +5,7 @@ import {
   unwrapEventData,
   getFillDeadline,
   getTimestampForSlot,
+  findNearestProducedSlot,
   getStatePda,
   SvmCpiEventsClient,
   findDeposit,
@@ -34,6 +35,10 @@ import { SVM_SPOKE_POOL_CLIENT_TYPE } from "./types";
  */
 export class SVMSpokePoolClient extends SpokePoolClient {
   readonly type = SVM_SPOKE_POOL_CLIENT_TYPE;
+  // Memoise per-block timestamp lookups for the lifetime of this client. Bundle data loading
+  // resolves the same boundary slots repeatedly across propose/validate/execute passes; storing
+  // the Promise (rather than the resolved value) also dedupes concurrent lookups.
+  private timestampForBlockCache: Map<number, Promise<number>> = new Map();
   /**
    * Note: Strongly prefer to use the async create() method to instantiate.
    */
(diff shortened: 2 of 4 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -203,6 +203,58 @@ async function _callGetTimestampForSlotWithRetry(
...
+  const window = BigInt(opts.window ?? DEFAULT_GET_BLOCKS_WINDOW);
+  const maxIterations = opts.maxIterations ?? DEFAULT_GET_BLOCKS_ITERATIONS;
+  assert(window > 0n, `findNearestProducedSlot: window must be > 0 (got ${window})`);
+  assert(maxIterations > 0, `findNearestProducedSlot: maxIterations must be > 0 (got ${maxIterations})`);
+  const commitmentArg = opts.commitment ? { commitment: opts.commitment } : undefined;
+
+  let upper = targetSlot;
+  for (let i = 0; i < maxIterations; i++) {
+    const lower = upper >= window ? upper - window + 1n : 0n;
+    const blocks = await provider.getBlocks(lower, upper, commitmentArg).send();
```
pxrl: Is it guaranteed that any elements in the response are actually blocks?

src/arch/svm/SpokeUtils.ts
```diff
@@ -203,6 +203,58 @@ async function _callGetTimestampForSlotWithRetry(
...
+    if (blocks.length > 0) {
+      return BigInt(blocks[blocks.length - 1]);
+    }
+    if (lower === 0n) return undefined;
```
pxrl:
```suggestion
    if (lower === 0n) {
      break;
    }
```

src/arch/svm/SpokeUtils.ts
```diff
@@ -203,6 +203,58 @@ async function _callGetTimestampForSlotWithRetry(
...
+  } = {}
+): Promise<bigint | undefined> {
+  const window = BigInt(opts.window ?? DEFAULT_GET_BLOCKS_WINDOW);
+  const maxIterations = opts.maxIterations ?? DEFAULT_GET_BLOCKS_ITERATIONS;
+  assert(window > 0n, `findNearestProducedSlot: window must be > 0 (got ${window})`);
+  assert(maxIterations > 0, `findNearestProducedSlot: maxIterations must be > 0 (got ${maxIterations})`);
+  const commitmentArg = opts.commitment ? { commitment: opts.commitment } : undefined;
+
+  let upper = targetSlot;
+  for (let i = 0; i < maxIterations; i++) {
```
pxrl:
```suggestion
  for (let i = 0; i < maxIterations; ++i) {
```

src/arch/svm/SpokeUtils.ts
```diff
@@ -203,6 +203,58 @@ async function _callGetTimestampForSlotWithRetry(
...
+const DEFAULT_GET_BLOCKS_WINDOW = 32;
+const DEFAULT_GET_BLOCKS_ITERATIONS = 32;
+
+/**
+ * Find the largest slot ≤ targetSlot that produced a block. Issues one getBlocks() call per
+ * window, iterating backwards in fixed-size windows when the initial range contains no produced
+ * slots (e.g. when targetSlot lands inside a skipped-slot run).
+ *
+ * Prefer this over scanning slot-by-slot with getBlockTime(): a single getBlocks() round-trip
+ * resolves the entire window at once, and the cost is bounded even for long skipped-slot runs.
```
pxrl: Comments very verbose; try to be concise.

test/Solana.findNearestProducedSlot.ts
pxrl, line 1: Should this file path be scoped more tightly? Do we have test/svm or something?

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -190,19 +195,34 @@ export class SVMSpokePoolClient extends SpokePoolClient {
...
+    const timestamp = await getTimestampForSlot(rpc, producedSlot, undefined, this.logger);
+    if (!isDefined(timestamp)) {
+      // findNearestProducedSlot reported a produced block at this slot, but getBlockTime came back
+      // empty. Treat this as a transient RPC inconsistency rather than caching a bad answer.
+      this.timestampForBlockCache.delete(slot);
+      throw new Error(
+        `Missing block time for produced ${getNetworkName(this.chainId)} slot ${producedSlot} (queried for ${slot})`
+      );
+    }
+    return timestamp;
```
pxrl:
```suggestion
    if (isDefined(timestamp)) {
      return timestamp;
    }
    
    // findNearestProducedSlot reported a produced block at this slot, but getBlockTime came back
    // empty. Treat this as a transient RPC inconsistency rather than caching a bad answer.
    this.timestampForBlockCache.delete(slot);
    throw new Error(
      `Missing block time for produced ${getNetworkName(this.chainId)} slot ${producedSlot} (queried for ${slot})`
    );
```

src/arch/svm/SpokeUtils.ts
droplet-rl: Yes — by the Solana JSON-RPC contract for `getBlocks`, every element of the response is a slot in the requested `[lower, upper]` range that produced a confirmed block (returned ascending). The `@solana/kit` typing reflects that as `Slot[]`, so `blocks[blocks.length - 1]` is guaranteed to be a produced slot once we've gated on `blocks.length > 0`.

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -190,19 +195,34 @@ export class SVMSpokePoolClient extends SpokePoolClient {
...
-      const timestamp = await getTimestampForSlot(this.svmEventsClient.getRpc(), _slot, maxRetries, this.logger);
-      if (isDefined(timestamp)) {
-        return timestamp;
-      }
-    } while (--_slot > 0);
+  public override getTimestampForBlock(slot: number): Promise<number> {
+    let cached = this.timestampForBlockCache.get(slot);
+    if (!isDefined(cached)) {
+      cached = this._resolveTimestampForBlock(slot);
+      this.timestampForBlockCache.set(slot, cached);
```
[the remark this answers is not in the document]
droplet-rl: Good catch — fixed in 73ae3a6. The cached promise now has a `.catch()` attached when it's stored, which clears the cache entry on rejection so subsequent `getTimestampForBlock(slot)` calls can retry after the RPC recovers. The caller's own `await` still observes the original rejection. Dropped the now-redundant explicit `delete()` from the `_resolveTimestampForBlock` undefined-timestamp path.

droplet-rl:
Addressed both review comments in 73ae3a6:

- **pxrl** (`SpokeUtils.ts:248`, "is it guaranteed that any elements in the response are actually blocks?") — replied inline. By the Solana RPC contract for `getBlocks` the response is a list of produced-block slots in `[lower, upper]` (ascending), typed as `Slot[]` in `@solana/kit`, so after the `blocks.length > 0` gate `blocks[blocks.length - 1]` is guaranteed to be a produced slot. No code change needed.
- **codex** (`SVMSpokePoolClient.ts:205`, P2, evict failed timestamp lookups) — fixed. Attached `cached.catch(() => this.timestampForBlockCache.delete(slot))` when the promise is first stored, so any rejection (RPC error from `findNearestProducedSlot`/`getBlocks`, undefined `producedSlot`, undefined block time) evicts the entry. Dropped the now-redundant explicit `delete()` from `_resolveTimestampForBlock`'s undefined-timestamp branch — the catch handles it.

`yarn build:types` and `yarn hardhat test test/Solana.findNearestProducedSlot.ts` both pass. Commit pushed via checkpoint.

src/arch/svm/SpokeUtils.ts
droplet-rl: Applied in c612662.

src/arch/svm/SpokeUtils.ts
droplet-rl: Applied in c612662.

src/arch/svm/SpokeUtils.ts
droplet-rl: Trimmed in c612662 — collapsed both block comments on `findNearestProducedSlot` and its defaults to one line each, and tightened the `.catch()` comment in `SVMSpokePoolClient` as well.

test/Solana.findNearestProducedSlot.ts
droplet-rl: There's no `test/svm/` directory today — existing SVM tests use the `Solana.*` filename prefix at `test/` root (`Solana.EventData.test.ts`, `Solana.SvmCpiEventsClient.Integration.test.ts`, `Solana.setup.ts`, etc.), so `Solana.findNearestProducedSlot.ts` follows that convention. Happy to move all of them into `test/svm/` as a follow-up if you'd prefer to standardise on subdirectories, but it felt out of scope for this PR.

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
droplet-rl: Inverted the condition in c612662 — positive early return now, throw falls through. I left out the explicit `this.timestampForBlockCache.delete(slot)` here because 73ae3a6 (your earlier review's eviction concern) already attached `cached.catch(() => this.timestampForBlockCache.delete(slot))` at the call site in `getTimestampForBlock`, so any rejection (including this one) evicts the entry. Happy to add it back as a defensive belt-and-suspenders if you'd rather have the explicit call alongside the catch.

droplet-rl:
Addressed all five inline comments from this review in c612662:

- **`if (lower === 0n) { break; }`** — applied.
- **`for (let i = 0; i < maxIterations; ++i)`** — applied.
- **"Comments very verbose"** — collapsed the block comments on `findNearestProducedSlot` and its defaults to one line each, and tightened the new `.catch()` comment in `SVMSpokePoolClient`.
- **Test path scoping** — replied; there's no `test/svm/` today, existing SVM tests use the `Solana.*` filename prefix at `test/` root (`Solana.setup.ts`, `Solana.EventData.test.ts`, etc.) and the new file follows that convention. Offered to move them all as a follow-up if the team wants to standardise on subdirectories.
- **Invert `isDefined(timestamp)` check** — applied (positive early return, throw falls through). The explicit `timestampForBlockCache.delete(slot)` from the suggestion was based on the pre-73ae3a6 code; the eviction is now handled by `cached.catch(...)` attached in `getTimestampForBlock`, so I left the redundant delete out and called that out in the reply.

`yarn build:types` and the `findNearestProducedSlot` tests still pass. Pushed via checkpoint.

src/arch/svm/SpokeUtils.ts
```diff
@@ -203,6 +203,45 @@ async function _callGetTimestampForSlotWithRetry(
...
+// 32 slots ≈ 13s at 400ms/slot; 32 iterations caps total coverage at 1024 slots.
+const DEFAULT_GET_BLOCKS_WINDOW = 32;
+const DEFAULT_GET_BLOCKS_ITERATIONS = 32;
+
+/**
+ * Find the largest slot ≤ targetSlot that produced a block, iterating backwards in fixed-size
+ * windows via getBlocks(). Returns undefined if no produced slot is found within the bounded
+ * search range.
+ */
+export async function findNearestProducedSlot(
```
pxrl: I think this belongs in src/arch/svm/utils.ts.

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -190,19 +195,40 @@ export class SVMSpokePoolClient extends SpokePoolClient {
...
+      // Evict on failure so a transient RPC error doesn't poison the cache.
+      cached.catch(() => this.timestampForBlockCache.delete(slot));
```
pxrl, line 207: Are there additional concurrency concerns here? Could the value for the same slot end up being accessed & overwritten concurrently?

src/arch/svm/SpokeUtils.ts
droplet-rl: Moved in 204db04 — `findNearestProducedSlot` and its two `DEFAULT_GET_BLOCKS_*` constants now live in `src/arch/svm/utils.ts` next to `getNearestSlotTime`, which was the main consumer. Updated `utils.ts`'s internal import from `./SpokeUtils` (dropping the now-stale `findNearestProducedSlot` entry) and updated the test import. `SVMSpokePoolClient` pulls it through the `arch/svm` barrel so that import was unaffected.

droplet-rl:
:hourglass_flowing_sand: Stopped 1 background subagent(s) at the 30m drain budget. Their results were not gathered:
  • yarn install for sdk

:white_check_mark: 3 background subagent(s) finished after the main reply:
  • Run typecheck in sdk dir
  • Run findNearestProducedSlot tests
  • Background command "Reinstall" completed (exit code 0)

_If you need the stopped results, ask again — the agent can re-run those subagents._

pxrl:
@codex review

droplet-rl:
No action needed from me — that comment is a `@codex review` invocation aimed at the chatgpt-codex-connector bot, which will post its own review pass. I'll wait for any actionable feedback from codex (or a follow-up from you) before making further changes.

droplet-rl:
Codex came back clean ("Didn't find any major issues") on the current HEAD (204db04). No further code changes warranted from this trigger. PR is ready for human review.

pxrl (approved):

droplet-rl:
Approval received from pxrl on review 4371702294 — no inline comments, empty body. PR is approved at HEAD 204db04. No code changes needed; ready for merge whenever you'd like.

src/clients/SpokePoolClient/SVMSpokePoolClient.ts
```diff
@@ -190,19 +195,33 @@ export class SVMSpokePoolClient extends SpokePoolClient {
...
+    }
```
pxrl, line 208: Are there additional concurrency concerns here? Could the value for the same slot end up being accessed & overwritten concurrently?

droplet-rl: No — JS is single-threaded, so the `get` → `set` pair runs atomically inside one microtask. A second `getTimestampForBlock(slot)` invoked while the first is still in flight runs after that synchronous body completes, sees `cached` already populated, and reuses the same Promise (so the underlying RPC fetch fires exactly once even under heavy concurrent dedupe).

The one subtle window is on rejection: between the moment the cached Promise settles as rejected and the moment the attached `.catch()` evicts the entry, another caller could grab the rejected Promise and observe the same error. That's intentional — the caller still gets the correct rejection, and the next call after eviction triggers a fresh resolver. So no `set`-vs-`set` overwrite race, no double-resolve, no stale-success cache.

droplet-rl:
Question, not a change request — replied inline on the PR. Short version: JS is single-threaded so the `get`/`set` pair runs atomically inside one microtask; concurrent callers dedupe onto the same in-flight Promise. The only nuance is a small window between rejection settling and the attached `.catch()` evicting the entry, during which another caller can grab the rejected Promise — that's intentional (caller still observes the correct error, next post-eviction call triggers a fresh resolver). No additional code change.

PR #1443, opened 2026-05-24
fix(svm): preserve SolanaError through QuorumFallback for shouldFailImmediate codes

droplet-rl:
## Summary

- `QuorumFallbackSolanaRpcFactory` wrapped the underlying `SolanaError` in a plain `Error` at two points (final aggregation via `createSendErrorWithMessage`, and the `quorumThreshold === 1 && shouldFailImmediate` short-circuit). Both wrappers erased the `SolanaError` type, so callers' `isSolanaError(err)` checks fell through to default error handling.
- The dataworker hit this on Solana mainnet: `provider.getBlockTime(421829272)` returned `SVM_SLOT_SKIPPED` (network-wide skipped slot, reproducible against the public RPC), but `_callGetTimestampForSlotWithRetry` saw a generic `Error` and re-threw instead of returning `undefined`. That broke `getNearestSlotTime`'s slot walk-back and crashed `zion-across-l2-executor-solana`.
- Both sites now rethrow the original error when `shouldFailImmediate(method, error)` is true. The final aggregation step still wraps with the descriptive `Not enough providers succeeded...` context whenever at least one rejection is a non-deterministic provider failure — that's the case where surfacing the provider URLs and successful peers is genuinely useful.

## Test plan

- [x] New unit tests in `test/SolanaQuorumFallbackProvider.ts` cover: single-rejection skipped slot rethrows the `SolanaError`; multi-provider all-skipped rethrows the first rejection; non-`shouldFailImmediate` failures keep the wrapping path with `cause`; `shouldFailImmediate` codes on unrelated methods still wrap; mixed failures fall through to the wrapper; successful responses still pass through.
- [x] `npx hardhat test test/SolanaQuorumFallbackProvider.ts test/providers/utils.test.ts test/providers/solana/utils.test.ts` — 22 passing.
- [x] `npx hardhat test test/SolanaCachedProvider.ts test/SolanaRateLimitedProvider.ts test/SolanaRetryProvider.ts` — 15 passing (no regressions in the surrounding factory chain).
- [x] `npx tsc --project tsconfig.build.json --noEmit` clean. `prettier --check` and `eslint` clean.

## Background

[...]

PR #1450, opened 2026-05-28
fix(svm): treat BLOCK_NOT_AVAILABLE like SLOT_SKIPPED for getBlock/getBlockTime

droplet-rl:
## Summary

- The relayer's `inventoryManager` crashed on a skipped Solana slot with `Not enough providers succeeded on getBlockTime call. ... Block not available for slot 422715124`. The root cause is an RPC error-code mismatch: third-party RPCs (QuickNode, Chainstack) return `-32004 BLOCK_NOT_AVAILABLE` for skipped slots whose blockstore marker has been pruned locally, while canonical Solana RPCs surface the same physical condition as `-32007 SLOT_SKIPPED`. The SDK only treated the `SLOT_SKIPPED` codes as "no block, move on", so the back-walk in `getNearestSlotTime` threw and the quorum layer wrapped the failure instead of short-circuiting.
- Fix: extend `shouldFailImmediate` (`getBlock` / `getBlockTime` arms) and the `_callGetTimestampForSlotWithRetry` switch to also recognize `SVM_BLOCK_NOT_AVAILABLE`. Constant is already re-exported from `arch/svm/provider`.
- Scoped change: only these two RPC methods. The "slot not yet rooted" semantics of `BLOCK_NOT_AVAILABLE` are preserved everywhere else (it can also be raised by `check_blockstore_root` when `slot >= blockstore.max_root()` per [agave/rpc/src/rpc.rs](https://github.com/anza-xyz/agave/blob/master/rpc/src/rpc.rs)), but for `getBlock` / `getBlockTime` lookups the SDK's callers (`getNearestSlotTime` back-walk, `SVMBlockFinder.getBlock`) treat "no block" as expected and walk back a slot.

## Context

- Triggering incident: relayer error in Slack, slot 422715124 on Solana mainnet.
- Verified directly against `api.mainnet-beta.solana.com`: that slot returns `-32007 SLOT_SKIPPED` ("Slot 422715124 was skipped, or missing due to ledger jump to recent snapshot") — confirming the slot is genuinely skipped, not pruned/missing.
- Canonical error definitions: [`anza-xyz/agave/rpc-client-api/src/custom_error.rs`](https://github.com/anza-xyz/agave/blob/master/rpc-client-api/src/custom_error.rs). `JSON_RPC_SERVER_ERROR_BLOCK_NOT_AVAILABLE = -32004`, `JSON_RPC_SERVER_ERROR_SLOT_SKIPPED = -32007`, `JSON_RPC_SERVER_ERROR_LONG_TERM_STORAGE_SLOT_SKIPPED = -32009`. They are not synonyms canonically — but on the lookup paths that hit `get_block` / `get_block_time`, both `SLOT_SKIPPED` and `BLOCK_NOT_AVAILABLE` indicate "no block here" from the SDK's perspective.
- Companion to #1443, which previously added the `shouldFailImmediate` plumbing for `SLOT_SKIPPED` so callers can branch on `isSolanaError(...)`.

## Test plan

- [x] `yarn hardhat test test/providers/solana/utils.test.ts test/providers/solana/quorumFallbackRpcFactory.test.ts` — 20 passing including three new cases (`shouldFailImmediate` short-circuits on `BLOCK_NOT_AVAILABLE` for `getBlock` and `getBlockTime`; quorum fallback rethrows the underlying `SolanaError` for `BLOCK_NOT_AVAILABLE` on `getBlockTime`).
- [x] `yarn lint-check` clean.
- [x] `yarn build` clean.
- [ ] Bump SDK in `across-protocol/relayer` after release to land in `zion-across-inventory-manager-primary`.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -86,7 +86,12 @@ import {
   toSvmRelayData,
 } from "./";
 import { SvmCpiEventsClient } from "./eventsClient";
-import { SVM_LONG_TERM_STORAGE_SLOT_SKIPPED, SVM_SLOT_SKIPPED, isSolanaError } from "./provider";
+import {
+  SVM_BLOCK_NOT_AVAILABLE,
+  SVM_LONG_TERM_STORAGE_SLOT_SKIPPED,
+  SVM_SLOT_SKIPPED,
+  isSolanaError,
+} from "./provider";
 import { AttestedCCTPMessage, SVMEventNames, SVMProvider, LatestBlockhash, SolanaTransaction } from "./types";
 import {
   getEmergencyDeleteRootBundleRootBundleId,
@@ -182,7 +187,12 @@ async function _callGetTimestampForSlotWithRetry(
     switch (code) {
       case SVM_SLOT_SKIPPED:
       case SVM_LONG_TERM_STORAGE_SLOT_SKIPPED:
+      case SVM_BLOCK_NOT_AVAILABLE:
         // No block available for this slot; caller must decide on how to handle this.
+        // BLOCK_NOT_AVAILABLE is treated as equivalent to SLOT_SKIPPED here because third-party
+        // RPCs (QuickNode, Chainstack) return -32004 for skipped slots once the local blockstore
+        // marker has been pruned, where canonical Solana RPCs return -32007. The back-walk in
+        // getNearestSlotTime relies on undefined being returned for either case.
         return undefined;
 
       default: {
--- a/src/providers/solana/utils.ts
+++ b/src/providers/solana/utils.ts
@@ -2,6 +2,7 @@ import { RpcTransport } from "@solana/rpc-spec";
 import { SolanaRpcApi } from "@solana/kit";
 import {
   isSolanaError,
+  SVM_BLOCK_NOT_AVAILABLE,
   SVM_SLOT_SKIPPED,
   SVM_LONG_TERM_STORAGE_SLOT_SKIPPED,
   SVM_TRANSACTION_PREFLIGHT_FAILURE,
@@ -79,8 +80,12 @@ export function shouldFailImmediate(method: string, error: unknown): boolean {
   switch (method) {
     case "getBlock":
     case "getBlockTime":
-      // No block at the requested slot. This may not be correct for blocks > 1 year old.
-      return [SVM_SLOT_SKIPPED, SVM_LONG_TERM_STORAGE_SLOT_SKIPPED].includes(code);
+      // No block at the requested slot. Some third-party RPCs (e.g. QuickNode, Chainstack) return
+      // BLOCK_NOT_AVAILABLE for skipped slots whose marker has been pruned from the local blockstore
+      // — the same physical condition canonical Solana RPCs surface as SLOT_SKIPPED. Treat all three
+      // codes as deterministic across providers so the quorum layer short-circuits instead of
+      // re-trying every fallback and wrapping the result in a "Not enough providers succeeded" error.
+      return [SVM_BLOCK_NOT_AVAILABLE, SVM_SLOT_SKIPPED, SVM_LONG_TERM_STORAGE_SLOT_SKIPPED].includes(code);
     case "sendTransaction":
       // Preflight failures are deterministic across providers; falling back only succeeds on RPCs that skip preflight.
       return code === SVM_TRANSACTION_PREFLIGHT_FAILURE;
--- a/test/providers/solana/quorumFallbackRpcFactory.test.ts
+++ b/test/providers/solana/quorumFallbackRpcFactory.test.ts
@@ -1,7 +1,11 @@
 import { RpcResponse, RpcTransport } from "@solana/kit";
 import { expect } from "chai";
 import winston from "winston";
-import { SVM_SLOT_SKIPPED, SVM_LONG_TERM_STORAGE_SLOT_SKIPPED } from "../../../src/arch/svm/provider";
+import {
+  SVM_BLOCK_NOT_AVAILABLE,
+  SVM_SLOT_SKIPPED,
+  SVM_LONG_TERM_STORAGE_SLOT_SKIPPED,
+} from "../../../src/arch/svm/provider";
 import { CachedSolanaRpcFactory } from "../../../src/providers/solana/cachedRpcFactory";
 import { QuorumFallbackSolanaRpcFactory } from "../../../src/providers/solana/quorumFallbackRpcFactory";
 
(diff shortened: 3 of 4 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -182,7 +187,12 @@ async function _callGetTimestampForSlotWithRetry(
     switch (code) {
       case SVM_SLOT_SKIPPED:
       case SVM_LONG_TERM_STORAGE_SLOT_SKIPPED:
+      case SVM_BLOCK_NOT_AVAILABLE:
         // No block available for this slot; caller must decide on how to handle this.
+        // BLOCK_NOT_AVAILABLE is treated as equivalent to SLOT_SKIPPED here because third-party
+        // RPCs (QuickNode, Chainstack) return -32004 for skipped slots once the local blockstore
+        // marker has been pruned, where canonical Solana RPCs return -32007. The back-walk in
+        // getNearestSlotTime relies on undefined being returned for either case.
```
pxrl, line 195: This seems dangerous to me. What if we're querying historical state and we just conclude it doesn't exist because of this?

droplet-rl: Fair concern, and worth pinning down before merging.

The architectural backstop is that `shouldFailImmediate` only changes outcomes in two narrow places, neither of which actually concludes "no block" on its own:

1. **`retryRpcFactory`** (`shouldFailImmediate` at `retryRpcFactory.ts:71`) — skip retries against a single provider that returned `BLOCK_NOT_AVAILABLE`. The quorum layer's fallback chain still tries the other providers.
2. **`quorumFallbackRpcFactory`** — the early-throw at line 71 is gated on `quorumThreshold === 1`, and `_getQuorum` returns `nodeQuorumThreshold` (typically >1) for `getBlock`/`getBlockTime`, so that branch never fires for these methods. The only effective path is line 102: rethrow the original `SolanaError` **only when every required provider rejected with a `shouldFailImmediate` code**. A single pruned provider in a multi-provider quorum doesn't trigger it; the other providers' results still win.

So the historical-state risk reduces to: every provider in the quorum returns `BLOCK_NOT_AVAILABLE` for the same slot. At that point `_callGetTimestampForSlotWithRetry` translates it to `undefined`, and `getNearestSlotTime` walks back one slot. For the callers that actually use this:
- `SVMBlockFinder.getBlock` already has "the resolved slot may be rotated backwards if no timestamp exists at the requested slot" baked into its semantics — the interpolation search converges either way.
- `getNearestSlotTime` callers (`relayFillStatus`, `findDeposit`, `BaseAbstractClient`, `PoolRebalanceUtils`) treat the returned slot as "nearest preceding block," which is exactly what canonical `SLOT_SKIPPED` already produces.

The pre-PR behavior on the incident slot was a hard crash (`Not enough providers succeeded on getBlockTime call` propagating up through the relayer's `inventoryManager`). The new behavior on the same code from a single provider is "quorum from the other providers wins"; on the same code from *every* provider, it's a 1-slot back-walk (~400ms drift) rather than a crash.

That said, if you'd prefer to be stricter for historical paths, the alternatives I see:
- (a) Keep as-is — accept the unanimous-pruning edge as a silent back-walk.
- (b) Drop `BLOCK_NOT_AVAILABLE` from `shouldFailImmediate` and only add it to `_callGetTimestampForSlotWithRetry`'s switch. Downside: the quorum layer wraps the error in `Not enough providers succeeded…`, so `isSolanaError(err)` fails at the caller and we're back to the original crash. So (b) doesn't actually help — it just re-introduces the bug for the unanimous-`BLOCK_NOT_AVAILABLE` case (which is the incident scenario, since QuickNode + Chainstack both returned it).
- (c) Log a `warn` (with the originating method, slot, and providers) inside `_callGetTimestampForSlotWithRetry` when we swallow `BLOCK_NOT_AVAILABLE` specifically — so operationally we see if this is firing on truly-historical lookups vs. the recent-skipped-slot case from the incident. Cheap, leaves behavior unchanged, gives us a signal to revisit if the edge becomes real.

I'd lean (a) + (c). Want me to add (c)?

droplet-rl:
Replied to pxrl's review on https://github.com/across-protocol/sdk/pull/1450#pullrequestreview-4380800258.

His concern (inline on `SpokeUtils.ts:190`): treating `BLOCK_NOT_AVAILABLE` like `SLOT_SKIPPED` could silently conclude a historical block doesn't exist.

My reply (posted as a thread reply at https://github.com/across-protocol/sdk/pull/1450#discussion_r3317672395) makes three points:

1. **`shouldFailImmediate` doesn't conclude "no block" on its own.** In `quorumFallbackRpcFactory`, the early-throw (line 71) is gated on `quorumThreshold === 1`; `_getQuorum` returns `nodeQuorumThreshold` (>1) for `getBlock`/`getBlockTime`, so that branch never fires for these methods. The only effective path is line 102: rethrow the original `SolanaError` *only when every required provider rejects with a `shouldFailImmediate` code*. A single pruned provider doesn't trigger it.
2. **The caller-side fallback is a 1-slot back-walk, not "block doesn't exist."** `_callGetTimestampForSlotWithRetry` → `getNearestSlotTime` walks back one slot. `SVMBlockFinder.getBlock` already documents "the resolved slot may be rotated backwards if no timestamp exists" — interpolation search converges. Other callers treat the result as "nearest preceding block," matching canonical `SLOT_SKIPPED` semantics.
3. **The naïve narrower fix doesn't work.** Removing `BLOCK_NOT_AVAILABLE` from quorum's `shouldFailImmediate` while keeping it in the callsite switch re-introduces the crash, because the quorum layer wraps the error and `isSolanaError(err)` at the callsite no longer matches.

Proposed mitigation if pxrl wants a paper trail: add a `warn` log inside `_callGetTimestampForSlotWithRetry` when we swallow `BLOCK_NOT_AVAILABLE` (vs `SLOT_SKIPPED`) so we can tell if it fires on truly-historical lookups vs. the recent-skipped-slot incident scenario. Asked for confirmation before making the code change.

Awaiting pxrl's response. No commits pushed yet.

droplet-rl:
Closing per discussion: empirical probing of canonical Solana RPCs (Solana Labs mainnet-beta, publicnode, **Alchemy**) all return SLOT_SKIPPED / LONG_TERM_STORAGE_SLOT_SKIPPED for the failing slot. Only QuickNode and Chainstack return BLOCK_NOT_AVAILABLE, which is non-canonical behavior — on a canonical Agave node, `check_blockstore_root` short-circuits to `SlotSkipped` before `get_block_time` can fall through to BLOCK_NOT_AVAILABLE. Treating the two codes as equivalent in the SDK would mask legitimate BLOCK_NOT_AVAILABLE failures (slot-not-yet-rooted, bigtable outages). Fix belongs at the operational layer — swap or augment the relayer's Solana RPC quorum with at least one canonical-conforming provider (Alchemy is already wired in for EVM).

PR #1453, opened 2026-05-29
chore(svm): surface JSON-RPC error code in QuorumFallback wrap messages

droplet-rl:
## Summary

Surface the JSON-RPC error code in the wrap message produced by `QuorumFallbackSolanaRpcFactory` when no provider succeeds. Today the per-provider error is stringified with `error.stack || error.toString()`. For a `@solana/kit` `SolanaError`, `toString()` only carries the server message — the numeric code in `context.__code` is dropped, because the Kit's message template for `BLOCK_NOT_AVAILABLE` / `SLOT_SKIPPED` / `LONG_TERM_STORAGE_SLOT_SKIPPED` is the verbatim `$__serverMessage` pass-through ([`anza-xyz/kit/packages/errors/src/messages.ts`](https://github.com/anza-xyz/kit/blob/main/packages/errors/src/messages.ts)).

That means an operator reading a wrapped `Not enough providers succeeded` log can't tell `-32004` (generic "couldn't return a block", inconclusive) apart from `-32007` / `-32009` (canonical positive determinations of skip) when the server message overlaps or is otherwise ambiguous.

This is purely diagnostic — no behavior change.

## Motivation

We recently spent a thread debating whether two providers were systematically returning the wrong error code for a particular slot, based only on the server message text we had in a relayer log. Both providers turned out to be canonical when re-probed directly; the failing slot had hit a transient state-transition window. We'd have skipped the wrong-headed analysis if the wrap had carried the `__code` alongside the message.

## Changes

- New helper `formatRpcError(error: unknown): string` in `src/providers/solana/utils.ts`:
  - For `SolanaError`-shaped values, renders as `SolanaError [<code>]: <server-message>`.
  - Else falls back to the existing `error.stack || error.toString()` behavior.
- `src/providers/solana/quorumFallbackRpcFactory.ts`: use `formatRpcError` at both error-push sites (the per-slot fallback walk and the parallel quorum-fallback path). No other behavior change.

## Test plan

- [x] `yarn hardhat test test/providers/solana/utils.test.ts test/providers/solana/quorumFallbackRpcFactory.test.ts` — 21 passing.
  - Three new `formatRpcError` cases (code+server-message rendering for canonical-skip codes and BLOCK_NOT_AVAILABLE; fallback to `.message` when `__serverMessage` is absent; fallback to stack/toString for non-Solana errors).
  - One new regression in the quorum factory test that exercises the prod-shape scenario (two providers both rejecting with a `SolanaError`, quorum=2 → wrap path) and asserts the wrap message contains `SolanaError [<code>]: <server-message>`.
- [x] `yarn lint-check` clean.
- [x] `yarn build` clean.

### DIFF
--- a/src/providers/solana/quorumFallbackRpcFactory.ts
+++ b/src/providers/solana/quorumFallbackRpcFactory.ts
@@ -57,9 +57,7 @@ export class QuorumFallbackSolanaRpcFactory extends SolanaBaseRpcFactory {
           .transport<TResponse>(...args)
           .then((result): [SolanaClusterRpcFactory, RpcResponse<TResponse>] => [factory.rpcFactory, result])
           .catch((error) => {
-            // Append the provider and error to the error array. For SolanaErrors this surfaces
-            // the JSON-RPC error code so wrap-message readers can tell -32007/-32009 (canonical
-            // skip determinations) apart from -32004 (generic "couldn't return a block").
+            // Preserve the underlying JSON-RPC error code in the wrap message.
             errors.push([factory.rpcFactory, formatRpcError(error)]);
 
             // If all fallback providers fail, then return the last received error.
--- a/src/providers/solana/quorumFallbackRpcFactory.ts
+++ b/src/providers/solana/quorumFallbackRpcFactory.ts
@@ -4,7 +4,7 @@ import { isPromiseFulfilled, isPromiseRejected } from "../../utils/TypeGuards";
 import { compareSvmRpcResults, createSendErrorWithMessage } from "../utils";
 import { CachedSolanaRpcFactory } from "./cachedRpcFactory";
 import { SolanaBaseRpcFactory, SolanaClusterRpcFactory } from "./baseRpcFactories";
-import { shouldFailImmediate } from "./utils";
+import { formatRpcError, shouldFailImmediate } from "./utils";
 
 // This factory stores multiple Cached RPC factories so that users of this factory can specify multiple RPC providers
 // and the factory will fallback through them if any RPC calls fail. This factory also implements quorum logic amongst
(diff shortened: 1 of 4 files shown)


src/providers/solana/utils.ts
```diff
@@ -63,6 +63,24 @@ export interface JitoInterface extends SolanaRpcApi {
...
 }
 
+/**
+ * Render an RPC error for inclusion in log/wrap messages. For SolanaErrors, surfaces the numeric
+ * JSON-RPC error code alongside the server message so downstream readers can distinguish canonical
+ * positive determinations (SLOT_SKIPPED, LONG_TERM_STORAGE_SLOT_SKIPPED) from generic catch-all
+ * codes (BLOCK_NOT_AVAILABLE) without parsing the message string. @solana/kit's default message
+ * template for these codes is `$__serverMessage` — pass-through — so the code is otherwise lost.
+ * Falls back to the stack/toString for non-Solana errors.
+ */
```
pxrl, line 69: Comment too verbose.

droplet-rl:
Trimmed the `formatRpcError` JSDoc down to a two-line one-liner — kept the SolanaError-vs-fallback distinction, dropped the canonical-code rationale (that lives in the PR description / `shouldFailImmediate` table). New comment:

```ts
/**
 * Render an RPC error for log/wrap messages. For SolanaErrors, includes the numeric JSON-RPC
 * code alongside the server message; otherwise falls back to stack/toString.
 */
```

Pushed as `70f57de` on top of the existing commit. `yarn lint-check` clean, 21/21 tests still passing.

src/providers/solana/quorumFallbackRpcFactory.ts
```diff
@@ -57,9 +57,10 @@ export class QuorumFallbackSolanaRpcFactory extends SolanaBaseRpcFactory {
           .transport<TResponse>(...args)
           .then((result): [SolanaClusterRpcFactory, RpcResponse<TResponse>] => [factory.rpcFactory, result])
           .catch((error) => {
-            // Append the provider and error to the error array.
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            errors.push([factory.rpcFactory, (error as any)?.stack || error?.toString()]);
+            // Append the provider and error to the error array. For SolanaErrors this surfaces
+            // the JSON-RPC error code so wrap-message readers can tell -32007/-32009 (canonical
+            // skip determinations) apart from -32004 (generic "couldn't return a block").
```
pxrl: Avoid identifying specific scenarios here. Just note that it aims to preserve the underlying JSON-RPC error code.

droplet-rl:
Trimmed the wrap-site comment in `quorumFallbackRpcFactory.ts:60` per the feedback — dropped the `-32007/-32009` vs `-32004` examples, now just states the intent:

```ts
// Preserve the underlying JSON-RPC error code in the wrap message.
errors.push([factory.rpcFactory, formatRpcError(error)]);
```

Pushed as `9692232`. `yarn lint-check` clean, 8/8 `quorumFallbackRpcFactory` tests passing.

pxrl (approved):

droplet-rl:
Thanks for the review and approval — no further inline feedback to address. Leaving the merge to a human per policy.

PR #1457, opened 2026-06-09
chore: Report invalidFills for determenistic deposits

dijanin-brat:


src/utils/SpokeUtils.ts
```diff
@@ -188,9 +193,13 @@ export async function findInvalidFills(spokePoolClients: {
...
       fills.map(async (fill) => {
-        // Skip fills with unsafe deposit IDs
+        // Fills from unsafeDepositV3() use uint256 deposit IDs. Allow time for the origin-chain
+        // deposit to be indexed before treating the fill as invalid.
         if (isUnsafeDepositId(fill.depositId)) {
-          return null; // Return null for unsafe deposits
+          const fillTimestamp = await spokePoolClient.getTimestampForBlock(fill.blockNumber);
+          if (getCurrentTime() - fillTimestamp < unsafeDepositGracePeriodSec) {
+            return null;
+          }
```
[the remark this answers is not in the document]
dijanin-brat: [a39686f](https://github.com/across-protocol/sdk/pull/1457/commits/a39686f99f0e122cbe817ce8c0508fb195a61f5d) you are right 

@codex can you re-review again. We should look for deposits only in memory now

src/utils/SpokeUtils.ts
```diff
@@ -188,16 +193,29 @@ export async function findInvalidFills(spokePoolClients: {
...
+        const originClient = spokePoolClients[fill.originChainId];
+        const unsafeDeposit = isUnsafeDepositId(fill.depositId);
+
+        // Fills from unsafeDepositV3() use uint256 deposit IDs. Allow time for the origin-chain
+        // deposit to be indexed before treating the fill as invalid.
+        if (unsafeDeposit) {
+          const fillTimestamp = await spokePoolClient.getTimestampForBlock(fill.blockNumber);
+          if (getCurrentTime() - fillTimestamp < unsafeDepositGracePeriodSec) {
+            return null;
+          }
```
pxrl: It's relatively expensive to do this lookup if we already have the deposit - wdyt about deferring the timestamp resolution until after we've verified that we don't actually have a deposit for this fill?

src/utils/SpokeUtils.ts
```diff
@@ -188,16 +193,29 @@ export async function findInvalidFills(spokePoolClients: {
...
-        if (isUnsafeDepositId(fill.depositId)) {
-          return null; // Return null for unsafe deposits
+        const originClient = spokePoolClients[fill.originChainId];
+        const unsafeDeposit = isUnsafeDepositId(fill.depositId);
+
+        // Fills from unsafeDepositV3() use uint256 deposit IDs. Allow time for the origin-chain
+        // deposit to be indexed before treating the fill as invalid.
+        if (unsafeDeposit) {
+          const fillTimestamp = await spokePoolClient.getTimestampForBlock(fill.blockNumber);
+          if (getCurrentTime() - fillTimestamp < unsafeDepositGracePeriodSec) {
```
pxrl: Specifically on timestamp resolution itself, I'm also wondering if there's something cheaper that we can do - i.e. could we reference the `fillDeadline` instead? I'm tbf not sure if this works in all cases. An alternative could be to infer the age of the fill based an interpolation of the delta between `spokePoolClient.latestHeightSearched` and `fill.blockNumber`, where we multiply by the average block time (we should have that on-hand).

dijanin-brat: I don't think its smart to use `fillDeadline` for this.
I think its much better to calculate the timestamp as you have described 
[a66093f](https://github.com/across-protocol/sdk/pull/1457/commits/a66093f5ab5aa34b085144a9fae3594280491626)

src/utils/SpokeUtils.ts
dijanin-brat: yup, make perfect sense
[a66093f](https://github.com/across-protocol/sdk/pull/1457/commits/a66093f5ab5aa34b085144a9fae3594280491626)

dijanin-brat:
@codex re-review

PR #1460, opened 2026-06-11
fix(relayFeeCalculator): keep minDeposit finite when aux native cost is non-zero

droplet-rl:
## Summary

\`relayerFeeDetails\` derived \`minDeposit\` from \`maxGasFeePercent = feeLimitPercent − capitalFeePercent − auxNativeFeePercent\`, where \`auxNativeFeePercent = auxNativeCost / outputAmount\` scales as 1 / outputAmount. Once 4.3.166 added Tron bandwidth via [\`TvmQuery\`](https://github.com/across-protocol/sdk/pull/1459), callers that pass a tiny probe amount started seeing the auxiliary percent explode, \`maxGasFeePercent\` clamp to 0, and \`minDeposit\` return \`MAX_BIG_INT\` for routes that were otherwise feasible.

Reported by @melisaguevara on across-protocol/quote-api#2779:

> GET /api/limits?...&destinationChainId=728126428          (no amount)
>   preview:  minDeposit = 508,066,351,383  ==  maxDeposit  ← route effectively closed
>   prod:     minDeposit =          22,068,563

Root cause: \`/api/limits\` falls back to \`simulationAmount = 100\` (1e-4 USDT) when no \`amount\` is provided. Against Tron's ~3 USDT bandwidth cost that drove \`auxNativeFeePercent\` past \`feeLimitPercent\`, then the closed-form for \`minDeposit\` collapsed.

## Fix

Solve for \`minDeposit\` directly from the absolute aux cost, in input-token decimals:

\`\`\`
minDeposit = (gasFeeTotal + auxFeesInInputDecimals)
           / max(feeLimitPercent − capitalFeePercent, 0)
\`\`\`

The result no longer depends on the probe \`outputAmount\`. When capital fee alone exhausts the fee budget, \`minDeposit = MAX_BIG_INT\` as before.

\`maxGasFeePercent\` is unchanged (still the per-amount remaining budget), so downstream consumers that surface it see no semantic shift.

Refactored aux cost computation into a \`resolveAuxNativeFee\` helper that returns both the absolute and percent forms; \`auxNativeFeePercent\` delegates to it.

## Why not patch quote-api alone

Quote-api shipped a workaround (across-protocol/quote-api#2779 commit \`76317ff8\`) that switches the simulation fallback to a 1000-token-worth amount for TVM destinations. That fixes the immediate \`/api/limits\` symptom but leaves the SDK's \`minDeposit\` math sensitive to caller-side probe sizing — any other consumer that probes with a small amount has the same trap. Fixing the SDK eliminates the foot-gun and lets quote-api drop the special case in a follow-up.

## Test plan

- [x] New unit test \`relayerFeeDetails: minDeposit stays finite when aux native cost dominates a small simulation amount\`:
  - With a query returning non-zero \`getAuxiliaryNativeTokenCost\` and a tiny probe amount, \`minDeposit\` is finite (not \`MAX_BIG_INT\`).
  - \`minDeposit\` returned for a tiny probe equals \`minDeposit\` returned for a real-sized probe — the value is invariant in the caller's amount.
- [x] All existing \`relayFeeCalculator.test.ts\` cases (incl. the canonical 10% fee-limit / 1000 USDC test) still pass — the zero-aux paths behave identically to the old formula.
- [x] \`yarn build\` clean, \`eslint\` + \`prettier\` clean.

### DIFF
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -594,10 +594,11 @@ export class RelayFeeCalculator {
     // fees are taken out. Kept in per-amount terms for backwards compatibility with
     // callers that display it as a live percentage.
     const maxGasFeePercent = max(
-      toBNWei(this.feeLimitPercent / 100)
+      toBNWei(this.feeLimitPercent)
+        .div(100)
         .sub(capitalFeePercent)
         .sub(auxNativeFeePercent),
-      toBN(0)
+      bnZero
     );
 
     // `minDeposit` is the smallest `outputAmount` for which gas, capital, and
@@ -617,7 +618,7 @@ export class RelayFeeCalculator {
     // to `MAX_BIG_INT` for any small simulation amount on chains with non-zero
     // auxiliary cost (e.g. Tron bandwidth) — even though a real deposit would have
     // been feasible. Computing it directly from absolute aux cost avoids that.
-    const minDepositBudget = max(toBNWei(this.feeLimitPercent / 100).sub(capitalFeePercent), bnZero);
+    const minDepositBudget = max(toBNWei(this.feeLimitPercent).div(100).sub(capitalFeePercent), bnZero);
     let minDeposit: BigNumber, isAmountTooLow: boolean;
     if (minDepositBudget.eq(bnZero)) {
       // Capital fee alone has exhausted the budget; no deposit amount can satisfy
--- a/src/relayFeeCalculator/relayFeeCalculator.ts
+++ b/src/relayFeeCalculator/relayFeeCalculator.ts
@@ -299,17 +299,19 @@ export class RelayFeeCalculator {
   }
 
   /**
-   * Calculate the auxiliary native token fee as a % of the amount to relay.
+   * Resolve the auxiliary native token fee in absolute output-token units and as a
+   * percentage of `outputAmount`. The absolute form is amount-invariant (it depends
+   * only on the deposit and price), so callers reasoning about `minDeposit` should
+   * prefer it over the percent — the percent diverges as `outputAmount → 0`.
+   *
    * Treats auxiliary native outlay as value forwarded to user, reported separately.
    */
-  async auxNativeFeePercent(
+  protected async resolveAuxNativeFee(
     deposit: RelayData & { destinationChainId: number },
     outputAmount: BigNumberish,
     outputTokenInfo: TokenInfo,
     _tokenPrice?: number
-  ): Promise<BigNumber> {
-    if (toBN(outputAmount).eq(bnZero)) return MAX_BIG_INT;
-
+  ): Promise<{ auxFeesInToken: BigNumber; auxNativeFeePercent: BigNumber }> {
     let auxNativeCost = bnZero;
     try {
       auxNativeCost = this.queries.getAuxiliaryNativeTokenCost(deposit);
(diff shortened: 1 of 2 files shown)


src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -563,26 +589,43 @@ export class RelayFeeCalculator {
...
+    // The earlier closed-form derived `minDeposit` from the per-amount
+    // `maxGasFeePercent`, which incorporated `auxNativeFeePercent = auxCost /
+    // outputAmount`. That ratio diverges as `outputAmount → 0`, pinning `minDeposit`
+    // to `MAX_BIG_INT` for any small simulation amount on chains with non-zero
+    // auxiliary cost (e.g. Tron bandwidth) — even though a real deposit would have
+    // been feasible. Computing it directly from absolute aux cost avoids that.
+    const minDepositBudget = max(toBNWei(this.feeLimitPercent / 100).sub(capitalFeePercent), toBN(0));
     let minDeposit: BigNumber, isAmountTooLow: boolean;
-    if (maxGasFeePercent.eq(toBN(0))) {
+    if (minDepositBudget.eq(toBN(0))) {
```
pxrl:
```suggestion
    if (minDepositBudget.eq(bnZero)) {
```

src/relayFeeCalculator/relayFeeCalculator.ts
```diff
@@ -563,26 +589,43 @@ export class RelayFeeCalculator {
...
+    //   ⇒ minDeposit = (gasFeeTotal + auxFeesInInputDecimals) /
+    //                  (feeLimitPercent − capitalFeePercent)
+    //
+    // The earlier closed-form derived `minDeposit` from the per-amount
+    // `maxGasFeePercent`, which incorporated `auxNativeFeePercent = auxCost /
+    // outputAmount`. That ratio diverges as `outputAmount → 0`, pinning `minDeposit`
+    // to `MAX_BIG_INT` for any small simulation amount on chains with non-zero
+    // auxiliary cost (e.g. Tron bandwidth) — even though a real deposit would have
+    // been feasible. Computing it directly from absolute aux cost avoids that.
+    const minDepositBudget = max(toBNWei(this.feeLimitPercent / 100).sub(capitalFeePercent), toBN(0));
```
pxrl:
```suggestion
    const minDepositBudget = max(toBNWei(this.feeLimitPercent / 100).sub(capitalFeePercent), bnZero);
```

pxrl: Can the immediate division by 100 produce truncation before the promotion to 18 decimals?

src/relayFeeCalculator/relayFeeCalculator.ts
droplet-rl: Done — switched both `minDepositBudget` references to `bnZero` in the follow-up commit.

droplet-rl:
Thanks for the review. Addressed the inline suggestion in bc09a77 — switched both `minDepositBudget` references (the `max` floor and the `.eq` check) to `bnZero` for consistency with the rest of the file. Left the pre-existing `toBN(0)` on the `maxGasFeePercent` `max` call alone to keep the diff focused; happy to fold that in too if you'd prefer.

src/relayFeeCalculator/relayFeeCalculator.ts
droplet-rl: Good question — short answer is "in theory yes, but in practice the same float-division pattern is already used one block up on the `maxGasFeePercent` line (pre-existing), and for realistic `feeLimitPercent` inputs it round-trips cleanly."

Walking through it: `feeLimitPercent` is a JS `number` (validated `0..100`). `toBNWei(x)` does `parseUnits(x.toString(), 18)`, so the path is `number → number/100 → .toString() → parseUnits` rather than a direct truncation. For typical configured values (e.g. `25`, `0.25`, `0.5`), `(x/100).toString()` produces the shortest decimal that round-trips the float, so `parseUnits` sees a clean string well under 18 decimals. Where it *could* bite is adversarial inputs — e.g. `0.1 + 0.2`-style accumulations — that surface trailing `…0004` digits which then get parsed faithfully. None of that should hit our path today.

If you'd like to remove the float divide entirely (and harden both this line and the pre-existing `maxGasFeePercent` site at the same time), I can switch them to `toBNWei(this.feeLimitPercent).div(100)` so the `/100` happens in BN arithmetic after the 18-decimal promotion. Happy to fold that in here or split it — let me know.

droplet-rl:
Replied on the inline thread for the float-precision question. Short version: the `toBNWei(this.feeLimitPercent / 100)` expression mirrors the pre-existing pattern on the `maxGasFeePercent` line just above; for typical configured values it round-trips cleanly through `parseUnits`, and the realistic precision risk is low. Offered to swap both call sites to `toBNWei(this.feeLimitPercent).div(100)` if you'd prefer to eliminate the float divide entirely — let me know and I'll fold it in.

src/relayFeeCalculator/relayFeeCalculator.ts
pxrl: Yes, apply this change to all places where it can be used in the relayfeecalculator.

droplet-rl: Done in 7510369 — both `maxGasFeePercent` and `minDepositBudget` now use `toBNWei(this.feeLimitPercent).div(100)`, so the divide happens in BN arithmetic after the 18-decimal promotion. Also swapped the remaining `toBN(0)` floor on `maxGasFeePercent` to `bnZero` while in the area. Confirmed `yarn hardhat test test/relayFeeCalculator.test.ts` passes (20/20).

droplet-rl:
Pushed 7510369 — moved the `/100` into BN arithmetic at both `maxGasFeePercent` and `minDepositBudget` call sites (`toBNWei(this.feeLimitPercent).div(100)`), and swapped the remaining `toBN(0)` floor to `bnZero` while in the neighborhood. `yarn hardhat test test/relayFeeCalculator.test.ts` passes 20/20. Replied on the inline thread.

melisaguevara (approved):

droplet-rl:
Thanks for the approval, Melisa!

melisaguevara:
@droplet-rl fix the linter issues

droplet-rl:
Fixed in 4df24c4 — prettier wanted the `toBNWei(this.feeLimitPercent).div(100).sub(capitalFeePercent).sub(auxNativeFeePercent)` chain on a single line. Re-ran `npx prettier --check` and `npx eslint`; both clean now.

PR #1491, opened 2026-07-21
feat(HubPoolClient): expose proposed poolRebalanceLeafCount on PendingRootBundle

droplet-rl:
`PendingRootBundle` only carried `unclaimedPoolRebalanceLeafCount`, the live on-chain counter that decrements as leaves execute. Consumers that need the leaf count the proposer originally committed had to re-resolve it from the `ProposeRootBundle` events.

Surface it directly: add `poolRebalanceLeafCount` to the interface and populate it in `HubPoolClient` from the same `ProposeRootBundle` event the pending bundle is already built from (`mostRecentProposedRootBundle`) — no extra lookup, and it maps to the latest proposal by event log order, so propose→dispute→propose resolves to the live proposal.

Enables `across-protocol/relayer#3610` to read `pendingRootBundle.poolRebalanceLeafCount` and drop the explicit `ProposedRootBundle` parameter threaded through `validateRootBundle`.

Verified: full `build:{cjs,esm,types}` clean, `HubPoolClient.*` suites green (18 passing), lint/prettier clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.4.13",
+  "version": "4.4.14",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "repository": {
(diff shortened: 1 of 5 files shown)


pxrl (approved):

droplet-rl:
Thanks for the review, @pxrl! No inline comments to address, so this is ready to merge whenever you are — I won't self-merge.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=3b56662c59ed6b015511886f52744ebd&trace_start_ts=1784644681&trace_end_ts=1784731381)

PR #1496, opened 2026-07-27
fix(quicknode): Append Avalanche C-Chain path by default

droplet-rl:
## Motivation

QuickNode serves the Avalanche C-Chain EVM RPC beneath the avalanchego API path: `/ext/bc/C/rpc` (HTTPS) and `/ext/bc/C/ws` (WSS). `quicknode.getURL()` currently emits the bare endpoint root (`https://<prefix>.avalanche-mainnet.quiknode.pro/<key>`), which QuickNode answers with **HTTP 404** for every request.

Any deployment relying on SDK URL generation for chain 43114 (i.e. without an explicit `RPC_PROVIDER_QUICKNODE_43114` override) therefore has a permanently dead QuickNode provider on Avalanche. This silently reduced quorum headroom for the Across dataworker (`across-config-15m.json` was missing the override) and was the root cause of the `zion-across-arweave` `Not enough providers agreed to meet quorum` crashes on 2026-07-27, once a Chainstack response-shape change removed the remaining 2-provider agreement. Companion config fix: UMAprotocol/bot-configs#4229.

## Change

- Add a `PATHS` map to `src/providers/quicknode.ts` appending the chain-specific API path per transport; populated for Avalanche (`https` → `/ext/bc/C/rpc`, `wss` → `/ext/bc/C/ws`). All other chains are unchanged.
- Unit tests covering the Avalanche path suffix (both transports) and the no-suffix behavior for mainnet/OP.

Explicit env URL overrides are unaffected — they bypass `getURL()` entirely, so configs that already carry the full URL keep working as before.

## Verification

- `POST eth_blockNumber` against the generated URL **without** the path → HTTP 404 "404 page not found"; **with** `/ext/bc/C/rpc` → valid result.
- WSS handshake against `/ext/bc/C/ws` → `HTTP/1.1 101 Switching Protocols`.
- `hardhat test test/ProviderUrls.test.ts` → 2 passing; eslint + prettier clean on changed files.

### DIFF
--- a/src/providers/quicknode.ts
+++ b/src/providers/quicknode.ts
@@ -1,4 +1,5 @@
 import { CHAIN_IDs, MAINNET_CHAIN_IDs as _MAINNET_CHAIN_IDs, PUBLIC_NETWORKS } from "../constants";
+import { isDefined } from "../utils/TypeGuards";
 import { RPCTransport } from "./types";
 
 const SNOWFLAKES = {
@@ -14,7 +15,7 @@ const MAINNET_CHAIN_IDs = Object.values(_MAINNET_CHAIN_IDs).map(Number);
 
 /* Some chains nest their EVM RPC beneath a chain-specific path (i.e. the avalanchego API structure). */
 const PATHS: { [chainId: number]: { [transport in RPCTransport]: string } } = {
-  [CHAIN_IDs.AVALANCHE]: { https: "/ext/bc/C/rpc", wss: "/ext/bc/C/ws" },
+  [CHAIN_IDs.AVALANCHE]: { https: "ext/bc/C/rpc", wss: "ext/bc/C/ws" },
 };
 
 export function getURL(chainId: number, apiKey: string, transport: RPCTransport): string {
@@ -41,7 +42,9 @@ export function getURL(chainId: number, apiKey: string, transport: RPCTransport)
   }
   chain = chain.toLowerCase().replace(" ", "-");
 
-  const path = PATHS[chainId]?.[transport] ?? "";
+  const path = PATHS[chainId]?.[transport];
 
-  return `${transport}://${prefix}.${chain}.${domain}/${apiKey}${path}`;
+  return isDefined(path)
+    ? `${transport}://${prefix}.${chain}.${domain}/${apiKey}/${path}`
+    : `${transport}://${prefix}.${chain}.${domain}/${apiKey}`;
 }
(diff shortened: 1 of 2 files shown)


src/providers/quicknode.ts
```diff
@@ -36,5 +41,7 @@ export function getURL(chainId: number, apiKey: string, transport: RPCTransport)
   }
   chain = chain.toLowerCase().replace(" ", "-");
 
-  return `${transport}://${prefix}.${chain}.${domain}/${apiKey}`;
+  const path = PATHS[chainId]?.[transport] ?? "";
+
+  return `${transport}://${prefix}.${chain}.${domain}/${apiKey}${path}`;
```
pxrl: Prefer the following approach. Note this implies stripping the leading `/` on linr 17.

```suggestion
  const path = PATHS[chainId]?.[transport];

  return isDefined(path)
    ? `${transport}://${prefix}.${chain}.${domain}/${apiKey}/${path}`
    : `${transport}://${prefix}.${chain}.${domain}/${apiKey};
```

droplet-rl: Done in f10d83a — switched to the `isDefined()` ternary and stripped the leading `/` from the `PATHS` entries. One small deviation from the suggestion: imported `isDefined` from `../utils/TypeGuards` rather than the `../utils` barrel, to avoid pulling the full utils index (incl. typechain) into this module — matching the direct-submodule imports already used in `retryProvider.ts`/`rateLimitedProvider.ts`. `ProviderUrls.test.ts` passes unchanged.

droplet-rl:
Addressed the review feedback in f10d83a: `getURL` now uses `isDefined(path)` with an explicit ternary, and the leading `/` was stripped from the `PATHS` entries as suggested. `isDefined` is imported from `../utils/TypeGuards` directly (rather than the `../utils` barrel) to keep this module's dependency footprint small — details on the inline thread. `test/ProviderUrls.test.ts` passes with the URLs unchanged.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=a3d6c484c5e8a139995b37fe9bf4ad8e&trace_start_ts=1785157706&trace_end_ts=1785244406)

pxrl (approved):

droplet-rl:
Thanks for the review. All checks are green and the PR is mergeable — leaving the merge to you.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=0f8b5d5e4b0853626802925906bd3a4b&trace_start_ts=1785160983&trace_end_ts=1785247683)

pxrl (approved):

droplet-rl:
Thanks — noting this is now merged. Nothing further from my side.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=9672502a5b70be6b74006d720a14ec4a&trace_start_ts=1785161973&trace_end_ts=1785248673)

PR #1498, opened 2026-07-27
fix(address): guard the TVM branch of toAddressType against malformed input

droplet-rl:
## Summary

Hardens the TVM branch of `toAddressType()` so that a malformed address value routed to it cannot throw during decoding.

`toAddressType()` is applied to address fields decoded from on-chain event data, which are not validated before being passed in. The EVM and SVM branches already validate the decoded value and fall through to a non-throwing `RawAddress` when it isn't a valid address. The **TVM branch did not** — it called `TvmAddress.from()` unconditionally, which throws for anything that isn't a valid 20-byte / zero-padded 32-byte address. As a result a single malformed value routed to the TVM branch could abort decoding by throwing.

This came out of a security triage.

## Change

- `toAddressType()`'s TVM branch now follows the same shape as its EVM/SVM neighbours — `chainIsTvm(chainId) && TvmAddress.validate(address)`, no try/catch. A malformed value falls through to `RawAddress`.
- `TvmAddress.validate()` accepts `string | Uint8Array`:
  - **`Uint8Array`** — unchanged from `master`: 20 bytes, or 32 with the upper 12 zeroed. This is the branch the `TvmAddress` constructor and `Address.isValidOn()` use, so their behaviour is untouched.
  - **string, Base58Check (`T…`) or 41-prefixed hex** — verified via `TronWeb.isAddress` (prefix + 4-byte checksum).
  - **string, `0x`-hex** — held to the raw-byte rule; undecodable hex returns `false` rather than throwing.
- The decode of `rawAddress` now happens **after** the TVM branch concludes, rather than ahead of every branch. This is a behaviour fix, not a tidy-up: the 41-prefixed hex form of a TRON address is accepted by `TvmAddress.validate()`, but is usually not valid Base58 (the alphabet excludes `0`), so decoding eagerly threw `Non-base58 character` on a perfectly valid TVM address. Both accepted forms now resolve to the same `TvmAddress`.
- `TvmAddress.validate()`'s `try/catch` wraps only the `utils.arrayify()` call, which is the sole operation that can throw; the byte-branch recursion it guards cannot.
- This also lands the checksum validation flagged as a future nicety when TRON support was added (#1364).
- Adds regression tests: malformed input falls through (no throw), valid 20-byte / zero-padded inputs coerce to `TvmAddress`, and the string forms (valid Base58Check, valid hex, malformed hex, undecodable hex, failed checksum, garbage strings).

### Why the TVM branch validates the string, not `rawAddress`

For a TRON address, the `bs58.decode` at the top of `toAddressType` yields the 25-byte Base58Check payload — `[0x41][20-byte address][4-byte checksum]` — not the canonical address. So `rawAddress` is simply the wrong input for the TVM branch, and the string is the form the checksum is meaningful on.

Note the union does **not** widen what the constructor accepts. `new TvmAddress(...)` passes a `Uint8Array`, hits the byte branch, and still rejects the 25-byte payload — verified directly. An earlier revision of this PR argued otherwise and split the string check into a separate helper on that basis; that reasoning was wrong and has been reverted.

An earlier revision also claimed the Base58Check checksum "cannot be verified from the plain `bs58.decode`". That is **false** — the checksum does survive it, and sha256d over the leading 21 bytes reproduces the stored 4 exactly (verified). Corrected here for the record.

Separately, `TvmAddress.from()` does throw on malformed input (TronWeb's decoder raises `Non-base58 character` and `Invalid address provided`), so the guard is load-bearing rather than belt-and-braces.

## Scope of the guarantee (deliberately narrow)

The fix is placed in `toAddressType()`, so it covers **all six** call sites in `unpackFillEvent()`, not just the one reported. `repaymentChainId` drives the family selection for `relayer` (`SpokeUtils.ts:108`), but `originChainId` is equally attacker-supplied in `relayData`, making `depositor` and `inputToken` (lines 103, 105) an equivalent entry point. Both are closed.

`toAddressType()` is **not** a total function after this change, and intentionally so. Two throw paths remain, both predating this PR:

1. `AddressUtils.ts:136` — `bs58.decode` / `utils.arrayify` throw on a non-`0x` string containing non-Base58 characters, or on odd-length hex. (These run after the TVM branch as of this PR, so they are no longer reachable for a valid TVM address; they still gate the EVM/SVM/`RawAddress` paths.)
2. `new RawAddress(rawAddress)` — a *valid* Base58 string that decodes to more than 32 bytes throws in the `Address` base constructor (`Address … cannot be longer than 32 bytes`).

Neither is reachable from event ingestion: ethers renders `bytes32` event args as well-formed `0x` + 64 hex, and SVM addresses arrive as Base58 of exactly 32 bytes. Both verified by direct probe.

They are left throwing on purpose. Making this function total would require inventing an `Address` for input that has no address in it, and the obvious candidate — a zero/empty `RawAddress` — is actively unsafe, because the zero address is a load-bearing sentinel:

- `SpokePoolClient.ts:540` (and `EVMSpokePoolClient.ts:197`) treat a zero `outputToken` as "resolve the destination token", substituting a **real** token address;
- `DepositUtils.ts:220 invalidOutputToken()` treats zero as "invalid";
- `HubPoolClient.ts:218/262` treat a zero `l2Token` as unsupported.

So a fabricated fallback would silently reinterpret "undecodable" as a meaningful value — and for `relayer`, would put a fabricated address into repayment logic. That is strictly worse than throwing.

The correct boundary for "no single undecodable event can stall ingestion" is therefore the event layer, not the address constructor: drop and log the offending event rather than fabricating a field. That is the follow-up below.

## Follow-ups (not in this PR)

Defense-in-depth, deliberately separate because it is a behaviour change worth its own review — silently dropping events can mask decode regressions and create data gaps in the dataworker:

- wrap per-event decoding (`unpackFillEvent`) so an undecodable event is dropped/logged rather than rejecting the batch (`SpokePoolClient.ts:673` currently maps over every event with no `try/catch` — the file has none);
- give the fleet updater per-chain error isolation instead of a single `Promise.all`.

#1518 (open, stacked on this branch) adds `tryToAddressType()` — a total counterpart returning `Address | undefined` — so callers handling unverified strings can express decode failure without `toAddressType()` fabricating an address. It deliberately does not make `toAddressType()` itself total, for the sentinel reasons set out above.

## Test

`yarn hardhat test test/AddressUtils.ts` — 61 passing. Against `master` at `6fc3b53` (merged in, clean): 87 passing across `AddressUtils` + `Tvm.TransactionUtils` + `SpokeUtils` + `BlockUtils`; lint, prettier and `build:types` clean. Full `yarn test` requires `solana-test-validator` and was left to CI.

Behaviour confirmed by direct probe: on `master`, a `bytes32` of `0xff…ff` with `chainId=728126428` throws; on this branch the same input returns a `RawAddress`, a valid zero-padded value still coerces to `TvmAddress`, and so does a Base58Check `T…` string.

### DIFF
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -127,14 +127,17 @@ export function isValidEvmAddress(address: string): boolean {
  * @todo: Change this to `toAddress` once we remove the other `toAddress` function.
  */
 export function toAddressType(address: string, chainId: number): Address {
-  const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
-
   // Validate before constructing so a malformed value (which may originate from unverified on-chain event data)
   // falls through to a non-throwing RawAddress, matching the EVM/SVM branches. The TVM branch validates the input
   // string rather than rawAddress: for a TRON address bs58.decode yields the 25-byte Base58Check payload
   // (0x41 prefix + address + checksum), not the canonical address, so rawAddress is the wrong input here.
   if (chainIsTvm(chainId) && TvmAddress.validate(address)) return TvmAddress.from(address);
-  else if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
+
+  // Decode only after the TVM case has concluded. TVM never consumes rawAddress, and decoding eagerly throws on the
+  // 41-prefixed hex form of a TRON address: it is valid per TvmAddress.validate(), but is not valid base58, since
+  // the alphabet excludes '0'.
+  const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
+  if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
   else if (chainIsSvm(chainId) && SvmAddress.validate(rawAddress)) return new SvmAddress(rawAddress);
 
   return new RawAddress(rawAddress);
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -119,6 +119,29 @@ export function isValidEvmAddress(address: string): boolean {
   }
 }
 
+// Checks whether a string is a valid TVM address, in either of the two forms `TvmAddress.from()` accepts.
+// Deliberately separate from `TvmAddress.validate()`, which takes bytes like its EVM/SVM counterparts:
+//   - the 4-byte Base58Check checksum is carried by the encoded form, not by the canonical 20-byte (or zero-padded
+//     32-byte) representation that `validate()` governs and the `TvmAddress` constructor consumes, so `validate()`
+//     is not the right place to check it;
+//   - `TvmAddress.from()` throws on malformed input (TronWeb's decoder raises on non-Base58 characters and on bad
+//     lengths), so callers need a total predicate to gate it.
+// Never throws.
+export function isValidTvmAddress(address: string): boolean {
+  // Base58Check (`T…`) or 41-prefixed hex: TronWeb verifies the prefix and the checksum. `TronWeb.isAddress()`
+  // wraps all decoding in try/catch and returns false for any malformed input (verified against tronweb@6.2.2).
+  if (!address.startsWith("0x")) {
+    return TronWeb.isAddress(address);
+  }
+  // 0x-hex: hold to the same rule as the decoded representation. arrayify throws on malformed hex (odd length,
+  // non-hex characters), which means the value is not an address at all.
+  try {
+    return TvmAddress.validate(utils.arrayify(address));
+  } catch {
+    return false;
+  }
+}
+
 /**
  * Creates the proper address type given the input chain ID corresponding to the address's origin network.
  * @param address Stringified address type to convert. Can be either hex encoded or base58 encoded.
@@ -130,9 +153,10 @@ export function toAddressType(address: string, chainId: number): Address {
   const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
 
   // Validate before constructing so a malformed value (which may originate from unverified on-chain event data)
-  // falls through to a non-throwing RawAddress, matching the EVM/SVM branches. The TVM branch validates the input
-  // string rather than rawAddress: TRON uses Base58Check, which the plain bs58.decode above cannot verify.
-  if (chainIsTvm(chainId) && TvmAddress.validate(address)) return TvmAddress.from(address);
+  // falls through to a non-throwing RawAddress, matching the EVM/SVM branches. TVM validates the input string:
+  // for a TRON address the bs58.decode above yields the 25-byte Base58Check payload rather than the canonical
+  // address, so it is not the right input to `TvmAddress.validate()`.
+  if (chainIsTvm(chainId) && isValidTvmAddress(address)) return TvmAddress.from(address);
   else if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
   else if (chainIsSvm(chainId) && SvmAddress.validate(rawAddress)) return new SvmAddress(rawAddress);
 
@@ -349,24 +373,12 @@ export class TvmAddress extends Address {
     super(rawAddress);
   }
 
-  // Validates either a raw address (the 20-byte internal form, or a 32-byte value with the upper 12 bytes zeroed)
-  // or an address string. String validation is checksum-aware: TRON Base58Check addresses are verified via
-  // TronWeb (prefix + checksum), and 0x-hex addresses are checked against the raw-byte rule. Never throws.
-  static validate(address: string | Uint8Array): boolean {
-    if (typeof address !== "string") {
-      return address.length === 20 || (address.length === 32 && address.slice(0, 12).every((field) => field === 0));
-    }
-
-    if (!address.startsWith("0x")) {
-      // TRON Base58Check (or 41-prefixed hex): TronWeb verifies the prefix and 4-byte checksum.
-      return TronWeb.isAddress(address);
-    }
-
-    try {
-      return TvmAddress.validate(utils.arrayify(address));
-    } catch {
-      return false;
-    }
+  // Validates the decoded representation: the 20-byte address, or a 32-byte value with the upper 12 bytes zeroed.
+  // String inputs are checked by `isValidTvmAddress()`, which can additionally verify the Base58Check checksum.
+  static validate(rawAddress: Uint8Array): boolean {
+    return (
+      rawAddress.length == 20 || (rawAddress.length === 32 && rawAddress.slice(0, 12).every((field) => field === 0))
+    );
   }
 
   override isTVM(): this is TvmAddress {
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -129,17 +129,11 @@ export function isValidEvmAddress(address: string): boolean {
 export function toAddressType(address: string, chainId: number): Address {
   const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
 
-  // Address inputs may originate from unverified on-chain event data, so a malformed value must never abort
-  // decoding by throwing here. Mirror the EVM/SVM branches: attempt the typed parse and fall through to a
-  // non-throwing RawAddress on failure. The TVM branch uses TvmAddress.from() (rather than validating rawAddress
-  // above) to preserve TRON base58check parsing, guarded so it cannot throw.
-  if (chainIsTvm(chainId)) {
-    try {
-      return TvmAddress.from(address);
-    } catch {
-      // Malformed TVM address; fall through to RawAddress.
-    }
-  } else if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
+  // Validate before constructing so a malformed value (which may originate from unverified on-chain event data)
+  // falls through to a non-throwing RawAddress, matching the EVM/SVM branches. The TVM branch validates the input
+  // string rather than rawAddress: TRON uses Base58Check, which the plain bs58.decode above cannot verify.
+  if (chainIsTvm(chainId) && TvmAddress.validate(address)) return TvmAddress.from(address);
+  else if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
   else if (chainIsSvm(chainId) && SvmAddress.validate(rawAddress)) return new SvmAddress(rawAddress);
 
   return new RawAddress(rawAddress);
@@ -355,10 +349,22 @@ export class TvmAddress extends Address {
     super(rawAddress);
   }
 
-  static validate(rawAddress: Uint8Array): boolean {
-    return (
-      rawAddress.length == 20 || (rawAddress.length === 32 && rawAddress.slice(0, 12).every((field) => field === 0))
-    );
+  // Validates either a raw address (the 20-byte internal form, or a 32-byte value with the upper 12 bytes zeroed)
+  // or an address string. String validation is checksum-aware: TRON Base58Check addresses are verified via
+  // TronWeb (prefix + checksum), and 0x-hex addresses are checked against the raw-byte rule. Never throws.
+  static validate(address: string | Uint8Array): boolean {
+    if (typeof address === "string") {
+      if (!address.startsWith("0x")) {
+        // TRON Base58Check (or 41-prefixed hex): TronWeb verifies the prefix and 4-byte checksum.
+        return TronWeb.isAddress(address);
+      }
+      try {
+        return TvmAddress.validate(utils.arrayify(address));
+      } catch {
+        return false;
+      }
+    }
+    return address.length == 20 || (address.length === 32 && address.slice(0, 12).every((field) => field === 0));
   }
 
   override isTVM(): this is TvmAddress {
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -353,18 +353,20 @@ export class TvmAddress extends Address {
   // or an address string. String validation is checksum-aware: TRON Base58Check addresses are verified via
   // TronWeb (prefix + checksum), and 0x-hex addresses are checked against the raw-byte rule. Never throws.
   static validate(address: string | Uint8Array): boolean {
-    if (typeof address === "string") {
-      if (!address.startsWith("0x")) {
-        // TRON Base58Check (or 41-prefixed hex): TronWeb verifies the prefix and 4-byte checksum.
-        return TronWeb.isAddress(address);
-      }
-      try {
-        return TvmAddress.validate(utils.arrayify(address));
-      } catch {
-        return false;
-      }
+    if (typeof address !== "string") {
+      return address.length === 20 || (address.length === 32 && address.slice(0, 12).every((field) => field === 0));
     }
-    return address.length == 20 || (address.length === 32 && address.slice(0, 12).every((field) => field === 0));
+
+    if (!address.startsWith("0x")) {
+      // TRON Base58Check (or 41-prefixed hex): TronWeb verifies the prefix and 4-byte checksum.
+      return TronWeb.isAddress(address);
+    }
+
+    try {
+      return TvmAddress.validate(utils.arrayify(address));
+    } catch {
+      return false;
+    }
   }
 
   override isTVM(): this is TvmAddress {
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -363,6 +363,10 @@ export class TvmAddress extends Address {
       return TronWeb.isAddress(address);
     }
 
+    // 0x-hex: hold it to the same rule as the decoded representation. The try/catch is strictly for arrayify(),
+    // which throws on odd-length ("0xabc") and non-hex ("0xnothex") input; the recursive validate() call operates
+    // on bytes and cannot throw. utils.isHexString() would not be a sufficient guard on its own, since it accepts
+    // odd-length hex that arrayify still rejects. This mirrors isValidEvmAddress() above.
     try {
       return TvmAddress.validate(utils.arrayify(address));
     } catch {
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -133,9 +133,6 @@ export function toAddressType(address: string, chainId: number): Address {
   // (0x41 prefix + address + checksum), not the canonical address, so rawAddress is the wrong input here.
   if (chainIsTvm(chainId) && TvmAddress.validate(address)) return TvmAddress.from(address);
 
-  // Decode only after the TVM case has concluded. TVM never consumes rawAddress, and decoding eagerly throws on the
-  // 41-prefixed hex form of a TRON address: it is valid per TvmAddress.validate(), but is not valid base58, since
-  // the alphabet excludes '0'.
   const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
   if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
   else if (chainIsSvm(chainId) && SvmAddress.validate(rawAddress)) return new SvmAddress(rawAddress);
@@ -366,15 +363,15 @@ export class TvmAddress extends Address {
       return TronWeb.isAddress(address);
     }
 
-    // 0x-hex: hold it to the same rule as the decoded representation. The try/catch is strictly for arrayify(),
-    // which throws on odd-length ("0xabc") and non-hex ("0xnothex") input; the recursive validate() call operates
-    // on bytes and cannot throw. utils.isHexString() would not be a sufficient guard on its own, since it accepts
-    // odd-length hex that arrayify still rejects. This mirrors isValidEvmAddress() above.
+    // arrayify throws on odd-length and non-hex input.
+    let rawAddress: Uint8Array;
     try {
-      return TvmAddress.validate(utils.arrayify(address));
+      rawAddress = utils.arrayify(address);
     } catch {
       return false;
     }
+
+    return TvmAddress.validate(rawAddress);
   }
 
   override isTVM(): this is TvmAddress {
--- a/src/utils/AddressUtils.ts
+++ b/src/utils/AddressUtils.ts
@@ -119,29 +119,6 @@ export function isValidEvmAddress(address: string): boolean {
   }
 }
 
-// Checks whether a string is a valid TVM address, in either of the two forms `TvmAddress.from()` accepts.
-// Deliberately separate from `TvmAddress.validate()`, which takes bytes like its EVM/SVM counterparts:
-//   - the 4-byte Base58Check checksum is carried by the encoded form, not by the canonical 20-byte (or zero-padded
-//     32-byte) representation that `validate()` governs and the `TvmAddress` constructor consumes, so `validate()`
-//     is not the right place to check it;
-//   - `TvmAddress.from()` throws on malformed input (TronWeb's decoder raises on non-Base58 characters and on bad
-//     lengths), so callers need a total predicate to gate it.
-// Never throws.
-export function isValidTvmAddress(address: string): boolean {
-  // Base58Check (`T…`) or 41-prefixed hex: TronWeb verifies the prefix and the checksum. `TronWeb.isAddress()`
-  // wraps all decoding in try/catch and returns false for any malformed input (verified against tronweb@6.2.2).
-  if (!address.startsWith("0x")) {
-    return TronWeb.isAddress(address);
-  }
-  // 0x-hex: hold to the same rule as the decoded representation. arrayify throws on malformed hex (odd length,
-  // non-hex characters), which means the value is not an address at all.
-  try {
-    return TvmAddress.validate(utils.arrayify(address));
-  } catch {
-    return false;
-  }
-}
-
 /**
  * Creates the proper address type given the input chain ID corresponding to the address's origin network.
  * @param address Stringified address type to convert. Can be either hex encoded or base58 encoded.
@@ -153,10 +130,10 @@ export function toAddressType(address: string, chainId: number): Address {
   const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
 
   // Validate before constructing so a malformed value (which may originate from unverified on-chain event data)
-  // falls through to a non-throwing RawAddress, matching the EVM/SVM branches. TVM validates the input string:
-  // for a TRON address the bs58.decode above yields the 25-byte Base58Check payload rather than the canonical
-  // address, so it is not the right input to `TvmAddress.validate()`.
-  if (chainIsTvm(chainId) && isValidTvmAddress(address)) return TvmAddress.from(address);
+  // falls through to a non-throwing RawAddress, matching the EVM/SVM branches. The TVM branch validates the input
+  // string rather than rawAddress: for a TRON address bs58.decode yields the 25-byte Base58Check payload
+  // (0x41 prefix + address + checksum), not the canonical address, so rawAddress is the wrong input here.
+  if (chainIsTvm(chainId) && TvmAddress.validate(address)) return TvmAddress.from(address);
   else if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
   else if (chainIsSvm(chainId) && SvmAddress.validate(rawAddress)) return new SvmAddress(rawAddress);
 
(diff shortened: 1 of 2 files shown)


src/utils/AddressUtils.ts
```diff
@@ -346,10 +349,22 @@ export class TvmAddress extends Address {
...
-      rawAddress.length == 20 || (rawAddress.length === 32 && rawAddress.slice(0, 12).every((field) => field === 0))
-    );
+  // Validates either a raw address (the 20-byte internal form, or a 32-byte value with the upper 12 bytes zeroed)
+  // or an address string. String validation is checksum-aware: TRON Base58Check addresses are verified via
+  // TronWeb (prefix + checksum), and 0x-hex addresses are checked against the raw-byte rule. Never throws.
+  static validate(address: string | Uint8Array): boolean {
+    if (typeof address === "string") {
+      if (!address.startsWith("0x")) {
+        // TRON Base58Check (or 41-prefixed hex): TronWeb verifies the prefix and 4-byte checksum.
+        return TronWeb.isAddress(address);
```
pxrl: Can `TronWeb.isAddress()` throw on a bad address? Confirm this.

droplet-rl: Confirmed — `TronWeb.isAddress()` cannot throw, in tronweb@6.2.2 (the version resolved by our lockfile).

**Source:** in `lib/commonjs/utils/address.js`, `isAddress()` returns `false` for falsy/non-string inputs up front, and both decode paths (42-char hex → `getBase58CheckAddress`, and Base58Check → `isAddressValid`) are individually wrapped in `try/catch` returning `false`. The only internal thrower — `decode58`'s `Non-base58 character` error — is inside that catch.

**Empirical:** ran 19 adversarial inputs against it (bad checksum, non-Base58 characters, wrong lengths, empty/whitespace/unicode strings, `null`, `undefined`, numbers, objects, arrays, symbols, functions) — every one returned `false` without throwing, and a genuinely valid Base58Check address returns `true`.

Also added a regression test ("Rejects garbage strings without throwing") so a future tronweb upgrade can't silently regress this guarantee.

droplet-rl:
Thanks for the review. Addressed the one question inline: I verified `TronWeb.isAddress()` cannot throw in tronweb@6.2.2 — both by reading its source (every decode path is wrapped in try/catch returning `false`, including the internal `decode58` that throws on non-Base58 characters) and by running a battery of adversarial inputs against it (garbage strings, wrong lengths, `null`/`undefined`/non-strings — all returned `false`, none threw). Details in the inline reply: https://github.com/across-protocol/sdk/pull/1498#discussion_r3658879194

Pushed one commit adding a regression test ("Rejects garbage strings without throwing") so a future tronweb upgrade can't silently break this. Full AddressUtils suite passes (59 tests), lint and prettier clean.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=bfd5147a0bbbb57658f9add9b07fdfc6&trace_start_ts=1785167649&trace_end_ts=1785254349)

droplet-rl:
Brought up to date with current `master` (`6fc3b53`) — merged in rather than rebased, so @pxrl's review anchor at `3926b6f` stays intact. Clean merge, no conflicts, and re-verified locally: 86 passing across `AddressUtils` + `Tvm.TransactionUtils` + `SpokeUtils` + `BlockUtils`, lint and prettier clean. Worth a fresh CI run since the previous one was against two-week-old `master`, before the TVM work in #1505 landed.

Marking ready for review. @pxrl — your `TronWeb.isAddress()` question is answered [above](https://github.com/across-protocol/sdk/pull/1498#discussion_r3658879194) (cannot throw in tronweb@6.2.2, confirmed by source reading and 19 adversarial inputs, plus a regression test).

Two things I tightened in the description rather than in code, flagging them explicitly since they change what this PR claims:

1. **The old title was overbroad.** It said "malformed inputs cannot throw"; the guarantee is specifically that the *TVM branch* cannot throw. `toAddressType` is still not total — `AddressUtils.ts:130` (`bs58.decode`/`arrayify`) throws on non-Base58 or odd-length hex, and `new RawAddress()` throws on a valid Base58 string decoding to >32 bytes. Both predate this PR; I confirmed by probe that neither is reachable from event ingestion, since ethers renders `bytes32` args as `0x`+64 hex and SVM addresses arrive as 32-byte Base58.

2. **I deliberately did not "fix" those two paths**, because the only way to make the function total is to invent an `Address` for input containing none — and a zero/empty `RawAddress` is unsafe: `SpokePoolClient.ts:540` treats a zero `outputToken` as "resolve the destination token" and substitutes a real token, `invalidOutputToken()` treats zero as invalid, and `HubPoolClient.ts:218` treats a zero `l2Token` as unsupported. A fabricated fallback would reinterpret "undecodable" as a meaningful value, and for `relayer` would feed a fabricated address into repayment logic. Strictly worse than throwing.

So the right boundary for "no single undecodable event stalls ingestion" is the event layer, not the address constructor — the per-event `try/catch` follow-up, kept separate because silently dropping events is a behaviour change that can mask decode regressions and create dataworker data gaps. Happy to open it if you want it in the same batch.

Note the underlying issue is still live on `master`: `AddressUtils.ts` hasn't changed since #1437 (2026-05-19), and neither defense-in-depth follow-up has landed, so this PR is currently the only thing between the bug and prod.

src/utils/AddressUtils.ts
```diff
@@ -346,10 +349,22 @@ export class TvmAddress extends Address {
...
+      }
+      try {
+        return TvmAddress.validate(utils.arrayify(address));
+      } catch {
+        return false;
+      }
+    }
+    return address.length == 20 || (address.length === 32 && address.slice(0, 12).every((field) => field === 0));
```
pxrl: Suggest to invert the logic, such that we impose this branch on `typeof address !== "string"` (implying Uint8Array). Then we can drop a level of indentation from most of the method.

droplet-rl: Good call — done in `aaceec7`. Branching on `typeof address !== "string"` and returning the raw-bytes check early leaves both string paths at a single level of indentation:

```ts
static validate(address: string | Uint8Array): boolean {
  if (typeof address !== "string") {
    return address.length === 20 || (address.length === 32 && address.slice(0, 12).every((field) => field === 0));
  }

  if (!address.startsWith("0x")) {
    // TRON Base58Check (or 41-prefixed hex): TronWeb verifies the prefix and 4-byte checksum.
    return TronWeb.isAddress(address);
  }

  try {
    return TvmAddress.validate(utils.arrayify(address));
  } catch {
    return false;
  }
}
```

Behavior is unchanged — TS still narrows to `Uint8Array` in the early-return branch. Since the length check moved anyway, I also tidied its `== 20` to `=== 20` to match the adjacent `=== 32`; shout if you'd rather keep that diff minimal.

droplet-rl:
Thanks — applied the one suggestion in `aaceec7`.

Inverted the guard in `TvmAddress.validate` to branch on `typeof address !== "string"` and return the raw-bytes check early, so both string paths drop to a single level of indentation. Behavior is unchanged (TS still narrows to `Uint8Array` in the early-return branch); while moving the length check I also tidied `== 20` to `=== 20` to match the adjacent `=== 32`. Detail in the [inline reply](https://github.com/across-protocol/sdk/pull/1498#discussion_r3747877906).

Note I fast-forwarded my local branch onto the master merge you pushed (`fb366f3`) before committing, so my earlier test commit is intact and this sits on top. Re-verified after the merge: AddressUtils suite passes (59 tests), eslint + prettier clean, `tsc` on `tsconfig.build.json` clean.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=34a272570024c0419d52d753f08f64db&trace_start_ts=1786350443&trace_end_ts=1786437143)

droplet-rl:
Good catch — done in aa96566. `TvmAddress.validate()` is back to `(rawAddress: Uint8Array)`, byte-identical to `master` and to `EvmAddress.validate`/`SvmAddress.validate`. The string-level check moved to a module-level `isValidTvmAddress(address: string)`, which follows the `isValidEvmAddress` precedent already in this file — bytes go to `Class.validate`, strings to `isValidXAddress`. Net effect: the diff against `master` is now purely additive, and no `validate()` signature changes at all.

Worth keeping them separate rather than just widening the union, for a reason I hadn't articulated properly: **`validate()` is also the `TvmAddress` constructor's gate.** Accepting the 25-byte Base58Check payload there would let `new TvmAddress(payload)` succeed and store TRON's `0x41` prefix and 4-byte checksum *as address bytes*. So it needs to stay scoped to the canonical 20-byte / zero-padded-32-byte form.

**Also: my original justification in the description was wrong, and I've corrected it.** I'd claimed the Base58Check checksum "cannot be verified from the plain `bs58.decode`". It can — `bs58.decode` yields the whole 25-byte payload `[0x41][20-byte address][4-byte checksum]`, and I confirmed sha256d over the leading 21 bytes reproduces the stored 4 exactly. The checksum was there all along. The real reasons for validating the string are the constructor-gate issue above, plus the fact that `bs58.decode`'s output simply isn't the canonical form for TRON, so it was never the right input to `validate()`.

One thing that fell out of checking this: `TvmAddress.from()` genuinely throws on malformed input — TronWeb's decoder raises `Non-base58 character` and `Invalid address provided` — so the guard is load-bearing, not belt-and-braces.

This supersedes aaceec7, which flattened the nesting inside the widened `validate()`. That function body no longer exists, so the reformatting and its `==`→`===` tidy are moot; I rebased on top of it rather than force-pushing it away. The length check stays as `master` wrote it, which also keeps it byte-identical to `EvmAddress.validate`.

60 passing in `test/AddressUtils.ts` (added one case for hex that can't be decoded at all), 87 across the address/TVM/spoke/block suites, lint + prettier + `build:types` green.

droplet-rl:
Reverted in 274a4b8 — `isValidTvmAddress` is gone, and the TVM branch reads `chainIsTvm(chainId) && TvmAddress.validate(address)` again, matching its EVM/SVM neighbours. This also restores aaceec7's flattened form of `validate()`, which my revert-of-a-revert had trampled.

**My objection to the union was wrong, and it's worth being precise about why** so it doesn't get re-litigated. I argued that widening `validate()` would let the `TvmAddress` constructor accept the 25-byte Base58Check payload and store TRON's `0x41` prefix and checksum as address bytes. That hazard belongs to a *different* design — one where the **byte** branch is widened to accept 25-byte payloads. The union doesn't do that: the constructor passes a `Uint8Array`, hits the byte branch, and behaves exactly as it does on `master`. Verified rather than assumed — `TvmAddress.validate(payload)` is `false` and `new TvmAddress(payload)` still throws.

Two things I kept from the reverted commit, since they're independent of the design question:

- **A factual correction.** An earlier revision of the description claimed the Base58Check checksum "cannot be verified from the plain `bs58.decode`". That's false — it yields the full 25-byte `[0x41][address][checksum]` payload, and sha256d over the leading 21 bytes reproduces the stored 4 exactly. The actual reason the TVM branch validates the string is narrower: `bs58.decode` gives that payload rather than the canonical address, so `rawAddress` is the wrong input here. Comment and description now say that.
- **One extra regression test**, for hex that can't be decoded at all (`0xabc`, `0xnothex`, `0x`) — now targeting `TvmAddress.validate`.

Verified on this revision: poison `bytes32` still falls through to `RawAddress`, and valid 20-byte, zero-padded, and Base58Check `T…` inputs all still coerce to `TvmAddress`. 61 passing in `test/AddressUtils.ts`, 87 across the address/TVM/spoke/block suites, lint + prettier + `build:types` green.

src/utils/AddressUtils.ts
```diff
@@ -346,10 +350,24 @@ export class TvmAddress extends Address {
...
+    if (!address.startsWith("0x")) {
+      // TRON Base58Check (or 41-prefixed hex): TronWeb verifies the prefix and 4-byte checksum.
+      return TronWeb.isAddress(address);
+    }
+
+    try {
+      return TvmAddress.validate(utils.arrayify(address));
+    } catch {
+      return false;
+    }
```
pxrl, line 372: What's the try/catch guard for? We already validated that TvmAddress.validate() doesn't throw. Is it for `arrayify()` ?

droplet-rl: Yes — it is exclusively for `arrayify()`. You are right that the recursive call is not the concern: the byte branch is `length === 20 || (length === 32 && slice().every())`, all total operations on a `Uint8Array`.

`utils.arrayify()` does throw, on two classes of input that reach this line:

```
"0xabc"     => THREW: hex data is odd-length
"0xnothex"  => THREW: invalid arrayify value
"0x"        => OK, len 0   (falls through to a false return)
```

I looked at whether a total predicate could replace the guard, since that would read better. It cannot, quite: `utils.isHexString("0xabc")` returns `true` while `arrayify("0xabc")` still throws, so it would need to be `isHexString(address) && address.length % 2 === 0` — duplicating arrayify's internal rules in a way that can drift. The try/catch is the cleaner total predicate here, and it matches what `isValidEvmAddress()` does a few lines up for the same reason.

The behaviour is already pinned by the "Rejects hex that cannot be decoded at all" test. What was missing is that the code never said any of this, hence the question — fixed in `e48e142`, which names the thrower and the isHexString caveat inline. Comment-only, no behaviour change.

droplet-rl:
Answered the one inline question: yes, the `try/catch` in `TvmAddress.validate` is strictly for `utils.arrayify()`, not the recursion. `arrayify` throws on odd-length (`"0xabc"`) and non-hex (`"0xnothex"`) input; the byte branch it recurses into is only length checks plus `.every()` on a `Uint8Array` and cannot throw.

I checked whether a total predicate could replace the guard and it can't cleanly — `utils.isHexString("0xabc")` is `true` while `arrayify("0xabc")` still throws, so it'd need `isHexString(address) && address.length % 2 === 0`, duplicating arrayify's rules. The try/catch matches what `isValidEvmAddress()` does just above for the same reason.

The behaviour was already pinned by the "Rejects hex that cannot be decoded at all" test; what was missing is that the code never said why the guard exists — that's `e48e142`, comment-only. Detail in the [inline reply](https://github.com/across-protocol/sdk/pull/1498#discussion_r3748390652). 60 passing in `test/AddressUtils.ts`, eslint + prettier clean.

Two things from the branch state worth flagging, since they happened outside this review round:

- `aa96566` (moving the string check into a module-level `isValidTvmAddress`) and its revert `274a4b8` are both mine from an earlier round. Net effect is zero — HEAD is back to the flattened `validate` with the union signature, plus a corrected comment on the `toAddressType` TVM branch and one extra regression test. Nothing for you to re-review there, but the revert message records that my stated objection to the union was wrong, in case it comes up again.
- The PR description still carries the original claim that the Base58Check checksum "cannot be verified from the plain `bs58.decode`". That's false — the checksum does survive the decode. The real reason the TVM branch validates the string is that `bs58.decode` yields the 25-byte payload rather than the canonical address. The code comment is corrected; want me to fix the PR body too?

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=d82050fd198d87e1a7b872ccbaf9de09&trace_start_ts=1786355000&trace_end_ts=1786441700)

src/utils/AddressUtils.ts
```diff
@@ -129,7 +129,11 @@ export function isValidEvmAddress(address: string): boolean {
 export function toAddressType(address: string, chainId: number): Address {
   const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
```
pxrl: Drop this to line ~136, after the TVM case. It's not used by TVM so there's no point in doing it here. This will change the TVM case to be an initial if statement that concludes; the EVM &SVM case will be a subsequent if statement.

src/utils/AddressUtils.ts
pxrl: OK, instead of bundling the arrayify call directly into the validate call, split them apart. Only apply the try/catch wrapper to the arrayify call.

src/utils/AddressUtils.ts
droplet-rl: Done in `ada4a47` — and this turned out to be more than a tidy-up. Moving the decode below the TVM branch fixes a live throw.

The 41-prefixed hex form of a TRON address passes `TvmAddress.validate()`, but is usually *not* valid base58, because the alphabet excludes `0`. Since `rawAddress` was computed eagerly, `bs58.decode()` raised before the TVM branch was ever reached:

```
// before
toAddressType('41e552f6487585c2b58bc2c9bb4492bc1f17132cd0', CHAIN_IDs.TRON)
  => THREW: Non-base58 character

// after
  => OK, isTVM=true, native=TWsm8HtU2A5eEzoT8ev8yaoFjHsXLLrckb
```

So both accepted forms now resolve to the same `TvmAddress`. Shape is as you described — TVM concludes first, then EVM/SVM share the decode:

```ts
if (chainIsTvm(chainId) && TvmAddress.validate(address)) return TvmAddress.from(address);

const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
if (chainIsEvm(chainId) && EvmAddress.validate(rawAddress)) return new EvmAddress(rawAddress);
else if (chainIsSvm(chainId) && SvmAddress.validate(rawAddress)) return new SvmAddress(rawAddress);

return new RawAddress(rawAddress);
```

Nothing else moves: non-TVM chains, and TVM chains given an address `validate()` rejects, still decode exactly as before.

Added a regression test. It pins a fixed address whose hex ends in `0` rather than reusing the random fixture — a random 20-byte address yields base58-safe hex about 8% of the time, so a random fixture would have let the old ordering pass intermittently. Confirmed the test fails with `Non-base58 character` when reverted against the previous ordering.

One case I deliberately left alone: for a TVM chain with a genuinely malformed string (`"garbage!!!"`), the decode on the next line still throws. That behaviour is identical across EVM and SVM and predates this PR, so it felt out of scope here — happy to take it in a follow-up if you'd like `toAddressType` to be total for all inputs.

droplet-rl:
Applied the reorder in `ada4a47`. Worth flagging that it was not just a tidy-up — it fixes a live throw.

The 41-prefixed hex form of a TRON address passes `TvmAddress.validate()`, but is usually not valid base58 (the alphabet excludes `0`). Because `rawAddress` was decoded eagerly, `bs58.decode()` raised before the TVM branch was ever reached:

```
toAddressType('41e552f6487585c2b58bc2c9bb4492bc1f17132cd0', CHAIN_IDs.TRON)
  before => THREW: Non-base58 character
  after  => OK, isTVM=true, native=TWsm8HtU2A5eEzoT8ev8yaoFjHsXLLrckb
```

Both accepted address forms now resolve to the same `TvmAddress`. Nothing else moves — non-TVM chains, and TVM chains given an address `validate()` rejects, decode exactly as before.

Added a regression test pinning a fixed address whose hex ends in `0`, not the random fixture: a random 20-byte address produces base58-safe hex ~8% of the time, so a random one would have let the old ordering pass intermittently. Verified the test fails with `Non-base58 character` against the previous ordering. 61 passing in `test/AddressUtils.ts`, 80 across the address/TVM/spoke suites; eslint, prettier and `build:types` clean. Detail in the [inline reply](https://github.com/across-protocol/sdk/pull/1498#discussion_r3755964839).

Two open items for you, neither blocking:

- For a TVM chain with a genuinely malformed string (`"garbage!!!"`), the decode still throws. That is identical across EVM/SVM and predates this PR, so I scoped it out — say the word if you want `toAddressType` total for all inputs.
- The PR description still carries the incorrect claim that the Base58Check checksum "cannot be verified from the plain `bs58.decode`" (it does survive the decode; the real reason is that the decode yields the 25-byte payload rather than the canonical address). I flagged this last round and left the body alone since it is your call — happy to correct it if you want.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=71f0404b6e8ab4f2476314b08f3c8f4d&trace_start_ts=1786432096&trace_end_ts=1786518796)

src/utils/AddressUtils.ts
pxrl: > One case I deliberately left alone: for a TVM chain with a genuinely malformed string ("garbage!!!"), the decode on the next line still throws. That behaviour is identical across EVM and SVM and predates this PR, so it felt out of scope here — happy to take it in a follow-up if you'd like toAddressType to be total for all inputs.

Yes, follow-up PR.

src/utils/AddressUtils.ts
```diff
@@ -127,10 +127,17 @@ export function isValidEvmAddress(address: string): boolean {
...
 export function toAddressType(address: string, chainId: number): Address {
+  // Validate before constructing so a malformed value (which may originate from unverified on-chain event data)
+  // falls through to a non-throwing RawAddress, matching the EVM/SVM branches. The TVM branch validates the input
+  // string rather than rawAddress: for a TRON address bs58.decode yields the 25-byte Base58Check payload
+  // (0x41 prefix + address + checksum), not the canonical address, so rawAddress is the wrong input here.
+  if (chainIsTvm(chainId) && TvmAddress.validate(address)) return TvmAddress.from(address);
+
+  // Decode only after the TVM case has concluded. TVM never consumes rawAddress, and decoding eagerly throws on the
+  // 41-prefixed hex form of a TRON address: it is valid per TvmAddress.validate(), but is not valid base58, since
+  // the alphabet excludes '0'.
```
pxrl: Squash this comment - not needed.

src/utils/AddressUtils.ts
droplet-rl: Opened as #1518. This PR needs no further changes.

While scoping it I confirmed the remaining throw is not TVM-specific at all — all three families share the decode step, so it reproduces identically on `MAINNET`, `SOLANA` and `TRON`:

| input | `toAddressType(input, …)` |
|---|---|
| `"garbage!!!"` | throws `Non-base58 character` |
| `"0xabc"` | throws `hex data is odd-length` |
| `"0xnothex"` | throws `invalid arrayify value` |

#1518 adds `tryToAddressType()` returning `Address | undefined` and leaves `toAddressType()` untouched, so it is additive.

Two calls in there I want to flag rather than bury, since "total for all inputs" admits a reading I deliberately did not implement:

- I did not widen `toAddressType()` itself to `Address | undefined`. It is exported from the SDK and has 47 internal call sites, none handling `undefined`.
- I did not make it total by coercing undecodable input to an empty `RawAddress`. In code that moves funds, silently turning garbage into an address seems worse than throwing — so the PR makes failure representable rather than making every input yield an address. Say the word if you meant the stronger version.

Also surfaced while probing: `""` and `"0x"` both decode to zero bytes today, so `toAddressType("", chainId)` already returns a **zero-byte RawAddress** instead of throwing. That is the inverse of the bug we have been fixing, and arguably the more dangerous direction. Left it alone since it changes existing behaviour — noted in #1518 if you want it tightened.

droplet-rl:
Understood — opened the follow-up as #1518. **No code changes to this PR this round**; it stays at `ada4a47`.

While scoping the follow-up I confirmed the remaining throw is not TVM-specific: all three families share the decode step, so `"garbage!!!"`, `"0xabc"` and `"0xnothex"` throw identically on `MAINNET`, `SOLANA` and `TRON`. #1518 adds `tryToAddressType(address, chainId): Address | undefined` and leaves `toAddressType()` untouched, so it is additive — 64 passing in `test/AddressUtils.ts` (3 new), lint/prettier/`build:types` clean.

Two judgement calls in #1518 that I flagged rather than decided quietly, because "total for all inputs" admits a stronger reading I chose not to implement:

- I did not widen `toAddressType()` itself to `Address | undefined` — it is exported from the SDK with 47 internal call sites, none handling `undefined`.
- I did not make it total by coercing undecodable input into an empty `RawAddress`. In code that moves funds, silently turning garbage into an address looks worse than throwing, so the PR makes failure *representable* instead. If you meant the stronger version, it is a small change from here.

One thing that fell out of the probing and is worth your attention independently: `""` and `"0x"` both decode to zero bytes today, so `toAddressType("", chainId)` already returns a zero-byte `RawAddress` rather than throwing. That is the inverse of the bug we have been fixing, and arguably the more dangerous direction. I left it alone because it changes existing behaviour, but it is documented in #1518.

#1518 is based on `droplet/harden-toaddresstype-tvm` rather than `master` so its diff shows only the new work; GitHub will retarget it to `master` when this PR merges.

Still outstanding from earlier rounds, your call: the PR description here retains the incorrect claim that the Base58Check checksum "cannot be verified from the plain `bs58.decode`". The code comment is fixed; I have not touched the body.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=a3340aa426686635186d8d3addd175c7&trace_start_ts=1786444654&trace_end_ts=1786531354)

src/utils/AddressUtils.ts
pxrl: @droplet-rl Pinging on this - did you forget?

droplet-rl: Not deliberately — I genuinely missed it, apologies. Done in `0e49ef6`.

Cause, since it will otherwise happen again: I fetch each review with the review-scoped comments endpoint, and your `07:14` comment here landed in a different review batch from the `07:12` one I was answering. So it never appeared in what I pulled, and I did not think to reconcile against the PR-wide comment list. I have now diffed every inline thread on the PR against my replies — this one and `3757228830` ("Squash this comment") were the only two outstanding, and both are fixed in that commit.

On the substance, you are right that splitting them is better:

```ts
// arrayify throws on odd-length and non-hex input.
let rawAddress: Uint8Array;
try {
  rawAddress = utils.arrayify(address);
} catch {
  return false;
}

return TvmAddress.validate(rawAddress);
```

The guard now wraps only the call that can throw, which is what my previous four-line comment was talking around — so that comment collapses to one line. `Rejects hex that cannot be decoded at all` still passes for `0xabc`, `0xnothex` and `0x`; 61 passing, no behaviour change.

src/utils/AddressUtils.ts
droplet-rl: Also missed this one on the last pass — squashed in `0e49ef6`. The three-line block above `rawAddress` is gone; `toAddressType()` now reads:

```ts
if (chainIsTvm(chainId) && TvmAddress.validate(address)) return TvmAddress.from(address);

const rawAddress = address.startsWith("0x") ? utils.arrayify(address) : bs58.decode(address);
```

The ordering constraint it described is not lost — it is pinned by the regression test `Returns a TvmAddress for 41-prefixed hex input that is not valid base58`, whose comment explains why the decode has to stay below the TVM branch. If a future refactor hoists it back up, that test fails with `Non-base58 character` rather than the reasoning having to survive in a comment.

droplet-rl:
You're right, I did forget — apologies. Two comments were outstanding, both now fixed in `0e49ef6`:

- **Split `arrayify` out of the validate call** ([r3755945778](https://github.com/across-protocol/sdk/pull/1498#discussion_r3755945778)) — the try/catch now wraps only `utils.arrayify()`, with the result passed to `TvmAddress.validate()` separately. That is what my earlier four-line comment was talking around, so it collapses to a single line.
- **Squashed the comment above `rawAddress`** ([r3757228830](https://github.com/across-protocol/sdk/pull/1498#discussion_r3757228830)) — gone. The ordering constraint stays pinned by the regression test `Returns a TvmAddress for 41-prefixed hex input that is not valid base58`, so hoisting the decode back up fails a test rather than only contradicting a comment.

No behaviour change — 61 passing in `test/AddressUtils.ts`, eslint/prettier/`build:types` clean.

Worth naming the cause so it doesn't repeat: I fetch each review through the review-scoped comments endpoint, and you posted these in different review batches from the ones I was answering (07:14 vs the 07:12 I replied to, and 10:42:56 vs 10:42:04). They never appeared in what I pulled. I have now reconciled every inline thread on the PR against my replies — those two were the only gaps, and both are closed. Going forward I'll diff against the PR-wide comment list rather than trusting the per-review fetch.

Still open for you, unchanged from earlier rounds: the PR description retains the incorrect claim that the Base58Check checksum "cannot be verified from the plain `bs58.decode`". Happy to correct the body, just haven't touched it without your say-so.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=8e06358ccaf860abeff24d567d21c4b0&trace_start_ts=1786445234&trace_end_ts=1786531934)

pxrl (approved):

droplet-rl:
Thanks for the approval. No code changes this round — HEAD stays at `0e49ef6`, which is pushed, and all four inline threads are resolved with my reply last in each.

**I owe you a correction.** In my last three replies I kept saying the PR description still contained the false claim that the Base58Check checksum "cannot be verified from the plain `bs58.decode`", and offered to fix it. That was wrong: the body already carried the correction, and had for several rounds. I never re-read it after round 3 and kept repeating a stale note. Sorry for the noise.

Re-reading it properly did turn up two things that were genuinely stale, both of which I've now fixed in the body, since a squash-merge makes it the permanent commit message:

- The "throw paths that remain" section said `bs58.decode`/`arrayify` "run before any branch" at `AddressUtils.ts:130`. `ada4a47` moved them *after* the TVM branch and they're now at line 136 — so that paragraph described code this PR had already changed.
- The body never mentioned the 41-prefixed-hex fix at all, which is the one real behaviour fix in here (eager decoding threw `Non-base58 character` on a valid TVM address). Now documented, along with the `try/catch` narrowing.

Also linked #1518 from the follow-ups section so the trail is intact.

Merge-ready state: `mergeable=MERGEABLE`, `mergeState=CLEAN`, 61 passing in `test/AddressUtils.ts` re-verified just now at `0e49ef6`. I'm not merging it — that's yours to do.

#1518 is open and ready whenever you want it. It's based on this branch so its diff shows only the new work; GitHub will retarget it to `master` once this merges.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=4ba0cbf99b82e56b48212330d71cb209&trace_start_ts=1786704435&trace_end_ts=1786791135)

PR #1499, opened 2026-07-28
fix(HubPoolClient): handle zero-address SVM pool rebalance routes

droplet-rl:
## Motivation

A queued HubPool admin batch prunes pool rebalance routes by setting their destination token to the zero address, including the route for USDC on Solana (`setPoolRebalanceRoute(34268394551451, USDC, 0x0)`).

`HubPoolClient`'s `SetPoolRebalanceRoute` event processing requires any SVM-destination event to carry (truncated) SVM USDC as the destination token and **throws** otherwise. A zero-address disable event for Solana therefore makes `HubPoolClient.update()` throw — permanently, for every consumer (dataworker, relayer, finalizer), from the moment the event is mined — since `SetPoolRebalanceRoute` is searched from deep history on every update.

## Changes

- Exempt the zero address from the SVM USDC translation/validation and store it untranslated, matching EVM route-disable semantics. Downstream lookups already handle it: `getL2TokenForL1TokenAtBlock` returns `undefined`, `l2TokenEnabledForL1TokenAtBlock` returns `false`, and `getL1TokenForL2TokenAtBlock` no longer resolves the token from the disable block onwards.
- The mismatch throw is preserved for genuine non-USDC SVM routes (existing test still passes).

## Testing

New test `correctly disables an SVM rebalancing route` covers: update does not throw on the disable event, the route still resolves at pre-disable blocks, and all three lookups correctly report the route as removed at the disable block. `yarn hardhat test test/HubPoolClient.DepositToDestinationToken.ts`: 9 passing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@across-protocol/sdk",
   "author": "UMA Team",
-  "version": "4.4.15",
+  "version": "4.4.16",
   "license": "AGPL-3.0",
   "homepage": "https://docs.across.to/reference/sdk",
   "repository": {
--- a/src/clients/HubPoolClient.ts
+++ b/src/clients/HubPoolClient.ts
@@ -1027,9 +1027,10 @@ export class HubPoolClient extends BaseAbstractClient {
 
         // If the destination chain is SVM, then we need to convert the destination token to the Solana address.
         // This is because the HubPool contract only holds a truncated address for the USDC token and currently
-        // only supports USDC as a destination token for Solana.
+        // only supports USDC as a destination token for Solana. The zero address is exempt from this conversion:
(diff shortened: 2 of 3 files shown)


pxrl (approved):

droplet-rl:
Thanks for the review, @pxrl. No inline comments to address, so no further changes from my side — CI (Builds/Lint/Test) is still running on 37f0b7e; once it's green this is ready to merge.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=703c784da95b3a0d395abbc0b0248e40&trace_start_ts=1785249780&trace_end_ts=1785336480)

PR #1510, opened 2026-08-10
improve(svm): Type SVM events as a name-discriminated union

pxrl:
Map each SvmSpoke event name to its generated event data type and type
EventWithData as a name-discriminated union, so checking event.name statically
narrows event.data. queryEvents() returns EventWithData<T> for the requested
name, and SVMEventNames members are interchangeable with their string literals.
The decoder table's type annotation is tightened so each entry's name/data
correlation is proven at compile time. queryDerivedAddressEvents() deliberately
stays untyped (eventName: string, raw events): it is the generic query for
non-SvmSpoke programs, and the relayer's CCTP flows call it with plain string
event names a regression test pins that call pattern. findDeposit()'s
depositId cast is replaced by typed access, and mock event construction is
checked against the generated types, which surfaced a missing repaymentChainId
field in mocked FilledRelay events.

### DIFF
--- a/src/arch/svm/eventsClient.ts
+++ b/src/arch/svm/eventsClient.ts
@@ -11,7 +11,7 @@ import {
   Signature,
 } from "@solana/kit";
 import { bs58, chainIsSvm, getMessageHash, isDefined, toAddressType } from "../../utils";
-import { EventName, EventWithData, SVMProvider } from "./types";
+import { DecodedEvent, EventName, EventWithData, SVMEventNames, SVMProvider } from "./types";
 import { decodeEvent, isDevnet } from "./utils";
 import { Deposit, DepositWithTime, Fill, FillWithTime } from "../../interfaces";
 import { unwrapEventData } from "./";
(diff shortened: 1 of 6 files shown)


src/arch/svm/eventsClient.ts
```diff
@@ -91,20 +91,27 @@ export class SvmCpiEventsClient {
...
+  public async queryEvents<T extends EventName>(
+    eventName: T,
     fromSlot?: bigint,
     toSlot?: bigint,
     options: GetSignaturesForAddressConfig = { limit: 1000, commitment: "confirmed" }
-  ): Promise<EventWithData[]> {
+  ): Promise<EventWithData<T>[]> {
     const events = await this.queryAllEvents(fromSlot, toSlot, options);
-    return events.filter((event) => event.name === eventName);
+    return events.filter((event): event is EventWithData<T> => event.name === eventName);
```
Reinis-FRP, line 101: Blocking: queryAllEvents() returns raw events because createFor(...) can bind this client to an arbitrary IDL, but this predicate treats name equality alone as proof of the SvmSpoke payload layout. An unrelated program can emit an event named FundsDeposited with different data and TypeScript will expose it as SvmSpokeClient.FundsDeposited. I reproduced this with raw data { fromAnotherProgram: true }: TypeScript accepted data.inputAmount as bigint, while it was undefined at runtime. Could queryEvents() require the exact bundled SvmSpokeIdl before returning typed events, or live on a Spoke-specific client instead?

---

PR #1511, opened 2026-08-10
improve(svm): Compute relay data hashes from relay events

pxrl:
Split getRelayDataHash() into a canonical-input wrapper over a core that does not accept the message: the SvmSpoke hashes relay data with the message replaced by messageHash, precisely so the hash is reconstructible from FilledRelay/RequestedSlowFill events, which carry only the hash. Add getRelayDataHashFromEvent() to do that reconstruction from typed event data, and pin the cross-domain invariant in tests: the hash computed from canonical relay data equals the hash computed from the event a fill of that relay emits. This is the identity that event-to-relay association will be built on.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -972,29 +972,48 @@ export async function getAssociatedTokenAddress(
   return associatedToken;
 }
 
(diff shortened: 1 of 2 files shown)


Reinis-FRP (approved):
LGTM - event-derived relay hashing matches the on-chain construction while preserving existing hash behavior.

PR #1512, opened 2026-08-10
improve(svm): Associate fillStatus PDA events with their relay

pxrl:
Add SvmCpiEventsClient.queryEventsForRelay(), which queries a relay's
fillStatus PDA and associates each returned event back to the relay by
its relay data hash - the same hash the PDA is derived from. Solana only
indexes transactions by account, so PDA queries return all program
events from any transaction touching the PDA, including events for other
relays (e.g. from batched fills); consuming them without association is
the source of the fill status misattribution bug. Signatures are
paginated to exhaustion within the slot bounds, so spamming transactions
at a PDA adds latency but cannot displace the relay's own events.
queryDerivedAddressEvents() remains the generic derived-address query
for non-SvmSpoke programs; its documentation now points SvmSpoke
relay-event callers at queryEventsForRelay(). The SpokePool callers
migrate together with the fill-status fix.


Reinis-FRP (approved):
LGTM - relay events are now safely associated by relay-data hash, preventing cross-relay misattribution from batched transactions.

PR #1513, opened 2026-08-10
fix(svm): Never attribute another relay's events to a fill status

pxrl:
Solana indexes transactions by account, so querying a relay's fillStatus
PDA returns all program events from any transaction touching that PDA -
including other relays' events from batched fills. relayFillStatus() and
findFillEvent() previously consumed those events unfiltered, so a
batched fill could trip findFillEvent's uniqueness assertion or resolve
another relay's event into a false Filled/RequestedSlowFill status.
Thread the relay data through to event reconstruction and resolve events
via queryEventsForRelay(), which associates each event to its relay by
relay data hash. The SpokePool paths no longer issue unassociated PDA
queries; queryDerivedAddressEvents() remains for non-SvmSpoke programs
(the relayer's CCTP flows), documented as requiring caller-side
association.

Commitment is caller-selectable (default confirmed) and applies to the 
signature listing and transaction reads alike. Concurrent transaction
reads are bounded by the provider's rate-limiting queue
(RateLimitedSolanaRpcFactory), mirroring the EVM provider layer. A
caller-supplied fillStatus PDA is only a transaction locator:
association is by relay data hash, so a stale or incorrect PDA can only
yield missing events, never another relay's.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -1036,19 +1036,14 @@ async function resolveFillStatusFromPdaEvents(
     { toSlot, fillStatusPda }
   );
 
-  // Sort events in ascending order of slot number
-  relevantEvents.sort((a, b) => Number(a.slot - b.slot));
-
-  // At this point we have an ordered array of only fill and requested slow fill events and
-  // since it's not possible to submit a slow fill request once a fill has been submitted,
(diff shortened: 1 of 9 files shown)


Reinis-FRP (approved):
LGTM - relay-specific event association and monotonic fill precedence correctly prevent cross-relay and same-slot status misattribution.

PR #1514, opened 2026-08-10
improve(svm): Convert typed events directly to SDK domain types

pxrl:
Add depositFromEvent()/fillFromEvent() to convert typed
FundsDeposited/FilledRelay event data directly to SDK Deposit/Fill
types, replacing the unwrapEventData claim-and-convert chains in
findDeposit, findFillEvent and getDeposit/FillEventsFromSignature. This
fixes three bugs the generic claims hid: the signature helpers converted
addresses with chain IDs read from fields that do not exist on the event
(originChainId on FundsDeposited, destinationChainId on FilledRelay),
yielding RawAddress instead of the native address types, and the fill
helper never converted relayer/updatedRecipient at all. Retire the
legacy unwrap remnants (the number[] byte-array heuristic, the fillType
variant-object branch, isUint8Array); parseEventData stays, since the
generic Anchor path (non-bundled IDLs) still decodes through it. The
differential validation script (committed in the decode PR) keeps the
retired unwrap semantics frozen verbatim as its legacy reference.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -40,7 +40,7 @@ import {
 } from "@solana/kit";
 import assert from "assert";
 import winston from "winston";
-import { arrayify } from "ethers/lib/utils";
+import { arrayify, hexlify } from "ethers/lib/utils";
 import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "../../constants";
 import { DepositWithBlock, FillStatus, FillWithBlock, RelayData, RelayDataWithMessageHash } from "../../interfaces";
 import {
@@ -57,8 +57,7 @@ import {
   isUnsafeDepositId,
   keccak256,
   mapAsync,
-  unpackDepositEvent,
-  unpackFillEvent,
+  toAddressType,
 } from "../../utils";
 import {
   createDefaultTransaction,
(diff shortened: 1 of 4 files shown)


src/arch/svm/SpokeUtils.ts
```diff
@@ -1021,6 +1018,83 @@ export function getRelayDataHashFromEvent(event: SvmRelayEventData, destinationC
...
+  { slot, signature }: EventEnvelope
+): Omit<DepositWithBlock, "quoteBlockNumber" | "fromLiteChain" | "toLiteChain"> {
+  const destinationChainId = Number(data.destinationChainId);
+  const message = hexlify(data.message);
+  return {
+    originChainId,
+    destinationChainId,
+    depositId: BigNumber.from(data.depositId),
+    depositor: toAddressType(data.depositor, originChainId),
+    recipient: toAddressType(data.recipient, destinationChainId),
```
Reinis-FRP, line 1040: Codama decodes every event `Pubkey` as Solana base58, but `toAddressType(..., CHAIN_IDs.TRON)` treats non-hex input as native TRON Base58Check. For a valid SVM→TRON deposit, `data.recipient` is the Solana-base58 encoding of the zero-padded TVM bytes (for example `1111111111113KCdB5mQUd99fY7YzEp79PEQi1ST`), so this line throws `Invalid address provided`; I reproduced that directly through `depositFromEvent`. The same issue affects `outputToken`/`exclusiveRelayer`, plus fill fields when the origin or repayment chain is TRON. The legacy path worked because `unwrapEventData` hexlified the event Pubkey before chain-specific typing. Please normalize these event addresses to raw bytes/hex first and add TVM deposit/fill coverage.

---

PR #1519, opened 2026-08-11
improve(svm): make fill-status resolution commitment caller-selectable

pxrl:
Consumers differ on finality requirements: the dataworker must not act on rollback-able state, while the relayer wants responsiveness. Thread a commitment level (default: confirmed, preserving current behaviour) through relayFillStatus, fillStatusArray and findFillEvent, applied uniformly to the PDA account reads, the slot-time lookups and the event reconstruction, so a single resolution never mixes finality levels. Exposed via SVMSpokePoolClient as a trailing optional, leaving the chain-agnostic SpokePoolClient signature untouched. The dataworker change to pin finalized at its call sites lands in its own repo.

### DIFF
--- a/src/arch/svm/SpokeUtils.ts
+++ b/src/arch/svm/SpokeUtils.ts
@@ -299,8 +299,12 @@ export async function findDeposit(
   >;
 }
 
+// Commitment levels at which fill status may be resolved. "processed" is excluded: the signature listing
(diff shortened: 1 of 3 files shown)


Reinis-FRP (approved):
Looks good - the selected commitment is propagated consistently through all fill-status resolution paths while preserving confirmed as the default.

PR #1530, opened 2026-09-02
feat(providers): move RPC throttling into RateLimitedProvider

droplet-rl:
**RFC / illustrative — not intended to merge as-is.** Replaces an earlier, much heavier sketch of this idea.

Two changes, both in `rateLimitedProvider.ts`:

1. **Migrate** the relayer's `ProviderUtils#rpcRateLimited` (exponential backoff + sampled logging) into the provider that actually owns the queue.
2. **Close the feedback loop** on the queue: a 429 halves `queue.concurrency` as well as backing off, and a successful send widens it by one once the backlog has cleared.

```ts
private async onRateLimited(attempt: number): Promise<boolean> {
  this.queue.concurrency = Math.max(1, Math.floor(this.queue.concurrency / 2));
  // ...existing backoff + sampled log...
}

private restoreConcurrency(): void {
  if (this.queue.concurrency < this.maxConcurrency && this.queue.length() === 0) {
    this.queue.concurrency += 1;
  }
}
```

## Why the backlog is the recovery signal

`queue.length()` counts only *waiting* tasks, so an empty queue means we are no longer outpacing the provider. That makes the loop self-damping without any tuning constants: under sustained load the backlog stays non-empty and the reduced concurrency holds, while an idle provider ramps back one slot per successful send. No cooldown, no config object, no success counters.

## Notes for review

- **The callback is installed ahead of `super()`.** ethers `Object.freeze`s `ConnectionInfo` and defines `connection` as non-writable/non-configurable, so it genuinely cannot be attached to a constructed provider — hence the small `throttle` indirection (`this` doesn't exist pre-`super`).
- **Attempt count is now ethers'.** `connection.throttleLimit` replaces the bespoke `return attempt < retries` cap, so the callback just asks for another attempt. Callers wanting N retries set `throttleLimit: N + 1`.
- **Callers must drop their own throttle config.** The provider now sets both `throttleCallback` and `throttleSlotInterval`, so the relayer's copies should be deleted in a companion PR — these want to land together, or the relayer's logging silently disappears.
- `maxConcurrency` becomes a `readonly` field and there's a public `concurrency` getter, mostly so the behaviour is observable and testable.
- **Rate-limit logging is no longer sampled by default.** Each 429 is a concurrency change, so all of them are logged; `NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS` (the env var the relayer already read) is retained purely as a sampling escape hatch.

## Tests

Four tests driving a real local 429 server through the provider: starts at `maxConcurrency`; narrows to 8 then widens to 9 across a single send that is throttled once and then succeeds; never narrows below 1; and logs both of two consecutive 429s. The existing `test/providers.test.ts` async/queue test still passes.

## Open question

Whether halving is right at all, versus stepping down by one. Halving reacts fast but overshoots, and the ramp back is `maxConcurrency / 2` successful sends — from 25 that's a while. Worth deciding against real 429 rates.

### DIFF
--- a/src/providers/rateLimitedProvider.ts
+++ b/src/providers/rateLimitedProvider.ts
@@ -9,7 +9,12 @@ import { getOriginFromURL } from "../utils/NetworkUtils";
 import { delay } from "../utils/common";
 import winston, { Logger } from "winston";
 
-const LOG_EVERY_N_RATE_LIMIT_ERRORS = 100;
+// Log every Nth rate-limit (429) response. A 429 now also narrows the queue, so each one marks a concurrency change
+// that's worth seeing and the default logs them all. Sampling is retained behind the env var the relayer already
+// used, in case the volume proves excessive. Non-numeric, zero and negative values fall back to logging everything
+// rather than silently muting the log.
+const { NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS } = process.env;
+const logEveryNRateLimitErrors = Math.max(1, Number(NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS) || 1);
 
 // Attaches the provider's own throttleCallback. ethers freezes ConnectionInfo and defines `connection` as
 // non-writable, so this can only be installed ahead of super() — it cannot be patched onto a built provider. The
@@ -87,7 +92,7 @@ export class RateLimitedProvider extends ethers.providers.StaticJsonRpcProvider 
     const baseDelay = Math.pow(2, attempt); // seconds; attempt = [0, 1, 2, ...]
     const retryAfter = baseDelay + baseDelay * Math.random();
 
-    if (this.rateLimitLogCounter++ % LOG_EVERY_N_RATE_LIMIT_ERRORS === 0) {
+    if (this.rateLimitLogCounter++ % logEveryNRateLimitErrors === 0) {
       this.logger.debug({
         at: "RateLimitedProvider#onRateLimited",
         message: `Got rate-limit (429) response on attempt ${attempt}.`,
--- a/src/providers/rateLimitedProvider.ts
+++ b/src/providers/rateLimitedProvider.ts
@@ -6,7 +6,34 @@ import { QueueObject, queue } from "async";
 import { ethers } from "ethers";
 import { RateLimitTask } from "./utils";
 import { getOriginFromURL } from "../utils/NetworkUtils";
+import { delay } from "../utils/common";
 import winston, { Logger } from "winston";
+
+const LOG_EVERY_N_RATE_LIMIT_ERRORS = 100;
+
+// Attaches the provider's own throttleCallback. ethers freezes ConnectionInfo and defines `connection` as
+// non-writable, so this can only be installed ahead of super() — it cannot be patched onto a built provider. The
+// callback is routed via `onThrottle` because `this` does not exist yet at that point.
+function withThrottleCallback(
+  params: ConstructorParameters<typeof ethers.providers.StaticJsonRpcProvider>,
+  onThrottle: (attempt: number) => Promise<boolean>
+): ConstructorParameters<typeof ethers.providers.StaticJsonRpcProvider> {
+  const [connection, network] = params;
+  if (typeof connection !== "object" || connection === null) {
+    return params;
+  }
+
+  return [
+    {
+      ...connection,
+      // Effectively disables ethers' internal backoff algorithm in favour of the one below. Note that ethers still
+      // honours a `Retry-After` header when the provider sends one, independently of this.
+      throttleSlotInterval: 1,
+      throttleCallback: (attempt: number) => onThrottle(attempt),
+    },
+    network,
+  ];
+}
 
 // This provider is a very small addition to the StaticJsonRpcProvider that ensures that no more than `maxConcurrency`
 // requests are ever in flight. It uses the async/queue library to manage this.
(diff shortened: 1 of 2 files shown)


src/providers/rateLimitedProvider.ts
```diff
@@ -6,25 +6,58 @@ import { QueueObject, queue } from "async";
 import { ethers } from "ethers";
 import { RateLimitTask } from "./utils";
 import { getOriginFromURL } from "../utils/NetworkUtils";
+import { delay } from "../utils/common";
 import winston, { Logger } from "winston";
 
+const LOG_EVERY_N_RATE_LIMIT_ERRORS = 100;
```
pxrl: Given that we'll now have an actual feedback loop, I wonder if we should just drop this and log each instance? Perhaps if we default to 1 and permit env-based override in case it causes problems.

droplet-rl: Agreed — done in c125eec. Default is now 1 (every 429 logged), with `NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS` (the same env var the relayer already read) retained as a sampling override:

```ts
const { NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS } = process.env;
const logEveryNRateLimitErrors = Math.max(1, Number(NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS) || 1);
```

The `Math.max`/`|| 1` is there so a non-numeric, empty, zero or negative value degrades to logging everything rather than to `% 0` → `NaN`, which would silently mute the log entirely.

Worth noting the volume this now implies is bounded by attempts, not requests: each logged line is a halving of the queue, so a sustained 429 storm on a `maxConcurrency` of 25 gets ~5 lines before concurrency pins at 1, and then one per retry after that. Also added a test asserting both of two consecutive 429s are logged.

src/providers/rateLimitedProvider.ts
```diff
@@ -6,25 +6,63 @@ import { QueueObject, queue } from "async";
...
 import { RateLimitTask } from "./utils";
 import { getOriginFromURL } from "../utils/NetworkUtils";
+import { delay } from "../utils/common";
 import winston, { Logger } from "winston";
 
+// Log every Nth rate-limit (429) response. A 429 now also narrows the queue, so each one marks a concurrency change
+// that's worth seeing and the default logs them all. Sampling is retained behind the env var the relayer already
+// used, in case the volume proves excessive. Non-numeric, zero and negative values fall back to logging everything
+// rather than silently muting the log.
+const { NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS } = process.env;
```
pxrl:
```suggestion
const { NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS = "1" } = process.env;
```

src/providers/rateLimitedProvider.ts
```diff
@@ -6,25 +6,63 @@ import { QueueObject, queue } from "async";
...
       transports: [new winston.transports.Console()],
     }),
     ...cacheConstructorParams: ConstructorParameters<typeof ethers.providers.StaticJsonRpcProvider>
   ) {
-    super(...cacheConstructorParams);
+    const throttle: { onThrottle: (attempt: number) => Promise<boolean> } = {
+      onThrottle: () => Promise.resolve(true),
+    };
+    super(...withThrottleCallback(cacheConstructorParams, (attempt) => throttle.onThrottle(attempt)));
+    throttle.onThrottle = (attempt) => this.onRateLimited(attempt);
```
pxrl: This is implemented differently to how it works in the relayer repo (ProviderUtils.ts). Can it be aligned with that implementation, or the approach incompatible?

src/providers/rateLimitedProvider.ts
droplet-rl: Applied in 98ff315 — matches `addressAggregator/adapters/abstract.ts`, and the default reads better where you look for it.

Kept the clamp on the next line, since the two cover different inputs. The destructure default only fires on `undefined`, so an explicitly-set-but-bad value still needs handling:

| `NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS` | `Number(...)` | result |
| --- | --- | --- |
| unset | `1` | 1 (destructure default) |
| `""` | `0` | 1 (clamp) |
| `"0"` | `0` | 1 (clamp) |
| `"abc"` | `NaN` | 1 (clamp) |
| `"-5"` | `-5` | 1 (clamp) |
| `"50"` | `50` | 50 |

The `""` and `"0"` rows are the ones worth keeping it for — `n % 0` is `NaN`, so a `NODE_LOG_EVERY_N_RATE_LIMIT_ERRORS=0` meaning "no sampling" would silently mute the log entirely, which is the opposite of what someone setting it to zero would expect. Happy to drop the clamp if you would rather the config just be trusted.

src/providers/rateLimitedProvider.ts
droplet-rl: Partly alignable, and I had made it gratuitously different — fixed in 7f27532.

**What I removed.** The mutable box was just me being wrong about a language rule:

```ts
// before
const throttle = { onThrottle: () => Promise.resolve(true) };
super(...withThrottleCallback(params, (attempt) => throttle.onThrottle(attempt)));
throttle.onThrottle = (attempt) => this.onRateLimited(attempt);

// after
super(...withThrottleCallback(params, (attempt) => this.onRateLimited(attempt)));
```

An arrow captures the `this` *binding*, not its value, so it can call an instance method provided it is not invoked before `super()` returns — and a `throttleCallback` never is, since it needs a request in flight. I checked that this both typechecks and runs rather than assuming it. The `() => Promise.resolve(true)` seed in the old version was dead code for a window that cannot occur.

**What is left, and why it is inherent.** The relayer builds the callback in `getProvider`, before any provider exists:

```ts
const config = { url, timeout, throttleSlotInterval: 1, throttleCallback: rpcRateLimited({ nodeMaxConcurrency, logger }) };
const provider = new RetryProvider(constructorArgumentLists, ...);
```

That works because `rpcRateLimited` only closes over a logger and `nodeMaxConcurrency` as a plain number. The callback here has to *resize the queue that the provider owns*, and ethers gives no way to reach the instance from config built beforehand — confirmed rather than assumed:

```
isFrozen(connection): true
connection desc: {"writable":false,"enumerable":true,"configurable":false}
```

So it cannot be patched on post-construction, which is what forces `withThrottleCallback`. If you would rather keep the relayer's exact shape, the way to get it is to drop the feedback loop and land only the migration — the sampled-logging half ports over as a free function unchanged. Happy to split it that way; the loop is the part I think earns the divergence, but it is your call.

**One real bug found on the way.** The old helper early-returned on a non-object connection, so `new RateLimitedProvider(25, 0, logger, "https://…", 1)` installed no callback at all and silently ran with no feedback loop. ethers accepts `string | ConnectionInfo`; both forms are normalised now, with a test.

PR #1532, opened 2026-09-08
fix(tvm): never lose the txid on a Tron broadcast failure

droplet-rl:
`arch/tvm/submitTransaction()` signs before it broadcasts, so the txid is known locally before `sendRawTransaction()` is called. Two exits threw it away. Tron has no nonce, so a resubmission is a second, independent transaction — the txid is the only handle a caller has, and losing it is a double-execution risk.

1. **A throw from the send discarded the txid.** A transport error or timeout doesn't prove the node refused the transaction. The send now rethrows a `TronBroadcastError` carrying `txid`, so the caller can verify on-chain instead of resubmitting.
2. **`DUP_TRANSACTION_ERROR` was reported as a failed send.** That's Tron's "already known" — the node holds this transaction. Now a successful send with the txid, as the EVM path does.
3. **Rejections carry `code` and `message`.** A bare boolean can't distinguish a definite rejection (`TAPOS_ERROR`) from an ambiguous one (`SERVER_BUSY`). `message` is utf8-decoded, since Tron hex-encodes it.
4. **The native-transfer path had the same defect** via `trx.sendTransaction()`, which fuses build/sign/broadcast. Now three steps, sharing the broadcast handling.

Notes:

- `TronBroadcastError` ships with `isTronBroadcastError()`. Prefer the guard over `instanceof`: the CJS and ESM builds don't share class identity.
- Success shape is unchanged and `code`/`message` are omitted when absent, so existing result assertions hold. The behavioural changes are the new throw type and the duplicate-as-success verdict.
- Not included: a table of which `response_code` values prove a transaction is *not* on-chain. Exposing `code` leaves that policy to the caller; say the word if it belongs here.

Testing: `test/Tvm.TransactionUtils.test.ts`, 13 passing — txid surviving a throw and duplicate-as-success on both the call and transfer paths, decoded `code`/`message`, and the type guard.


src/arch/tvm/TransactionUtils.ts
```diff
@@ -1,10 +1,44 @@
...
+const DUP_TRANSACTION_ERROR = "DUP_TRANSACTION_ERROR";
 
 export interface TronTransactionResult {
   txid: string;
   result: boolean;
+  /** TRON `response_code` for a rejected broadcast (e.g. "TAPOS_ERROR"); absent on success. */
+  code?: string;
+  /** The node's rejection reason, utf8-decoded where TRON hex-encoded it; absent on success. */
+  message?: string;
+}
```
pxrl: If these fields are absent on success, should we instead return a discriminated type on failure vs. success?

droplet-rl: Agreed — done in `832e5bf`.

```ts
export interface TronTransactionSuccess {
  txid: string;
  result: true;
}

export interface TronTransactionFailure {
  txid: string;
  result: false;
  code?: string;
  message?: string;
}

export type TronTransactionResult = TronTransactionSuccess | TronTransactionFailure;
```

`txid` stays common to both branches — it is fixed at signing and reported whichever way the node decides, which is the whole point of the PR.

I checked the narrowing rather than assuming it, since the payoff here is entirely compile-time:

```
error TS2339: Property 'code' does not exist on type 'TronTransactionResult'.      // unnarrowed
error TS2339: Property 'code' does not exist on type 'TronTransactionSuccess'.     // if (r.result)
```

and `if (!r.result) { r.code }` resolves to `string | undefined`.

Both branches are exported so a consumer can name the one it handles rather than reaching for `Extract<...>`.

Two notes on the shape:

- `code` and `message` stay optional *on the failure branch*. A node need not send either, so making them required would be a lie one level down; a rejection is still only ever inferred from `result: false`, never from their absence.
- On blast radius: nothing in-repo consumes this type, and `code`/`message` were introduced by this PR, so no existing reader gets narrowed. The one theoretical break is a caller *constructing* a `TronTransactionResult` with a non-literal `result: boolean` — that no longer assigns. Worth knowing, but there is no such caller today.

I also added a test that reads `code` through the narrowing. It is really a compile-time assertion wearing a runtime test's clothes: flattening the union back to one optional-field interface would still satisfy the `deep.equal` assertions in the neighbouring tests, but would stop that one from type-checking.

One thing I deliberately left alone, since it is adjacent to your question rather than in it. A `DUP_TRANSACTION_ERROR` rejection is mapped to `{ txid, result: true }`, so the success branch cannot express "the node already had this". I think that is right — the contract is "the node holds this txid", and a duplicate satisfies it — but it does mean a caller cannot distinguish a fresh accept from a duplicate. Adding `duplicate?: true` to the success branch later is a non-breaking change, so I would rather not guess at it now. Say the word if you want it.

src/arch/tvm/TransactionUtils.ts
```diff
@@ -1,10 +1,88 @@
...
   txid: string;
-  result: boolean;
+  result: false;
+  /**
+   * TRON `response_code`, resolved to its name (e.g. "TAPOS_ERROR"). Optional because a node need
+   * not send one — a rejection is never inferred from its absence, only from `result: false`.
+   */
+  code?: string;
+  /** The node's rejection reason, utf8-decoded where TRON hex-encoded it. */
+  message?: string;
```
pxrl, line 52: Are these still optional even when returning a discriminated failure type?

droplet-rl: Good push — you were right to ask twice. My previous answer ("a node need not send one") was accurate but it was defending a weak type instead of asking why the weak case existed. Digging into it turned up a bug. `code` is now required, in `d312788`.

The reason a code-less failure was representable at all is that `sendRawTransaction` returns the HTTP body **verbatim**:

```ts
const result = await this.tronWeb.fullNode.request<...>('wallet/broadcasttransaction', signedTransaction, 'post');
return { ...result, transaction: signedTransaction };
```

No normalisation, and notably no `Error` check — TronWeb does guard `if (transaction.Error) throw ...` on its transaction-building paths (`helper.ts:17`, `TransactionBuilder.ts:1075`, `trx.ts:1549`) but *not* here. So TRON's bare `{ Error: "..." }` shape, an empty body, or anything a proxy mangles all reach us as `result` falsy with no `response_code`.

The type question turned out to be the smaller half of it. **None of those responses establish that the transaction was rejected.** The node may have accepted it and failed to say so. Reporting `result: false` there asserts a verdict we do not have, and on TRON a false "it failed" invites the resubmit that — with no nonce to replace through — executes a second time. That is the exact hazard this PR exists to close, and the original code walked into it one branch over from where I was looking.

So rather than make `code` required by inventing a filler value (`OTHER_ERROR` would have been actively harmful — indistinguishable from the node actually saying `OTHER_ERROR`), I routed the case to the error that already means "may or may not have been sent" and carries the txid:

```ts
if (!isDefined(code)) {
  throw new TronBroadcastError(
    `TRON broadcast for ${txID} returned no response code, so its outcome is unknown: ${describeCodelessRejection(broadcast)}`,
    txID
  );
}
```

`describeCodelessRejection` surfaces the `Error` string so the diagnostic is not swallowed. Three outcomes now, and only the middle one is safe to retry blindly: **accepted** (including a duplicate the node already holds), **definitely rejected with a reason**, **unknown**.

With that, a failure is only ever constructed when the node named a reason, so `code: string` is a guarantee rather than a hope. Verified, since the payoff is compile-time:

- `if (!r.result) { const code: string = r.code }` — compiles clean
- `const message: string = r.message` — still `error TS2322: Type 'string | undefined' is not assignable`

`message` I did keep optional, and I want to be explicit that this is a considered split rather than the same hand-wave as before: the protocol's `Return.message` is genuinely empty for some rejections, and it is human prose, not something to branch on — `code` is the machine-readable half, and that is the one now guaranteed.

Behavioural change worth flagging on review: a `{ result: false }` response with no code used to return; it now throws. Only reachable via malformed/proxy responses, and callers already have to handle `TronBroadcastError`, but it is a contract change rather than pure tightening. Test updated accordingly, plus new coverage for the code-less and bare-`Error` shapes. 20 passing, lint and all three `tsc` targets clean.

pxrl:
@droplet-rl Address merge conflicts.

droplet-rl:
@pxrl Merged `master` in — `8131de9`. Worth flagging though: there were no conflicts to resolve.

A trial merge applied cleanly with zero conflicted paths, and GitHub agrees — the PR reports `mergeable: MERGEABLE`. The branch was simply 6 commits behind, which shows up in the UI as *"This branch is out-of-date with the base branch"* and is easy to read as a conflict.

The thing actually holding it is `reviewDecision: REVIEW_REQUIRED`. All four checks pass (Builds, Lint, Test, dependency-review), so the `BLOCKED` in the merge box is the required-approval gate, not a code problem. It needs an approving review from someone — I can't self-approve.

Merging was still worth doing, so I did it: the branch is current and CI now validates against current `master` rather than a two-week-old base.

Your #1539 landed in the middle of this and touches TVM, so I checked for interference rather than assuming the clean merge meant nothing to look at:

- **No file overlap.** #1539 is in `src/arch/tvm/MessageUtils.ts`; mine is `src/arch/tvm/TransactionUtils.ts`. Same directory, disjoint files.
- **No semantic overlap either.** I checked specifically whether `master` had added any consumer of the type I tightened last round, since making `TronTransactionFailure.code` required would break one — `git grep` for `TronTransactionResult` / `TronTransactionFailure` / `submitTransaction` / `isTronBroadcastError` across `src/` on `master` returns nothing outside the file itself. Still no in-repo consumers.
- **Ran your rewritten tests.** `test/relayFeeCalculator.test.ts` — 24 passing, including the new measured-calldata bandwidth cases.

Verified on the merged tree: 20 passing in `test/Tvm.TransactionUtils.test.ts`, 24 in `test/relayFeeCalculator.test.ts`, `yarn lint-check` clean, all three `tsc` targets clean. Effective diff against `master` is unchanged in scope — the same two files.

One deliberate choice: I merged rather than rebased. A rebase would rewrite the commit SHAs that your inline review threads are anchored to and orphan the discussion on `d312788` / `832e5bf`. Happy to rebase and squash instead if you'd rather have linear history, now that the threads are resolved — just say which you prefer.

Still outstanding from the last round, since it's your call and not mine: a `{ result: false }` broadcast response carrying no `response_code` now throws `TronBroadcastError` instead of returning a definite failure. Only reachable via malformed or proxied responses, but it is a contract change rather than pure tightening.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=befe23d52b89de41f57a24c05f6bbd03&trace_start_ts=1790073186&trace_end_ts=1790159886)

PR #1533, opened 2026-09-09
fix: apply configured gas and capital discount percents

### LINKED ISSUE #1449: Configured discount percentages are validated and returned but do not appear applied in fee calculations
### Summary

`gasDiscountPercent` and `capitalDiscountPercent` are accepted as `RelayFeeCalculatorConfig` options, validated in the constructor, and returned in `RelayerFeeDetails`, but they do not appear to be applied to any of the computed fee values. Configuring a non-zero discount therefore has no observable effect on `gasFeeTotal`, `capitalFeeTotal`, or `relayFeeTotal`.

I'm not sure whether this is an oversight (the discount is meant to reduce the fee) or intended (the fields are informational and applied by the caller). Opening this to clarify the intended behavior.

### What I observed

In `src/relayFeeCalculator/relayFeeCalculator.ts`:

- Config options `gasDiscountPercent?` / `capitalDiscountPercent?` (interface, ~L69–70).
- Stored and validated in the constructor (~L157–167): `this.gasDiscountPercent = config.gasDiscountPercent || 0`, with `assert(this.gasDiscountPercent >= 0 && this.gasDiscountPercent <= 100, "gasDiscountPercent must be between 0 and 100 percent")` (and the analogous check for `capitalDiscountPercent`).
- Returned in `RelayerFeeDetails` (~L595 / L601), along with `auxNativeDiscountPercent` (~L561, L598) which is assigned from `this.gasDiscountPercent`.

However, the fee computations don't appear to reference them:

- `gasFeeTotal = gasFeePercent.mul(outToInDecimals(outputAmount)).div(fixedPointAdjustment)` (~L548)
- `capitalFeeTotal = capitalFeePercent.mul(outToInDecimals(outputAmount)).div(fixedPointAdjustment)` (~L555)
- `relayFeePercent` / `relayFeeTotal` are sums of the above, with no discount term (~L563–564)
- `gasFeePercent`, `capitalFeePercent`, and `auxNativeFeePercent` don't reference the discount fields either.

A repo-wide search shows the only references to `gasDiscountPercent` / `capitalDiscountPercent` are the config interface, the stored private fields, the constructor validation, and the echoed output — none inside a fee calculation. I didn't find a test that sets a non-zero discount and asserts a reduced fee, nor documentation describing the fields as pass-through.

### Question

Is the intent that these discounts are applied to the returned fees, or that they're informational (applied downstream by the caller)?

- If they're meant to reduce the fee, the application looks like it may be missing.
- If they're informational, validating them as `0–100` calculator config and returning them without applying them may be a little misleading — documenting that they're pass-through (or removing them) would make the contract clearer.

Happy to open a PR once the intended behavior is confirmed.

devtechedge:
## Summary

`gasDiscountPercent` and `capitalDiscountPercent` were accepted on `RelayFeeCalculatorConfig`, validated in the constructor, and echoed on `RelayerFeeDetails`, but the fee path never multiplied them in. A non-zero config therefore left `gasFeeTotal`, `capitalFeeTotal`, and `relayFeeTotal` unchanged.

This applies the discounts multiplicatively as `(100 - discount) / 100` with BigNumber integer division (round down) on the gas and capital fee percents before totals are derived from them. Auxiliary native fees share the gas discount, matching `auxNativeDiscountPercent`. Downstream aggregates (`relayFee*`, `minDeposit`, `maxGasFeePercent`) then use the discounted values.

## Test plan

- [x] Added unit coverage for mid-range (50%) and full (100%) gas discounts, plus a 50% capital discount against a flat capital fee config
- [x] `yarn hardhat test test/relayFeeCalculator.test.ts` (21 passing)

Fixes #1449


devtechedge:
Hi - soft follow-up on this fee-calculator fix whenever review time opens up.

`gasDiscountPercent` and `capitalDiscountPercent` were validated and echoed on `RelayerFeeDetails` but never applied; this multiplies them in as `(100 - discount) / 100` with BigNumber integer division before totals are derived (`Fixes #1449`).

Auxiliary native fees share the gas discount to match `auxNativeDiscountPercent`, and there is unit coverage for 50% / 100% gas and 50% capital cases (`yarn hardhat test test/relayFeeCalculator.test.ts` was green when I last ran it).

Happy to rebase onto `master`, tighten the rounding notes, or split the capital path if that would ease review.

Whenever you get to it is perfectly fine - the branch is still ready from my side.

PR #1538, opened 2026-09-16
fix: read Solana v1 transactions in SvmCpiEventsClient

droplet-rl:
## Why

Solana activated the **v1 transaction format** on mainnet at epoch 1035 (~01:04 UTC, 2026-09-15). `getTransaction` rejects the *entire request* with JSON-RPC error `-32015` when the transaction it would return is newer than the caller's `maxSupportedTransactionVersion`. This client hardcoded `0` in three places.

**This is already breaking production.** The V4 indexer's `SvmCCTPIndexerDataHandler` has been wedged since 15:11 UTC today:

```
SolanaError: Transaction version (1) is not supported by the requesting client.
  at SvmCpiEventsClient.readEventsFromSignature (src/arch/svm/eventsClient.ts:195)
  at SvmCCTPIndexerDataHandler.fetchEventsByRange (packages/indexer/.../SvmCCTPIndexerDataHandler.ts:411)
```

| indexer | last finalised slot | stalled for |
|---|---|---|
| `cctp:v2:34268394551451` | 447553242 | **~45 min and widening** |
| `DLv3Ngg…:34268394551451` (spoke) | 447561495 | 1s |

It does not self-heal: `Indexer.start` only saves progress after `processBlockRange` resolves, so it retries the same range forever, and the offending signature is permanently in history.

## The key point

**The version ceiling is a property of what we can *read*, not of what we choose to send.**

`#1537` bumped `@solana/kit` to 8.x as the prerequisite for *building* v1 transactions, and was explicit that "no v1 is adopted in this PR" — V4 deliberately stays on v0 because the dataworker refund path needs ALTs. That reasoning is correct for the send path and unchanged by this PR.

But reading isn't opt-in. These clients query **shared** programs — Circle's `TokenMessengerMinter` and `MessageTransmitter`, whose signature history contains every CCTP user's transactions, not just ours. A v1 transaction sent by an unrelated third party is enough to fail the call. That's why CCTP broke first and the Across spoke indexer didn't: the spoke program only ever sees our own still-v0 transactions.

`#1537` is also what makes this fix expressible — kit 5 typed `TransactionVersion` as `'legacy' | 0`, so `: 1` wouldn't have compiled. kit 8.3 types it `'legacy' | 0 | 1`.

## What changed

Three literals → a named constant set to `1`, in `src/arch/svm/eventsClient.ts` (`readEventsFromSignature`, `getDepositEventsFromSignature`, `getFillEventsFromSignature`). These are the only `getTransaction` call sites in `src/`; the SVM block-time path uses `getBlockTime`, which returns no transaction body and is unaffected.

Decoding needs no changes: v1 removes address lookup tables, so `meta.loadedAddresses` comes back empty and the existing `?? []` in `processEventFromTx` already covers it — the static account keys are the whole set. Added a test that locks that in.

## Tests

- `Solana.SvmCpiEventsClient.TransactionVersion.unit.test.ts` — asserts the config we hand the RPC. Verified as a real regression guard: reverting the constant to `0` fails it with `expected +0 to equal 1`.
- `Solana.SvmCpiEventsClient.ForgedEvent.unit.test.ts` — added a case decoding a genuine event when `loadedAddresses` is absent (the v1 shape).

## Verification

- ✅ `tsc --project tsconfig.build.json --noEmit` — 0 errors
- ✅ `eslint src test` + `prettier --check` — clean
- ✅ **376 passing** non-SVM tests, 0 failing
- ✅ **39 passing** SVM unit tests, 0 failing
- ⚠️ **SVM integration tests could not run** — `solana-test-validator` isn't in my sandbox, same gap as #1537. **CI must confirm these.**

One thing a reviewer should weigh in on: `CachedSolanaRpcFactory` keys its Redis cache on `JSON.stringify(params)`, so changing the parameter orphans every existing `getTransaction` entry. No stale-data risk (cold cache, then refill), but those entries are written with `TTL = POSITIVE_INFINITY` and will never be evicted. Worth a manual prefix cleanup if that memory matters.

## Follow-up

`indexer` pins `^4.4.16` with a lockfile at 4.4.16, so it needs a bump to pick this up once published — bumping it *without* this fix would have been a no-op. Happy to open that PR the moment this lands and publishes.


droplet-rl:
Codex finished at 16:12 UTC with **zero findings** — no review body, no inline comments, and a 👍 on the PR, which per its own legend means "all reviews finish with no findings." So there's nothing from it to address, and I've pushed no code changes.

Since that was the one automated round and it came back empty, I verified the change myself rather than take a clean pass at face value.

**The ⚠️ in the description is resolved — CI ran the SVM integration tests and they pass.** The `Test` job installs the Solana toolchain via `solana-developers/github-actions/setup-all`, so `solana-test-validator` is available there (it wasn't in my sandbox). From the job log:

```
SvmCpiEventsClient (integration)                  ✔ 7/7  (real validator)
Svm Cctp Messages (integration)                   ✔ 2/2
SvmCpiEventsClient (transaction version ceiling)  ✔ asks the RPC to deserialize v1 transactions
SvmCpiEventsClient (forged event rejection)       ✔ 3/3, incl. loadedAddresses absent
cached solana provider                            ✔ 6/6
```

All four checks green (Test 5m52s, Builds, Lint, dependency-review).

**Independent checks:**
- The ALT-removal claim is the load-bearing safety argument, so I checked it outside the repo: v1 (SIMD-0385, alongside SIMD-0296's 1232→4096 byte bump) does remove address lookup tables and inline all addresses, activated epoch 1035 / 2026-09-15. The comment is accurate.
- Decoding is actually robust *either way*, which is a stronger claim than the description makes: `messageAccountKeys.push(...(meta?.loadedAddresses?.writable ?? []))` appends whatever is present, so it's correct whether `loadedAddresses` is absent (v1) or populated (v0) — it isn't contingent on v1 emptying it.
- Confirmed those 3 are the only `getTransaction` callers in `src/`, and the block-time path (`BlockUtils`/`SpokeUtils`) uses `getBlockTime`, which returns only slot + timestamp — no transaction body, so no ceiling applies.

**On the cache question you raised:** real, and benign. Caching still *functions* — `isGetTransactionParams` validates the config as `optional(object())` with no shape, so the new param passes validation; only the keys change. No stale-read risk, and not merely because the cache goes cold: `requestAndCacheFinalized` caches only finalized transactions, which are immutable, so both key generations can coexist safely. The orphans are also cheap to isolate — the config serializes in the same key order as before, so the only delta is the literal, and a `SCAN` for `*:getTransaction,*maxSupportedTransactionVersion":0*` under the namespace prefix matches exactly the dead set. Worth a cleanup only if that memory matters. Longer term the key could drop the config and use the signature alone (a finalized tx body is identical for any ceiling ≥ its version, and error responses aren't cached), which would make the cache immune to future ceiling bumps — separate PR, not this one.

**One deliberate non-change:** four test call sites still hardcode `0` — `SVMSpokePoolClient.fills.ts:303`, `Solana.SvmCctpMessages.ts:76`, `Solana.SvmCpiEventsClient.Integration.test.ts:96,113`. Each reads back a v0 transaction the test itself just sent to a local validator, so `0` is correct there and they pass. I left them rather than widen a production hotfix's diff — say the word if you'd rather sweep them for consistency.

Only thing outstanding is human approval (`mergeStateStatus: BLOCKED`); @melisaguevara is already requested. Version bump 4.4.20 → 4.4.21 is clean against master.

Sources for the v1 format claim: [solana.com](https://solana.com/news/transaction-v1-and-the-alt-trade-off), [QuickNode](https://www.quicknode.com/blog/solana-v1-transactions-explained)

dohaki (approved):

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=746eb0d50a807b7f72781e198a8e8fe4&trace_start_ts=1789574712&trace_end_ts=1789661412)

pxrl (approved):

droplet-rl:
Thanks @dohaki — approval noted (and @pxrl's at 16:17). Empty review body and no inline comments on either, so there's nothing to address; no further commits from me. This landed as `b9bd3f5`.

**⚠️ Merging this did not unwedge production. The indexer is still stalled and the gap is widening.**

Current state of `indexer_progress_info` (read just now):

| indexer | last finalised slot | last updated | stalled for |
|---|---|---|---|
| `cctp:v2:34268394551451` | 447553242 | 2026-09-16 15:11:17Z | **1h 22m** |
| `DLv3Ngg…:34268394551451` (spoke) | 447568783 | 2026-09-16 16:33:13Z | 2s |

Still the exact slot and timestamp from the PR description — it has not advanced once. Gap is now ~15.5k slots against the healthy spoke indexer, growing at ~1 slot/400ms. As described, it won't self-heal.

**v4.4.21 is published and live.** Tagged off `b9bd3f5`, published with signed provenance, `dist-tags.latest = 4.4.21` as of ~16:31 UTC. Heads-up: there was ~9 minutes of registry propagation lag (npm's own "may take a few minutes to become available") — the workflow went green at 16:23 but the registry 404'd on 4.4.21 until 16:31. If you checked in that window you'd have seen 4.4.20.

**The remaining fix is narrower than I wrote in the description.** I said the indexer "needs a bump" — it doesn't need a `package.json` change. All three importers (`packages/indexer`, `indexer-api`, `indexer-database`) specify `^4.4.16`, which **already permits 4.4.21**. What's pinning it is `pnpm-lock.yaml`, which resolves the specifier to exactly `4.4.16`. So the whole change is a lockfile re-resolution plus redeploy — no code, no range edit.

Two things I verified so nobody has to re-derive them:

- **The read path is clean.** `RetryProvidersFactory.ts` builds the SVM provider via the SDK's own `providers.CachedSolanaRpcFactory` and types it `arch.svm.SVMProvider`, both imported from `@across-protocol/sdk`. So the fixed `getTransaction` call site ships entirely inside the SDK — the lockfile refresh is sufficient for the fix to take effect, with no indexer code change.
- **One caveat for whoever does it.** 4.4.16 → 4.4.21 also picks up #1537, which moves the transitive `@solana/kit` from 5.5.1 to 8.x. `packages/indexer` separately carries a *direct* `@solana/kit ^2.3.0` (so the tree already double-installs 2.3.0 + 5.5.1 today; this makes it 2.3.0 + 8.x). Not in the CCTP read path, so I don't expect a runtime issue, but worth confirming the workspace still typechecks after re-resolution rather than assuming a patch-level bump.

I've deliberately not touched the indexer repo — it's outside this PR's scope and I'd rather not push cross-repo commits unasked. **Say the word and I'll open that PR immediately**; given it's a lockfile-only change against an actively stalled indexer it should be quick to review. Happy to hand it to whoever owns the redeploy instead if that's faster.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=3f790629e8bb41e0a002d6ef27ba28de&trace_start_ts=1789575151&trace_end_ts=1789661851)

PR #1539, opened 2026-09-18
fix(tvm): correct bandwidth-cost estimation and speed-up signature handling

pxrl:
Corrects `TVM_RAW_DATA_OVERHEAD_BYTES` from 290 to 280 bytes (measured against real onchain fills; the old value overestimated the empty-message case by ~1.4%), then replaces the modeled Tron bandwidth-cost estimate with one measured from the real ABI-encoded calldata.

`TvmQuery.getAuxiliaryNativeTokenCost` now builds the actual `fillRelay`/`fillRelayWithUpdatedDeposit` calldata (via a new synchronous `getV3RelayCalldata`, sharing `resolveV3RelayCall` with the existing `populateV3Relay`) and measures its real length, instead of re-deriving an equivalent byte count from a hand-counted ABI layout. This removes `TVM_FILL_RELAY_FIXED_CALLDATA_BYTES`, `TVM_SPEED_UP_FIXED_CALLDATA_EXTRA_BYTES`, and the padding-length helper entirely — the estimate is exact by construction rather than a model that can drift from the real ABI, and it generalizes to any TVM call shape without its own per-function constant.

`TVM_RAW_DATA_OVERHEAD_BYTES` stays: it's the raw_data envelope + signature, which isn't part of the calldata and isn't known ahead of signing, so it remains a measured constant.

Also tightens a related correctness gap in the shared `resolveV3RelayCall`: a `speedUpSignature` of `"0x"`/`""` can never be a valid ECDSA signature, so it's now rejected as a malformed speed-up — alongside the existing `updatedRecipient`/`updatedOutputAmount`/`updatedMessage` asserts — rather than silently falling back to a plain `fillRelay` on the original terms. Applies to every EVM chain, not just Tron.

`QueryInterface#getAuxiliaryNativeTokenCost`'s parameter type is widened to require `destinationChainId`; EVM and SVM keep their existing synchronous logic unchanged.


dijanin-brat (approved):

test/relayFeeCalculator.test.ts
```diff
@@ -844,10 +844,12 @@ describe("getAuxiliaryNativeTokenCost", function () {
...
-    // Tron bandwidth burns at 1000 SUN/byte. The fixed-tx cost (no message) is the
-    // raw-data envelope + the static portion of the fillRelay ABI calldata: 290 + 4 +
-    // 96 + 384 + 32 = 806 bytes → 806,000 SUN.
-    const FIXED_BANDWIDTH_SUN = 806_000;
+    // Fixed-tx cost (no message) is the raw-data envelope + the static portion of the
+    // fillRelay ABI calldata, at Tron's per-byte bandwidth rate. Derived from the same
+    // constants MessageUtils.ts uses, so this stays in sync if either one changes.
+    const FIXED_BANDWIDTH_SUN =
+      (tvmArch.TVM_RAW_DATA_OVERHEAD_BYTES + tvmArch.TVM_FILL_RELAY_FIXED_CALLDATA_BYTES) *
+      tvmArch.TVM_BANDWIDTH_SUN_PER_BYTE;
```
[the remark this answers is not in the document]
pxrl: Good catch — this was a real gap. Deriving `FIXED_BANDWIDTH_SUN` from the same exported constants made the fixed-cost/scaling tests unable to catch a regression in those constants, and the on-chain calibration test's 1% tolerance also admitted the previous 290-byte value (2,918,000 vs. the observed 2,908,000 is only 0.34% off).

Fixed in 88d06c6e: the calibration test now asserts exact equality against the independent, hardcoded on-chain observation (2,908,000 SUN) instead of a percentage tolerance. Verified it fails against the old 290-byte constant and passes against the corrected 280. That test is now the actual regression guard for this constant; the other tests stay derived since they're checking the arithmetic/formula, not the specific byte value.

src/arch/tvm/MessageUtils.ts
```diff
@@ -4,36 +4,42 @@ import { BigNumber, bnZero, isDefined, isMessageEmpty, toBN } from "../../utils"
...
 // rate is set by Tron governance; refresh from `wallet/getchainparameters` if it drifts.
-const TVM_BANDWIDTH_SUN_PER_BYTE = 1000;
+export const TVM_BANDWIDTH_SUN_PER_BYTE = 1000;
 
 // Tron raw_data envelope (contract type, ref_block, timestamp, expiration, fee_limit)
-// plus a single ECDSA signature. Measured empirically against `fillRelay` transactions
-// emitted by the production relayer; varies by ~10 bytes between calls.
-const TVM_RAW_DATA_OVERHEAD_BYTES = 290;
+// plus a single ECDSA signature. Measured at 280 bytes.
+export const TVM_RAW_DATA_OVERHEAD_BYTES = 280;
```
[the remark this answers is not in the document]
pxrl: Thanks for flagging this — worth clarifying the basis for 280, since the old comment's "varies by ~10 bytes" claim (which this PR removes) is exactly what motivated a closer look.

This wasn't a single sampled transaction: it's measured against the full on-chain history of this contract's `fillRelay` calls (838 fills spanning ~4 months, every sender address that has ever used it). Empty-message overhead is 279 bytes with zero variance across 670 samples; message-bearing fills show 279–280 across 168 samples. So the real variance is ≤1 byte, not ~10 — the old comment's uncertainty was overstated. 280 (not 279) is used specifically to keep the message-bearing case covered with a byte of margin, and the on-chain calibration test (now tightened per the other thread) confirms it matches an independent observed fill exactly.

An upper-bound-style estimate would reintroduce the same problem this PR fixes: the old 290 was already an attempt at a conservative bound, and it turned out to be wrong by enough to cause real false declines downstream. Given the measured distribution is this tight, I'd rather trust it than pad further without evidence of a case it doesn't cover.

pxrl:
@droplet-rl PR #1540 has been squash-landed onto this branch (`509cf503`), so this PR now carries the full calldata-measurement refactor on top of the original 290->280 byte fix, and it's this PR (targeting `master`) that actually ships. Could you take a pass here?

Context: `509cf503` went through 4 rounds of your review on #1540 before it landed — found and fixed: an async/sync interface risk, `message: ""` and `speedUpSignature: "0x"`/`""` throwing in the real ABI encoder, a `hexDataLength` null-vs-throw correction, a factually wrong code comment about the 796,000 SUN test value (and a follow-up precision fix on the 279-vs-280 envelope variance), and — the last one — `updatedMessage: ""` crashing the same way as the other two string fields, plus making the fail-soft-vs-fail-hard choice on a missing `updatedMessage` deliberate. Your last comment on #1540 was "Ship this."

Since this is the branch that ships to master, flagging in case anything reads differently in that context (e.g. interaction with the two earlier commits it's now stacked on).

src/arch/tvm/MessageUtils.ts
```diff
@@ -1,72 +1,26 @@
...
-//   96   65-byte ECDSA signature padded to 3 EVM words
-const TVM_SPEED_UP_FIXED_CALLDATA_EXTRA_BYTES = 128 + 32 + 32 + 96;
-
-function paddedMessageByteLength(message: string | undefined): number {
-  if (!isDefined(message) || isMessageEmpty(message)) return 0;
-  const bytes = Math.max(0, Math.floor((message.length - 2) / 2));
-  return Math.ceil(bytes / 32) * 32;
-}
+// Tron raw_data envelope + one ECDSA signature. Varies with protobuf varint widths: 279
+// bytes for 670 empty-message fills, 280 for the 2112-byte test fill — 280 is the observed max.
```
droplet-rl: What actually drives the 279 → 280 difference? It isn't calldata size via nested length prefixes — both 516 and 2628 bytes sit in the 2-byte varint range (128..16383), so those prefixes are the same width in both samples. The most likely candidate is `fee_limit`, whose varint widens with the energy estimate rather than with calldata.

If that's the cause, "280 is the observed max" is weaker than it reads: it's the max over two buckets, one of which is a single transaction, and a relayer running a higher `fee_limit` than the sampled fills would serialize wider still. Also worth noting that the old code was explicitly conservative at every size, whereas 280 is now *exact* for the 2112-byte fill — so there's no headroom left at the top end.

The residual error is ~1000 SUN either way, so I wouldn't change the number. Just suggest softening the comment to say what varies and why, rather than asserting a max from n=2.

src/relayFeeCalculator/chain-queries/tvmQuery.ts
```diff
@@ -1,20 +1,32 @@
...
+  override getAuxiliaryNativeTokenCost(
+    deposit: RelayData & Partial<SpeedUpCommon> & { destinationChainId: number; speedUpSignature?: string }
+  ): BigNumber {
+    // The real ABI encoder rejects "" where this codebase treats it as a no-op: message,
+    // updatedMessage (isMessageEmpty), and a present-but-empty speedUpSignature. updatedMessage
+    // skips undefined so a malformed partial speed-up still fails the isDefined assert below.
+    const message = isMessageEmpty(deposit.message) ? "0x" : deposit.message;
+    const updatedMessage =
+      isDefined(deposit.updatedMessage) && isMessageEmpty(deposit.updatedMessage) ? "0x" : deposit.updatedMessage;
+    const speedUpSignature = !isMessageEmpty(deposit.speedUpSignature) ? deposit.speedUpSignature : undefined;
```
droplet-rl: This normalization diverges from the path it's meant to measure. Both `resolveV3RelayCall` (`src/arch/evm/SpokeUtils.ts:57`) and `isDepositSpedUp` (`src/utils/DepositUtils.ts:251`) branch on a bare `isDefined(speedUpSignature)`, so `"0x"` counts as sped up for them. Here it's normalized to `undefined` and estimated as a plain `fillRelay` — under-estimating by at least 288 bytes relative to the `fillRelayWithUpdatedDeposit` that would actually be encoded at fill time. The test on line 956 locks that divergence in.

It's unreachable today (`SpokePoolClient.ts:297` always sets a real `depositorSignature`), so this isn't a live bug. But this line is the one remaining *model* in a PR whose stated point is to stop modeling, and it models the fill path incorrectly. Cleaner either way:

- reuse `isDepositSpedUp()` here so the estimator and the fill path share one predicate, or
- fix `resolveV3RelayCall` to treat `"0x"`/`""` as absent, and drop the normalization entirely.

The second is better if a `"0x"` signature reaching `fillRelayWithUpdatedDeposit` would revert onchain anyway.

src/relayFeeCalculator/chain-queries/tvmQuery.ts
```diff
@@ -1,20 +1,32 @@
...
+
+    const calldata = getV3RelayCalldata(
+      this.spokePool,
+      { ...deposit, message, updatedMessage, speedUpSignature },
+      getDefaultRelayer(deposit.destinationChainId)
+    );
+    return arch.tvm.bandwidthCostForCalldata(calldata);
```
droplet-rl, line 30: Worth confirming deliberately: for TVM this function was total before this PR (pure arithmetic over string lengths), and now it can throw — via `resolveV3RelayCall`'s asserts on `updatedRecipient` being defined and non-zero, and via the ABI encoder on a malformed `message`. `resolveAuxNativeFee` logs and rethrows, and `relayer/src/clients/ProfitClient.ts` calls this with no try/catch, so a single bad Tron-destined deposit now propagates out of profitability evaluation rather than returning a number.

The SVM path already throws on a malformed AcrossPlus message (with a test asserting it), so there's precedent and I don't think this needs changing here. Flagging it because it's a behavior change the description doesn't mention, and because the relayer-side blast radius differs from the quote-API one. A depositor speeding up to `address(0)` is the realistic trigger — worth a glance at whether the SpokePool permits that.

test/relayFeeCalculator.test.ts
```diff
@@ -844,94 +844,150 @@ describe("getAuxiliaryNativeTokenCost", function () {
...
-      // the original `message` field inside fillRelayWithUpdatedDeposit.
-      const originalMessage = "0x" + "cd".repeat(32);
-      const both = { ...withUpdatedMessage, message: originalMessage };
-      const bothFee = tvmArch.getAuxiliaryNativeTokenCost(both);
-      expect(bothFee.eq(FIXED_BANDWIDTH_SUN + SPEED_UP_FIXED_EXTRA_SUN + 64 * 1000 + 32 * 1000)).to.equal(true);
+      // 796,000 is 1 byte over the 795,000 observed onchain (670 fills). "" and "0x" both
+      // mean "no message" (isMessageEmpty), so both cost the same here.
+      it("costs an empty or blank message the fixed envelope, matching onchain history", function () {
+        const emptyFee = tvmQuery.getAuxiliaryNativeTokenCost(makeDeposit(EMPTY_MESSAGE));
+        expect(emptyFee.eq(796_000)).to.equal(true);
```
droplet-rl: `expect(fee.eq(796_000)).to.equal(true)` fails as `expected false to equal true` — no numbers. That's tolerable for the relative assertions elsewhere in this block, but these two are exact onchain-calibration anchors (this line and 909), so they're precisely the ones where you want the delta printed. They'll also fire on any unrelated SpokePool ABI change, and whoever hits that will be staring at a boolean.

`expect(fee.toNumber()).to.equal(796_000)` gets you `expected 806000 to equal 796000` for free.

Separately: these are the hardcoded sums the PR description says tests should stop duplicating. Keeping the literals is the right call here — they're the onchain anchor, and deriving them from `TVM_RAW_DATA_OVERHEAD_BYTES` would make the test tautological — but that's another sign the description needs updating.

droplet-rl (commented):
The constant correction is right, and the refactor it now carries is a real improvement. Nothing blocking, but the PR description no longer describes the diff — please fix that before merge — plus four smaller notes inline.

## What I verified

Checked out `pxrl/tvmBuffer` and ran the full local gate:

- `tsc --noEmit -p tsconfig.build.json` — clean
- `eslint src test` and `prettier --check` — clean
- `npx hardhat test test/relayFeeCalculator.test.ts` — 24 passing

Cross-checked the arithmetic independently:

| case | calldata | + overhead | est. SUN | onchain |
|---|---|---|---|---|
| empty message | `4 + 96 + 384 + 32` = 516 | 796 B | 796,000 | 795,000 |
| 2112-byte multicall | 516 + 2112 = 2628 | 2908 B | 2,908,000 | 2,908,000 |

280 reproduces the cited `net_fee` for deposit #3984372 exactly and overshoots empty-message fills by one byte. The old 290 gave 806,000 against 795,000 actual (+1.4%). The correction is well-supported.

Downstream is safe: `relayer/src/clients/ProfitClient.ts` passes a full `Deposit` (already carries `destinationChainId`) and its mocks already type the param as `{ destinationChainId: number }`, so widening `QueryInterface.getAuxiliaryNativeTokenCost` doesn't break it. I also searched the org for importers of the two symbols this deletes — `arch.tvm.getAuxiliaryNativeTokenCost` and `tvmBandwidthCostZero` — and found none outside this repo.

## The description doesn't match the diff

#1540 landed into this branch ~30 min ago, so #1539 now carries three commits. The body still describes the first one: "name the ABI-encoding components ... and sum them explicitly" and "export the byte/rate constants ... so tests can derive their expected values." #1540 then *deleted* every one of those named constants (`TVM_FILL_RELAY_FIXED_CALLDATA_BYTES`, `TVM_SPEED_UP_FIXED_CALLDATA_EXTRA_BYTES`, `paddedMessageByteLength`) and replaced the model with real ABI encoding.

That matters beyond tidiness: as it stands the diff also changes the `QueryInterface` signature, removes two exported symbols, and moves the TVM cost logic out of `arch/tvm` into `TvmQuery` — none of which a reader of the description would expect from a one-constant fix. Worth rewriting the body around the final state.

## Smaller architectural note

`arch/evm` and `arch/svm` both export `getAuxiliaryNativeTokenCost`; `arch/tvm` now doesn't, because TVM needs the `spokePool` Contract that only lives on the query object. Reasonable trade, but the per-arch symmetry is gone and `src/arch/tvm/MessageUtils.ts` no longer has anything to do with messages — consider renaming it (`BandwidthUtils.ts`) while the file is small.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=c6cde1ba68aa012b4dab59c73be6363d&trace_start_ts=1789770279&trace_end_ts=1789856979)

pxrl:
@droplet-rl another pass please — addressed everything from the last review, in `ff21b0b4`:

1. Rewrote the PR description around the final diff (the calldata-measurement refactor and the constants it removes, plus the fix below), replacing the stale text that only described the original 290->280 commit.
2. Fixed the `speedUpSignature: "0x"`/`""` divergence at the root: `resolveV3RelayCall` itself now treats it as absent, not just `TvmQuery`'s local estimate. I checked `SpokePool.sol` directly — `fillRelayWithUpdatedDeposit` re-verifies the signature via `_verifyUpdateV3DepositMessage`, so a `"0x"` signature reaching it would revert onchain regardless, confirming your suggested fix is safe for every EVM chain, not just Tron. This also let me delete `TvmQuery`'s now-redundant local normalization.
3. Two on-chain-calibration assertions now use `.toNumber()).to.equal(...)` for a real diff on failure instead of a bare boolean.
4. Softened the `TVM_RAW_DATA_OVERHEAD_BYTES` comment — verified your fee_limit-varint theory is the more honest framing than an "observed max" claim from n=2, reworded accordingly.

Left as-is, for your awareness: the `arch/tvm/MessageUtils.ts` naming/architecture note (non-blocking, said "consider") and the uncaught-throw-in-ProfitClient note (you said doesn't need changing). 24/24 tests passing, lint clean, tsc clean.

src/arch/tvm/MessageUtils.ts
pxrl: Verified independently before fixing: temporarily instrumented the 2112-byte test to measure the real calldata length (2628 bytes) and confirmed 2908 - 2628 = 280 exactly, so that sample's true envelope is 280, not 279 — your fee_limit-varint theory checks out over the alternative (nested length-prefix width, which I ruled out the same way you did, since both samples sit in the same 2-byte varint bucket). Reworded the comment in ff21b0b4 to state what varies and that 280 isn't a hard ceiling, rather than asserting an "observed max" from two samples. Left the constant itself unchanged, per your call.

src/relayFeeCalculator/chain-queries/tvmQuery.ts
pxrl: Took your second option, fixed at the root in ff21b0b4: `resolveV3RelayCall` (src/arch/evm/SpokeUtils.ts) now treats `speedUpSignature: "0x"`/`""` as absent itself, rather than TvmQuery normalizing its own copy before calling in. Checked `SpokePool.sol` directly to confirm this is safe: `fillRelayWithUpdatedDeposit` re-verifies the signature via `_verifyUpdateV3DepositMessage`, so a `"0x"` signature reaching it would revert onchain regardless — there's no legitimate case this forecloses. This fixes the divergence for every EVM chain, not just Tron, and let me delete TvmQuery's now-redundant local normalization, so it's a net simplification.

src/relayFeeCalculator/chain-queries/tvmQuery.ts
pxrl: No code change here, per your own read (matches SVM precedent). On the "worth a glance" question: checked `_verifyUpdateV3DepositMessage` in SpokePool.sol directly — it only verifies the depositor's signature over the hashed fields, with no zero-address check on `updatedRecipient`. So a depositor genuinely can sign and emit a real `RequestedSpeedUpDeposit` with `updatedRecipient = address(0)` on Mainnet today; it's not just a malformed-input hypothetical. I didn't trace whether the actual `fillRelayWithUpdatedDeposit` execution path would separately revert on that recipient (e.g. via an ERC20 transfer-to-zero revert) — that's beyond what I verified, so I can't say whether such a deposit is fillable at all, only that it can exist. Leaving this as an FYI, not blocking.

test/relayFeeCalculator.test.ts
pxrl: Applied in ff21b0b4 — both onchain-calibration anchors now use `.toNumber()).to.equal(...)` so a future failure shows the actual mismatched value instead of `expected false to equal true`. Also folded your point about these being the hardcoded onchain anchors (deliberately not derived from the constants under test) into the rewritten PR description, so that intent isn't only in a review comment.

droplet-rl:
Re-ran the gate on `ff21b0b`: `tsc --noEmit -p tsconfig.build.json` clean, `eslint src test` clean, 24/24 in `relayFeeCalculator.test.ts`, 61 passing across `Tvm.TransactionUtils` + `relayFeeCalculator` + `common` + `validatorUtils` + `multicall`. (Full `hardhat test` needs `solana-test-validator`, which isn't in this sandbox, so I ran the affected files individually rather than the root suite.)

Items 1, 3 and 4 are good. The description now matches the diff. On the `fee_limit` rewording — the arithmetic backs it: every size-driven varint in the serialization (`data` length, the `Any` value length, `contract` length, `raw_data` length) lands in the 2-byte range for both 516 and 2628 bytes, so `fee_limit` is the only field that plausibly differs between the two samples. "Not a hard max" is the honest framing.

## On #2 — I under-specified my original suggestion, and I'd like to reopen it

I offered "fix `resolveV3RelayCall` to treat `"0x"`/`""` as absent" conditional on a `"0x"` signature reverting onchain anyway. You verified that premise and it holds. But I stated the conclusion too loosely: **treating it as absent doesn't prevent a transaction, it produces a different, successful one.**

Confirmed against the real ABI (scratch test, since removed) with `speedUpSignature: "0x"` plus a full set of `updated*` fields:

```
isDepositSpedUp:        true
encoded method:         fillRelay        (was: fillRelayWithUpdatedDeposit)
recipient in calldata:  0x1111...1111    (original — not updatedRecipient 0x2222...2222)
outputAmount:           1000             (original — not updatedOutputAmount 900)
```

So the failure mode moved in the wrong direction:

- **before** — build `fillRelayWithUpdatedDeposit` → reverts onchain / fails simulation → loud and recoverable
- **after** — build `fillRelay` → **succeeds** on the original terms → the depositor's speed-up is silently dropped, and the fill is final

It also re-creates the divergence one layer up. `isDepositSpedUp` (`src/utils/DepositUtils.ts:251`) still uses a bare `isDefined` and wasn't updated, so it and the encoder now disagree on the same input. Two live consumers:

- `sdk/src/utils/DepositUtils.ts:266` — `resolveDepositMessage` returns `updatedMessage`
- `relayer/src/relayer/Relayer.ts:314` — the recipient allow-list gate, whose own comment reads *"Gate on the effective recipient the fill pays out to — updatedRecipient for sped-up deposits, which fill via fillRelayWithUpdatedDeposit — so a speed-up can't redirect funds off the list."* That premise is exactly what this change breaks: the gate would check `updatedRecipient` while the tx pays `recipient`.

**Not reachable today, and not a security finding** — I chased that down before writing it up. `SpokePool.speedUpDeposit` calls `_verifyUpdateV3DepositMessage` *before* emitting (`SpokePool.sol:833` → `:846`), so `RequestedSpeedUpDeposit` can never carry an empty `depositorSignature`, and `SpokePoolClient.ts:297` only ever reads a verified one. Neither `scripts/spokepool.ts` nor `scripts/simulateFill.ts` sets the field either. So it's a latent invariant break, not a live bypass.

Cheap fix that keeps the loud failure on the same premise you verified:

```ts
if (isDefined(relayData.speedUpSignature)) {
  assert(!isMessageEmpty(relayData.speedUpSignature), "Empty speedUpSignature on a sped-up deposit");
  assert(isDefined(relayData.updatedRecipient) && !relayData.updatedRecipient.isZeroAddress());
  ...
}
```

That names an empty signature for what it is — a malformed speed-up — right next to the `updatedRecipient`/`updatedMessage` asserts already there, and keeps `isDepositSpedUp` and the encoder on one rule. The TvmQuery test `"treats a present-but-empty speedUpSignature as not sped up"` would flip to expecting a throw, pairing it with its sibling `"still throws on a malformed speed-up missing updatedMessage entirely"` — which is the more coherent pair anyway, since both are malformed-shape cases.

The alternative, if empty-means-absent really is the intended convention for signatures and not just messages, is to push it into `isDepositSpedUp` so every consumer agrees. Bigger blast radius; I'd default to the assert.

Your call either way — it's unreachable, so I'm not blocking on it. I left the code untouched since this is a convention decision on shared EVM semantics that's yours to make; say the word and I'll push the assert version plus the test flip.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=b1c2c4062d5535f74b6fcfaf69947e1f&trace_start_ts=1789771447&trace_end_ts=1789858147)

pxrl:
Good catch, and correctly reopened — you're right that "treat as absent" was the wrong fix. Fell into exactly the trap you named: it doesn't stop a bad transaction, it builds a different, successful one on the wrong terms.

Took your suggested fix as-is: reverted the branch condition back to bare `isDefined(relayData.speedUpSignature)` (matching `isDepositSpedUp`) and added `assert(!isMessageEmpty(relayData.speedUpSignature), "Empty speedUpSignature on a sped-up deposit")` as the first check inside that branch, right alongside the existing `updatedRecipient`/`updatedOutputAmount`/`updatedMessage` asserts. An empty signature now fails loudly as a malformed speed-up instead of silently degrading to plain `fillRelay`.

Flipped the TvmQuery test that was locking in the old (wrong) behavior — renamed to `"still throws on a malformed speed-up with an empty speedUpSignature"`, now paired with `"still throws on a malformed speed-up missing updatedMessage entirely"` as you suggested, since both are the same class of malformed-shape case. Also ran `SpokePoolClient.SpeedUp.ts` (the other test file touching this code path) to check for regressions there — none.

This is verified locally (24/24 passing, tsc/lint clean) but not pushed yet — will ping again once it's up.

pxrl:
@droplet-rl the empty-speedUpSignature fix is pushed in `3b878ab5` — matches your suggested patch exactly: bare `isDefined` on the branch condition, `assert(!isMessageEmpty(relayData.speedUpSignature), "Empty speedUpSignature on a sped-up deposit")` as the first check inside it, and the TvmQuery test flipped to expect a throw (renamed to pair with the missing-updatedMessage one). Also checked `SpokePoolClient.SpeedUp.ts` for regressions on this path — none. 24/24 in `relayFeeCalculator.test.ts`, tsc/lint clean.

Another pass whenever you get a chance.

droplet-rl:
`3b878ab` is exactly right, and I verified it does what it says rather than just reading it.

**Verification on `3b878ab`:** `tsc --noEmit` clean, `eslint src test` clean, `prettier --check` clean, 79 passing across `SpokePoolClient.SpeedUp` + `SpokePoolClient.v3Events` + `relayFeeCalculator` + `Tvm.TransactionUtils` + `common` + `validatorUtils` + `multicall`. (The root suite still needs `solana-test-validator` via the `Solana.setup.ts` mocha hook, absent here, so I ran the affected files directly.)

Behaviour confirmed against the real ABI with a scratch test — the thing I actually wanted to check was that the new assert didn't cost you the happy path:

```
sig="0x"  isDepositSpedUp=true  -> threw: Empty speedUpSignature on a sped-up deposit
sig=""    isDepositSpedUp=true  -> threw: Empty speedUpSignature on a sped-up deposit

real 65-byte sig -> fillRelayWithUpdatedDeposit
  relayData.recipient   0x1111…1111   updatedRecipient      0x2222…2222
  relayData.outputAmount 1000         updatedOutputAmount   900

no speedUpSignature -> fillRelay
```

Updated terms still carried correctly, plain path untouched, and `isDepositSpedUp` and the encoder now agree on one rule: both call the empty-sig deposit sped up, and the encoder rejects it as malformed instead of reinterpreting it. That's the divergence closed properly. `SpokePoolClient.SpeedUp.ts` passes 3/3, matching what you found.

Two doc-only leftovers, both describing the behaviour that `3b878ab` reverted:

**1. I pushed a fix for one** — `acbc05f`. `tvmQuery.ts:19` still read "*speedUpSignature is normalized there*", which was true of `ff21b0b` but not of `3b878ab`. Now:

```
// fails the isDefined assert in resolveV3RelayCall (which also rejects an empty
// speedUpSignature outright, so it needs no normalization here).
```

Comment-only, tsc/lint/prettier clean, 27/27 still passing. Revert it if you'd rather word it yourself.

**2. The PR description needs the same correction** — I can't edit it without taking over your description, so here's paragraph 4 as a paste. It currently still says "*treated as absent (plain `fillRelay`) rather than attempted*":

> Also tightens a related correctness gap in the shared `resolveV3RelayCall`: a `speedUpSignature` of `"0x"`/`""` can never be a valid ECDSA signature, so it's now rejected as a malformed speed-up — alongside the existing `updatedRecipient`/`updatedOutputAmount`/`updatedMessage` asserts — rather than silently falling back to a plain `fillRelay` on the original terms. Applies to every EVM chain, not just Tron.

With those two, this is good to merge from my side. The 290→280 correction is sound (2,908,000 SUN reproduces the onchain `net_fee` for #3984372 exactly), the measure-don't-model refactor removes a class of drift rather than patching an instance of it, and the speed-up path now fails loudly on malformed input instead of quietly filling the wrong terms.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=ca2dfac24e80c99bfd0708f34b696ff1&trace_start_ts=1789807350&trace_end_ts=1789894050)

pxrl:
Description updated with your paragraph 4 (verbatim). Pulled `acbc05f` and re-ran the full local gate against it: 27/27 passing, tsc/lint clean. Nothing else pending — thanks for the thorough pass.

droplet-rl (approved):
Approving. Everything from the previous three rounds is resolved, and I re-verified the final state rather than assuming the last round carried.

## Verification on `acbc05f`

- `tsc --noEmit -p tsconfig.build.json`, `eslint src test`, `prettier --check` — all clean
- **87 passing** across `relayFeeCalculator`, `SpokePoolClient.SpeedUp`, `SpokePoolClient.v3Events`, `Tvm.TransactionUtils`, `common`, `validatorUtils`, `multicall`, `BlockUtils`

The root `hardhat test` can't run here — the `Solana.setup.ts` mocha hook spawns `solana-test-validator`, which isn't in this sandbox, and the root `before all` bails before any test executes. So the above is the affected files run directly, not a full-suite pass. Worth confirming CI is green on the SVM files before merge, since `svmQuery.ts` is in the diff (signature-only change, but I couldn't exercise it).

## What I checked, and how

**The constant.** Re-derived independently: empty-message `fillRelay` calldata is `4 + 96 + 384 + 32` = 516 B, so 280 + 516 = 796 B → 796,000 SUN; the 2112-byte case is 2628 + 280 = 2908 B → 2,908,000 SUN, reproducing the cited onchain `net_fee` for #3984372 exactly. The old 290 gave 806,000 against 795,000 actual. Both calibration assertions pass.

**The `fee_limit` framing.** Checked the arithmetic rather than taking it on faith: every size-driven varint in the Tron serialization — `data` length, the `Any` value length, `contract` length, `raw_data` length — sits in the 2-byte range (128..16383) for both 516 and 2628 bytes, so `fee_limit` is the only field that can plausibly differ between the two samples. "Not a hard max" is the honest wording.

**The speed-up fix.** Exercised against the real ABI, since the thing worth checking was whether the new assert cost you the happy path:

```
sig="0x"  isDepositSpedUp=true  -> throws: Empty speedUpSignature on a sped-up deposit
sig=""    isDepositSpedUp=true  -> throws: Empty speedUpSignature on a sped-up deposit

real 65-byte sig -> fillRelayWithUpdatedDeposit
  relayData.recipient   0x1111…1111   updatedRecipient      0x2222…2222
  relayData.outputAmount 1000         updatedOutputAmount   900

no speedUpSignature -> fillRelay
```

Updated terms carried correctly, plain path untouched. `isDepositSpedUp` and the encoder now agree on one rule — both classify the empty-sig deposit as sped up, and the encoder rejects it as malformed rather than reinterpreting it as a plain fill on the original terms. That closes the divergence properly rather than relocating it.

**Reachability, for the record.** `SpokePool.speedUpDeposit` calls `_verifyUpdateV3DepositMessage` *before* emitting (`SpokePool.sol:833` → `:846`), so `RequestedSpeedUpDeposit` can never carry an empty `depositorSignature`, and `SpokePoolClient.ts:297` only ever reads a verified one. The case this hardens against was never live — the value is that the encoder and `isDepositSpedUp` can't drift apart later.

**Downstream.** `relayer/src/clients/ProfitClient.ts` passes a full `Deposit`, which already carries `destinationChainId`, and its mocks already type the param that way — the widened `QueryInterface` signature doesn't break it. No importers anywhere in the org of the two removed symbols (`arch.tvm.getAuxiliaryNativeTokenCost`, `tvmBandwidthCostZero`).

## One note, not blocking

I pushed `acbc05f` myself — a comment-only fix to `tvmQuery.ts:19`, which still read *"speedUpSignature is normalized there"* after `3b878ab` changed that path from normalizing to rejecting. Revert or reword it if you'd rather own the phrasing; it's your file and the fix is cosmetic.

The shape of this change is the right one: replacing a hand-counted ABI model with a measurement removes a whole class of drift instead of correcting one instance of it, and the 280 correction is the smaller half of the win.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=666282e60079edb1f18fad2fbd3ae499&trace_start_ts=1789987835&trace_end_ts=1790074535)

PR #1540, opened 2026-09-18
refactor(tvm): measure real calldata for bandwidth cost instead of modeling it

pxrl:
TvmQuery.getAuxiliaryNativeTokenCost now builds the actual fillRelay/ fillRelayWithUpdatedDeposit transaction (the same one already built for energy estimation) and measures its real calldata length, instead of re-deriving an equivalent byte count from a hand-counted ABI layout. This removes TVM_FILL_RELAY_FIXED_CALLDATA_BYTES, TVM_SPEED_UP_FIXED_CALLDATA_EXTRA_BYTES, and the padding-length helper entirely — the estimate is now exact by construction rather than a model that can silently drift from the real ABI, and it generalizes to any TVM call shape without needing its own per-function constant.

TVM_RAW_DATA_OVERHEAD_BYTES stays: it's the raw_data envelope + signature, which isn't part of the calldata and isn't known ahead of signing, so it remains a measured constant rather than something computed on the fly.

This changes QueryInterface#getAuxiliaryNativeTokenCost from sync to async across all three chain implementations (EVM and SVM keep their existing synchronous logic, just wrapped in a resolved/rejected promise) and its one caller in relayFeeCalculator.ts. Test mocks and the TVM/SVM/EVM test suites are updated to match; the TVM tests now exercise the real populated-transaction path end to end rather than checking a parallel model against itself.

Extracted the EVM/TVM address-narrowing assert (needed before calling getUnsignedTxFromDeposit, which rejects SVM addresses) into a single shared assertEvmOrTvmRelayData helper in baseQuery.ts. It was previously duplicated verbatim in getGasCosts and getNativeGasCost; adding a third copy for getAuxiliaryNativeTokenCost was the trigger to deduplicate all three instead. A single assertion on the whole relayData object (rather than one assert per field) is also what lets TypeScript's control-flow narrowing apply to the object as a whole, since narrowing doesn't propagate out of a loop or callback back to the variable it closed over.


droplet-rl (commented):
Nice direction — replacing a hand-counted ABI model with a measurement of the real encoding is the right call, and the numbers line up: `4 + 3*32 + 12*32 + 32 = 516` for plain `fillRelay` and `4*32 + 32 + 32 + 96 = 288` extra for `fillRelayWithUpdatedDeposit` both reproduce the deleted constants exactly, so this is value-preserving. Deduplicating the three copies of the address-narrowing assert is a clear win, and the docstring on `assertEvmOrTvmRelayData` explaining *why* it's one object-level assert rather than three field-level ones is genuinely useful.

**Verified locally** (fresh clone at `8056e7c`): `yarn build` ✅, `yarn lint-check` ✅, `npx hardhat test test/relayFeeCalculator.test.ts` → 22 passing ✅. The new TVM tests don't touch the network (`populateTransaction` is a local encode), so they stay fast — 2s for the file.

Not blocking, but a few things worth a look before merge:

**1. The TVM aux estimate can now throw where it previously always returned a number.** `populateV3Relay` validates its inputs; the deleted arithmetic didn't. I confirmed empirically against the built output on this branch:

| input | before | after |
|---|---|---|
| `message: ""` | finite (guarded by `isMessageEmpty`) | throws `invalid arrayify value` |
| `speedUpSignature: "0x"` | finite (guarded by `!== "0x"`) | throws `updatedRecipient` assert |
| no `destinationChainId` | finite | throws `invalid BigNumber value` |

`resolveAuxNativeFee` rethrows, so each of these now fails the whole quote. Usually harmless because `getGasCosts` would hit the same assert first — **except** in `relayerFeeDetails`, where `tokenGasCost && gasUnits` short-circuits `getGasCosts` entirely (and `_tokenGasCost` short-circuits `gasFeePercent`). On that precomputed-gas path this PR makes the aux estimate the only caller of `populateV3Relay`, so it becomes a newly-reachable failure. The deleted code carried an explicit `speedUpSignature !== "0x"` guard — worth confirming that guard wasn't load-bearing for some upstream producer before dropping it.

**2. Breaking change to a published interface.** `QueryInterface#getAuxiliaryNativeTokenCost` goes sync→async *and* narrows its parameter to require `destinationChainId`. Downstream implementers break at compile time; downstream callers that use the result directly break at runtime (`BigNumber` → `Promise<BigNumber>`, so `.mul` is `undefined`). `package.json` is still at `4.4.22` here — presumably the bump lands separately, but this looks like it needs a coordinated release with relayer/api rather than a patch.

Everything else below is small.

src/relayFeeCalculator/chain-queries/tvmQuery.ts
```diff
@@ -10,11 +11,13 @@ import { CustomGasTokenQueries } from "./customGasToken";
...
+ * measure it here directly from the real populated transaction's calldata, rather than
+ * modeling a per-function calldata size that could drift from the actual ABI.
  */
 export class TvmQuery extends CustomGasTokenQueries {
-  override getAuxiliaryNativeTokenCost(deposit: RelayData): BigNumber {
-    return arch.tvm.getAuxiliaryNativeTokenCost(deposit);
+  override async getAuxiliaryNativeTokenCost(deposit: RelayData & { destinationChainId: number }): Promise<BigNumber> {
+    assertEvmOrTvmRelayData(deposit);
+    const { data } = await this.getUnsignedTxFromDeposit(deposit);
+    return arch.tvm.getAuxiliaryNativeTokenCost(data ?? "0x");
```
droplet-rl: `data ?? "0x"` fails quietly in the wrong direction. If `data` is ever undefined this returns the bare 280-byte envelope cost (~0.28 TRX) instead of the real ~0.8 TRX floor for a plain fill, or ~2.9 TRX for a message-bearing one — a silent 3–10x underestimate rather than a visible failure.

`populateTransaction.fillRelay(...)` always populates `data`; the `?? "0x"` is only there to satisfy `PopulatedTransaction.data?: string`. Since the whole point of this PR is that the estimate is "exact by construction", an assert expresses that better than a fallback:

```ts
const { data } = await this.getUnsignedTxFromDeposit(deposit);
assert(isDefined(data), "populateV3Relay produced no calldata");
return arch.tvm.getAuxiliaryNativeTokenCost(data);
```

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -172,8 +169,8 @@ export class QueryBase implements QueryInterface {
    * @param deposit RelayData associated with Deposit we're estimating for
    * @returns Native token cost
    */
-  getAuxiliaryNativeTokenCost(_deposit: RelayData): BigNumber {
-    return arch.evm.getAuxiliaryNativeTokenCost(_deposit);
+  getAuxiliaryNativeTokenCost(_deposit: RelayData): Promise<BigNumber> {
```
droplet-rl: This override's parameter is still `RelayData`, but `QueryInterface` (and the `TvmQuery` override) now require `RelayData & { destinationChainId: number }`. TypeScript's method-parameter bivariance lets the narrowing override compile, so nothing complains — but it means a caller holding a `QueryBase`-typed reference can pass a deposit without `destinationChainId` and get a runtime failure instead of a type error.

Confirmed on this branch: calling `TvmQuery#getAuxiliaryNativeTokenCost` with `destinationChainId` deleted throws `invalid BigNumber value (argument="value", value=undefined)` from deep inside `populateV3Relay`, because `repaymentChainId` defaults to `relayData.destinationChainId`.

Widening this signature (and `SvmQuery`'s at svmQuery.ts:207) to `RelayData & { destinationChainId: number }` to match the interface would surface the requirement at every call site. Neither implementation reads the field, so it costs nothing.

src/relayFeeCalculator/chain-queries/svmQuery.ts
```diff
@@ -204,8 +204,14 @@ export class SvmQuery implements QueryInterface {
    * @throws If deposit.message is malformed (unable to be deserialized into `AcrossPlusMessage`)
    * @returns Native token cost
    */
-  getAuxiliaryNativeTokenCost(deposit: RelayData): BigNumber {
-    return arch.svm.getAuxiliaryNativeTokenCost(deposit);
+  getAuxiliaryNativeTokenCost(deposit: RelayData): Promise<BigNumber> {
+    // Not `async`: `arch.svm.getAuxiliaryNativeTokenCost` can throw synchronously (malformed
+    // message), and this converts that into a rejection rather than a synchronous throw.
```
droplet-rl: The comment has the reasoning backwards. An `async` function converts a synchronous throw in its body into a rejected promise automatically — that's exactly the behaviour being hand-rolled here. Being *non*-`async` is what would let a sync throw escape, which is why the try/catch became necessary in the first place.

The whole body reduces to:

```ts
async getAuxiliaryNativeTokenCost(deposit: RelayData): Promise<BigNumber> {
  return arch.svm.getAuxiliaryNativeTokenCost(deposit);
}
```

Same rejection semantics, and it matches how `TvmQuery` is written one file over. The new `assertPromiseError` test at test/relayFeeCalculator.test.ts:843 covers it either way.

src/arch/tvm/MessageUtils.ts
```diff
@@ -10,67 +9,25 @@ export const TVM_BANDWIDTH_SUN_PER_BYTE = 1000;
...
-  const paddedMessageBytes = paddedMessageByteLength(deposit.message);
-
-  let speedUpExtraBytes = 0;
-  if (isDefined(deposit.speedUpSignature) && deposit.speedUpSignature !== "0x") {
-    speedUpExtraBytes = TVM_SPEED_UP_FIXED_CALLDATA_EXTRA_BYTES + paddedMessageByteLength(deposit.updatedMessage);
-  }
-
-  const calldataBytes = TVM_FILL_RELAY_FIXED_CALLDATA_BYTES + paddedMessageBytes + speedUpExtraBytes;
+export function getAuxiliaryNativeTokenCost(calldata: string): BigNumber {
+  const calldataBytes = Math.max(0, Math.floor((calldata.length - 2) / 2));
```
droplet-rl: Now that the input is an opaque hex string rather than a `message` field the caller controls, it's worth using the validating helper instead of the arithmetic:

```ts
const calldataBytes = utils.hexDataLength(calldata);
```

`hexDataLength` throws on a non-`0x`-prefixed or odd-length string rather than silently undercounting by one byte, and it makes the `Math.max(0, ...)` clamp unnecessary. `ethers/utils` is already imported throughout the arch layer.

src/arch/tvm/MessageUtils.ts
```diff
@@ -10,67 +9,25 @@ export const TVM_BANDWIDTH_SUN_PER_BYTE = 1000;
...
 /**
- * Estimate the bandwidth cost (in SUN) of a Tron fill for `deposit`.
+ * Estimate the bandwidth cost (in SUN) of a Tron transaction carrying `calldata`.
  *
  * On Tron, transaction bytes consume bandwidth — distinct from energy (paid via the tx
  * fee limit and modeled in `nativeGasCost`). When the relayer has no free bandwidth
- * staked, every byte burns 1,000 SUN. For message-bearing fills this is non-trivial:
- * the message dominates calldata size, so the cost scales linearly with message length.
+ * staked, every byte burns 1,000 SUN.
  *
```
droplet-rl: Minor: "`calldata` is measured directly from the real populated transaction" describes what `TvmQuery` does, not what this function does — this function just takes a hex string and can't know its provenance. The paragraph would sit better on `TvmQuery#getAuxiliaryNativeTokenCost` (where an equivalent note already exists in the class docstring), leaving this one to document the actual contract: bytes in → SUN out.

test/relayFeeCalculator.test.ts
```diff
@@ -840,108 +840,130 @@ describe("getAuxiliaryNativeTokenCost", function () {
...
-      ).to.equal(true);
+      it("matches the full onchain history for an empty-message fillRelay", async function () {
+        // 670 real fillRelay fills, zero variance: overhead 279, calldata 516 -> 795 bytes
+        // total. 280 is used for the overhead constant (a one-byte margin over 279), so
+        // this reproduces 796,000 SUN exactly.
+        const fee = await tvmQuery.getAuxiliaryNativeTokenCost(makeDeposit(EMPTY_MESSAGE));
+        expect(fee.eq(796_000)).to.equal(true);
+      });
+
+      it("returns the same cost for 0x and fully-empty messages", async function () {
```
droplet-rl: This test can't fail. `EMPTY_MESSAGE` is `"0x"` (src/constants.ts:65), so `makeDeposit(EMPTY_MESSAGE)` and `makeDeposit("0x")` build identical objects and the assertion is `x.eq(x)`.

The pre-PR version had the same redundancy but at least pinned both cases to `FIXED_BANDWIDTH_SUN`, so each independently asserted a value. Suggest just deleting this one — the preceding test already pins the empty-message case to `796_000`.

test/relayFeeCalculator.test.ts
```diff
@@ -840,108 +840,130 @@ describe("getAuxiliaryNativeTokenCost", function () {
...
-      expect(
-        bothFee.eq(
-          FIXED_BANDWIDTH_SUN +
-            SPEED_UP_FIXED_EXTRA_SUN +
-            64 * tvmArch.TVM_BANDWIDTH_SUN_PER_BYTE +
-            32 * tvmArch.TVM_BANDWIDTH_SUN_PER_BYTE
-        )
-      ).to.equal(true);
+      it("matches the full onchain history for an empty-message fillRelay", async function () {
+        // 670 real fillRelay fills, zero variance: overhead 279, calldata 516 -> 795 bytes
```
droplet-rl: These two ground-truth comments disagree by a byte, and a future reader reconciling them will lose some time. Here: 670 fills at 795 bytes total − 516 calldata ⇒ **279** envelope, with 280 described as "a one-byte margin". At line 914, the 2112-byte fill: 2908 − 516 − 2112 ⇒ **280** envelope, described as matching "exactly".

Both can't be exact. Most likely the raw_data protobuf varints (timestamp/expiration/fee_limit) shift by a byte between samples, which would mean `TVM_RAW_DATA_OVERHEAD_BYTES` is inherently ±1 rather than a fixed measurement. 280 is the right conservative pick either way — but it'd help to say that once, rather than have one comment claim "zero variance" and the other claim an exact match at a different value.

src/relayFeeCalculator/chain-queries/baseQuery.ts
```diff
@@ -32,6 +32,29 @@ export type SymbolMappingType = Record<
...
+  exclusiveRelayer: EvmAddress | TvmAddress;
+};
+
+/**
+ * Asserts that `relayData`'s address fields are EVM- or TVM-like, narrowing them for
+ * `getUnsignedTxFromDeposit`/`populateV3Relay`, which don't accept SVM addresses. A single
+ * assertion on the whole object (rather than one per field) is what lets TypeScript narrow
+ * `relayData` itself, since narrowing doesn't propagate out of a loop or callback back to the
+ * variable it closed over.
+ */
```
droplet-rl: The consolidation loses the `getGasCosts:` / `getNativeGasCost:` prefixes the three call sites used to carry, so a production assert now reports only `recipient not an EVM-like address (0x…)` with no indication of which estimate path failed. With three callers (and the aux-cost path newly able to surface these), an optional context arg would restore it cheaply:

```ts
export function assertEvmOrTvmRelayData<T extends RelayData>(
  relayData: T,
  context = ""
): asserts relayData is T & EvmOrTvmRelayData {
```

Stack traces make it recoverable, so purely a log-ergonomics nit.

src/relayFeeCalculator/chain-queries/tvmQuery.ts
```diff
@@ -10,11 +11,13 @@ import { CustomGasTokenQueries } from "./customGasToken";
...
- * calldata footprint.
+ * measure it here directly from the real populated transaction's calldata, rather than
+ * modeling a per-function calldata size that could drift from the actual ABI.
  */
 export class TvmQuery extends CustomGasTokenQueries {
-  override getAuxiliaryNativeTokenCost(deposit: RelayData): BigNumber {
-    return arch.tvm.getAuxiliaryNativeTokenCost(deposit);
+  override async getAuxiliaryNativeTokenCost(deposit: RelayData & { destinationChainId: number }): Promise<BigNumber> {
+    assertEvmOrTvmRelayData(deposit);
+    const { data } = await this.getUnsignedTxFromDeposit(deposit);
```
droplet-rl: Design note, not a defect: in the full quote path the fill transaction now gets ABI-encoded up to three times for the same deposit — once in `getGasCosts`, once in `getNativeGasCost`, and once here. It's a local encode so the cost is negligible even for a 2KB message, and the independence keeps this method's contract simple.

Still, since `getGasCosts` already has the `PopulatedTransaction` in hand, threading the calldata through (or accepting an optional pre-populated tx) would both eliminate the repeat work and guarantee the aux estimate measures precisely the transaction that was gas-estimated, rather than one rebuilt with the default relayer instead of the caller-supplied one. Size is invariant to the relayer address today, so it makes no numerical difference — worth considering only if this grows another caller.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=14539fec2b5e99d4435350f8110b56f2&trace_start_ts=1789762327&trace_end_ts=1789849027)

pxrl:
@droplet-rl could you take another look? Since your last review:

- Reverted the sync→async interface change entirely (kept the calldata-measurement idea, but via a new synchronous `getV3RelayCalldata` in `arch/evm/SpokeUtils.ts` instead of threading through `populateV3Relay`/`getUnsignedTxFromDeposit`). `QueryInterface`, `baseQuery.ts`, `svmQuery.ts`, and `relayFeeCalculator.ts` are all back to their pre-PR shape except for one widened parameter (see below).
- Fixed the two edge cases you reproduced empirically (`message: ""` and `speedUpSignature: "0x"`), normalized locally in `TvmQuery` rather than in the shared `resolveV3RelayCall`, with regression tests for both.
- Widened `QueryInterface#getAuxiliaryNativeTokenCost` to require `destinationChainId`, matching `getGasCosts`/`getNativeGasCost`'s existing signature.
- Fixed `hexDataLength` usage per your suggestion, but corrected an assumption in it — it returns `null` on invalid input rather than throwing, so there's an explicit check now.
- Renamed `arch.tvm.getAuxiliaryNativeTokenCost` (calldata → SUN) to `bandwidthCostForCalldata` to stop it colliding in name with `TvmQuery.getAuxiliaryNativeTokenCost` (deposit → SUN), and cut an unused export (`tvmBandwidthCostZero`).
- General comment trim throughout.

Commits `1260ec94` onward are new; `8056e7c1` is what you already reviewed.

droplet-rl:
Re-reviewed `c537302` (commits `1260ec9`..`c537302` on top of what I saw at `8056e7c`). This is a better design than my review assumed was available — a synchronous `getV3RelayCalldata` sharing `resolveV3RelayCall` with `populateV3Relay` keeps the measurement idea while leaving the interface alone, which is strictly better than the async propagation. All six items are addressed.

**Verified locally** (fresh install at `c537302`): `yarn build` ✅, `yarn lint-check` ✅, `npx hardhat test test/relayFeeCalculator.test.ts` → 22 passing ✅.

I checked the thing the whole refactor rests on — that `encodeFunctionData` and `populateTransaction` agree — by diffing the two paths directly against the built output:

| case | `getV3RelayCalldata` vs `populateV3Relay(...).data` | bytes |
|---|---|---|
| plain `fillRelay` | identical | 516 |
| 2112-byte message | identical | 2628 |
| sped-up | identical | 900 |

And the two edge cases I'd reproduced now return finite values instead of throwing: `message: ""` → `796000`, `speedUpSignature: "0x"` → `796000`.

You're also right about `hexDataLength` and I was wrong — it returns `null` rather than throwing. I exercised the new guard across `"0x"`, `"0xab"`, `"ab"`, `"0xabc"`, `""`, `"0xzz"` and it rejects exactly the four malformed ones. Renaming to `bandwidthCostForCalldata` is worth more than the collision avoidance you cite: `arch.evm`/`arch.svm`'s `getAuxiliaryNativeTokenCost` take a deposit and this one takes calldata, so the shared name implied a uniformity that never existed.

Two things left, one of which I'd genuinely fix before merge.

---

**1. The `796_000` comment became factually wrong during the comment trim.** This is the one I'd act on.

Tracing it through the fixups:

- `8056e7c` — *"670 real fillRelay fills, zero variance: overhead 279, calldata 516 -> 795 bytes total. 280 is used for the overhead constant (a one-byte margin over 279), so this reproduces 796,000 SUN exactly."*
- `db84982` — *"796,000 matches 670 real empty-message fillRelay fills exactly (see `TVM_RAW_DATA_OVERHEAD_BYTES`'s comment for why it's 280, not the on-chain 279)."*
- `c537302` — *"796,000 matches 670 real empty-message fillRelay fills exactly."*

The final form asserts onchain exactness that your own earlier note contradicts: the fills came in at 795,000, and 796,000 is the deliberately-conservative model output. Two compounding problems:

- The forward reference added in `db84982` pointed somewhere that never gained the explanation — `TVM_RAW_DATA_OVERHEAD_BYTES` still just reads *"Measured at 280 bytes."* So `c537302` dropping the pointer removed the last trace of 279 from the repo.
- Within the same `describe` block, this test now implies a 279-byte envelope while the 2112-byte test (`2,908,000 = 280 + 516 + 2112`) implies 280, with nothing left explaining why. A future reader reconciling those two has every reason to "correct" 280 → 279 and start silently underestimating every fill.

One clause restores it, e.g. on the constant itself:

```ts
// Tron raw_data envelope (contract type, ref_block, timestamp, expiration, fee_limit)
// plus a single ECDSA signature. 670 empty-message fills measured 279 bytes with zero
// variance; rounded up to 280 so the estimate stays conservative.
export const TVM_RAW_DATA_OVERHEAD_BYTES = 280;
```

and then the test comment can honestly say *"796,000 — one byte conservative over the 795,000 observed onchain."*

**2. `speedUpSignature: ""` still throws.** The new guard is `!== "0x"`, so an empty-string signature is treated as a real one and fails in the encoder. Confirmed against the build. It's asymmetric with the line directly above it, which uses `isMessageEmpty` precisely because this codebase treats `""` and `"0x"` as the same "absent" value:

```ts
const speedUpSignature = !isMessageEmpty(deposit.speedUpSignature) ? deposit.speedUpSignature : undefined;
```

`isMessageEmpty` already defaults `undefined` → `EMPTY_MESSAGE`, so this subsumes the `isDefined` check too. Lower stakes than `message: ""` (the validators type this as hex), but it's a one-line fix for free symmetry.

---

Two smaller notes, neither blocking:

- Partial speed-up fields (signature present, `updatedRecipient` missing) still throw, and a missing `destinationChainId` throws an opaque `invalid BigNumber value`. Both are fine by me — the first matches what `getGasCosts` already does, and the widened signature makes the second unreachable from TypeScript.
- `populateV3Relay` has no test coverage anywhere in `test/`, so the equivalence I measured above isn't pinned by anything. `resolveV3RelayCall` guarantees the two paths agree on method and args, so they can't drift on *your* code — but nothing catches an ethers-side change to how `populateTransaction` encodes. A three-line `expect(getV3RelayCalldata(...)).to.equal((await populateV3Relay(...)).data)` would lock it. Optional.

I haven't pushed anything — you're actively iterating on the branch and both changes are one-liners in files you're editing. Say the word if you'd rather I commit them.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=cba6c6605da91cfc93e2dc0b38c1b1ed&trace_start_ts=1789766620&trace_end_ts=1789853320)

pxrl:
@droplet-rl thanks for the review — both items addressed in `6d79627b`:

1. Fixed the `TVM_RAW_DATA_OVERHEAD_BYTES` comment and the `796,000` test comment to state the true numbers: 670 empty-message fills measured 279 bytes with zero variance; 280 (and therefore 796,000) is the deliberately-conservative rounding, not an exact onchain match.
2. Replaced the `isDefined(...) && !== "0x"` speedUpSignature guard with `!isMessageEmpty(...)`, matching the `message` handling one line above and closing the `speedUpSignature: ""` gap you found. Extended the regression test to cover `""` alongside `"0x"`.

Both verified: 22/22 tests passing, lint clean. Another pass whenever you get a chance.

droplet-rl:
Reviewed `6d79627`. Both fixes are correct and this is good to merge from my side.

**Verified locally** (fresh install at `6d79627`): `yarn build` ✅, `yarn lint-check` ✅, 22/22 passing ✅.

I re-ran the probe against the built output to confirm the `isMessageEmpty` swap closes the gap without over-broadening — every "absent" form now converges, and a genuine speed-up still takes the `fillRelayWithUpdatedDeposit` path:

| input | SUN |
|---|---|
| baseline | 796,000 |
| `message: ""` | 796,000 |
| `speedUpSignature: "0x"` | 796,000 |
| `speedUpSignature: ""` | 796,000 (was throwing) |
| `speedUpSignature: undefined` | 796,000 |
| real speed-up | 1,084,000 = 796,000 + 288×1,000 ✓ |

Calldata equivalence with `populateV3Relay` still holds byte-for-byte (516 plain, 804 sped-up).

---

One residual observation, **optional and not worth another round** — I'm signing off either way.

The constant's comment now reads as though 280 is pure safety margin ("measured 279 … rounded up to 280 so the estimate stays conservative"), but the 2112-byte test one file over asserts `2_908_000` as a real onchain `net_fee`, and `2908 − 516 − 2112 = 280`. So that fill's envelope genuinely *was* 280, not 279 — the margin is one byte for empty-message fills and zero for that one. The envelope varies across the population (protobuf varint widths in `timestamp`/`expiration`/`fee_limit` would do it); "zero variance" is true only within the 670-fill empty-message sample.

Worth a clause mainly because it changes what happens if the envelope ever reaches 281 — at that point 280 under-estimates rather than eating into slack. If you want it:

```ts
// Tron raw_data envelope (contract type, ref_block, timestamp, expiration, fee_limit)
// plus a single ECDSA signature. Varies slightly with protobuf varint widths: 279 bytes
// across 670 empty-message fills, 280 on the 2112-byte fill in the tests. 280 is the
// observed max — exact for the latter, one byte conservative for the former.
export const TVM_RAW_DATA_OVERHEAD_BYTES = 280;
```

Take it or leave it — happy for this to merge as-is.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=80515a56a0acab85f3b3eda672a24243&trace_start_ts=1789768183&trace_end_ts=1789854883)

pxrl:
@droplet-rl another pass please — one more real bug found and fixed since your last review, in `b8843dba`:

An independent audit caught that `updatedMessage: ""` on a sped-up fill still crashed the real ABI encoder with `invalid arrayify value` — the same "" vs "0x" gap you found for `message` and `speedUpSignature`, but on a third, structurally identical field that got missed. Fixed by normalizing `updatedMessage` through `isMessageEmpty` the same way as `message`. Verified: reverted the fix and confirmed the new regression test fails with that exact error, then restored it — 23/23 tests passing, lint clean.

Also trimmed the comments further per a fresh minimalism pass (no cuts to test coverage — a parallel adversarial audit and a parallel minimalism audit both ran against the full diff; the minimalism one found nothing else worth cutting, and re-verified two of your/our earlier keep-decisions still hold).

droplet-rl:
Reviewed `b8843db`. Good catch, and it's mine to own — I probed `message` and `speedUpSignature` in my round-2 verification and never thought to check the third field.

**Verified locally** (fresh install at `b8843db`): `yarn build` ✅, `yarn lint-check` ✅, 23/23 passing ✅.

I reproduced the bug independently rather than taking the test's word for it — passing `updatedMessage: ""` straight into `getV3RelayCalldata`, bypassing the normalization, still throws `invalid arrayify value (argument="value", value="")`. So the new test exercises a real failure, and the fix genuinely closes it: `""` and `"0x"` both yield 1,084,000 now, and real speed-ups are unaffected (1,084,000 = 796,000 + 288×1,000).

**The class is now provably closed.** Rather than eyeball it, I enumerated the dynamic params off the actual ABI:

```
fillRelay(...)                     bytes: relayData.message
fillRelayWithUpdatedDeposit(...)   bytes: relayData.message, updatedMessage, depositorSignature
```

Three dynamic `bytes` params across both methods, all three now normalized. There is no fourth field with this shape — worth stating plainly since this is the third time the same bug surfaced.

---

Two residual notes. Neither blocks merge.

**1. Partial speed-ups are now handled inconsistently.** Normalizing unconditionally means `isMessageEmpty(undefined)` → `"0x"`, which lands *before* `resolveV3RelayCall`'s `assert(isDefined(relayData.updatedMessage))` can fire. Measured:

| malformed speed-up | before | now |
|---|---|---|
| missing `updatedMessage` | threw | **1,084,000** |
| missing `updatedRecipient` | threw | threw |
| missing `updatedOutputAmount` | threw | threw |

So one missing field silently defaults while the other two still reject. Fail-soft is defensible for a fee estimator — arguably better than failing a quote over a malformed deposit — but it should be a choice, not a side effect of the `""` fix. If you'd rather keep the validation:

```ts
const updatedMessage =
  isDefined(deposit.updatedMessage) && isMessageEmpty(deposit.updatedMessage) ? "0x" : deposit.updatedMessage;
```

That normalizes `""` → `"0x"` while leaving `undefined` alone so the assert still fires. Either direction is fine; picking one deliberately is the point.

**2. Follow-up, out of scope here:** the normalization is `TvmQuery`-local by design, so `getGasCosts`/`getNativeGasCost` still hand raw deposits to `populateV3Relay` and still crash on `message: ""` — I confirmed that path throws the same `invalid arrayify value`, **on every chain, not just Tron**. Entirely pre-existing and this PR doesn't worsen it, but the asymmetry is sharper now: the same deposit produces a clean bandwidth estimate and a crashed gas estimate. Might be worth a ticket to normalize in `resolveV3RelayCall` instead, which would collapse all three fixes into one place and cover the EVM path too.

Nothing here needs another round from me — happy for this to land.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=65bb870e418770e83d2d4432dbc57f4e&trace_start_ts=1789768967&trace_end_ts=1789855667)

pxrl:
@droplet-rl another pass please — addressed your residual note from the last round, in `ab5400ba`:

Made the fail-soft-vs-fail-hard choice on `updatedMessage` deliberate rather than incidental: switched to `isDefined(deposit.updatedMessage) && isMessageEmpty(deposit.updatedMessage) ? "0x" : deposit.updatedMessage`, exactly as you suggested. "" still normalizes to "0x", but `undefined` is left alone so a malformed partial speed-up (missing `updatedMessage` but not the other `updated*` fields) still fails `resolveV3RelayCall`'s `isDefined` assert, matching how `updatedRecipient`/`updatedOutputAmount` already behave. Added a regression test for this, verified it fails without the fix (same `AssertionError` you'd expect) and passes with it.

Also cut the new comments down further per your earlier size cop. 24/24 tests passing, lint clean.

droplet-rl:
Reviewed `ab5400b`. Correct, and the PR looks done to me.

**Verified locally** (rebuilt at `ab5400b`): `yarn build` ✅, `yarn lint-check` ✅, 24/24 passing ✅.

The malformed-speed-up handling is now uniform, which was the whole point of the note:

| malformed speed-up | `b8843db` | `ab5400b` |
|---|---|---|
| missing `updatedMessage` | 1,084,000 | **AssertionError** |
| missing `updatedRecipient` | AssertionError | AssertionError |
| missing `updatedOutputAmount` | AssertionError | AssertionError |

And nothing regressed on the way: `""` still normalizes on all three `bytes` fields, plain fills are 796,000, real speed-ups 1,084,000, and the 2112-byte onchain ground truth still reproduces at 2,908,000.

One case worth calling out because the new `isDefined` guard makes `updatedMessage` reachable as `undefined` in the spread: I checked that this can't leak into the plain `fillRelay` path. A non-sped-up deposit returns 796,000 whether `updatedMessage` is absent, explicitly `undefined`, or `""` — the speed-up branch is gated on `speedUpSignature`, so the assert never sees those. No change there.

You also picked up the envelope comment from two rounds back — *"279 bytes for 670 empty-message fills, 280 for the 2112-byte test fill — 280 is the observed max"* — which closes the last thing that could have misled someone into "correcting" 280 down to 279.

Nothing further from me. The only open thread is the follow-up from last round, deliberately out of scope here: `getGasCosts`/`getNativeGasCost` still hand raw deposits to `populateV3Relay`, so `message: ""` crashes the gas path on every chain while the bandwidth path handles it. Worth a ticket to move the normalization into `resolveV3RelayCall`, which would fold all three fixes into one place and cover EVM too — but that's its own PR. Ship this.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=06d63a7d46c1e962dbec833722b0837f&trace_start_ts=1789769622&trace_end_ts=1789856322)

